PR #178, opened 2014-11-17
Updated required ruby version in readme to 2.1.2

hklebs:


jywarren:
Hi hklebs - just confirming you've seen that we've included a GPLv3 license in the latest release; we'd described it as GPLv3 since the beginning but did not have a license file.

PR #201, opened 2015-02-11
first stab at uploading images via data url; useful for some external apps

jywarren:
This is not ready for merge; it's not working yet. But for setting up the ability to post images from Infragram.org, this would be nice.


jywarren:
Crap, that was a mistake.

jywarren:
I had meant to close this, to be opened later again if the feature is picked back up. My bad; reverted.

PR #235, opened 2015-02-23
adding license file, readme tweaks

jywarren:
Hi, @btbonval, @sirMackk, @aawray, @btakita, @BobStevens and @manleyjster -- embarrassingly we don't have a license file in this repo, although it's marked as GPLv3 all over the Public Lab site. No changes; just adding the GPLv3 file itself to the repo. 

I just want to get +1s from all contributors while that's still easy to do. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ A complete rewrite of the Public Lab website on a new platform, with a whole new
 * simplify and refine common tasks like posting research notes and filtering spam
 * enable faster development (based on Ruby on Rails and Twitter's Bootstrap frameworks)
 * completely revise and streamline "following" other contributors or specific keywords, with email alerts
-work on tablets, smartphones, and in recent versions of Internet Explorer
+* work on tablets, smartphones, and in recent versions of Internet Explorer
 
 ###Key new features:
 
@@ -27,7 +27,7 @@ work on tablets, smartphones, and in recent versions of Internet Explorer
 
 You'll need: 
 
-* ruby (1.9.3 recommended; try using http://rvm.io)
+* ruby (2.1.2 recommended; try using http://rvm.io)
 * mysql (not adverse to others, but this is what we run)
 
 Installation steps:
@@ -42,7 +42,7 @@ Installation steps:
     * run `rake db:migrate`
     * OR
     * in MySQL, `drop database XXX;` for each database in `config/database.yml` and then try `rake db:setup` again
-* `rake db:seed` to populate it with initial data
+* `rake db:seed` to populate it with initial dummy data
 * `passenger start` to start up the app
 * in a browser, navigate to http://localhost:3000
 * wheeeee!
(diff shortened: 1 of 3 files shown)


sirMackk:
Definitely +1! Thanks for the shout out!

btbonval:
Since plots2 is a web service, AGPLv3 would be more appropriate.

README.md
```diff
@@ -42,7 +42,7 @@ Installation steps:
     * run `rake db:migrate`
     * OR
     * in MySQL, `drop database XXX;` for each database in `config/database.yml` and then try `rake db:setup` again
-* `rake db:seed` to populate it with initial data
+* `rake db:seed` to populate it with initial dummy data
```
btbonval: is it dummy data? I thought it was just the most basic data required to get up and running. "dummy data" makes it sound like it'd be for running tests only, but not for bootstrapping a brand spanking new and potentially production-ready system.

jywarren: I believe it's a wiki page and research note and user for devs to see --
not automated testing data, but not real data either. I said dummy data in
order to distinguish from data in the test db and so people know it's fake.

re: AGPL, I don't disagree, but we've always said GPLv3, so that's what the
assumption's been for folks who've contributed.

Can we cross license, or re-license under AGPL? I folks are OK with AGPL,
can you say so in this thread too? I'm fine with either or both.

btbonval: woops we should bring the AGPL discussion back to the group. This is just in the diffs.

jywarren: oh sorry it showed up in the same email thread for me. will do.

jywarren:
re: AGPL, I don't disagree, but we've always said GPLv3, so that's what the
assumption's been for folks who've contributed.

Can we cross license, or re-license under AGPL? I folks are OK with AGPL,
can you say so in this thread too? I'm fine with either or both.

btbonval:
Cross licensing GPLv3 and AGPLv3 would boil down to licensing as GPLv3. When dual licensing, the user gets to choose which license, in which case the AGPL license is freely ignored if someone wants to host the web service with code changes but not share code.

btbonval:
In case people are not aware of the difference:

* GPLv3 says that any executables that are shared must have the code shared with it, more or less. If you change code and run it, but never give your executable to anyone, you don't have to share code. Running a web service, for example, is not sharing the executable in the rules of GPLv3.
* AGPLv3 says that any executable running as a service that others may access must have its code shared. Thus by running the executable yourself as a web service other users might use, you are obliged to share code.

manleyjster:
+1!

Justin

BobStevens:
+1

aawray:
+1

ebarry:
+1

--

txt only: +1 336-269-1539
@lizbarry <http://twitter.com/lizbarry>

btbonval:
more +1 for AGPLv3, less +1 for GPLv3

PR #261, opened 2015-03-28
Solved issue 259. Wiki indexing is taken care of

Sreyanth:


jywarren:
I'm going to just pull in yours and take a look on the live site; I may have misunderstood the cause of this bug or there may be two overlapping bugs. But your PR has other things I'd like to merge so thanks!

Running tests first.

Issue #259, opened 2015-03-27
wiki index ordering -- with optimization difficulties


jywarren:
**Update:** see [this comment below](https://github.com/publiclab/plots2/issues/259#issuecomment-101333050) for the latest -- subtle ordering vs. optimization issue still persists. We could/should create a functional test to reproduce the ordering issue. 

I'm trying to nail that bug where on the http://publiclab.org/wiki listing page, some wiki pages show up as recent but don't actually have a recent edit. The crux is that we sort these wiki pages by `node.changed` -- a timestamp which gets updated any time a wiki page is saved. So, **starring/liking a page bumps it to the top**

```
Node.includes(:drupal_node_revision,:drupal_node_counter).where("status = 1").order("changed DESC")
```

Since sometimes pages are saved without a new revision being made (like starring/liking), when we display the edit date of the last revision, it shows up as being quite old, actually: 

```
Last edit / 7 months ago by Shannon
```

Can anyone help me craft a query that sorts DrupalNodes by the timestamp of their most recent associated node_revision, rather than by their own timestamp? Note that nodes have multiple node_revisions, linked by the nid.

---

Still not making much progress on this; the resulting query is now:

```
SELECT DISTINCT `node`.nid FROM `node` LEFT OUTER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` ORDER BY node_revisions.timestamp DESC
```

but somehow I'm not getting all recent revisions. Timestamps (of assoc. revisions) are:

```
 => [1427465867, 1427218775, 1426896777, 1426896105, 1426895691, 1426895121, 1426894917, 1426894230, 1426894036, 1426893636] 
```

I can confirm that 

```
SELECT `node`.* FROM `node` WHERE `node`.`nid` = 10229 LIMIT 1
SELECT `node_revisions`.* FROM `node_revisions` WHERE `node_revisions`.`nid` IN (10229) ORDER BY timestamp

 => 1427412864 
```

Which means node 10299 should appear in results from the top query, but it doesn't. I think I'm missing something obvious here. 

My best guess is that since each node has multiple node_revisions, we are not ensuring that the revision joined is the most recent one, although we are sorting the resulting list by recency (timestamp) once the join has been made. 

Is there a way to sort before the join happens? 

I'm going to abandon this for now; An alternate solution could be to display **either** the last edited date, or the last liked/starred date, whichever is more recent, in the list view.

Sreyanth:
Solved in the latest pull request.

[...]

PR #262, opened 2015-03-30
Another attempt at solving #259

Sreyanth:


jywarren:
wow that did work. Odd -- that's the old Rails 2 syntax -- i guess we just hadn't fully reproduced it. Maybe there's some ordering issue with the chained method syntax we missed? Anyhow, very happy to see this working, thanks Sreyanth!!! Pulling this in now.

jywarren:
and now live at publiclab.org!

PR #269, opened 2015-04-13
password reset visible to admin

btbonval:
I had been mixing up pushes to my repo and the Public Lab repo, so this particular commit didn't rebase cleanly.

I had to do a seemingly pointless merge to reconcile the differences, although the PR summary below shows it is just the same commit twice. Dunno. The diff log has everything it should, not sure why this got so messy.


jywarren:
Yeah this looks pretty straightforward; go ahead and merge!

Issue #203, opened 2015-02-12
links in notification emails have are relative or incorrect


jywarren:
> Hi! There's been a response to your research note "Planning for Spectral Workbench 2.0". Do NOT reply to this email; click this link to respond:

leads to simply `http://notes/warren/02-10-2015/planning-for-spectral-workbench-2-0`

jywarren:
via Mathew's merged-in issue: 

![image](https://cloud.githubusercontent.com/assets/24359/10293218/55b3a42e-6b81-11e5-9983-c3670fcff976.png)

icarito:
It is important that if there is an explicit address in the text, the anchor link points to the same address in the text, otherwise spam filters will definitively take notice. I would prioritize this.

icarito:
Note this is part of the problem in issue https://github.com/jywarren/web/issues/5 (Publiclab.org mail issues)

jywarren:
Also using this as a catchall issue for the problems in callouts -- where we say things like `@icarito` in comments and notes. @icarito has been testing this in jywarren/web#5.

jywarren:
I believe this is confirmed; I sent a callout at: http://publiclab.org/notes/warren/12-12-2010/front-illumination-ccds-used-digital-cameras-dont-perform-well-uv#c12653

And the log shows:

````
Completed 200 OK in 1.2ms (Views: 0.1ms | ActiveRecord: 0.3ms)
  Rendered comment_mailer/notify_callout.html.erb (3.9ms)
Sent mail to xxxxxx@xxxxxx.com (66.8ms)
Sent mail to xxxxxxx@xxxxxx.org (77.7ms)
````

Issue #229, opened 2015-02-21
fix factory_girl sequences, which are broken


jywarren:
they're repeating across tests and forcing test_public_pages and other tests to destroy their test data after each test, as well as manually specify primary keys (nid) and titles.

jywarren:
removed factory_girl entirely

PR #327, opened 2015-10-07
Notification email fixes & callout tests, test overhaul

jywarren:
I broke a bunch of tests getting this new test for comment callouts working, but all is passing now. We're using a hybrid of FactoryGirl and Fixtures, because FactoryGirl is a little illegible (reparable, i guess) but fixtures isn't working for UserSession.create "login faking" so FG is needed for login-dependent tests. 

* Should close https://github.com/publiclab/plots2/issues/203
* related FactoryGirl issue: https://github.com/publiclab/plots2/issues/229

### DIFF
--- a/app/mailers/comment_mailer.rb
+++ b/app/mailers/comment_mailer.rb
@@ -26,4 +26,5 @@ class CommentMailer < ActionMailer::Base
     @comment = comment
     mail(:to => user.email, :subject => "You were mentioned in a comment.").deliver
   end
+
 end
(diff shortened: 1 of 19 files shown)


jywarren:
@btbonval @icarito @ebarry  - this should fix #203, and institute an automated test for callouts so we'll know if it breaks again in the future. Look good? I can post in an hour or so.

jywarren:
This is now deployed.

PR #386, opened 2016-03-04
Projects dropdown Safari fix

jiteshjha:
"Projects" dropdown is working on iOS devices now.

Tested on  Safari 9.0.3.

@jywarren


jywarren:
Hi, jiteshjha - this is a great fix, but I'm about to merge #381, which should resolve this. Thanks for the catch and if you want to double check in the new Bootstrap 3 version, it's possible that your fix will need to be applied there. 

However, for the time being, I'm going to close this and not merge. Thanks again and reopen if you still see the issue.

jiteshjha:
Sure, I'll track. :smile:

jywarren:
Hi, Jiteshjha - I think this fix is needed on the new code, having just borrowed an iPad to test. Can you rebase or adapt it and re-submit? Thanks for finding a fix! It'll have to apply to all menus, I believe.

PR #388, opened 2016-03-04
Safari fix

jiteshjha:
Works on Safari 9.0.x.
Checked on an iPhone 6S.

@jywarren


jywarren:
Fantastic, thanks. In the future if you could confirm that you ran tests too, that'd be extra helpful. Although I can see the changes in this one and that they involve no Ruby code. 

Thanks, merging!

jiteshjha:
Sure thing! :smile:

Issue #401, opened 2016-03-09
Root url for wiki pages fails to load


ananyo2012:
##What happened just before the problem occurred
When trying to go to the /about page or any root wiki page from localhost:3000 the page fails to load
![screenshot from 2016-03-09 12 04 42](https://cloud.githubusercontent.com/assets/10486343/13627263/3f78eb72-e5ef-11e5-9a74-e14e190995db.png)

##Relevant URLs

localhost:3000/about

##PublicLab.org username
ananyo2012
(to help reproduce the issue)

##Browser, version, and operating system
Google Chrome 47.0.2526.106 (64 bit)
OS: Ubuntu 14.04 LTS

A simple change in root method in wiki_controller is solving this problem. Should i add a commit for it?

For bug reports, fill out the above template; for feature requests, you can delete the template.

jywarren:
Hi, I believe this is not a bug - on closer look, the correct address for the about page in the seed data is http://localhost:3000/wiki/about -- I'm aware that the header shows http://localhost:3000/about, but unless you want to correct the seed data at https://github.com/publiclab/plots2/blob/master/db/seeds.rb#L30, this is just a mismatch between seed data and production data which does not result in a bug for a user of the live site. Thanks!

PR #408, opened 2016-03-11
Fix bug in wiki root link

ananyo2012:
This solves the problem mentioned in Issue #401

### DIFF
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -194,7 +194,7 @@ class WikiController < ApplicationController
 
   # wiki pages which have a root URL, like //publiclab.org/about
   def root
-    @node = DrupalNode.find_root_by_slug(params[:id])
+    @node = DrupalNode.find_by_slug(params[:id])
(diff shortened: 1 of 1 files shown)


jywarren:
Hello, have you run `rake test` to confirm that this doesn't break anything? 

Thanks for your contribution!

ananyo2012:
@jywarren Some tests fail for the notes controller which were also failing for the original code. But all tests for wiki controller pass.

jywarren:
Which tests? All are passing on master, so perhaps you could pull in the
latest master and rebase your changes on top of it.

Thanks!

jywarren:
Also, are you sure that these pages have paths like `/about` ? the difference by find_root_by_slug and find_by_slug is that find_root only finds those which have the url without "/wiki/xxx" for example -- just "/xxx". So I'm not sure this is a bug.

jywarren:
Hi, ananyo - I think this can safely be called not a bug, as my response to #401 explains. Thanks!

Issue #309, opened 2015-08-24
Banned user accounts still show as having liked things


ebarry:
Hi, this is a `first-timers-only` issue. This means we've worked to make it more legible to folks who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**. If you have contributed before, consider leaving this one for someone new, and looking through our general [help wanted issues](https://github.com/publiclab/plots2/labels/help-wanted). Thanks!

### The problem

We ban spam users by marking their user record and their posted notes with `status: 0`. But users can "like" things using the star button on most content, as in this screenshot:

![screenshot 2016-03-16 at 11 46 30 am](https://cloud.githubusercontent.com/assets/24359/13818579/c56349da-eb6c-11e5-8ada-d4813a71ab76.png)

For an example, see [FlorentTecumsah](http://publiclab.org/profile/FlorentTecumseh), who is banned (so only mods and admins will be able to navigate to their profile page), on http://publiclab.org/wiki/nonprofit-initiatives:

![screenshot 2016-03-16 at 11 47 57 am](https://cloud.githubusercontent.com/assets/24359/13818620/fb9fad18-eb6c-11e5-870f-465c123eef37.png)

We don't want banned users showing up there. 

### Solution

The query for "likers" of a given node is here: https://github.com/publiclab/plots2/blob/master/app/models/drupal_node.rb#L107-L118

We want to filter with `.where(status: 1)`, but probably will need to `joins` the user model, so we can check the `status` field. That'll probably look like this:

````rails
self.node_selections.where(:liking => true)
    .joins(:drupal_users)
    .where('users.status = 1')
    .collect(&:user)
````

To try this out, try:

* logging in as a user
* liking something
* logging in as admin (using the [seed data's password](https://github.com/publiclab/plots2/blob/master/db/seeds.rb#L13))
* banning the user using the "spam" button (see it in the popover pictured above)
* checking the "likers" list as shown above, to see if the now-banned user

Know about automated testing? There is a spam user (uid #3) in the fixtures, so this can be tested for, too! See https://github.com/publiclab/plots2/blob/master/test/fixtures/users.yml

### Steps to fix

* [ ] claim this issue with a comment
* [ ] set up a repository locally and make sure that all tests pass
* [ ] start a pull request (see [contributing to Public Lab software](https://publiclab.org/wiki/contributing-to-public-lab-software)) but mark it as "in progress"
* [ ] fix the issue following the steps above
* [ ] try the manual test listed above
[...]

PR #439, opened 2016-03-22
fixed likes to be removed when spammers are banned

franktank:

* [ ] tests pass -- `rake test`
* [ ] code has been rebased on top of latest master (check if another pull request was added recently, and please rebase)
* [x] pull request is descriptively named and, if possible, multiple commits squashed if they're smaller changes

Thanks!

### DIFF
--- a/app/models/drupal_node.rb
+++ b/app/models/drupal_node.rb
@@ -110,7 +110,7 @@ class DrupalNode < ActiveRecord::Base
 
   # users who like this node
   def likers
-    self.node_selections.where(:liking => true).collect(&:user)
+    self.node_selections.where(:liking => true).joins(:drupal_users).where('users.status = 1').collect(&:user)
   end
 
   def liked_by(uid)
--- a/app/views/like/_like.html.erb
+++ b/app/views/like/_like.html.erb
@@ -3,7 +3,7 @@
 
     <li class="btn btn-default btn-sm" data-html="true" rel="popover" data-content="<%= "No tags" if tagnames.nil? || tagnames.length == 0 %><% if tagnames %><% tagnames.each do |tagname| %><p style='margin-bottom:3px;'><a href='/subscribe/tag/<%= tagname %>' class='btn btn-default btn-mini'><%= tagname %></a></p><% end %><% end %><hr /><a class='btn btn-sm'><i class='fa fa-user'></i> <%= node.author.name %></a>" title="Follow by tag or author" data-placement="left"><i class="fa fa-eye"></i><span class="hidden-xs hidden-sm"> Follow</span></li>
 
-    <li rel="tooltip" title="Helpful? Follow it!" class="btn btn-default btn-sm" node-id="<%= node.id %>" id="like-button-<%= node.id %>"><span id="like-star-<%= node.id %>" class="fa fa-star<% if current_user && !node.liked_by(current_user.uid) %>-o<% end %>"></span> <span id="like-count-<%= node.id %>"><%= node.likers.length %> </span></li>
+    <li rel="tooltip" title="Helpful? Follow it!" class="btn btn-default btn-sm" node-id="<%= node.id %>" id="like-button-<%= node.id %>"><span id="like-star-<%= node.id %>" class="fa fa-star<% if current_user && !node.liked_by(current_user.uid) %>-o<% end %>"></span> <span id="like-count"><%= node.likers.length %> </span></li>
 
     <li class="btn btn-default btn-sm" rel="popover" data-placement="left" data-html="true" data-title="Tools" data-content="
 <p><b>Users who liked this</b></p>
--- a/app/views/notes/_notes.html.erb
+++ b/app/views/notes/_notes.html.erb
@@ -23,7 +23,7 @@
             <%= time_ago_in_words(Time.at(node.latest.timestamp)) %> ago
           <% end %>
           | <i class="fa fa-eye"></i> <%= number_with_delimiter(node.totalcount) %> <span class="hidden-xs hidden-sm">views</span>
-          | <i style="<% if node.likes > 0 %>color:#db4;<% else %>color:#888;<% end %>" class="fa fa-star-o"></i> <%= node.likers.length %>
+          | <i style="<% if node.likers.length > 0 %>color:#db4;<% else %>color:#888;<% end %>" class="fa fa-star-o"></i> <%= node.likers.length %>
         </small></p>
 
         <% if @compact.nil? %>
--- a/app/views/notes/_notes.html.erb
+++ b/app/views/notes/_notes.html.erb
@@ -23,7 +23,7 @@
             <%= time_ago_in_words(Time.at(node.latest.timestamp)) %> ago
           <% end %>
           | <i class="fa fa-eye"></i> <%= number_with_delimiter(node.totalcount) %> <span class="hidden-xs hidden-sm">views</span>
-          | <i style="<% if node.likes > 0 %>color:#db4;<% else %>color:#888;<% end %>" class="fa fa-star-o"></i> <%= node.likes %>
+          | <i style="<% if node.likes > 0 %>color:#db4;<% else %>color:#888;<% end %>" class="fa fa-star-o"></i> <%= node.likers.length %>
         </small></p>
 
         <% if @compact.nil? %>


jywarren:
Hi @franktank, I see you removed the ID from the id name of the like button, `like-button-<%= node.ID %>`. I think this may break some behaviors when multiple like menus are displayed, such as on /blog when there are multiple posts tagged "blog". Can you confirm, comment, or fix? Thanks!

franktank:
When I tested it manually on /blog with multiple test posts and on /notes/popular, nothing breaks.

Also, to further enhance the ban feature, do we want banned users to be deleted or have their spam posts be deleted?

jywarren:
Hi, @franktank - ok, i'll test this out and try to merge. In the future, please submit pull requests from a named feature branch -- this one could be, for example, "franktank-liker-spam". This helps you because if the feature needs to be rebased over master, you have a copy of master to rebase it onto. It also helps us keep things tidy on our end. Thanks!

jywarren:
I tested it and the like count is not incremented/decremented correctly, which I believe relates to the issue I was concered about. See this line in the like.js file:

https://github.com/publiclab/plots2/blob/master/app/assets/javascripts/like.js#L5

I think if you add those back in, it'll be fine. 

Then, you'll have to rebase your code -- I think that if you're currently working in your master branch, you could:

* [ ] make your fixes/additions, commit them
* [ ] run `git branch franktank-liker-spam` to move it to a new branch
* [ ] rewind your master to the last spot you began working on top of with `git reset --hard aeb9db472bee3d8db08285b4850959f351d376cf` (i got this hash from the commit log using `git log`)
* [ ] then get the latest commits to master with `git pull https://github.com/publiclab/plots2.git master`
* [ ] then `git checkout franktank-liker-spam` to go back to your new code
* [ ] run `git rebase master` to rebase all your changes on top of master again
* [ ] run `git push origin franktank-liker-spam` and you'll push up a named feature branch with your new code

Then, go to  https://github.com/franktank/plots2 and it should prompt you to open a new pull request with the named feature branch, against publiclab/master. Then we meet up in a new PR and go through tests again. 

I know this is a long process I've outlined, but it'll help us keep everything in order and it's not hard once you get the hang of it. The typical process is outlined in https://publiclab.org/wiki/contributing-to-public-lab-software. 

Thanks, Franky! I really appreciate your contributions!

franktank:
Hmm, interesting. How come in the views, the number next to the fa-star increments after a like? 

I'll work on adding it back in and testing with the console as well to see if like count increments. 

Thanks for the feedback!

jywarren:
Watch out for the big commit I just merged into master -- i sent an email
to plots-dev about the changes it may cause.

The number next to the star is indicating the # of people who've liked that
post. Thanks, Franky!

franktank:
Yes, I meant on my local server, the # of people next to the star increments and decrements with likes while having the node.id code removed.

However, when I had the node.id left in, the # did not decrease after banning a user.

jywarren:
Ah, interesting. Why didn't it decrease -- did the `where()` filter not
affect how you're counting?

There's this code in
https://github.com/publiclab/plots2/blob/master/app/controllers/like_controller.rb#L35-L66
which also counts likes and caches the value (line 43). I bet you have to
run that to get it to re-sum. I think this functionality could be moved
into a model, and run on save, using `before_save`, as we do in
drupal_node.rb:
https://github.com/publiclab/plots2/blob/master/app/models/drupal_node.rb#L58

That way developers don't have to remember to increment it; it's just done
automatically on saving.

franktank:
Not sure why it didn't decrease, as I mentioned in #309, the display with "Users who liked this" would remove the spammer from the list, but next to the fa-star, the number remained the same, and when I removed node.id, the number decremented with it. Also, I ran DrupalNode.all in the console after each like/unlike and the numbers change with the current code that I have.

Where is it that you see the counts not incrementing/decrementing properly?

Thanks!

jywarren:
Hi, Franky - take a look at line 54 here: 
https://github.com/publiclab/plots2/blob/master/app/controllers/like_controller.rb#L35-L66

There's a column, where the # of likers is counted up and stored in the record, and that's the value that's used in the HTML. The cached value is:

              node.cached_likes = node.cached_likes + 1

It's used here, in `/app/models/drupal_node.rb`:

https://github.com/publiclab/plots2/blob/master/app/models/drupal_node.rb#L107

So, you have to update that value by recounting to get it to increment in the view. As this is relatively confusing, I'm proposing that the updater to cached_likes gets automated somehow.

franktank:
So I think I am misunderstanding, but after taking out the `node.id`, I tested to see if the liking system was still working. I made a blog post with a test account. Posted a note. Liked the note. Checked rails console with DrupalNode.all, found that the cached_likes number changed. Logged into the admin account, banned the test user account. Checked the cached_likes, and the number changed again. 

I am not quite sure where the problem currently is. 

Thanks.

jywarren:
Ah, i think we're not being specific enough, and I may have misunderstood as well, please forgive me. It sounds like we want:

1. the cached_likes to decrease when a user who's liked content is banned
2. the likes displayed on the initial pageload to be correct
3. the likes displayed to correctly increment based on the javascript for "liking" a post in a list view

anything else? 

This sounds like we need some tests to establish as many of these as we can, and then (since we don't yet have a JavaScript test suite) we can manually test the javascript.  

I've created a new test suite for the like controller, in #453, which may help you to debug. If you pull the latest master, rebase your code on top, perhaps those tests will help you figure out what's wrong?

jywarren:
I realized there may be other tests that could help here. Perhaps adding a test for banning a user, which then checks the like count on a note they've liked, to ensure that it goes down by 1?

franktank:
Sure, I will start working on the tests! Is it okay if I write RSpec tests?

jywarren:
I think so -- the tests we currently have are very basic. What is the
prospect for rspec integrating with our current test suite, or running our
current tests? Would people have to run two different test suites?

franktank:
Yes, the two suites would have to be run separately. I think integrating the current test suite with RSpec would be fairly easy to do since there is not too much test code.

jywarren:
Short term (as in this pull request) we could perhaps modify the rakefile?

franktank:
Okay sure, we can add a task to the Rakefile and it will run RSpec tests as well.

franktank:
Hi Jeff,

How would I approach creating the users and note? Currently the code I have is `require 'rails_helper'

describe "banning a spammer" do

    it "decrements cached_likes on a note" do
        byebug
        spam_user = User.create(username:"spammer", email: "test@test.com", password: "publiclab")
        byebug
        spam_uid = spam_user.uid
        byebug
        writer_user = User.create(username:"writer", email: "writer@test.com", password: "publiclab")
        writer_uid = writer_user.uid
        byebug
        research_note = DrupalNode.create(type:"note", title: "NicePost", uid: writer_uid, path: "/wiki/donate")
        research_note.liked_by(spam_uid)
        expect { spam_user.status = 0}.to change {research_note.cached_likes.count}
    end    
end`

Currently I am getting errors when I create the user. I have byebug in to help me debug, but I am having trouble finding the cause.

Thanks,
Franky

jywarren:
Can you post the errors you're seeing? You can look at this example in the pull request I just opened: #459. There's an oddity we haven't figured out where to save the basic user model, you have to pass it an empty hash, like `user.save({})`. This is annoying, and if you can figure out a solution, that'd be welcome too.

I'm sorry this has dragged on. Do you think the tests you're writing will help answer your questions?

franktank:
The error is " Failure/Error: result = super
     
     NoMethodError:
       undefined method `[]' for true:TrueClass"

and this test will help us see if the cached_likes is changing when the spammer is banned.

jywarren:
On what line? Perhaps you're getting an array and trying to call cached_likes on the array, but you need to say array.first.cached_likes or something?

franktank:
Well, going into the console I accessed a note and tried to get to the cached_likes and this is what I am getting: 
`2.1.2 :016 > drup = DrupalNode.where(title: "Blog Post")
  DrupalNode Load (0.5ms)  SELECT "node".* FROM "node" WHERE "node"."title" = 'Blog Post'
 => [#<DrupalNode nid: 8, vid: 0, type: "note", language: "", title: "Blog Post", uid: 1, status: 1, created: 1458431639, changed: 1458431639, comment: 1, promote: 0, moderate: 0, sticky: 0, tnid: 0, translate: 0, cached_likes: 1, drupal_comments_count: 1, drupal_node_revisions_count: 1, path: "/notes/admin/03-19-2016/blog-post">] 
2.1.2 :017 > drup.cached_likes
NoMethodError: undefined method `cached_likes' for #<ActiveRecord::Relation:0x000000064a9a60>
        from /usr/local/rvm/gems/ruby-2.1.2/gems/activerecord-3.2.22.2/lib/active_record/relation/delegation.rb:45:in `method_missing'
        from (irb):17
        from /usr/local/rvm/gems/ruby-2.1.2/gems/railties-3.2.22.2/lib/rails/commands/console.rb:47:in `start'
        from /usr/local/rvm/gems/ruby-2.1.2/gems/railties-3.2.22.2/lib/rails/commands/console.rb:8:in `start'
        from /usr/local/rvm/gems/ruby-2.1.2/gems/railties-3.2.22.2/lib/rails/commands.rb:41:in `<top (required)>'
        from script/rails:6:in `require'
        from script/rails:6:in `<main>'`

jywarren:
Yes, the `where()` call returns an array of results, so you have to run
.cached_likes on one of the results. Even if there's only one result, it's
in an array, hence the `[#<DrupalNode... ..>]`

franktank:
Yes, I did drup = Drup.where()

jywarren:
Right, so you need drup[0].cached_likes or drup.first.cached_likes -- make
sense?

franktank:
Oh, that makes sense! Thanks!!

jywarren:
No prob!

franktank:
Also, how can I access the status of the user? Its not part of User? 

"2.1.2 :030 > User
 => User(id: integer, username: string, email: string, crypted_password: string, password_salt: string, persistence_token: string, login_count: integer, failed_login_count: integer, last_request_at: datetime, current_login_at: datetime, last_login_at: datetime, current_login_ip: string, last_login_ip: string, created_at: datetime, updated_at: datetime, openid_identifier: string, role: string, reset_key: string, photo_file_name: string, photo_content_type: string, photo_file_size: string) "

jywarren:
Ah, it's user.drupal_user.status, unfortunately because we haven't finished
merging those two models.

franktank:
Okay, and also where is the like being generated?

jywarren:
In the view? In /app/views/like/ is a good place to look, but also
/app/assets/javascript/like.js

franktank:
I'm looking for a way to set the value of liked_by() from a user to true.

jywarren:
That's what currently happens at the bottom of the likes controller. But I
think it should be tucked into the node_selection model, and accessible by
a user method like `user.add_like(node)` in the future.

franktank:
Hm, I will work on adding that since we won't be able to write tests to check for likes incrementing/decrementing without it.

jywarren:
Great!

franktank:
Hi Jeff,

Quick question, where is the association between DrupalNode and User for the cached_likes?

Thanks!

jywarren:
Hmm, not sure exactly what you're asking but it's be at the top of one of
those two files. There may not be a formal active record association until
we resolve the uid/ID issues but we will hopefully add one soon. For now
you can rely only on common user.username = drupal_user.name

franktank:
I mean, how can I set a 'like' on a note to be from a selected user?

jywarren:
Each like is a `node_selection` record -- which has a `node_selection.user_id` column. The user_id corresponds to the DrupalUser.uid field. Is that helpful?

franktank:
Perfect.

Thanks Jeff!

franktank:
Hi Jeff,
I was thinking for the add_like function, we can take in values of uid and nid and simply do NodeSelection.create(user_id: uid, nid: nid, following: true, liking: true). However I get the error ActiveModel::MassAssignmentSecurity::Error: Can't mass-assign protected attributes: user_id, nid. Should I try adding those values to the attr_accessible?

jywarren:
Sorry, @franktank, I was in staff meetings all last week and am a bit behind. Catching up now. 

for your create() function, yes, you can add them to attr_accessible, that is fine. 

`cached_likes` make it more efficient to sort large #s of nodes by their like count, without having to include or at a minimum `count()` potentially thousands of like records. It's intended to provide an optimization for sorting by like count.

jywarren:
Hi, Franky - shall we try and wrap up a portion of this? I'm concerned that leaving the PR open too long will make integration harder as the master branch shifts.

franktank:
Hi Jeff,

Yes for sure, I've been swamped with exams and projects lately. I will work on wrapping it up this weekend.

jywarren:
Oh, super! Thanks!

franktank:
Made new pull request #498

PR #443, opened 2016-03-23
fixes issues with css.map files

jywarren:
Small fix for .css.map files during `rake assets:precompile`


jywarren:
See fix copied from MapKnitter here: https://github.com/publiclab/mapknitter/issues/212

Issue #472, opened 2016-03-27
clean up deployment and legacy files


jywarren:
This is just a stub to keep track of this task; @icarito and I will be dealing with it. 

* [x] @icarito, do you still use the startup/restart/stop scripts in the home directory?
* [x] I also want to clean up db/empty.sql
* [x] @icarito, were there important reasons to remove `log/.gitkeep`? Or maybe I did that? do we need it?

Thanks!

icarito:
On 27/03/16 12:17, Jeffrey Warren wrote:
>
> This is just a stub to keep track of this task; @icarito
> <https://github.com/icarito> and I will be dealing with it.
>
>   * @icarito <https://github.com/icarito>, do you still use the
>     startup/restart/stop scripts in the home directory?
>
Nope, all starting and stopping is now done as part of the NGINX
initialization as described in our deployment docs:
https://github.com/jywarren/web/wiki/Deployment
>
>   * I also want to clean up db/empty.sql
>
That's not used as far as I can tell.
>
>   * @icarito <https://github.com/icarito>, were there important
>     reasons to remove |log/.gitkeep|? Or maybe I did that? do we need it?
>
It's only needed when cloning initially. I may have removed it when
deploying. I restored it with `git checkout log/.gitkeep` to keep git
happy.

PR #474, opened 2016-03-28
refinements to install instructions, development sqlite db, cleanup

### LINKED ISSUE #469: rename development.db to development.db.example
## Problem

People are accidentally committing changes to their development.db - we should fix this if we ship it.

@lalithr95 has suggested renaming to development.sqlite, but I don't think the extension is the problem. We can keep using development.db, and add development.db to the .gitignore file, I believe. Since development.db is already in a lot of peoples' repositories now, I'm inclined not to change it. This indicates that there's not a standard: http://stackoverflow.com/questions/808499/what-is-the-best-extension-name-sqlite-database-files . And if people choose to add .sqlite to their database.yml, then it won't be tracked, so we don't have to worry about it. 

So I think we just need to confirm that:

* [ ] Does .gitignore ignore new copies of the file when pulling from remote? 
* [ ] Does .gitignore prevent a copy of the file from being created locally on initial git clone?
* [ ] Does a .gitignore ignored file cause conflicts when pulling down a new version over a locally changed file? 

****

**Update:** Apparently, .gitignore will only stop files from becoming git-tracked. It does not, [according to this post](http://stackoverflow.com/questions/1274057/making-git-forget-about-a-file-that-was-tracked-but-is-now-in-gitignore?rq=1), stop tracking updates to files which are already in git. So it seems the above plan won't work. [More info here](http://stackoverflow.com/questions/1139762/ignore-files-that-have-already-been-committed-to-a-git-repository?rq=1)

****

The alternative is to add one more step to the setup instructions, and ship a development.db.example (at which point we might as well rename it development.sqlite.example) and tell people to copy it over if they intend to use the ready-to-go sqlite database.

jywarren:
Closes #469, #472. Pull request and issue template additions, README fixes/changes.

This PR will change the example sqlite database to development.sqlite, and remove it from git tracking, but provide a development.sqlite.example file. 

* [x] code has been rebased on top of latest master (check if another pull request was added recently, and please rebase)
* [x] pull request is descriptively named and, if possible, multiple commits squashed if they're smaller changes


jywarren:
Manually merged!

Issue #479, opened 2016-04-02
New changes in install_cloudnine.sh


ananyo2012:
### What happened just before the problem occurred

Now that there is no development.db database in the master branch of the repository the cloud9 deployment no longer works properly. A new user won't be able to see the the deployment running as there is no development.db file. 

We need to change the **database.yml.cloud9** file and also copy and rename the **development.sqlite.example** file to **development.sqlite**. Should I get this issue fixed?

jywarren:
Yes please! Very good catch, thank you. Please submit a PR with `database.yml.cloud9` changed to point at `development.sqlite` instead of `development.db`

But please do not change `development.sqlite.example` -- new instructions in the README tell users to make a copy themselves named `development.sqlite` -- because if we version track that database, it'll keep getting included in submitted pull requests. The actual used database file should be outside version tracking.

jywarren:
And yes, also we should include a line in `install_cloudnine.sh` as follows:

`cp development.sqlite.example development.sqlite`

Thanks!

ananyo2012:
@jywarren Yes I was going to make the same changes. I will make a PR soon with the changes mentioned.

jywarren:
Great, thanks!!!

ananyo2012:
@jywarren Please review the PR. I have made the cloud9 fixes. I have also added test.sqlite file in gitignore.

jywarren:
Merged, thanks!

PR #482, opened 2016-04-04
Cloud9 fix

ananyo2012:

#479 is solved with this PR. Also added test.sqlite to gitignore.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ data
 Gemfile.lock
 coverage*
 development.sqlite
+test.sqlite
 config/database.yml
 config/amazon_s3.yml
 config/initializers/recaptcha.rb
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ This is a quick installation for use with the cloud environment https://c9.io - 
 1. If you have a GitHub account, visit https://c9.io and log in with the GitHub button.
 2. Fork this repository to your own GitHub account, creating a `yourname/plots2` project.
 3. Name your project, then enter `yourname/plots2` in the "Clone from Git or Mercurial URL" field, and press **Create Workspace** 
-4. In the command line prompt at the bottom of the page, type `./install_cloudnine.sh` and press enter.
+4. In the command line prompt at the bottom of the page, type `. ./install_cloudnine.sh` and press enter.
 5. Enter your username when prompted, and run `rails s -b $IP -p $PORT` when it's done.
 6. You're done! Go to the URL shown!
 
--- a/config/database.yml.cloud9
+++ b/config/database.yml.cloud9
@@ -1,14 +1,14 @@
 development:
   adapter: sqlite3
-  database: development.db
+  database: development.sqlite
 
 production:
   adapter: sqlite3
-  database: production.db
+  database: production.sqlite
 
 # Warning: The contents of the database defined as "test" will be erased and
 # re-generated from your development database when you run "rake".
 # Do not set this db to the same as development or production.
 test:
   adapter: sqlite3
-  database: test.db
+  database: test.sqlite
--- a/install_cloudnine.sh
+++ b/install_cloudnine.sh
@@ -1,11 +1,12 @@
 read -p "Enter your cloud9 username: " un
 rvm install ruby-2.1.2
-rvm 2.1.2
+source $(rvm 2.1.2 do rvm env --path)
 sudo apt-get update
 sudo apt-get -y install imagemagick ruby-rmagick
 npm install -g bower
 bower install
 bundle install --without production
+cp development.sqlite.example development.sqlite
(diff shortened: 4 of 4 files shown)


ananyo2012:
@jywarren Please review the PR. I have made the cloud9 fixes. I have also added test.sqlite file in gitignore.

ananyo2012:
@jywarren I am facing a **rvm** related problem in cloud9 deployment. The `rvm 2.1.2` command won't work in the install_cloudnine.sh script. I am getting this error

`RVM is not a function, selecting rubies with 'rvm use ...' will not work.`

Though when I run the same command in the terminal it works fine and running the installation script runs the deployment.

I have looked up for the solutions and found it is happening because the console is not running as login shell. Solutions for gnome terminal is given [here](https://rvm.io/integration/gnome-terminal). But such a thing is not possible in cloud9(I guess so).

jywarren:
Hi, @ananyo2012 - good troubleshooting, this is a common issue with RVM. I believe it's possible that we could solve it by running:

    [[ -s "$HOME/.rvm/scripts/rvm" ]] && source "$HOME/.rvm/scripts/rvm" # Load RVM into a shell session *as a function*

This is what RVM puts into your `~/.profile` and seems relevant. Can you try this? Is this a concern before I merge this PR?

I also see that you haven't checked "passes tests" -- does it? I look for that to be checked before merging in PRs, so please check it if they do, even if this code isn't something that'd be expected to break tests. Thanks!

jywarren:
Also, I don't believe we should add `rake db:migrate` -- it's preferred to simply commit a freshly `rake db:setup`'ed database file with the latest schema. I can do this or you can, but please do remove the `migrate` command unless you have a reason to think it'd be better. Thanks!

ananyo2012:
@jywarren Actually I think `rake db:migrate` should be added since there may be migrations that will be added over course of time and the current sqlite database file doesn't contain all of them. I simply face this issue while running the installation script now as there is the last migration that is not present in the sqlite database file and I am prompted to run that migration while starting the server. For first timers I think it would be best if they would see the deployment running just after the running the installation script. So what do you think now? Should I keep that line?

ananyo2012:
@jywarren Fixed the problem of setting Ruby version using `source $(rvm 2.1.2 do rvm env --path)`. But for setting the ruby version after the script runs we have to run the script as `source ./install_cloudnine.sh` or `. ./install_cloudnine.sh` since it is an environment variable. Otherwise it won't persist after running the script and the server won't start. I have include it in the Readme instructions.

Optionally we could use the line `source "/usr/local/rvm/scripts/rvm"` since cloud9 has rvm installed in its root. `source "$HOME/.rvm/scripts/rvm"` won't work as rvm is not installed in cloud9  home. This also involves running the script in the same manner as stated above.

ananyo2012:
@jywarren I have also checked the tests passing. Actually the tests were not passing in the sqlite database. So I was waiting for your fixes before I checked. Now they will pass as expected.

jywarren:
I think that we could include the `rake db:migrate` as a backup, but that we should always include an up-to-date version of development.sqlite.example. But to prove your point, I forgot to do that in the sqlite test fixes I made -- so I'll do that now, and maybe we should add a checklist item to the pull request template for updating it? 

Anyhow, for your pull request, I'd like to see 

1. commits squashed
2. rebase over latest master
3. some explanation in the README for the option of `source` or not, because if we simply say "or", some folks may be confused as to why there is an option. Since you've included the `source` in the `install_cloudnine.sh` file, can't we skip it in the README?

Thanks, I appreciate the extra effort in this PR.

ananyo2012:
@jywarren Ok I will remove the `source` line from the README. Using `source` or `.` gives the same result. Since I have already mentioned unless we run the script with `source` or `.` the ruby version won't persist after execution of the script. The `source` included in the `install_cloudnine.sh` file ensures that the environment is set while running the script but it doesn't persist after the execution ends. If you run `ruby -v` after running the script(when you haven't use `source` or `.` to run the script) you will get the default ruby version.

ananyo2012:
@jywarren Ok then I will remove the `rake db:migrate` line from the script. Yes that is a better idea. We can tell people to include changes in development.sqlite.example when there is a new migration. But again there are always chances that there are additional changes in the database file while the user is working in the development environment and since it is a binary file we can't really see the changes made. So it is best for the maintainer to make such changes. What are your thoughts on it?

jywarren:
Ah, i didn't realize that it was `source` or `.` - i read too quickly and thought it was just `source` or nothing. Thanks for the clarification!

I think we just have to be rigorous about keeping the development.sqlite.example file up to date. I'll add it as a step in the PR template to remind ourselves. 

Great, I think this can be merged on Friday.

PR #548, opened 2016-05-27
/research based on dashboard code

jywarren:

### DIFF
--- a/'
+++ b/'
@@ -1,0 +1,325 @@
+'use strict';
+
+var fs = require('fs');
+var test = require('tape');
+var domador = require('..');
+
+function read (file) {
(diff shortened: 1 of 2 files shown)


ananyo2012:
@jywarren Perhaps you could keep a route for the `notes#index` action. There is presently no link that redirects to the notes index page now.

jywarren:
True; I guess we should have a dedicated /notes route but I'd like to
either use the new per-node templates from the dashboard design or display
a table listing. What do you think?

ananyo2012:
Yeah that's a good way to list all activities in the community. But it also contains comments so it would be nice to go to the notes index page to see all notes. Perhaps a link to /notes can be given in the community scientists link in the dashboard page which currently links to the same page.

Issue #549, opened 2016-05-27
Initial features and tasks of Expanded Q & A system


ananyo2012:
These are the initial tasks that are needed to be completed for the initial look of the Expanded Q & A project. The Design ideas are motivated from the research note posted on https://publiclab.org/notes/ananyo2012/05-11-2016/design-ideas-for-expanded-q-a-project

- [x] Write basic CRUD actions for Question Controller and corresponding Views
- [x] Write functional tests for Question Controller
- [x] Create Answer Model and migration
- [x] Write create, update, delete action for Answer Controller
- [x] Rendering Answers using ajax requests
- [x] Write unit tests for answer model
- [x] Write functional tests for answer controller 
 
I will keep on updating the task list as I go on implementing the features and the issue can be closed on completion of the tasks.

jywarren:
Ah, excellent -- feel free to merge my additions from #554 into here if it's easier for you.

ananyo2012:
#555 closes covers first two tasks and #566 covers the remaining tasks. Closing this issue. I will open a new one for implementing commenting system for Q & A page.

PR #550, opened 2016-05-27
Question Controller and views for Expanded Q & A system

ananyo2012:
This contains the first two tasks listed in Issue #549 

* [x] Write basic CRUD actions for Question Controller and corresponding Views
* [x] Write functional tests for Question Controller

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -80,7 +80,7 @@ class ApplicationController < ActionController::Base
       unless current_user
         store_location
         flash[:notice] = "You must be logged in to access this page"
-        redirect_to login_url+'?return_to=' + URI.encode(request.env['PATH_INFO'])
+        redirect_to login_url
         return false
       end
     end
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -103,7 +103,11 @@ class NotesController < ApplicationController
         else
           flash[:notice] = "Research note published. Get the word out on <a href='/lists'>the discussion lists</a>!"
         end
-        redirect_to @node.path
+        if params[:redirect] && params[:redirect] == 'question'
+          redirect_to '/questions/'+@node.author.name+'/'+Time.at(@node.created).strftime("%m-%d-%Y")+'/'+@node.title.parameterize
+        else
+         redirect_to @node.path
+        end
       else
         render :template => "editor/post"
       end
@@ -159,7 +163,11 @@ class NotesController < ApplicationController
         end
         @node.save!
         flash[:notice] = "Edits saved."
-        redirect_to @node.path
+        if params[:redirect] && params[:redirect] == 'question'
+          redirect_to '/questions/'+@node.author.name+'/'+Time.at(@node.created).strftime("%m-%d-%Y")+'/'+@node.title.parameterize
+        else
+         redirect_to @node.path
+        end
       else
         flash[:error] = "Your edit could not be saved."
         render :action => :edit
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -1,0 +1,82 @@
+class QuestionsController < ApplicationController
+
+  def index
+    @title = "Recent Questions"
+    @notes = DrupalNode.where(status: 1, type: 'note')
+                       .joins(:drupal_tag)
+                       .where('term_data.name LIKE ?', 'question:%')
+                       .order('node.nid DESC')
+                       .paginate(:page => params[:page], :per_page => 30)
+    @wikis = DrupalNode.limit(10)
+                       .where(type: 'page', status: 1)
+                       .order("nid DESC")
+  end
+
+  def show
+    if params[:author] && params[:date]
+      @node = DrupalNode.where(path: "/notes/#{params[:author]}/#{params[:date]}/#{params[:id]}").first
+      @node = @node || DrupalNode.where(path: "/report/#{params[:id]}").first
+    else
+      @node = DrupalNode.find params[:id]
+    end
+
+    unless @node.has_power_tag('question')
+      flash[:error] = "Not a question"
+      redirect_to "/"
+    end
+
+    if @node.author.status == 0 && !(current_user && (current_user.role == "admin" || current_user.role == "moderator"))
+      flash[:error] = "The author of that note has been banned."
+      redirect_to "/"
+    elsif @node.status == 4 && (current_user && (current_user.role == "admin" || current_user.role == "moderator"))
+      flash[:warning] = "First-time poster <a href='#{@node.author.name}'>#{@node.author.name}</a> submitted this #{time_ago_in_words(@node.created_at)} ago and it has not yet been approved by a moderator. <a class='btn btn-default btn-sm' href='/moderate/publish/#{@node.id}'>Approve</a> <a class='btn btn-default btn-sm' href='/moderate/spam/#{@node.id}'>Spam</a>"
+    elsif @node.status == 4 && (current_user && current_user.id == @node.author.id) && !flash[:first_time_post]
+      flash[:warning] = "Thank you for contributing open research, and thanks for your patience while your post is approved by <a href='/wiki/moderation'>community moderators</a> and we'll email you when it is published. In the meantime, if you have more to contribute, feel free to do so."
+    elsif @node.status != 1 && !(current_user && (current_user.role == "admin" || current_user.role == "moderator"))
+      # if it's spam or a draft
+      # no notification; don't let people easily fish for existing draft titles; we should try to 404 it
+      redirect_to "/"
+    end
+
+    @node.view
+    @title = @node.latest.title
+    @tags = @node.tags
+    @tagnames = @tags.collect(&:name)
+
+    set_sidebar :tags, @tagnames
+  end
+
+  def shortlink
+    @node = DrupalNode.find params[:id]
+    redirect_to '/questions/'+@node.author.name+'/'+Time.at(@node.created).strftime("%m-%d-%Y")+'/'+@node.title.parameterize
+  end
+
+  def popular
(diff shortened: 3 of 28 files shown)


jywarren:
Hi, Ananyo, thanks -- i'm guessing this is not ready for review? If it's not, can you uncheck some boxes, since I look at the # of steps complete (in list view) to tell if I should try merging? Thanks! And/or will you simply rename it and update when it becomes ready?

Thanks.

ananyo2012:
Here are the first looks of the Q & A Pages. Picture below shows the view of Questions index Page.
question_index

This is the view of the question show page
question_show

ananyo2012:
Before this could be merged i need to solve one issue which is, while the user is is not logged in Clicking the Ask a Question button redirects him to the login page but the redirection url after logging in doesn't contain the question parameters in the url.

ananyo2012:
@jywarren @liz Please have a look at the initial changes made.

ananyo2012:
Since question are actually a modified research notes so actions involving post requests are same as that of research notes just their redirection url will differ so I have modified notes controller so that is receives a `:redirect` parameter and redirects to the required url accordingly.

jywarren:
Thanks, do you think we could use the new question templates from the
dashboard redesign?

And we should have an area at the top of this page where it describes the
page and prompts you to ask a question using a button or form. Thanks!

ananyo2012:
@jywarren Yeah perhaps we could. They look good. Should I use them? I have already added a button for posting question. I think showing tags in the index page would be good too.

jywarren:
I guess I'd like to think about some different page designs, and I'll do
some sketching for Tuesday. But yes, in the meantime, do you think you
could rework your code to use the newer templates?

Thanks!

ananyo2012:
Ok Jeff I will use the templates and modify the code.

ananyo2012:
@jywarren Please review the changes made. It's almost ready for merge. Please suggest changes needed. I have used the question template used in dashboard page as suggested.

ananyo2012:
This is how the page listing all questions looks now.
questions_index_v2

jywarren:
@ebarry @steviepubliclab -- I think we can merge this as it will be visible to all, but not prominently linked-to from anywhere and we can continue to refine the design as we go. It'll be good to get some of this great work into the codebase, as we're encouraging students to merge in their new changes regularly instead of saving them all up.

jywarren:
I'm merging this now, and will post some additional sketches hopefully later tonight or tomorrow for next steps on design. Ananyo -- looks great and I think the design side can progress in parallel to some of the coding work you're continuing. Thanks!

app/controllers/application_controller.rb
```diff
@@ -80,7 +80,7 @@ def require_user
       unless current_user
         store_location
         flash[:notice] = "You must be logged in to access this page"
-        redirect_to login_url+'?return_to=' + URI.encode(request.env['PATH_INFO'])
```
jywarren: Actually, wait - why was this removed?

jywarren:
Hi, Ananyo -- I'm sorry, I just reverted this and wanted to ask for some changes. I'm very sorry to have to have entered the reversion in the history, and apologize -- I should've been a bit more careful. The code is close to ready but I want to ask for just a few more changes first. I'll leave inline questions.

app/controllers/notes_controller.rb
```diff
@@ -103,7 +103,11 @@ def create
         else
           flash[:notice] = "Research note published. Get the word out on <a href='/lists'>the discussion lists</a>!"
         end
-        redirect_to @node.path
+        if params[:redirect] && params[:redirect] == 'question'
```
jywarren: Instead of doing this here, can we do it inside the `@node.path` method, which we already use to switch between types? I think we should be able to based on whether there are tags on the note, since they'll have been added by this point.

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    @title = @node.latest.title
+    @tags = @node.tags
+    @tagnames = @tags.collect(&:name)
+
+    set_sidebar :tags, @tagnames
+  end
+
+  def shortlink
+    @node = DrupalNode.find params[:id]
+    redirect_to '/questions/'+@node.author.name+'/'+Time.at(@node.created).strftime("%m-%d-%Y")+'/'+@node.title.parameterize
```
jywarren: Is it possible to use `@node.path` here too?

jywarren: This helps centralize where paths are generated.

app/views/editor/post.html.erb
```diff
@@ -79,6 +79,10 @@
         <h3>Share your work</h3>
       <% end %>
   
+      <% if params[:redirect] == 'question' %>
+        <%= hidden_field_tag 'redirect', 'question' %>
```
jywarren: Can we instead simply rely on the provided tagging?

app/views/like/_like.html.erb
```diff
@@ -33,7 +33,7 @@
   <hr />
 
   <div class='btn-group'>
-    <% if node.type == 'note' %><a <% if current_user && current_user.uid != node.uid %>data-confirm='Please be aware that you are editing someone else\'s post, a privilege you have only because you are a moderator or admin.' <% end %>class='btn btn-default btn-sm' href='/notes/edit/<%= node.id %>?t=<%= Time.now.to_i %>'><i class='fa fa-pencil'></i><span class='hidden-xs hidden-sm'> Edit</span></a><% end %>
+    <% if node.type == 'note' %><a <% if current_user && current_user.uid != node.uid %>data-confirm='Please be aware that you are editing someone else\'s post, a privilege you have only because you are a moderator or admin.' <% end %>class='btn btn-default btn-sm' href='/notes/edit/<%= node.id %>?t=<%= Time.now.to_i %><% if params[:controller] == "questions" %>&redirect=question<% end %>'><i class='fa fa-pencil'></i><span class='hidden-xs hidden-sm'> Edit</span></a><% end %>
```
jywarren: We should do this at the controller level, as it's relatively hidden here. Can we check for tagging at the controller and redirect there?

app/views/sidebar/_post_button.html.erb
```diff
@@ -1,10 +1,22 @@
+<% if params[:controller] == 'questions' %>
```
jywarren: I'd like to keep this template file simpler. Can we put this question button at the top of the questions index page?

app/controllers/application_controller.rb
ananyo2012: Actually the `store_location` method already stores the full path of the request in the session. See reference here: https://github.com/publiclab/plots2/pull/550/files#diff-55c5b7aecfb519d0e4880eaf2788eb6eL98. The `URI.encode(request.env['PATH_INFO'])` was actually returning the parameters but it was only available to `\login` page. As soon it posted the login parameters of an user the parameters were no longer available and it didn't redirect to with no parameters in the `/post` url. So it was an issue that I solved it here using the changes done in the user_sessions_controller: https://github.com/publiclab/plots2/pull/550/files#diff-0770692f348632e109afe78e3585bfe0R35

app/controllers/notes_controller.rb
ananyo2012: Oh yes we can surely do that by changing the path for questions in the `generate_path` method. But that would mean users could no longer view the as normal research notes. Thus it would affect the present case if deployed on the main website, thus affecting the backward compatibility. So I devised a way so that question could co-exist in the research notes page as well as in the questions page.

app/controllers/questions_controller.rb
ananyo2012: Similar answer as the previous thread.

app/views/editor/post.html.erb
ananyo2012: This is added to provide redirect to the question show path only when a `:redirect` parameter is given in the url. Well I thought of redirecting the questions to the notes show page if questions are posted from post a research note link so that the user doesn't gets redirected to the questions show page which is currently under construction.

app/views/like/_like.html.erb
ananyo2012: Similar answer as the last thread.

app/controllers/application_controller.rb
jywarren: I see -- great, thanks!

jywarren: In the future, perhaps smaller, independent changes could be submitted in separate pull requests, however, where possible. Thanks, though, this is a good catch and a welcome contribution!

app/views/sidebar/_post_button.html.erb
ananyo2012: Ok Jeff I will add the **Ask a question** button at the top of the question show page and index page.

app/controllers/notes_controller.rb
jywarren: I see! Great to know this was considered. I'm interested in how this can be clearly communicated in how the code is implemented, and also keep path generation code centralized.

Do you think we could have `@node.path()` accept an optional parameter, like `def path(type = :default)` (best double check that for setting defaults as `:symbols` but I think it'd work) where using `@node.path(:question)` would generate the question version of the path? That also reduced redundancy in this controller.

app/views/editor/post.html.erb
jywarren: I see -- so this is temporary? If so, let's leave a clear comment that it should be removed once question show template is complete. Something like `<!-- Notice: Remove this ... -->`. Thanks!

app/controllers/application_controller.rb
ananyo2012: Yeah this is a issue I should have handled separately raising a github issue.Will do it as per from the next time. Thanks Jeff!

app/views/editor/post.html.erb
ananyo2012: Ok will add comments wherever necessary.

PR #558, opened 2016-06-03
foundation structure for searching feature 

ujithaperera:
Here I moved typeahead action body to separate service class. To keep search controller simple and readable, hope to move all other bulky search queries to the service class and optimise them within search_service class.

For JSON responses I selected json jbuilder. For the first step I create index.json.jbuilder to respond json requests. I found that this gem makes our REST calls simple. 

Hope to continue on this foundation with your approval. 

Thanks


ujithaperera:
sorry wrong has been selected !

PR #600, opened 2016-07-01
Add friendly_id gem and update node paths on changing title

ananyo2012:

## Important
To make use of the changes we need to populate the new `slug` column for the existing data in the database. So we have to run this query from the rails console.
```ruby
DrupalNode.find_each(&:save)
```

Thanks!


ananyo2012:
This solves the issue of updating node titles when titles are changed. Since wiki pages, notes, questions, and maps come under the same umbrella model of drupal_node with varying paths so I have continued using node paths as it is for nodes. Just adding a `update_path` method which will be executed before updating the node. This updates the node paths when title are changed.

ananyo2012:
Now coming the role of [friendly_id](https://github.com/norman/friendly_id) here. Once node paths are updated older path becomes obsolete. So people will land up with 404's if trying to visit the nodes will old urls. Here comes friendly_id's `:history` module which saves from any 404's and it is possible to redirect old url's to new ones.

app/models/drupal_node.rb
```diff
@@ -399,9 +418,9 @@ def totalcount
   # URL-related methods:
 
   # is this used anymore? deprecate?
-  def slug
-    self.path.split('/').last
-  end
+  # def slug
+  #  self.path.split('/').last
+  # end
```
ananyo2012: Need to comment this due to new `slug` column introduced in the node table.

test/functional/notes_controller_test.rb
```diff
@@ -235,45 +234,6 @@ def setup
...
-        author: rusers(:bob).username, 
-        date: Time.now.strftime("%m-%d-%Y"), 
-        id: title.parameterize
-
-    assert_equal flash[:notice], "Edits saved."
-    assert_select "h1", newtitle
-    # assert_select "span#teststring", "added content" # this test does not work!! very frustrating. 
-    # assert_select ".label", "meetup" # test for tag addition too, later
-  end
-
```
ananyo2012: Moved this part to integration tests since this test involves multiple requests. Functional tests are best for handling single requests since the request variable doesn't update with  new requests. If we wan't to test more than one request in a test we should use integration test.

ananyo2012:
@jywarren Please review these changes once you are back. I have added tests and also updated previous tests so that these new changes stand right. I have also left comments in the code that will make it easier to follow.

Thanks!

ananyo2012:
I haven't updated the sqlite and schema files for any of these changes. I will do it with a follow up PR once you merge in both my PR's

app/controllers/wiki_controller.rb
```diff
@@ -142,6 +138,7 @@ def update
           i.vid = @revision.vid 
           i.save
         end
+        @node.title = @revision.title
```
ananyo2012: This solves #7

app/models/drupal_node.rb
jywarren: How are slug and path different?

db/migrate/20160630133409_add_slug_to_nodes.rb
```diff
@@ -0,0 +1,11 @@
+class AddSlugToNodes < ActiveRecord::Migration
+  def up
+    add_column :node, :slug, :string
+    add_index :node, :slug
+  end
+
+  def down
+  	remove_index :node, :slug
```
jywarren: just two spaces, please, thanks!

jywarren:
Ah, so we'd originally decided that we should not ever change the slug once its saved, as that will break any previous links, and for wiki pages, it would move the wiki page itself. So I think we must ask around if this new behavior is OK (on the dev list, maybe, and also @ebarry)? Just to be clear, it would:

* change slugs to match title when title is updated
* keep old paths (urls) active and 301-redirect them to new path

What happens for wiki pages? What if you move a wiki page and then want to make a new wiki page where the old one was? What if you want to change it back to the original title but the redirect record exists? If you retitle it twice -- does it doubly redirect?

In general, I like the simpler `find` instead of `find_by_slug` methods, but I'm worried about the redundancy of the slug and path columns. If we do choose to go forward here (and I appreciate the effort!) I'd like to write a bit of documentation at the top of the drupal_node file to make it clear to other developers.

app/models/drupal_node.rb
ananyo2012: In the present context path is the full url of a node like `/notes/used/07-06-2016/some-test-url` while slug is only the parameterized version of title  like `some-test-url`

db/migrate/20160630133409_add_slug_to_nodes.rb
ananyo2012: @jywarren Ah overlooked it. Updated.

ananyo2012:
@jywarren I have made some changes to the code. Please review. Now the answers to your questions. Friendly_id has its own system to handle duplicate slugs. So all slugs are unique and no two slugs are the same. So if there is already a slug named some `some-test-url` then if there is a updated node having same title as Some test url the slug would be `some-test-url--2` i.e.  index no with  a separator is appended to the url.

- If we change the title of  wiki page and then create an wiki with the title similar to the old title of the previous wiki a new slug will be assigned which will be unique and different from the slug of the old url of the previous wiki according to the format stated above

- If we want to change the title of a wiki again to its original title and no similar title exists then it will just use the original url

- But if we want to change the title of  the wiki to its original title but there already exists another wiki with the old title then a new unique url will be generated as per the format stated above.

So using friendly_id ensures that no wiki pages are lost and also the slug generated is unique and older urls are redirected back to the newer urls.

ananyo2012:
@jywarren I will dd some more tests to cover the cases you mentioned.

ananyo2012:
Well we could use just friendly_id slugs for node paths as friendly_id itself handles paths  but I didn't want   to completely switch over to friendly_id as it would involve changing a lot of code and the current path column had to be discarded. So to keep the best of both worlds so that node paths update as well as old urls are not lost I used the current path column along with friendly_id. To use the friendly_id `:history` module we need to have a slug column.

jywarren:
Ah, but just to be clear, sorry -- if you have a page like `/wiki/barnraising` which you then change to `/wiki/barnraising-2015`, and then want to make a new page called simply `/wiki/barnraising` again, would the old URL be available again to make this new page? Or would the old URL persist and redirect to the new `/wiki/barnraising-2015`? 

Second, do I understand correctly then, that `node.path` is used as the primary path, and `node.slug` is used in some way for the history -- that is, preserving old URLs? 

Thanks, this is a bit subtle and we need to have a clear understanding -- and to ensure people can get the URL they need on a wiki page. Note that for research notes the URL must simply be permanent for linking to (and exactly what the URL becomes is not that important-- it can have "-2" at the end with no problem), but for wikis, having a "good" or recognizable short URL is important too, and "losing" the ability to put a page at a well-known URL is more serious.

ananyo2012:
if you have a page like /wiki/barnraising which you then change to
/wiki/barnraising-2015, and then want to make a new page called simply
/wiki/barnraising again, the old url won't be used instead a unique url
like /wiki/barnraising--2.

Yes its somewhat like that. The node.path is the primary path and the slug
is used by the history module.
ᐧ

jywarren:
Ah, hmmm. So you wouldn't be able to use /wiki/barnraising anymore? I think
we have to think of a way around that if we're going to incorporate this
code. @ebarry?

ananyo2012:
I am trying to figure out how to reuse the old urls. But do we really need
it? Since in the present situation too a new wiki page with the same url
can't be made
ᐧ

ananyo2012:
Currently friendly_id doesn't support this but I saw a PR
https://github.com/norman/friendly_id/pull/663 where this was mentioned. So
there is a possible solution. Though I am not yet convinced why to reuse
the old urls as it is not so even in the present state. But it would be
great to know your views on it.
ᐧ

jywarren:
Hi, @ebarry and @steviepubliclab - would like your input here on if and why it's important to be able to preserver old URLs in the (not common) case that:

1. a page is moved, (like for example `/wiki/barnraising` to `/wiki/barnraising-2016`) 
2. a new page is created with the title "Barnraising" with the intended (new or resuscitated) URL `/wiki/barnraising`)

Ananyo's done some great work here, but as it stands, it would make step 2 above impossible -- the `/wiki/barnraising` URL would always be locked to the renamed/moved original page, and redirected to it. 

If that's OK, then I can merge this, but it's not the current behavior, and I thought we'd like to retain more flexibility over URLs.

jywarren:
I think both Liz and Stevie are off today. I'll try to get their input on Monday; thanks for your patience! We just had the Regional Barnraising event in California this past weekend, so folks are still catching up (and catching their breath!)

ananyo2012:
No problem. Waiting for the reviews. @jywarren In the mean time I will add tests for the cases you pointed out. I am quite late in the schedule. Trying to manage it.

steviepubliclab:
Hello! Thanks for working on this and pinging me in. 

I have a couple thoughts on the URL/Redirect idea. I like the idea that the titles/urls would match if a title is changed, also like that there would be a redirect from previous URLs to the new page, but I'm worried about a situation where someone would want to make a new wiki with an old name. If there was a way to override the redirect in creating a new page, I think it's a great move.

ananyo2012:
@steviepubliclab As I have already discussed if we want to make a new wiki with a old name a url with an unique identifier will be used such `/wiki/barnraising--2` where `/wiki/barnraising` is the old title. There is currently no way in friendly_id to override a old url if it is used in a new item. But there is a method mentioned in https://github.com/norman/friendly_id/pull/663, but that will require making changes in the gem itself.

jywarren:
Is there any way to delete a redirect record? Or is that what the norman
mod does?

ananyo2012:
@jywarren Following your suggestion I found a solution for it. We can reuse the old slugs for new wiki pages if we delete the old slugs on new wiki creation. This way it works as expected. Now new wiki pages can use the old urls of another wiki. I have added some tests to justify it. Please have a look.

jywarren:
Oh i'm so excited! I've been eager to merge this. Just to confirm, the old
urls are redirected until a new wiki tries to use the same URL, right?

ananyo2012:
@jywarren Yes the old urls redirect to the new ones until a new wiki uses the old url.

app/controllers/map_controller.rb
```diff
@@ -16,7 +16,10 @@ def index
   end
 
   def show
-    @node = DrupalNode.find_map_by_slug(params[:name]+'/'+params[:date])
+    @node = DrupalNode.find params[:name]
```
jywarren: Sorry, i hadn't caught this -- I believe there are maps with the same name but different dates. This is for when people re-map the same site. So I believe we need the friendly_id of maps to include the date.

app/models/drupal_node.rb
```diff
@@ -412,17 +440,17 @@ def edit_path
     path
   end
 
-  def self.find_by_slug(title)
-    DrupalNode.where(path: ["/#{title}", "/tool/#{title}", "/wiki/#{title}", "/place/#{title}"]).first
-  end
+  # def self.find_by_slug(title)
```
jywarren: shall we simply delete these, then?

jywarren:
I left two more inline comments, and once we address those, this is ready for merge. Thanks, Ananyo!

jywarren:
Oh, and just be sure also to update the development.sqlite.example and confirm that to me. Thanks.

app/controllers/map_controller.rb
ananyo2012: @jywarren Ok for this friendly_id already has the solution. There will be no similar slugs for two nodes so a similar title will have a unique slug with a index appended to it like `NewYork-map--2` for a map that has a same name. So two maps with the same name won't conflict. If you say I can still modify the slug but then the url for maps would be like `/map/NewYork-map-07-22-2016/07-22-2016`

app/models/drupal_node.rb
ananyo2012: Yes I can delete it once you clarify the last point.

jywarren: Well, both because the map URLs are specifically designed to have dates,
and because we need to support legacy URLs for existing maps, I think we
need to match the existing map URL format exactly. Thanks, Ananyo, I
appreciate the extra effort.

app/controllers/map_controller.rb
ananyo2012: @jywarren A better option would be to use url like `/map/NewYork-map-07-22-2016`. Your views on it?

ananyo2012: @jywarren Same thing also holds for research notes. Should we use an url like `/notes/user-07-23-2016-test-note` for research notes?

jywarren: The URL formatting was carefully designed and I think we should stick to it
for API stability reasons; is it a problem to have a slash in a friendly_id
path?

ananyo2012: @jywarren Actually friendly_id parameterizes the whole string. Slashes don't get well with parameterize. They get converted to dashes. Do you have any other format in mind? Though you can always use the default behavior friendly_id uses that appends a unique index at the en of slugs with similar title.

jywarren: Gosh, I'm not sure what to do, then. But in the research notes, you were
able to not include the `/username/` in the slug, though it was in the URL.
Is nothing similar possible here, because the date comes *after* the slug?

ananyo2012: Yeah we can always use an url like `/map/NewYork-map-07-22-2016/07-22-2016` for maps (where `NewYork-map-07-22-2016` is the slug) and something similar to that in notes. Only thing is that it looks like overuse of the date in the slugs.

jywarren: Ah, you're saying that in research notes, it's a unique slug, but in maps,
it's not. I'm not sure what to do about that.

I'm a bit stumped... so just to throw out a possibility, could you (just
for maps, as a special case) string replace the "/" char with another char,
like "*" or something, so that the URL forms a unique friendly_id, or
perhaps have a custom route just for maps, with `find( params[:slug] + '/'
+ params[:date])` , just for this one case?

ananyo2012: Uh ok I think there is a bit misunderstanding in this part. The `params[:name]` basically contains the slug of the map not just the name. So the query searches for the map based on the friendly_id slug. So if there is a map with a similar name but a different date it will go as `/map/NewYork-map--2/07-22-2016`. So you see there are no two slugs which are same.

jywarren: I'm sorry, Ananyo -- i think it'll help to refer to things as they are now
vs. things as they are in this pull request. The word "slug" had been used
in our codebase already to mean a string, used within a longer URL, where
the slug does not have to be unique. So let's call that "legacy slug" vs.
"friendly_id slug". What I mean is that we have firmly established
conventions for URLs already on the site, and the cost of changing those
is, I believe, too high. So I'm trying to think of ways that we could
maintain our existing URL schema, where parameterized "map name" ("New
York" => "new-york") is followed by the date, separated by a slash, as
shown in the routes.rb file:
https://github.com/publiclab/plots2/blob/master/config/routes.rb#L175

I'm sure you see that changing URL schemas is quite a serious proposal, and
we should do all we can to avoid this, because it'd result in either a)
broken links to legacy pages, or b) different URLs for legacy vs. new pages
of a given type (map).

I know this is frustrating, and I thank you for your patience and
thoroughness. I think we have to consider all creative possibilities for
adapting friendly_id to our legacy URL schema as shown in the current
(master branch) routes.rb before we turn to the alternative -- making a
proposal to the community of changing the legacy URL schema and either
redirecting or simply allowing old map URLs to break. Make sense? Thanks
again for your careful consideration!

ananyo2012: Oh it's not a big deal Jeff! Thanks for your time and response on this. Every new feature takes a lot of discussion and thoughts before adaptation. And it is best to discuss it with the actual changes in the code. I am leaving this PR open in case you want to discuss it further.

jywarren: OK. So do you think there's a way to conform to the existing URL schema in routes.rb (which demands a non-unique "legacy slug") by somehow concatenating the slug with the date before making the friendly_id-based `find()` call? (and doing this only for maps?) I know this is difficult challenge, but I would really like to merge this new code.

jywarren: I further understood from the call (where it was see to meet you face to face!) that the history module of friendly_id will preserve (but redirect) old map URLs, which is a bit better, but still not ideal.

I also learned that you are asking for *some part* of the URL to be completely unique. Would it not be possible to do (pseudocode only, here):

```ruby
if url.match("/map/")
  @node = DrupalNode.find params[:name] + '/' + params[:date]
else
  @node = DrupalNode.find params[:slug]
end
```

ananyo2012: Yeah that's what I wanted to say that the history module will take care of old urls. The thing is the `find` method will work only for a valid friendly_id slug. Taking the first case where the date is appended with a '/' the find method will have to search something like `map-name/07-29-2016` which is not the a valid friendly_id slug in this case. The slug is just the parameterized version of the node title.

Not to contradict you Jeff, but now that it is clear to you that no old urls gets lost and get redirected to the new url is there any problem to use a map url like `/map/map-name--2/07-29-2016` for a repost of the map. I mean an identifier like `--2` beside the map name in the url won't be a big problem .

jywarren: I think @ebarry and @steviepubliclab may join me in wanting to avoid the need for `--2` to be added to URLs if possible. 

If the `find` method won't accept a '/' character (why won't it, is there a validation?) could we do:

```js
  @node = DrupalNode.find params[:name] + '#' + params[:date]
```

We'll surely then have to override the *generation* of the friendly_id similarly to substitute a '#' or another suitable character. But in any case, maps use a different route, so we can add logic of this kind before using the `find` method, or before creating the unique id for friendly_id's use. 

I know this is not ideal, but it is not also not ideal to change our map URLs (even if we do redirect). I also want to note that it's pretty typical of any web project to be extremely resistant to URL schema changes, especially if the reason is simply that a library doesn't support a schema by default. 

Do you see a problem with this (or a similar) approach, and would you be willing to give it a try? Otherwise I think you'll have to make a strong case to convince everyone that a URL change is worthwhile, for the benefits in this PR. 

Thanks.

ebarry:
Hey @Ananyo2012, thank you for working on this! i also think it's important to **not** change the URLs -- might there be a different way to arrive at the solution you're seeking?

ananyo2012:
Ok I have thought of something so that we won't have to change the format of the urls. I will making an update to this PR soon.

jywarren:
Thanks, Ananyo -- I'm happy to take a close look and offer input. I
appreciate your attention to this!

app/controllers/notes_controller.rb
```diff
@@ -90,8 +90,10 @@ def raw
 
   def show
     if params[:author] && params[:date]
-      @node = DrupalNode.where(path: "/notes/#{params[:author]}/#{params[:date]}/#{params[:id]}").first
+      finder = "#{params[:author]} #{params[:date]} #{params[:id]}".parameterize
```
ananyo2012: @jywarren This keeps the url format intact. I am just creating a finder string from the parameters that finds the appropriate node using the friendly_id slug. Thus we can use best of both worlds! Not modifying the url format as well as using the friendly_id for old url redirects. while updating the url as the title changes.

ananyo2012:
@jywarren I have solved the problem discussed earlier. Also rebased on the latest master. Please review the changes.

jywarren:
Wow, exciting; checking this out now.

app/models/drupal_node.rb
```diff
@@ -16,10 +16,28 @@ def validate(record)
...
   self.primary_key = 'nid'
 
+  extend FriendlyId
+  friendly_id :pretty_url, use: [:slugged, :history]
+
+  def should_generate_new_friendly_id?
+    slug.blank? || title_changed?
+  end
+
+  def pretty_url
```
jywarren: Perhaps this should be called "friendly_id_generator" or "friendly_id_adaptor" as the string it returns is not actually used as a URL.

app/controllers/map_controller.rb
```diff
@@ -16,7 +16,11 @@ def index
   end
 
   def show
-    @node = DrupalNode.find_map_by_slug(params[:name]+'/'+params[:date])
+    finder = "#{params[:name]} #{params[:date]}".parameterize
```
jywarren: Nice, this makes sense. Is there a way to create a new `.find` method such that the parameters may be passed as comma-delimited arguments, like `.find(params[:name], params[:date])`? This would be simpler and we could later generate the string in the new `.find` method and pass it to the old `.find` method (the `friendly_id` one which accepts a string). If that's too complicated, at least we could make a `.find_map` method or something which formats the string and submits a `.find` call. 

I ask because i see the string construction at multiple points here.

app/models/drupal_node.rb
jywarren: or `friendly_id_string`?

app/models/drupal_node.rb
jywarren: Let's just delete this method, sound good?

app/models/drupal_node.rb
jywarren: This can be deleted too now.

db/migrate/20160630133409_add_slug_to_nodes.rb
```diff
@@ -0,0 +1,11 @@
+class AddSlugToNodes < ActiveRecord::Migration
```
jywarren: Here can you update the timestamp of the migration so that it falls in the correct order? Just 201608009... would be perfect.

db/migrate/20160630134131_create_friendly_id_slugs.rb
```diff
@@ -0,0 +1,18 @@
+class CreateFriendlyIdSlugs < ActiveRecord::Migration
```
jywarren: same for timestamp here, though you can put it just *after* the previous one

db/migrate/20160715024414_rename_long_column_in_users.rb
```diff
@@ -4,5 +4,6 @@ def up
   end
 
   def down
+  	rename_column :location_tags, :lon, :long
```
jywarren: Why is this necessary? Lalith and I agreed to use the shorter `:lon` for this column? I think we can just leave this out of the PR.

jywarren:
This looks great. I just added a couple small requests and we're basically done. Thanks!

app/models/drupal_node.rb
ananyo2012: Yeah now we can remove this.

app/models/drupal_node.rb
ananyo2012: `friendly_id_string` sounds perfect

app/controllers/map_controller.rb
ananyo2012: Umm, well then we can make a `find_map` and `find_note` method for finding notes and maps. Both of them contain different format of friendly_id slug. Overriding the find method may create problems in finding nodes as the find method itself is a friendly_id finder method.

db/migrate/20160715024414_rename_long_column_in_users.rb
ananyo2012: You can see this is in the down method of the migration. Without this problems are caused while reversing the migration. An alternative is to use `change` method instead of `up` method in the migration. Will do as you recommend.

db/migrate/20160630133409_add_slug_to_nodes.rb
ananyo2012: Ok will change the timestamps.

ananyo2012:
@jywarren I have updated the PR as you requested. Please review.

db/migrate/20160715024414_rename_long_column_in_users.rb
jywarren: oh sorry, misread it! All good, thanks!

jywarren: My mistake, sorry! all good.

jywarren:
Running tests... this is a huge update, Ananyo!

jywarren:
We're going to have to watch out extra carefully for any issues that crop up, but your tests look quite thorough. 

Bundle updated, migrations run... tests running...

All good! Wow. Merging...

jywarren:
Congratulations, this was a big one. Thanks for sticking with it and for all your hard work on it!

ananyo2012:
@jywarren Thanks for patience and time Jeff. You also stuck to it till the end. Once again to remind you you need to run this command in the rails console to initialize all the slug column of the nodes.
```ruby
DrupalNode.find_each(&:save)
```
I will also look at any issues that may creep up after this and solve it asap. Thanks a lot Jeff!

jywarren:
Oh, I should've mentioned, you can add that command to the migration--
could you add a follow-up migration with that command and it'll run when I
next publish to the live server? It can contain just that line.

PR #609, opened 2016-07-12
add tagging nodes with hashtags text in body

### LINKED ISSUE #532: Tag nodes with any #hashtag used in body
## Proposal

As of #531, we've solved the core issue in #138, as any word preceded by a `#` mark like `#hashtag` will now link to https://publiclab.org/tags/hashtag (for example). 

Now I'd like to propose that on `after_save` of any node (not comments!), any hashtags also be collected into an array and each be used to try to create a new tag for the node. See syntax for `after_save` here: https://github.com/publiclab/plots2/blob/c2dda7dece061dd0a73efefc34e61dc881b2e4f2/app/models/drupal_node_revision.rb#L17

`drupalnode.add_tag()` checks for duplicates, so that should be fine: https://github.com/publiclab/plots2/blob/c2dda7dece061dd0a73efefc34e61dc881b2e4f2/app/models/drupal_node.rb#L590

We'd also want to have some tests:

* [ ] that adding a #hashtag inline generates a tag
* [ ] that adding a #hashtag twice, or one that already exists, doesn't add doubly
* [ ] that the added tags have the same author as the user making the edit or publishing

chongr:
Fixes #532

* [ x ] tests pass -- `rake test`
* [ x ] code is in uniquely-named feature branch, and has been rebased on top of latest master (especially if you've been asked to make additional changes)
* [ x ] pull request are descriptively named
* [ x ] if possible, multiple commits squashed if they're smaller changes
* [ x ] `development.sqlite.example` has been updated if any database migrations were added

Thanks!

### DIFF
--- a/app/models/drupal_node_revision.rb
+++ b/app/models/drupal_node_revision.rb
@@ -6,8 +6,8 @@ class DrupalNodeRevision < ActiveRecord::Base
   belongs_to :drupal_node, :foreign_key => 'nid', :dependent => :destroy, :counter_cache => true
   has_one :drupal_users, :foreign_key => 'uid'
 
-  validates :title, 
-    :presence => :true, 
+  validates :title,
+    :presence => :true,
     :length => { :minimum => 2, :maximum => 100 },
     :format => {:with => /[A-Z][\w\-_]*/i, :message => "can only include letters, numbers, and dashes"}
   validates :body, :presence => :true
@@ -15,6 +15,7 @@ class DrupalNodeRevision < ActiveRecord::Base
   validates :nid, :presence => :true
 
   before_save :inline_tags
+  after_save :inline_hashtags
   before_create :setup
 
   def setup
@@ -40,6 +41,13 @@ class DrupalNodeRevision < ActiveRecord::Base
   def inline_tags
     self.body.scan(/\[question(:[\w-]+)\]/).each do |match|
       self.parent.add_tag("prompt"+match.first,self.author)
+    end
+  end
+
+  # search for inline hashtags(such as #hashtag) and create a new tag
+  def inline_hashtags
+    self.body.scan(Callouts.const_get(:HASHTAG)).each do |match|
+      self.parent.add_tag(match.last, self.author)
     end
   end
 
--- a/test/fixtures/node_revisions.yml
+++ b/test/fixtures/node_revisions.yml
@@ -1,28 +1,28 @@
-one: 
+one:
   nid: 1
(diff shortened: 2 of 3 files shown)


app/models/drupal_node_revision.rb
```diff
@@ -43,6 +43,13 @@ def inline_tags
     end
   end
 
+  # search for inline hashtags(such as #hashtag) and create a new tag
+  def inline_hashtags
+    self.body.scan(/[^\n]\#([\w-]+)/).each do |match|
```
ananyo2012: Hey @chongr I think this regexp won't give correct results. It checks hashtags only at the beginning of new line. The correct hashtag would be as used in the constants file. Refer to this https://github.com/publiclab/plots2/blob/master/config/initializers/constants.rb#L3. The best way would be to use `Callouts.const_get(:HASHTAG)` for the hashtag regexp as used elsewhere in the application.

chongr: That makes sense, but I think that regex also doesn't work for things such as this: 
```
#Heading 1
Here is a #hashtag
#Heading 2
Here is another #hashtag
```
that regex will match heading 2 if I'm not mistaken. Let me know if you have an idea of how to avoid markdown headers.  Thanks!

jywarren: Actually headers in strict markdown require a space after the hash, but as
many people (and even our templates) don't follow that, we've made
single-hash headers read as hashtags. This is a compromise as there have
been very few single-hash headers on the site since they're not provided in
templates.

I believe we have a set of tests for this in comments, Ananyo - could you
link chongr to the relevant tests?

ananyo2012:
@chongr You can see all the tests related to hashtags here. It covers all the edge cases. https://github.com/publiclab/plots2/blob/master/test/unit/drupal_comment_test.rb#L43-L118

chongr:
Thanks for the feedback I've made the updates that were discussed!

jywarren:
Excellent, thanks! Just to be clear,  this happens on create or update? Yes (just checked: `after_save :inline_hashtags`). And can we rely on the `node.add_tag` method to eliminate duplicates, or do you all think we should add one more test for deduplication? (i'm happy to add one on top of this commit at merge time, if so)

chongr:
I added another test for if the body is updated, you can add another if that wasn't what you had in mind

jywarren:
Perfect! Tests passed; merged. Thanks, @chongr! Much appreciated.

PR #612, opened 2016-07-16
Q & A email notifications

ananyo2012:

I have used Active Support Concerns to maintain shared code across models to reduce duplicacy and keep the code DRY. I have used this link for reference: http://stackoverflow.com/questions/14541823/how-to-use-concerns-in-rails-4

Thanks!

### DIFF
--- a/app/mailers/answer_mailer.rb
+++ b/app/mailers/answer_mailer.rb
@@ -1,0 +1,30 @@
+class AnswerMailer < ActionMailer::Base
+  default from: "do-not-reply@publiclab.org"
+
+  def notify_question_author(user, answer)
+    subject = "[PublicLab] New answer to Question: " + answer.node.title
+    @user = user
+    @answer = answer
+    mail(:to => @user.email, :subject => subject)
+  end
+
+  def notify_answer_likers_author(user, answer)
+    subject = "[PublicLab] New answer to Question: " + answer.node.title
+    @user = user
+    @answer = answer
+    mail(:to => @user.email, :subject => subject)
+  end
+
+  def notify_answer_accept(user, answer)
+    @user = user
+    @answer = answer
+    mail(:to => @user.email, :subject => "[PublicLab] Your answer has been accepted")
+  end
+
+  def notify_answer_like(user, answer)
+    subject = "[PublicLab] #{user.username} liked your answer to: " + answer.node.title
+    @user = user
+    @answer = answer
+    mail(:to => @answer.author.email, :subject => subject)
+  end
+end
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -13,6 +13,10 @@ class Answer < ActiveRecord::Base
     finder.gsub(Callouts.const_get(:HASHTAG), Callouts.const_get(:HASHLINKMD))
   end
 
+  def body_email
+    self.body.gsub(/([\s|"|'|\[|\(])(\/\/)([\w]?\.?publiclab.org)/, '\1https://\3')
+  end
+
   def author
     DrupalUsers.find_by_uid self.uid
   end
@@ -41,5 +45,21 @@ class Answer < ActiveRecord::Base
   def comments
     self.drupal_comments
         .order('timestamp DESC')
+  end
+
+  def answer_notify(current_user)
+    # notify question author
+    if current_user.uid != self.author.uid
+      AnswerMailer.notify_question_author(self.author, self).deliver
+    end
+
+    uids =  (self.node.answers.collect(&:uid) + self.node.likers.collect(&:uid)).uniq
+
+    # notify other answer authors and users who liked the question
+    DrupalUsers.where("uid IN (?)", uids).each do |user|
+      if user.uid != (current_user.uid && self.author.uid)
+        AnswerMailer.notify_answer_likers_author(user.user, self).deliver
+      end
+    end
   end 
 end
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -1,4 +1,6 @@
 class Answer < ActiveRecord::Base
+  include NodeShared, CommentsShared
+
   attr_accessible :uid, :nid, :content, :cached_likes, :created_at, :updated_at
 
   belongs_to :drupal_node, foreign_key: 'nid', dependent: :destroy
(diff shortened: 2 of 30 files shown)


ananyo2012:
@jywarren I will add some tests for mailers before it can be merged. Apart from that the notification system for Q & A system is complete. Please review the changes.

ananyo2012:
@jywarren  @ ebarry For notifications when a new answer is posted for question notification is sent to the question author and other answer authors who posted an answer to the question(excluding the user who posted the answer). But I think you will have others views on it. So please give your suggestions on it. 

I was thinking of sending notifications to those who subscribed to the specific question tag i.e. suppose anyone who subscribed to a `question:spectrometer` tag will get notifications for answers posted for question related with that tag. What do you think about it?

ananyo2012:
@jywarren This can be merged now. I have added enough unit tests for the mailers. There is obviously more room for functional tests but this should be enough for now i guess. Please give your reviews.

jywarren:
OK, i'll check it out probably tomorrow morning. Thanks!

ananyo2012:
Rebased on the latest master.

app/mailers/answer_mailer.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def notify_answer_accept(user, answer)
+    @user = user
+    @answer = answer
+    mail(:to => @user.email, :subject => "[PublicLab] Your answer has been accepted")
+  end
+
+  def notify_answer_like(user, answer)
+    subject = "[PublicLab] #{user.username} liked your answer to " + answer.node.title
```
jywarren: I think we should either put the title in quotation marks, or at least separate it with a colon, as in the "new answer" email title. It's helpful for distinguishing the email title from the note title.

app/models/answer.rb
```diff
@@ -41,5 +45,21 @@ def liked_by(uid)
   def comments
     self.drupal_comments
         .order('timestamp DESC')
+  end
+
+  def answer_notify(current_user)
```
jywarren: This is a lot of code to duplicate. I know I've asked this in other ways, but just one more time to see what you think: is it possible for the Answer model to descend from the Comment model, or some other model, in a class-inheritance where it would inherit these methods instead of having to duplicate them?

app/models/drupal_comment.rb
```diff
@@ -103,4 +107,29 @@ def notify(current_user)
     end 
   end
 
+  def answer_comment_notify(current_user)
```
jywarren: Again, here is a lot of repeated code. I really like this PR and am excited to merge it, but I'm worried about maintainability. Think if someone tried to change this behavior in the future; would they know to change it in both places? Even if we can't get this to inherit from another model cleanly, could we create a super class that both models inherit from, or a mixin that gives both nodes and answers access to these methods?

app/views/answer_mailer/notify_answer_likers_author.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+<p><%= raw auto_link(RDiscount.new(@answer.body_email).to_html) %></p>
+
+<hr />
+
+<div style="color:#aaa;">
+
+<p>You can comment on the answer at: https://publiclab.org<%= @answer.node.path(:question) %>#answer-<%= @answer.id %>-comment</p>
+
+<p>Report abuse to: web@publiclab.org</p>
```
jywarren: This will have to change to moderators@publiclab.org throughout, sorry -- recent change!

jywarren:
Oh, wow! The tests here are really excellent. This is a very exciting PR. I've left some comments, the main one is a question about reducing repeated code (DRY) using some kind of class inheritance, to ensure that future coders will have an easier time of it. If you could take a look and see what the options are, I'd appreciate it. 

Thanks as always, Ananyo!

app/models/answer.rb
ananyo2012: @jywarren Yes you are right. There is lot of repetitions. Inheritance is the best solution. I am looking this now. Solving it asap.

ananyo2012:
@jywarren I have used Rails concerns for using mixins in the DrupalNode, DrupalComment and Answer models and also kept the code DRY as much possible. I also found one test failing in the users_controller  test when offline due to the Geocoder gem which was making calls using the API. So I stubbed it according to the [Geocoder testing documentation](https://github.com/alexreisner/geocoder#testing-apps-that-use-geocoder).

ananyo2012:
@jywarren Uh I just saw you reverted the `map` action in users controller. Was it affecting the page load? Should I then remove the last commit that stubs the Geocoder lookup?

jywarren:
It was the `DrupalUsers.all` call in that action. I'm not sure it was much
related to the Geocoder lookup, but can you explain more if you see a
connection?

I'm not familiar with Rails concerns, as they're fairly new, but that
sounds good. Can you leave a comment at the top of the Concern and in your
PR so I (and future coders) can follow?

Thanks!

jywarren:
ah this an interesting resource:
http://stackoverflow.com/questions/14541823/how-to-use-concerns-in-rails-4

ananyo2012:
@jywarren Yes I have indeed referred to that except for concerns in Rails 3 we need to load the concerns paths in application.rb. Ok I will add some comments at the beginning.

jywarren:
wow, with rails 5 coming out, we should consider porting :-/ (i just ported
to rails 3 last year, to be honest!!!!)

ananyo2012:
Yeah once we implement all the features we can port to Rails 4 or Rails 5. We can make a thorough code cleanup in the process.

ananyo2012:
@jywarren Ok then I am keeping the Geocoder test modifications. It will be helpful in Geocoder related tests.

app/models/concerns/comments_shared.rb
```diff
@@ -0,0 +1,19 @@
+# Active Support concerns are a good way to use modules that can used across different models
+# Refer to this link: http://stackoverflow.com/questions/14541823/how-to-use-concerns-in-rails-4
+module CommentsShared
```
ananyo2012: @jywarren I have include the link in the comments here.

ananyo2012:
@jywarren I have left comments in the code and in the PR as you asked. Please feel free to add a commit to modify the comments.

jywarren:
This looks great, thanks! Testing now.

jywarren:
Passed, merging!

PR #652, opened 2016-08-07
Restful API 

david-days:
Restful API for plots2 with basic functional tests.


config/routes.rb
```diff
@@ -274,4 +288,5 @@
   match ':controller(/:action(/:id))(.:format)'
```
david-days: ended up leaving the generic match rule in because over 20 test cases depended on this to pass.  As far as I can tell, it doesn't affect the functionality of the RESTful api, since the route is set up before all the others (and therefore is a priority match over the general rule)

test/functional/questions_search_controller_test.rb
```diff
@@ -8,15 +8,15 @@ def setup
...
-  test "should get questions_search and render template if question match found" do
-    get :index, id: 'How to'
-    assert_response :success
-    assert_not_nil :title
-    assert_not_nil :tagnames
-    assert_not_nil :users
-    assert_not_nil :notes
-    assert_template :index
-  end
+  #test "should get questions_search and render template if question match found" do
```
david-days: searches#index has changed significantly, and this particular test is not correct as far as I can tell.  At the moment. it's not relevant so I commented out the test.  We can go back and make more precise corrections in the cleanup round.

app/views/searches/dynamic.html.erb
```diff
@@ -0,0 +1,83 @@
+<% content_for (:head) { javascript_include_tag "dynamic_search" } %>
```
david-days: The dynamic search page is complete and functional, but I have not added a link to the page to any of the menus.  At this point, the page can be used as a beta-testing ground, or @jywarren can decide to add a link and put it into play.

config/routes.rb
jywarren: Yes, we could set an issue asking people to pitch in on those remaining 20 routes, but for the time being it is the lowest-priority route, so we should be fine. Thanks.

test/functional/questions_search_controller_test.rb
jywarren: That controller is marked as temporary until the search system is in place, so perhaps @ananyo2012 and you can coordinate on how to connect up that method? 

https://github.com/publiclab/plots2/blob/master/app/controllers/questions_search_controller.rb#L3-L27

app/views/searches/dynamic.html.erb
jywarren: Agreed, let's give it some time for beta testing, ensure we can get the whole thing running, etc. before exposing any interfaces to the public, so far as possible.

app/api/srch/api.rb
```diff
@@ -0,0 +1,35 @@
...
+        DocResult::Entity,
+        TagResult::Entity,
+        DocList::Entity,
+        TagList::Entity
+      ],
+      info: {
+        title: "RESTful API calls for Public Lab site",
+        description: "These API calls provide programmatic access to features and data resources within Public Lab",
+        contact_name: "Jeffrey Warren",
+        contact_email: "jeff@publiclab.org",
```
jywarren: Could we do "Public Lab Web Working Group" and `web@publiclab.org`?

app/assets/javascripts/advanced_search.js
```diff
@@ -0,0 +1,17 @@
...
+  This javascript contains the page-specific functionality for the advanced search page.  This script is included via the tag on the advanced.html.erb page as follows:
+
+  <% content_for (:head) { javascript_include_tag "advanced_search" } %>
+
+  To expand on the capabilities, either add more script here or add the functionality to another script and add the values to the tag above.
+
+**/
+
+jQuery(document).ready(function() {
+	//Set the search date format for the date pickers.
```
jywarren: Could I request two-space indentation here? Sorry to nitpick, just trying to keep it consistent throughout. Thanks!

app/api/srch/typeahead.rb
```diff
@@ -0,0 +1,124 @@
...
+require 'grape-entity'
+
+module Srch
+  class Typeahead < Grape::API
+
+    # Endpoint definitions
+    resource :typeahead do
+
+      # Request URL should be /api/typeahead/all?srchString=QRY&seq=KEYCOUNT
+      # Basic implementation from classic plots2 SearchController
```
jywarren: I'm not sure what the convention is here, but could we show or link to an example output, so it's a little easier for people to design their API calls against it?

app/assets/javascripts/dynamic_search.js
```diff
@@ -0,0 +1,164 @@
...
+  This script holds the functionality for the dynamic search forms, using the RESTful API and new search classes
+  under development by the Advanced Search Project.
+**/
+//URL for search queries
+var srchBase = "/api/srch/all";
+//Number of searches conducted--used to manage display flow
+var srchCount = 0;
+//array of mapping objects to coordinate displays
+var tabDataArray = [
+	{ qryType: "file", tabId: "#tabnotes", countspan: "#notecount"},
```
jywarren: double-spaced indents would be great throughout this file too, thanks.

app/assets/javascripts/dynamic_search.js
```diff
@@ -0,0 +1,164 @@
...
+	}
+}
+
+function resultContainer() {
+	return $("<div class='panel-group'></div>");
+}
+
+/**
+  Create a row result item from the document
+  TODO:  Better way to create these?
```
jywarren: AFAIK, basically not until we decide JS templating is worthwhile, which for the scope of JS HTML generation we currently do, is not quite yet.

app/assets/javascripts/dynamic_search.js
```diff
@@ -0,0 +1,164 @@
+/**
```
jywarren: We don't currently test the autocomplete search, but with this rollout it's probably a good idea. Ananyo and I put together a Jasmine-based test framework for JavaScript code you could use for a basic test of "typing in the box makes it show", following the format of existing tests here: https://github.com/publiclab/plots2/blob/master/spec/javascripts/plots2_spec.js

HTML fixtures (test markup templates) to work with are here: https://github.com/publiclab/plots2/tree/master/spec/javascripts/fixtures

app/helpers/searches_helper.rb
```diff
@@ -0,0 +1,2 @@
+module SearchesHelper
```
jywarren: I think we'd prefer not to include the stub files unless we have a specific need.

app/services/search_service.rb
```diff
@@ -0,0 +1,208 @@
+# The SearchService class is a utility class whose purpose is to provide detailed responses to queries within 
+# different categories (record types, functionality, subsystems, etc).
+# Though similar in operation to the TypeaheadService, the implementation is separate, in that the goal of the response
+# is to provide _detailed_ results at a deep level.  In effect, TypeaheadService provides pointers to 
+# better searches, while SearchService provides deep and detailed information.
+# TODO: Refactor TypeaheadService and SearchService so that common functions come from a higher level class?
```
jywarren: We recently saw Rails Concerns used to provide common methods to two related models -- this may only be viable for models: https://github.com/publiclab/plots2/tree/master/app/models/concerns but in general I like this idea.

app/services/search_service.rb
```diff
@@ -0,0 +1,208 @@
...
+        .where('type = "note" AND node.status = 1 AND title LIKE ?', '%' + input + '%')
+  end
+
+  def find_maps(input, limit=5)
+    DrupalNode.limit(limit)
+        .order('nid DESC')
+        .where('type = "map" AND node.status = 1 AND title LIKE ?', '%' + input + '%')
+  end
+  
+  # DEPRECATED
```
jywarren: Shall we have some idea of when we'd eliminate this -- say, when all referring code is eliminated, or something?

db/migrate/20160605140228_create_searches.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSearches < ActiveRecord::Migration
```
jywarren: If we could post-date the filename timestamp to fall after existing migrations, that'd be excellent; we may have to do this just before final merge in case other migrations are merged first.

app/assets/javascripts/restful_typeahead.js
```diff
@@ -0,0 +1,220 @@
...
+  Handle the return values for a typeahead call which returns a list of document data
+**/
+function typeaheadDocList(ielem, docList) {
+	var elemRef = $(ielem);
+	var elemList = $('#'+elemRef.attr('data-results'));
+	elemList.html('');
+	if (docList.items) {
+		for (var i=0;i<docList.items.length;i++) {
+			var listopt = $('<li data-text="'+docList.items[i].docTitle+'"><span class="fa fa-'+docList.items[i].docType+'"></span>&nbsp;'+docList.items[i].docTitle+'</li>');
+			listopt.mousedown(function(e) {
```
jywarren: Are we sure that `mousedown` will be triggered with a touch event on mobile? I think `cilck` may be a higher level catchall, but probably worth checking.

jywarren: And to trigger this in a Jasmine test, you'd use `$(el).trigger('event')` just FYI

app/models/search.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  attr_accessible :key_words, :title, :main_type, :note_type, :min_date, :created_by,
+                  :language, :max_date
+
+  ## current advanced search
+  def advanced_search(input, params)
+    all = !params[:notes] && !params[:wikis] && !params[:maps] && !params[:comments]
+    @nodes = []
+    unless input.blank?
+      @nodes += SearchService.new.find_notes(input, 25) if params[:notes] || all
```
jywarren: Here, would it be possible (or desirable?) to eliminate the `new` method by using a method like:

```ruby
def self.find_notes(input, num)
  ss = self.new
  # ...
end
```
or something like this? Not hugely important, but I often see calls like this without a specific `new` invocation and just curious if that's a helpful simplification?

jywarren:
Thanks, David - I left some comments, and it looks really good. Just a note -- this will need to be rebased on top of master once we're completely ready to go. 

Do I understand that this includes the Solr integration? (just to confirm) If we are including Solr-requiring features, we should alert @icarito for Solr installation roadmap. I believe @ujithaperera was working on a README addition to guide through these steps, but I don't see that in this PR yet. 

Thanks, I'm also thinking about how we might leave some helpful pointers to people for where future work would best be integrated. Let's say we add a new doc type -- Questions, for example. Or Answers. What integrations would we have to make to include those in search results?

Another thought -- though this may not be something to tackle right now -- is "related content" as is displayed around the site, generated from Ruby and not as a JS call. This sounds similar to the TypeaheadService, in that we find titles and pointers, but not full content, based on search material from a given page or node. We can think about this later, but just wanted to put in a bookmark on that idea. 

I also just want to say, I find this line really interesting, where the new `SearchService` is being called and added into the same array as a legacy search method: https://github.com/publiclab/plots2/pull/652/files#diff-6196e5bded94c3c41f52406b16c0bfa2R16 -- shows exactly the equivalency of the two methods in a succinct way! 

Thanks! This PR represents SO MUCH GREAT WORK -- I'm happy to go through any code in more detail, or answer any other followup questions.

app/models/search.rb
ujithaperera: @jywarren , 
Actually this line of code is not using in latest update. We can remove this piece of code. Now advanced search page completely rely on solr.

app/helpers/searches_helper.rb
ujithaperera: Yes, this was added by me. Let me remove this.

ujithaperera:
@jywarren initial setup changes of solr are only presenting here. Most of the solr updates are in my 'solr_search` branch. currently these changes are available in `david-days:master`. If we are planing to merge this PR to `publiclab:master`, I can send another PR with those changes to upstream, before we go for production deployment.

jywarren:
Whatever you think is best -- I guess this does not include the `solr`
changes which will make `solr` optional for new contributors, right? So
shall I wait until that code is merged before testing a merge of this into
publiclab:master?

ujithaperera:
Sure then I'll send another small PR to complete this PR. you an merge second PR after this one. Yes, separate gem group in `Gemfile` is not appear in this PR. Will send the PR including this Gemfile change.

app/api/srch/api.rb
david-days: Not a problem--I'll make that change in a few minutes.

app/api/srch/typeahead.rb
david-days: The Swagger doc endpoint (and the whole Swagger UI) is used to interpret this, though I've been having trouble with mine--the definitions section of the Swagger doc output is not describing the output classes completely, and I'm still trying to figure out the problem.

In the meantime, I was going to write up a how-to research note (actually, several notes) describing usage of these endpoints.  That would include description of the output and basic/advanced work with the included javascript code.

app/assets/javascripts/advanced_search.js
david-days: Can do.  Sorry I missed that.

app/assets/javascripts/dynamic_search.js
david-days: I'll read through and see if I can put together the testing for it.

app/assets/javascripts/dynamic_search.js
david-days: Will do.

app/assets/javascripts/restful_typeahead.js
david-days: I tried the "click" method, but it caused the display and follow-up javascript calls to fail because the elements on the page were being manipulated by several methods at once--lots of "not found" or "undefined" errors.  Once I switched to mousedown, that all worked (and was the recommended way of handling this).

I will try to perform a test with a couple of my mobile devices (Android tablet and IPad/IPhone) to verify that this works properly.

app/services/search_service.rb
david-days: Once the RESTful typeahead is up and running in production, @ujithaperera and I can go through all the code and move the calls to the RESTful API.  That can be one relatively simple PR.

I didn't want to complicate the issue by breaking other code not written by us but perfectly functioning for the moment.

app/assets/javascripts/restful_typeahead.js
jywarren: Ah, another option is to use multiple listeners, like:

```js
$(el).on("click,touch", fn());
```

I think that's the format but you may need to check.

test/functional/questions_search_controller_test.rb
ananyo2012: @david-days @ujithaperera The current question search in the questions uses this controller. You can remove this if you have implemented this part.

ananyo2012: @david-days @ujithaperera The current question search in the questions uses this controller. You can remove this if you have implemented this part.

app/assets/javascripts/dynamic_search.js
david-days: I agree--I just put in the TODO so that it was noted that this is a "manually" created div.

I didn't catch wind of any particular JS or dynamic templating (other than bootstrap layouts), but I was allowing for someone else to have a better idea.

david-days:
@jywarren Formatting changes for Javascript files completed.  I still need to test out the mobile mousedown to make sure it works.

Also, I'll need some guidance to make sure I do the rebase properly--It's been fairly smooth, but I want to make sure I do it properly from the branch.  If it's just like the merge/rebase from the publiclab/plots2 code, I can do that--just let me know.

jywarren:
Yes, I think @ujithaperera's been rebasing pretty regularly and could help out. I think you may need to just get a local branch from publiclab/plots2 master, and rebase everything on top of that. Unfortunately, we merge Tues/Fri, so that will change, so we'll try to ensure you don't have to do it too many times -- just give me a heads up if you want me to freeze merges for a day or two so we can work through this monster and get it merged before the master shifts again. Thanks!

jywarren:
And I wonder if we should ask other students on the plots2 system to attempt merging in and booting it up and trying it out -- could help us catch issues before doing the final merge.

david-days:
Rebase of the test-restful branch is completed.  I'm going to be re-running all the testing to make sure I didn't mess anything up, and verify those points that we have discussed.

jywarren:
Very excited. Ping me if you encounter issues and with your progress. I'm going to be merging a bunch of stuff tomorrow on our usual schedule, but @jiteshjha's work covers a lot of pages, since he's incorporating internationalization of the whole site. Will that cause trouble for merges/rebases/tests?

ujithaperera:
@david-days I added couple of commits to drop solr. Now this branch is solr free. I found few test errors in test/functional/typeahead_api_test.rb. All the others tests are successfully passed.

db/migrate/20160605140228_create_searches.rb
jywarren: more confidence in not changing timestamps due to this: http://stackoverflow.com/questions/36608063/merging-activerecord-migrations-out-of-order

jywarren:
also: reading through again now and will try a test merge...!

jywarren:
Also, if I'm following what the TypeaheadApiTest does, it makes sense as a controller test for individual actions, so a functional test sounds right... was there a reason to make it an integration test? I've only needed those in order to test routing and/or chained calls with multiple or cross-controller actions.

jywarren:
I'm seeing some of Jitesh's and Ananyo's commits in the commit log for this branch. I'd have thought all the changes introduced in this branch would be rebased to be *after* all existing changes on master. Given how much work has gone into this merge, I'm OK with it, but just wondered was happened there? Thanks!

jywarren:
Hmm, I'm having trouble with this -- I tried rebasing it on top of the latest master, but it seems to have changes intertwined with commits leading back some ways, and it wasn't simple. 

If the issue is that this was rebased on top of the state of `master` from early yesterday, I would've expected an additional rebase to be relatively simple, but I'm not sure of the conflicts I'm resolving. 

We don't have to be in a rush here -- it's Saturday, after all. Curious for your input, thanks!

jywarren:
My process was:

1. `git checkout master && git pull publiclab master` -- get the latest publiclab/plots2 master
2. `git checkout -b david-days-test-restful master` - check out a new named branch based on publiclab/plots2's latest master
3. `git reset --hard 0410186e4b1f8d1f7596f5c0c3d644fed83a7f8b` -- rewind the branch to before any commits were made on the `david-days-test-restful` branch
4. `git pull https://github.com/david-days/plots2.git test-restful` -- pull in all this branch's commits
5. `git rebase master` -- this begins the rebase process of attaching each commit to the end of the latest publiclab/plots2's master
6. resolve conflicts as you go
7. when complete, the entire branch has been added on top of the most recent commit of publiclab/plots2

This can be pretty painstaking, but once it's done, any additional rebasing need only be on top of the new commits, since the "origin point" has been moved forward in "time". I think this will make repeated rebasing a lot less painful, but we can still put a freeze on merging any new commits for a few days as this process happens, so that no additional rebasing is needed. How does that sound?

jywarren:
Actually, I was thinking -- if you have any time (ideally before Monday night 7pm), we could try screensharing and walking through the rebase synchronously, so we can discuss any ambiguous conflicts. I have time before 11am Monday and after 4pm, but before 7pm. Or Sunday, though I'm reluctant to offer, as I know you've done a huge amount of work and deserve a weekend!

ujithaperera:
@jywarren I fixed merge conflits in this branch. but still don't find a reason to appearing of Annanyo's and Jitesh's commit massages in the log. I followed same steps in PR #663 and I was able to rebase top of the master successfully. Sometimes @david-days has followed different steps to make this PR.

jywarren:
OK great -- i see it now:

```
git rebase master
Current branch master is up to date.
```

I just wanted to be sure, but this looks pretty good then. I'll run tests and see if I can help with the remaining failing tests.

jywarren:
And note the new CI tests ran, it's a good way to check that the same tests pass/fail on a "standard" machine: https://travis-ci.org/publiclab/plots2/builds/153802202

jywarren:
OK, it's showing a database table missing -- @icarito, not urgent, but does the Docker/Travis setup include a `rake db:migrate` step? Or maybe the timestamps on the migrations are causing them not to be run? But as per my inline comment, I believe it runs all that are not yet run, regardless of timestamp.

ujithaperera:
Yes, I was excited to try this out. But without running migrations in this branch we can't find `searches` table in database. So unfortunately all the tests got failed.

jywarren:
I'm running `rake db:migrate` locally to see how the tests run here. Pls hold.

ujithaperera:
no problem. Thanks!

jywarren:
oh you know what? we probably need a new `schema.rb.example` -- that's what generates the database in a new setup.

jywarren:
Yes, try copying it from your `db/schema.rb` and pushing once more.

jywarren:
And I do see a bunch of this error:

`Error: test_typeahead_questions_functionality(TypeaheadApiTest): Errno::ECONNREFUSED: Connection refused - connect(2) for "localhost" port 3000
test/functional/typeahead_api_test.rb:65:in `block in <class:TypeaheadApiTest>'`

I'll look into this in a few hours; I have to go out at the moment. But i'm *very glad* we're all the way here -- much better than I'd feared with a long rebase process. Thanks!!!

jywarren:
My first thought is that the tests are not running on port 3000, so this might have been passing in the past because the development server was still running. We need to point it at the service provided by the test server instead.

jywarren:
Hmm, this doesn't have a solution, but I think it's the right question: http://stackoverflow.com/questions/31881857/rails-rspec-test-api-with-restclient

ujithaperera:
Usually in these kind of libraries comes with built in test packages. If we have this kind of support from grape, we can override those tests in our end. since @david-days is using bunch of gems power the grape framework, I believe, there should be a well developed test suite also.

jywarren:
OK -- for `rspec`, [this post](http://stackoverflow.com/questions/14267373/stubbing-restclient-response-in-rspec?rq=1) recommends mocking a response, rather than letting it go through an http interface. This would separate out testing of the underlying methods from a test of the http interface itself, and if there's a way to test just the underlying methods, that might be more powerful. 

Just a possible suggestion!

jywarren:
Here are some other resources on testing a REST api: http://stackoverflow.com/questions/4812132/writing-unit-tests-in-ruby-for-a-rest-api

I'm going to freeze merges until this is figured out -- it seems possible we'd be able to resolve it pretty quickly (fingers crossed!).

ujithaperera:
yeah. I'll try to figure out a good way to fix these tests by going through this links. Thank you Jeff.

jywarren:
Oh wait, I had made a mistake in rebasing -- i was not actually rebasing onto master. When re-running that, I get:

```
(trusty)warren@localhost:~/sites/plots2$ git rebase master
First, rewinding head to replay your work on top of it...
Applying: add jbuilder gem
Using index info to reconstruct a base tree...
M       Gemfile
Falling back to patching base and 3-way merge...
Auto-merging Gemfile
CONFLICT (content): Merge conflict in Gemfile
Failed to merge in the changes.
Patch failed at 0001 add jbuilder gem
The copy of the patch that failed is found in:
   /home/warren/sites/plots2/.git/rebase-apply/patch

When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".
```

So I think we're back where I'd mentioned in my [earlier comment](https://github.com/publiclab/plots2/pull/652#issuecomment-241202434), although once we get past that, we do have more resources for fixing the API tests. Ujitha, can you confirm that the `master` you're rebasing on top of is exactly this commit? 0c73ae178d87ec010059dcc6bf93a9c5ea6d83a8

jywarren:
Honestly, though, before you do, make a copy of the current state with `git checkout -b test-restful-backup` -- before attempting a rebase. I think that it may be preferable to do an unrebased merge if the rebase process becomes too onerous.

ujithaperera:
Yes Jeff. this is the last commit

david-days:
Jeff,

I'm going to go through the rebase again to help ease the transition--the closer I can get it to the current code, the better.  I will also use your steps to run through the rebase as you would (outside the code submission), so I can see if there is anything I can do to make it better.

I'm guessing that you saw a lot of what I did, regarding the Gemfile, as well as config/routes.rb.  Those files were touched a lot in the GSoC projects, so the sooner we can get those settled down the better. 

I should have time to Monday after 12 PM Eastern to work with you on a screensharing.  Let me know what time you are thinking of, and I will make sure my schedule is clear.

I'll test out @ujithaperera 's changes and work on getting the CI build to work properly.  I think the typeahead issue might just be a choice in coding that I made--it shouldn't be too bad.

jywarren:
Great, thanks. If you complete the rebase in this way, and force push back
to this branch, I should be able to merge with a simple ff and no rebase
required. That's the ideal, so if we aren't there by Monday, we can go
through it at 4pm if that sounds good?

Thanks a lot!

david-days:
That's what I'll work towards.  If we don't meet sooner, then 4PM Monday will be our date.

david-days:
@jywarren and @ujithaperera  ,

The testing issue is that I was following guides for running tests against running, third-party instances of RESTful APIs.  Not what we have going on here.  So the tests "work" if you have the server running by `passenger start`, but I now understand that that is not a real or valid test.

I went back to the original grape documentation [here](https://github.com/ruby-grape/grape#writing-tests), and I'm reading through (and rewriting) the tests according to their recommendations.  I'm new to ruby/rails testing, so I'm still catching up, but I believe I've found the issue and I'm should have it working soon.

david-days:
also--placing the tests in 'functional' was probably a bad choice from the sound of it--my thought was that I was testing the functionality only, so it seemed appropriate.  Do you want me to move the tests somewhere else while I'm at it?

jywarren:
I guess I'm not sure of the convention for API tests. If it's kind of like
a controller, functional is fine. And it looked like you were testing the
routes too so that makes sense.

On Aug 20, 2016 7:17 PM, "David C Days" <notifications@github.com> wrote:

also--placing the tests in 'functional' was probably a bad choice from the
sound of it--my thought was that I was testing the functionality only, so
it seemed appropriate. Do you want me to move the tests somewhere else
while I'm at it?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/publiclab/plots2/pull/652#issuecomment-241229517>, or mute
the thread

david-days:
@ujithaperera and @jywarren 

I've finished the rework of the API testing.  Turns out that the best reference is the MiniTest example on the grape page, here:  https://github.com/ruby-grape/grape#minitest-1

All tests pass on my machine, though I had to truncate a couple because the data in the test database was different than the dev.  Being a "ruby newbie", I'll need to get up to speed on how to set that up--until then, I'm confident that it's working at the basic level we need to go forward (I've tested it manually against the publiclab_dev database, and all the API responses are what was to be expected.)

The CI build just failed because of the missing "searches" table.  I solved it on my end by moving over the schema.rb.example as you suggested, and it seemed to work fine.

I'll keep checking in, but unless I hear different, @jywarren and I will plan on meeting at 4PM tomorrow.

ujithaperera:
David,

Since we are introducing new RESTful API to the plots2 project, we have to overcome few first-time only problems here. If this mini-test based approach provides the solution, it wouldn’t be a bad decision to try this out. And also I think this won’t generate any conflicts with Rspec tests. If Jeff is okay with this attachment of mini-test, then we can wrap up this issue and go forward. 

Thanks,
Ujitha.

app/services/search_service.rb
```diff
@@ -0,0 +1,208 @@
...
+    return sresult
+  end  
+
+  # Search notes for matching strings
+  def textSearch_notes(srchString)
+    sresult = DocList.new
+    unless srchString.nil? || srchString == 0
+      # notes
+      find_notes(srchString,25).each do |match|
+        doc = DocResult.fromSearch(match.nid,"file",match.path,match.title,match.body.split(/#+.+\n+/,5)[1],0)
```
jywarren: Ah, this one here? Wait, remind me what it does?

david-days: the last line "match.body.split(/#+.+\n+/,5)[1]" pulls out the first
paragraph of the .md data of the research note. Or, more precisely, it
extracts the first lines without the '#' markup. I've been trying to use
this to give the Google-like "preview' of the data

jywarren:
Watching here for TravisCI passage. Thanks, fingers X!

david-days:
@jywarren and @ujithaperera 

Did a rebase, fixed a couple of errors that I (most likely) introduced, and ran the tests.  The RESTful tests are working just fine, but the SortByTagTest is failing (assert not null) and 7 of the SearchesControllerTest tests are failing with "Fixture class not found".  

Since it was working last night, I'm guessing that I introduced an issue with a bad choice on a merge conflict in regards to the above tests.  (Alternatively, something changed in the publiclab/master that affected this, but I think it's more likely me).

If one or both of you could take a look, I would appreciate it.  Also, the SortByTagTest is a new one on me, so I don't know how that fits into this.  

I don't know how this will look from @jywarren's perspective, but if we can get these issues corrected then I think our end is good to go for the first part of the final submission.

jywarren:
I see :

`ActiveRecord::StatementInvalid: Mysql2::Error: Table 'plots.searches' doesn't exist: DELETE FROM "searches"` 

So I think you need to either update the db/schema.rb.example file to include that table, or remove the tests or code that depend on that table.

jywarren:
I don't see it in schema.rb.example - so I think if you add one more commit, where you update that file by making a copy of your schema.rb, the tests may pass.

david-days:
Followed the instructions, and starting to see the same problems as on my system:

FixtureClassNotFound: No class attached to find.
test/functional/searches_controller_test.rb:7:in `setup

Failure: <nil> expected to not be nil

So at least it's not a config problem on my system (or it followed the push).  I haven't had time to catch up with fixtures, so I'm not familiar with how to correct this problem.  Since all 7 of these are the same, however, I bet it's something straightforward.

jywarren:
Could be this? http://stackoverflow.com/questions/7020494/how-do-you-solve-fixtureclassnotfound-no-class-attached-to-find#7029933

jywarren:
What about the nil not expected to be nil error here? Maybe it'll resolve if we fix the other: https://travis-ci.org/publiclab/plots2/builds/154336568#L1575

david-days:
@jywarren ,

I tried the fix, but as soon as I put in anything, it immediately fails with a bunch of other errors.  

@ananyo2012 rewrote the sort_by_tags_test.rb file 4 days ago, and the changes are beyond me on that one.  The test is failing at a session[:tags] call, so the data is an inherent part of the http session as part of rails testing, but the code on the back end is putting data into the session.  I understand the mechanics of HTTP, but not what the controller is doing or why it would fail.

As far as the searches_controller_test.rb, everything that my limited understanding of the fixtures and the setup sees _should_ work, but it obviously doesn't.  Adding the lines regarding the inherent fixtures doesn't seem to help, and the line "set_fixture_class :searches => Search just gives code failures.

Think we could have @ananyo2012 look at the sort_by_tags_test.rb issue and @ujithaperera look at the searches_controller_test.rb?  Help!

jywarren:
Yes, any help from you two appreciated. David, is the latest here the same
version you're looking at, and can we look at the errors in the test
output?

I'll also give it a try in the morning. Thanks for your hard work.

david-days:
Because my changes were introducing more errors, I just reset it back
(though I added a couple of new lines to the top of sort_by_tags_test.rb to
kick off a new Travis CI session).

So what you are seeing is the same errors we have been dealing with.
Didn't want to confuse the issue even more.

jywarren:
Hmm, this is odd -- i think it actually has to do with standard pluralization/singularization (an odd feature of Rails) where it wants to find the fixture called (we think) `searches.yml` based on the class name `Search` -- I'm going to try messing with the `test_helper.rb` entry to see what I can get working.

jywarren:
Hmm, it's really odd, but when I run:

```
git checkout -b david-days-test-restful master
git pull https://github.com/david-days/plots2.git test-restful
git rebase master
```

I get immediate merge conflicts. Is `https://github.com/david-days/plots2.git test-restful` the version you've rebased?

jywarren:
I first rewound my `david-days-test-restful` branch back way before the summer's work, just in case, to get it fully from your own history.

jywarren:
Hmm, odd. I think I'm not quite sure what my steps were leading to... when i simply run:

```
git checkout -b david-days-test-restful master
git pull https://github.com/david-days/plots2.git test-restful
```

from my latest master, all goes smoothly as a fast-forward, so I don't think there's a problem. Maybe I was trying to rebase all the changes one more time over master? I guess I assumed that running `git rebase master` would just say "up to date" if it were already rebased, but perhaps that's wrong. Sorry for the trouble! Checking the tests now.

jywarren:
Progress, I think -- I believe `Search` is somehow a reserved word, or ambiguous. I renamed that model `SearchRecord` and made the necessary adjustments (the table name remains `searches`). Got past the initial errors, solving followup errors now.

jywarren:
Working through the form helpers now; good docs here: http://stackoverflow.com/questions/341143/can-rails-routing-helpers-i-e-mymodel-pathmodel-be-used-in-models note that we can see them all in a Rails console with `Rails.application.routes.url_helpers` -- i.e. `Rails.application.routes.url_helpers.search` and then tab-complete to see them all.

jywarren:
Working through the `form_for` template helper options -- with this resource: http://api.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html -- but may just set the URL manually in `app/views/searches/_form.html.erb` for now. @ujithaperera would be glad if you revised that later.

jywarren:
Unit and functional tests passed. @ananyo2012 - this is such a big rebase, I'm going to disable the two tests in `sort_by_tags_test` for now, and ask you if you could please check that out. After the days of work from David, I'm eager to have this merged and not have to rebase again. I appreciate it!

jywarren:
Ran tests once more for good measure and merging manually; see 9467de0b047bd63d8fa717d4309139d990bc8e1e for my final fixes. 

WHEW!!!!! Congratulations, all!

jywarren:
Tests run here: https://travis-ci.org/publiclab/plots2/builds/154780731

Just as a followup, I wanted to check with you all on why others' commits are included in this one. I don't believe there should be any rebasing of others' work as part of this, but maybe there were some earlier merge commits made which were later rebased? I know this is an unusually large commit, but my understanding is that a clean rebase should take all of the new commits and play them on top of the latest master, not affecting any previous work. What do you think led to this here? Thanks again!

ananyo2012:
@jywarren No problem Jeff. I will look what is the problem. A lot of code changed with this. So I have to keep an eye on it.

jywarren:
Hmm, mostly I'm a little concerned about commits like 0898952cb1432265e891f6ecffb2e01e2200ceac -- which you can see above, which I believe was rebased/squashed to include a lot of code which Jitesh did not write. 

I think what may have been happening is that Ujitha's been regularly merging in changes, but somehow in the merges or rebases, this was incorporated. If rebasing happens correctly, there should never be rebasing of others' commits, and for example, in this PR, all the commits would be sequential and not mixed in the timeline with others' commits. In theory, the entire stack of new commits could be squashed and merged. 

I'm not exactly sure what to do here, because I don't want to mess with the big rebase, but we have some odd reassignment of authorship and timeline here.

jywarren:
Looking closer, the file changes (which has been erroneously attributed in 0898952 to Jitesh) are individually correctly attributed: https://github.com/publiclab/plots2/blame/master/app/models/search_record.rb

So perhaps this is OK -- and I tested rebasing #726 on top of this, and it was pretty easy. So maybe we're just OK here.

jywarren:
And same here -- `blame` seems to properly attribute all changes to the Gemfile: https://github.com/publiclab/plots2/blame/master/Gemfile

jywarren:
OK, thanks for bearing with me -- I think we're all good; there are some oddities and we'll try to be more careful in the future, with regard to rebasing and not doing merge commits or rebasing others' commits.

david-days:
@jywarren ,

With all else being equal, I'm going to say that any merge/rebase issues are coming from me.  I followed the instructions on github and associated sites to make sure that what I was doing was standard, but I suspect that there is a lot of leeway in what happens under the hood, especially when changes are making the round trip from publiclab:master to david-days:test-restful and back out again.  A "minor" difference in the way I do something is probably giving you headaches.

With that said, I'm glad to see that it went forward.  I'm looking forward to checking out the functionality and, if there aren't any major issues, spreading the typeahead/search capabilities around (such as the questions_search_controller and the type).  

Once solr is online and functional, I'll start working the search and RESTful capabilities into other places where it can be useful.  For example, this can add to tagging of research notes in the Rich Editor with a tiny bit of javascript making RESTful calls on the last whole word being entered (basically, split the tag input strings by space, take the last token in the resulting array, and bring back a list of matching tags.  When clicked, append that tag to the input values.)  It's a small thing, but might help people find appropriate tags, which helps the searches, which helps the community.

jywarren:
No major headaches -- I realized soon enough that the attributions were
correct in the `blame` interface that we don't have to worry -- i was just
trying to be thorough and by-the-book. Thanks again, and the Solr plan
sounds great. Did you see my note asking @icarito whether you and Ujitha
should try the solr installation as part of travis/docker configuration, to
see if it can pass tests that way?

david-days:
I saw that.  It seems like you would be able to add a line like "rake
solr:start" to the early part of the config, and that should take care of
it.  I'm not sure, but I'll dig into it more.

I read around on the Travis CI, finally--I'm really impressed with that
service.  I have a couple of Java projects and a C project that I'm working
on on my own time, and I think that Travis CI is the way to go--long term,
it'd be a big help.

jywarren:
Just a note that this code is now live on the site!

ananyo2012:
Yes I also see there are commits that have been previously merged in the
commit log. I think it's always good to make regular merges with the main
repo as this helps to maintain regular rebases and keep a cleaner commit
log. Also conflict management is easier in a small scale and it helps to
follow along. Just sharing my views :)
ᐧ

PR #659, opened 2016-08-10
Fixes #656

### LINKED ISSUE #656: Remove "moderators@publiclab.org" from mailers in dev mode
https://github.com/publiclab/plots2/blob/9cb76917efcff9c2d91ed2e403dc2417cd7c0bb1/app/mailers/admin_mailer.rb#L9

Otherwise running tests can inadvertently email the actual moderators list :-)

We could use a value set at the top of the page using `Rails.env`.

I think we should send to "test@publiclab.org" instead. 

http://api.rubyonrails.org/classes/ActionMailer/Base.html shows how to set a default, but I'm not sure if we can do this in a conditional loop, or set the default based on environment. It bears some looking into, and we will probably have to adjust the functional and unit tests, too:

https://github.com/publiclab/plots2/blob/master/test/functional/admin_controller_test.rb#L155

https://github.com/publiclab/plots2/blob/master/test/unit/admin_mailer_test.rb#L24

ananyo2012:

Thanks!

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -35,7 +35,7 @@ Plots2::Application.configure do
   # Expands the lines which load the assets
   config.assets.debug = true
 
-  config.action_mailer.delivery_method = :sendmail
+  config.action_mailer.delivery_method = :file
 
   # force SSL
   #config.force_ssl = true


lalithr95:
@ananyo2012 @jywarren 
How about using default localhost option
`  config.action_mailer.default_url_options = { host: 'localhost', port: 3000 }`

jywarren:
Ananyo, how does lalith's proposal compare to this PR's approach? Pros/cons? This PR seems pretty simple. @lalithr95 is your change in the development.rb file too, so would only affect development? I think even so, if localhost has an active mail server, it could cause issues. I think we better merge this, but will wait for a confirmation.

ananyo2012:
Ah ok I am just checking this out.

ananyo2012:
@lalithr95 Setting the default url to localhost just sets the host for the action mailer. And Jeff is right on it that if the localhost has an active mail server in some case then it will continue to send emails. So its better to just set the default delivery option to `:file`. This will ensure that all mails are saved to the tmp/mails directory and doesn't go out to the real world.

lalithr95:
I'm good with using file 👍

jywarren:
Great, we'll merge this on Tuesday with the next batch

jywarren:
Rebased, tested, and manually merged -- can you follow up to be sure all the related issues are closed? I think they are. Thanks both!

PR #664, opened 2016-08-16
Update on rich editor template

ananyo2012:

Add woofmark and banksy to bower.json

Update rich editor template and provide redirect parameter for question

Thanks!

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -7,6 +7,7 @@
     "bootstrap-lightbox": "jbutz/bootstrap-lightbox#~0.5.0",
     "bootstrap-typeahead": "bassjobsen/Bootstrap-3-Typeahead#~4.0.0",
     "bootstrap-datepicker": "~1.6.0",
+    "bootstrap-tokenfield": "~0.12.0",
     "jquery": "~1.11",
     "jquery-file-upload": "~8.1.0",
     "marked": "0.3.2",
@@ -17,7 +18,9 @@
     "moment": "~2.10.6",
     "junction": "theleagueof/junction",
     "publiclab-editor": "~0.1.0",
-    "leaflet-d3": "^0.3.8"
+    "leaflet-d3": "^0.3.8",
+    "woofmark": "~4.2.0",
+    "horsey": "3.0.0"
   },
   "resolutions": {
     "jquery": "~2.1.0",
--- a/app/controllers/editor_controller.rb
+++ b/app/controllers/editor_controller.rb
@@ -1,6 +1,6 @@
 class EditorController < ApplicationController
 
-  before_filter :require_user, :only => [:post]
+  before_filter :require_user, :only => [:post, :rich]
 
   # main image via URL passed as GET param
   def post
--- a/app/views/editor/rich.html.erb
+++ b/app/views/editor/rich.html.erb
@@ -21,7 +21,7 @@
           <div class="col-xs-10">
             <select class="form-control input-lg">
               <option selected value="">a research note</option>
-              <option value="">a question or request</option>
+              <option value="" <% if params[:redirect] == 'question' %>selected='selected'<% end %>>a question or request</option>
               <option value="">an event</option>
             </select>
           </div>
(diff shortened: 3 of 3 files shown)


ananyo2012:
@jywarren I have updated the rich editor template to keep it in sync with the **PublicLab.Editor** master. I have also added some erb code to support redirect to the question page.

bower.json
```diff
@@ -17,7 +18,9 @@
     "moment": "~2.10.6",
     "junction": "theleagueof/junction",
     "publiclab-editor": "~0.1.0",
-    "leaflet-d3": "^0.3.8"
+    "leaflet-d3": "^0.3.8",
+    "woofmark": "~4.2.0",
+    "horsey": "3.0.0"
```
jywarren: These are actually compiled into `publiclab-editor` using `browserify`, so they're not needed in this bower.json file.

ananyo2012: Yes the javascript modules are indeed complied into PublicLab.editor.js but
the css files aren't so we need these modules. The rich editor template
links the stylesheets of those packages without which the css gets broken.
ᐧ

jywarren: Haha, ok, excellent-- but what was not working? The rich editor works fine,
doesn't it?

ananyo2012: Yes the rich editor does work fine but some of the css was broken when i
was using it in my local machine. I figured out it was due to the missing
css files in the packages.
ᐧ

jywarren: great catch, thanks!

jywarren:
This looks great, thanks. What do you think about adapting the new Question form to use the rich editor itself, instead of redirecting? It need not be in this PR, but curious about your thoughts.

ananyo2012:
Well that can be obviously done. But the redirect is not for the editor
parameters but actually for the redirect that happens after the note
creation to the question page as the questions also use the same note
create action.
ᐧ

ananyo2012:
The redirect parameter introduces an extra hidden field that is passed
along with the form which is detected in the notes controller for redirect.
ᐧ

jywarren:
Thanks, perfect. Sorry, I'm merging a *lot* of code today, so haven't been able to follow everything as closely as I'd like, and I appreciate the help! 

Rebasing and testing this now.

jywarren:
Done! manually merged. Thanks!

ananyo2012:
@jywarren On close look I saw the quetions are not actually redirected to the questions page after creation since no flash message gets displayed on creation. It is first redirected to the notes page. The rich editor is not taking the redirect field data from the form. A extra redirect field needs to be added to the editor module of the rich editor. I tried adding a redirect property in the `_editor.data` seeing these lines: https://github.com/publiclab/PublicLab.Editor/blob/master/src/PublicLab.Editor.js#L65-L72. But it didn't work as i saw it takes values from `options.data` mentioned in this line: https://github.com/publiclab/PublicLab.Editor/blob/master/src/PublicLab.Editor.js#L76. Where does this `options.data come` from? Need some help in this. I am not quite familiar working with nodejs and its module based structure.

jywarren:
https://github.com/publiclab/PublicLab.Editor/blob/master/src/PublicLab.Editor.js#L80-L106 shows that each module has a `key` param which is used to add to `editor.data`, so `editor.data[key] = module.value()`. This happens when `editor.publish()` is run. I should add this to the README so people can create new modules. 

However, any options from the constructor are copied in at initialization time, so to add form data with out creating a new module, you can initialize the editor with a `data` option, as shown in the README: https://github.com/publiclab/PublicLab.Editor#setup

ananyo2012:
Oh thanks Jeff! I am trying to solve this asap

ananyo2012:
@jywarren I still can't figure this out. Can you have a look? I tried adding a `redirect` property using the `data` option. Also I added a `output.redirect` property in the `_formatter.schemas` as I found it returns the formatted value when the form submits. Initially I tried making changes in the `dist/Publiclab.Editor.js` file in the publiclab.editor directory in the `public/lib` folder. But it wasn't working so I tried adding the Publiclab.Editor.js directly in the assets directory and include it in the rich template. It did append the redirect parameter while submission but then I got a new problem. The form was getting submitted successfully but showed an error and it didn't redirect to the question page. I confirmed that the form created the question as I visited the dashboard and saw the question listed there.

jywarren:
Yes, i'll take a look after doing these merges -- probably in 3 hours or
so. Thanks!

PR #756, opened 2016-08-30
Grids fix 2

jywarren:
urgent fix


jywarren:
Tested locally, this was urgent so bypassed travis check but it'll continue running.

PR #770, opened 2016-08-31
Adds tooltips to tags

carolineh101:

Thanks!

### DIFF
--- a/app/views/tag/_tagging.html.erb
+++ b/app/views/tag/_tagging.html.erb
@@ -2,10 +2,10 @@
 
 <p>
   <span id="tags">
-    <% @node.tags.each do |tag| %>
-      <span id="tag_<%= tag.id %>" class="label label-primary">
+    <% @node.community_tags.each do |tag| %>
+      <span id="tag_<%= tag.id %>" class="label label-primary" data-toggle="tooltip" data-html="true" title="created by <strong><%= tag.author.username %></strong> <%= time_ago_in_words(Time.at(tag.date)) %> ago">
         <a href="<%= "/maps" if @node.type == "map" %>/tag/<%= tag.name %>"><%= tag.name %></a>
-        <% if current_user && (tag.is_community_tag(@node.id) && tag.belongs_to(current_user,@node.id) || current_user.role == "admin" || current_user.role == "moderator") %>
+        <% if current_user && (tag.uid == @node.uid || current_user.role == "admin" || current_user.role == "moderator") %>
           <a class="tag-delete" data-remote="true" href="/tag/delete/<%= @node.id %>/<%= tag.id %>" data-tag-id="<%= tag.id %>">x</a>
         <% end %>
       </span>


app/models/drupal_tag.rb
```diff
@@ -170,4 +170,9 @@ def self.find_research_notes(tagnames, limit = 10)
               .limit(limit)
   end
 
+  # Returns author of tag
+  def author
+    self.drupal_node_community_tag.first.author
```
jywarren: Actually, now that I think about this again -- i was wrong! This would return the author of the first person to ever have used a tag of this name. I think, unfortunately, we need to remove these four lines, and solve it another way. I'm so sorry!

app/views/tag/_tagging.html.erb
```diff
@@ -3,7 +3,7 @@
 <p>
   <span id="tags">
     <% @node.tags.each do |tag| %>
```
jywarren: So then here, we could use `@node.community_tags` instead of `@node.tags` -- that'd get us the /usage/ of the tag, which has an author. See https://github.com/publiclab/plots2/blob/master/app/models/drupal_node.rb#L414-L420 for the difference, and see how `drupal_tag` is associated via `drupal_node_community_tag`, which is a join table: https://github.com/publiclab/plots2/blob/master/app/models/drupal_node.rb#L50

app/views/tag/_tagging.html.erb
```diff
@@ -3,7 +3,7 @@
...
-      <span id="tag_<%= tag.id %>" class="label label-primary">
+      <span id="tag_<%= tag.id %>" class="label label-primary" data-toggle="tooltip" data-html="true" title="created by <strong><%= tag.author.username %></strong> <%= time_ago_in_words(Time.at(tag.drupal_node_community_tag.first.date)) %> ago">
```
jywarren: The first part of this line would stay the same. But `tag.drupal_node_community_tag.first.date` could then become `tag.date` since we'd already be inside the `community_tag`. Sorry this seems a bit convoluted... it's:

* `nodes` have many `community_tags` which have an `author` and `date`
* `community_tags` belong to `tags` which have a `name`

This way, we can easily collect unique uses of each tag name. But it made our lives more difficult here. Again, sorry for my oversight here, I had meant to help!

jywarren:
I'm doing some merges today, so remember to rebase your code on top of master when you update this. Also, I'm happy to help out if you get stuck, though there's no rush if you've been busy. 

Apologies again for the mistake in my advice!

carolineh101:
I've made the change to ```community_tags``` but am getting an exception since the ```belongs_to``` method (used in line 8) is now not one of its methods.  Is there a method that would get the same information?

jywarren:
Hmm, i guess I meant we should make these changes in the view, and to pull all our code out of the model. Can you push up a new commit with your changes so I can see? Thanks!

jywarren:
Hmm, I need to look closer at this but have a call at the moment. Give me 30 mins, thanks!

jywarren:
Ah, ok -- you can instead do `tag.uid == node.uid` maybe?

jywarren:
instead of the `belongs_to` test

carolineh101:
It seems to work now – thanks! I'll rebase the code.

jywarren:
Super!

jywarren:
The tests failed due to `phantomjs` not loading properly, which occasionally happens but isn't due to bad code. I restarted the tests and they should pass this time. If you get that (clicking on "Details" on the tests to check if it's a real failure) you can `git push -f origin branchname` to trigger them to re-run, or just close and re-open the PR to restart them. Sorry, we'll have that resolved soon and it won't happen anymore.

jywarren:
Hmm, that's odd it's not working.

jywarren:
We may have just gotten unlucky, unfortunately. I'll run it again.

jywarren:
Great, that worked! I'll merge on Tuesday, thanks!!!

jywarren:
Done, thanks!!!

jywarren:
Looks great online now;

screenshot 2016-09-07 at 4 25 04 pm

Super job, thanks again!

PR #780, opened 2016-09-02
body_preview method

nickstaggs:
* [ ] All tests pass -- `rake test:all`
* [ ] pull request are descriptively named

This is just the opening of the pull request. I will hopefully have the method done tomorrow along with a unit test.

### DIFF
--- a/app/models/drupal_node.rb
+++ b/app/models/drupal_node.rb
@@ -1,7 +1,7 @@
 class UniqueUrlValidator < ActiveModel::Validator
   def validate(record)
     if record.title == "" || record.title.nil?
-      #record.errors[:base] << "You must provide a title." 
+      #record.errors[:base] << "You must provide a title."
       # otherwise the below title uniqueness check fails, as title presence validation doesn't run until after
     elsif record.title == "new" && (record.type == "page" || record.type == "place" || record.type == "tool")
       record.errors[:base] << "You may not use the title 'new'." # otherwise the below title uniqueness check fails, as title presence validation doesn't run until after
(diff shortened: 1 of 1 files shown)


jywarren:
Great, thanks!

nickstaggs:
Hey @jywarren I have this which seems to work:

```ruby
def body_preview
	if !!(self.match(/#+.+\n+/))
		self.split(/#+.+\n+/,5)[1]
	else
		self
	end
end
```

Any thoughts and how would I go about creating the unit test?

jywarren:
Sure -- you could do a test like:

```ruby
revision = node(:one).latest
revision.body = "## Introduction\n\nThis is my post"
assert_nil revision.body_latest.match("Introduction")
```

although I'm not sure `assert_nil` works that way... or what `match` returns if there's no match, but you get the idea.

I do notice though that you're using `self.match` -- but you might want to do `self.body.match` and `self.body.split(...)` since `self` returns the whole record and not just the body text.

jywarren:
Hi, Nick - i also noticed you're working on your master branch -- it'll be easier if you move into a named feature branch -- [read more here](https://publiclab.org/wiki/contributing-to-public-lab-software). 

This will enable you to rebase changes on top of the latest master. It's not essential, but it's good practice.

nickstaggs:
@jywarren should i delete this pull request and create another with a feature branch, then?

jywarren:
If you're going to rebase it, which would be great, then yes.  But if you
get stuck, I'm happy to help you with the rebase using just this branch.
Thanks!

PR #781, opened 2016-09-02
Solr search implementation

ujithaperera:
* [ ] All tests pass -- `rake test:all`
* [x] code is in uniquely-named feature branch, and has been rebased on top of latest master (especially if you've been asked to make additional changes)
* [x] pull request are descriptively named
* [ ] if possible, multiple commits squashed if they're smaller changes
* [ ] reviewed/confirmed/tested by another contributor or maintainer
* [ ] `schema.rb.example` has been updated if any database migrations were added

PR #810, opened 2016-09-09
Solr

jywarren:
Based on #781 781

* [ ] All tests pass -- `rake test:all`


jywarren:
Oops, mistyped "icedtea-7-plugin". Trying again.

jywarren:
One more time with `RAILS_ENV=test rake sunspot:solr:start` in `.travis.yml`

jywarren:
Making:

```yml
env:
- RAILS_ENV=test
```

line in .travis.yml instead of prefixing the command with `RAILS_ENV=test `

jywarren:
It's been running 22 minutes!

jywarren:
Ok, well we probably have to turn off solr for tests at least if it's going to take that long... Resources:

https://github.com/collectiveidea/sunspot_test/blob/master/README.md

http://www.subelsky.com/2007_10_01_archive.html?m=1

https://github.com/dcrec1/acts_as_solr_reloaded/blob/master/README.markdown

jywarren:
I'm trying using `rake sunspot:solr:start RAILS_ENV=test` instead of a default var in an `env` setting in `.travis.yml`.

jywarren:
Hmm, so by seeding the database, we seem to need to start up Solr for `development` as well as for `test`. Do we really need to run `rake db:setup` and `rake db:migrate` if we're just running tests @icarito? Shouldn't the test db be able to install and run without the development db?

jywarren:
I pushed ddb14b4 which removes `rake db:setup` and `rake db:migrate` to see what happens.

jywarren:
Ok I get error `You have 47 pending migrations`

Taking a break!

icarito:
> Do we really need to run rake db:setup and rake db:migrate if we're just running tests @icarito? 

I assumed so because otherwise you won't have a fully working environment. There is no "testing" database, it is setup on the fly for testing.

jywarren:
Hmm, I keep getting `Errno::ECONNREFUSED: Connection refused - connect(2) for "localhost" port 8981` on `rake db:setup` -- but I've tried starting solr (and it says successful) in both development and test modes:

`  - docker-compose run web rake sunspot:solr:start RAILS_ENV=test`
`  - docker-compose run web rake sunspot:solr:start`

and it says `Successfully started Solr`:

https://travis-ci.org/publiclab/plots2/builds/158977363#L2297

Any ideas here?

jywarren:
Port 8981 is the test environment, as specified in `config/sunspot.yml`. @ujithaperera  -- any ideas here? If you do make a fix, please create it as a commit based off this PR, rather than your own, so that we can preserve the rebase.

jywarren:
This indicates that we might need to insert a "try" check to ensure Solr has had a chance to start up properly:

http://stackoverflow.com/questions/7325885/connection-rejected-from-solr-in-rspec#7327654

It's for rspec, but we could likely add a similar "rescue" for our own tests...

```ruby
before :all do
  `sunspot-solr start`
  begin
    Sunspot.remove_all!
  rescue Errno::ECONNREFUSED
    sleep 1 && retry
  end
end
```

jywarren:
Or perhaps we really do need to install https://github.com/collectiveidea/sunspot_test to automatically manage solr startup for tests.

icarito:
You can recreate the testing environment if you install `docker` and `docker-compose` in your environment.
 
Those are the only requirements for running the tests locally and interacting with the testing environment interactively and directly.

The commands for setting up the testing environment (which creates a Debian 8 container complete with dependencies), are detailed in the .travis.yml file. (Starting with `docker-compose build`).

You can enter the container with `docker-compose run web /bin/bash` and treat it like a server, locally.

jywarren:
Sebastian, do you happen to know if the `install` steps are run asynchronously, or if it really is returning `Successfully started Solr` before we're seeing the `Connection refused` error?

icarito:
It really is returning success. It might be that Solr is not ready to respond to clients yet. A similar issue happens with Mysql, that's why I had to introduce the `sleep 5` step in .travis.yml. Maybe try introducing a long sleep (try 20 or 30s first) between Solr startup and attempts at using it.

Another, cleaner option would be to use a "wait-for-port-to-open" script, such as this one https://github.com/vishnubob/wait-for-it

jywarren:
I tried sleep 25 to no avail. These are seconds, not ms, right?

icarito:
Yup. I suggest building the container locally and experimenting directly
with it.

ujithaperera:
Since this is a docker container based issue, Couldn't we use an image of solr or sunspot_solr from docker hub. after this PR solr is going to stay in the system for every build that we making. right ?
So just an idea, can we go for this kind of solution. @icarito what do think about this. Actually I didn't check for practical difficulties here.

jywarren:
Well, I think solr is actually running ok now. I think it's either a timing
issue or perhaps a port number issue?

jywarren:
Like, you could wget to the different ports and echo the response codes in
a new commit, but setting up local docker may really be easier.

icarito:
@ujithaperera Yes, using a prebuilt image from docker hub would be great. It would need to be defined in `docker-compose.yml`. My recommendation is, install Docker and Docker Compose locally and try it until it works, then setup `travis.yml` to replicate.

jywarren:
I'd do this myself, but I am running in Ubuntu on a Chromebook, and docker is not supported. @ujithaperera -- I'm going to try running:

`wget http://localhost:8982/solr/ && less index.html` at various points -- it took my chromebook 13 seconds to return an admin console at that address, even after it returned `successful`. So maybe we just need more waiting.

jywarren:
Still trying different variations; i'll probably squash all these commits at some point, but this preserves a record of everything I've tried. 

http://stackoverflow.com/questions/9041722/solr-reports-connection-refused-when-running-import#9041769 has some tips on unresponsive solr applications, but this one was pretty interesting -- it says it can't connect to mysql, while running the solr startup command with `RAILS_ENV=development` -- https://travis-ci.org/publiclab/plots2/builds/159440471#L2300

Does this indicate that solr is not running in development, which would be required for `rake db:setup`?

jywarren:
Hmm, actually, why is it trying to run `mysql` at all? The config at `config/database.yml.example` which is being used specifies sqlite3 for both `test` and `development` environments. @icarito or @ujithaperera -- any thoughts on that?

jywarren:
Oh, sorry, misread -- config/database.yml.example specifies mysql for both. Oops, never mind! But we only have one database set up on docker -- "plots" -- and `development` environment asks for one called `publiclaboratory`. Kind of confused how `rake db:setup` ever works without a matching database in the docker config?

icarito:
Currently travis ci tests use the rails test environment. It is defined
in docker-compose.yml

jywarren:
Ah hmm -- so `docker-compose.yml` sets `RAILS_ENV=test` after all, so the test I did which specified `RAILS_ENV=development` was the first time we'd tried to start up `solr` in development mode. So it makes sense that `rake db:setup` was failing on a `solr` related error -- since `solr` was never before running in development.

But this also means that all our `rake db:setup` and `rake db:migrate` commands are run in the test environment, not development. 

Hmm. I guess I'll try a much longer wait before running `wget` to test solr startup -- but we now know it should be running in `test` env at port 8981.

jywarren:
Still not a lot of progress here. I'm going to try installing docker on https://c9.io and testing in there.

icarito:
Sorry to hear that. You can try it on tycho server if you wish, Docker
is installed already. mapknitter_priv1 container runs there.

jywarren:
Thanks. I'll do it here so we can in theory all access the test site. But
will keep in mind as a backup.

jywarren:
Oh, haha -- of course c9.io is already running in a docker file, so you need the *outer* docker to have `--privileged` flag so it can contain another docker. Switching over to tycho, thanks!

http://stackoverflow.com/questions/28987324/could-i-run-docker-in-a-docker-container-ubuntu-14-04#28987415

jywarren:
Huh -- odd -- it gets all the way to tests in docker on the tycho server: https://gist.github.com/jywarren/cafca0753257bf8f7ea5cc9c0573ed20

I followed the current `.travis.yml` exactly. Very odd.

jywarren:
You can see that I can `wget http://localhost:8981` successfully here: https://gist.github.com/jywarren/cafca0753257bf8f7ea5cc9c0573ed20#file-solr-log-L633

jywarren:
Sebastian, would you be able to get into the container I made on Tycho and poke around? I'm not that familiar with docker, and wonder if you can see anything different from how TravisCI runs it?

icarito:
Okay I'll have a look.

jywarren:
I wonder if it's at all related to running steps sequentially in a shell, vs. how `.travis.yml` does:

`docker-compose run web rake sunspot:solr:start`

Are you in there now? I want to try shutting it down and restarting it, and running the steps in separate `docker-compose run web ...` commands.

icarito:
Not yet, no, go ahead.

jywarren:
Aha! Look at this:

```
warren@tycho:~/sites/plots2-solr-test$ docker-compose run web wget http://localhost:8981
converted 'http://localhost:8981' (ANSI_X3.4-1968) -> 'http://localhost:8981' (UTF-8)
--2016-09-14 16:03:58--  http://localhost:8981/
Resolving localhost (localhost)... ::1, 127.0.0.1
Connecting to localhost (localhost)|::1|:8981... failed: Connection refused.
Connecting to localhost (localhost)|127.0.0.1|:8981... failed: Connection refused.
```

This after the same command ran fine in shell mode. Progress!

jywarren:
gonna read up a bit about the `docker-compose run` command

icarito:
It may indicate that Solr is not listening to the loopback device
(localhost, 127.0.0.1).

Look for a way to configure the address to bind to 0.0.0.0 or 127.0.0.1.

jywarren:
but the same command run inside the shell worked fine. hmm.

icarito:
Right.

Another thing to look for:

Make sure you're not actually creating a wholly *new* container.

To run a command in an already existing container, you can also try the
`docker exec` command.

jywarren:
Yes, that may be what's happening; a separate session? 

```
warren@tycho:~/sites/plots2-solr-test$ docker-compose run web /bin/bash
root@d03b2c3016d0:/app# wget http://localhost:8981
converted 'http://localhost:8981' (ANSI_X3.4-1968) -> 'http://localhost:8981' (UTF-8)
--2016-09-14 17:01:29--  http://localhost:8981/
Resolving localhost (localhost)... ::1, 127.0.0.1
Connecting to localhost (localhost)|::1|:8981... failed: Connection refused.
Connecting to localhost (localhost)|127.0.0.1|:8981... failed: Connection refused.
```

I'll try that.

jywarren:
:-( i tried again enabling ports, and chaining the commands so they happen in the same... session?

```
warren@tycho:~/sites/plots2-solr-test$ docker-compose run -p 127.0.0.1:8981:8981 web rake sunspot:solr:start && sleep 60 && wget http://localhost:8981/ && cat index.html
Removing stale PID file at /app/solr/pids/test/sunspot-solr-test.pid
Successfully started Solr ...
--2016-09-15 22:55:56--  http://localhost:8981/
Resolving localhost (localhost)... ::1, 127.0.0.1
Connecting to localhost (localhost)|::1|:8981... failed: Connection refused.
Connecting to localhost (localhost)|127.0.0.1|:8981... failed: Connection refused.
```

i don't know why that didn't work -- those commands /do/ work when run in `docker-compose run web /bin/bash`... but at least we now have an interactive test environment where we can keep working on this. @icarito -- feel free to take over my docker setup on tycho -- i won't have time to work on this for the next 2-3 days, most likely, so if you have a chance to try, that would be super great. Thanks.

icarito:
I will have a look, sorry it has been such an ordeal.

docker-compose.yml
```diff
@@ -14,7 +14,7 @@ db:
 #
 web:
   build: .
-  command: /bin/bash -c "sleep 5 && bundle exec rails s -p 4000 -b '0.0.0.0'"
+  command: /bin/bash -c "sleep 5 && bundle exec rails s -p 4000 -b '0.0.0.0' && rake sunspot:solr:start"
```
icarito: "&&" means 'wait for process to finish and then start this...'. If you want simultaneous execution use only one amperstand. That said, this command is only run when you don't specify a command when you issue `docker run`.

icarito:
I am installing it locally to have a closer look.

icarito:
I was able to reproduce locally, while Solr appears to have started, I can't connect to the port... digging deeper.

icarito:
Hi @jywarren , @ujithaperera looks like I managed to run the tests (but one is failing). I'll pull request into this pull-request...

jywarren:
oh wow -- just one! super. I'll accept your PR and repush up a version
without all the intermediate commits.

icarito:
While this works and is fine for testing, ideally we should do as @ujithaperera proposed and actually use a third container for the Solr process (the philosophy of containers is "one process per  container")

jywarren:
OK -- I'd really like to merge this so we're not backed up on pull
requests, so we should open a new issue for that. Would you mind? Thanks
for your help on this, Sebastian, it's super.

jywarren:
Great, working on the failing tests now -- they seem pretty minor.

jywarren:
OK -- @ujithaperera -- can you take a look at the currently failing tests? They seem relatively minor, and we're almost there. Please make your changes to a new branch and pull request against this one using:

```
git checkout -b jywarren-solr master
git pull https://github.com/jywarren/plots2.git solr
```

Thanks!!

jywarren:
Actually I was able to fix the time-based errors using `Timecop` -- but there are two remaining errors (SO CLOSE!) to do with a mismatch between `search` and `search_record` -- this sounds like @david-days might be able to help?

```ruby
======================================================================================================================================================================
.....................E
======================================================================================================================================================================
Error: test_should_get_questions_search_and_render_template_if_question_match_found(QuestionsSearchControllerTest): ActionView::Template::Error: undefined method `len
gth' for nil:NilClass
app/views/searches/normal_search.html.erb:16:in `_app_views_searches_normal_search_html_erb__887431197_66770904'
app/controllers/questions_search_controller.rb:25:in `index'
test/functional/questions_search_controller_test.rb:12:in `block in <class:QuestionsSearchControllerTest>'
======================================================================================================================================================================
........E
======================================================================================================================================================================
Error: test_should_get_show(SearchesControllerTest): ArgumentError: wrong number of arguments (1 for 0)
app/models/search_record.rb:31:in `notes'
app/controllers/searches_controller.rb:44:in `show'
test/functional/searches_controller_test.rb:25:in `block in <class:SearchesControllerTest>'
======================================================================================================================================================================
...........................................................................................
```

We'd /really/ like to get this merged and running, if possible. David, are you at all able to take a brief look at this? I think it's probably very simple for you as you wrote this part of the API, I believe?

david-days:
I'll take a look right now.

david-days:
Finally able to get the tests running--for some reason, I had to change the port for config/sunspot.yml to 8982 for all instances.  I followed the instructions on your branch, but the port assignment was off.

Getting into the real tests, now, so we can figure out that issue (or not--just not include it in the update) when/if I get a fix put together.

jywarren:
Ah, hmm, for running tests I've been booting solr with "rake
sunspot:solr:start RAILS_ENV=test" once that's running, do the default
sunspot.yml settings work? Sorry, I should've made that more clear!!!

david-days:
Ok--think I've figured out the problem for the question_search_controller_test.rb:

1.  Call to questions_search_controller pulls up the information, then sends the call to searches/normal_search.html.erb as a template (":index" call in the test, line 12).
2.  However, a look at questions_search_controller index() shows that the value for @questions is not being filled--it is nil.  questions_search_controller is actually just setting @notes.
3.  Template gets to line 16, and tries to find the length of a nil, and craps out.
4.  Short-term fix is changing line 16 from "<% if @questions.length == 0 %>" to "<% if @questions.nil? || @questions.length == 0 %>"
5.  Long term fix is reworking questions_search_controller to fill @questions instead of (or in addition to) @notes.

david-days:
I'm testing a small fix now, but it looks like the problem is that the searches_controller.rb "def show" uses SearchRecord (app/models/search_record.rb), when you actually want your new class Search (app/models/search.rb).  

For a quick fix, I'm just adding "@search = Search.new" just under "def show", line 42.

Testing that now.

jywarren:
Odd, I get one more error, 

```
=======

Error: test_should_get_show(SearchesControllerTest): ActionView::Template::Error: undefined method `body' for nil:NilClass

app/views/searches/_node_result.html.erb:20:in `block in _app_views_searches__node_result_html_erb___252723600144578527_46994973454380'

app/views/searches/_node_result.html.erb:1:in `each'

app/views/searches/_node_result.html.erb:1:in `_app_views_searches__node_result_html_erb___252723600144578527_46994973454380'

app/views/searches/show.html.erb:21:in `_app_views_searches_show_html_erb__327083743399389248_46994983933580'

test/functional/searches_controller_test.rb:25:in `block in <class:SearchesControllerTest>'

====================================
```

jywarren:
OK - -changed line 20 of `_node_result.html.erb` to:

```ruby
<%= node.drupal_node_revision.try(:first).body.first(350) %>...
```

The `.try(:first)` should catch if there are no revisions.

jywarren:
TESTS PASSED!!!!!! WOOOOHOOOOOOOO!!!

jywarren:
OK, this is great, but when in development, I go to "advanced search" and type the name of a page and hit enter, i get a blank page showing "Required parameter missing: search".

That said, the Advanced Search page already doesn't return anything on the production server, so perhaps we should move forward with this. 

@icarito, i'm a little worried that if the Solr server goes down, we'll see the "Connection refused" error, as apparently that'll happen any time we even load the `/dashboard` page. I guess for every page on the site. Could/should we implement some kind of try/catch so that if the Solr service goes down, the site still mostly works?

Otherwise, all tests are passing and I'm ready to pull this in.

PR #861, opened 2016-10-02
moved solr-dependent tests into a managed rake test:solr task

jywarren:
Still need to confirm that solr/jasmine/rails tests are all run in TravisCI.

* [x] Solr dependence removed from README etc.


jywarren:
@ananyo2012, @icarito, @david-days, @ujithaperera - this looks good to me, can you try it out? It removes the Solr dependency for normal test/development use, and adds a `rake test:solr` to start up solr, run the solr tests, and shut it down. This new rake task is then run as part of `rake test:all`, which is what TravisCI uses. This passed, above.

ananyo2012:
Just trying this now Jeff!

jywarren:
Ok! I made some mistake though where I tried running rake test:solr and it
still passed all tests, which it shouldn't... not sure why but it needs
some looking into.

ananyo2012:
@jywarren Uh it did pass only solr tests for me. This is what I got
```
7 tests, 19 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
```
This is absolutely correct as there are 7 tests and 19 assertions in searches_controller_test :blush:

ananyo2012:
But don't you think we should use `rake test TEST=test/solr/*_test.rb` to run any file under solr directory. Since we may include more test files in future?

jywarren:
Yes, I agree. But try commenting out the solr startup line and rerunning
the tests. For me they ran... but shouldn't have of course. Im going to try
it again in a cleanly restarted environment so I'm sure no solr instances
are running.

ananyo2012:
@jywarren Ok double-check to be sure. And one more thing you should remove the `disabled:true` line from development in the sunspot config. Else it throws an error while running `rake sunspot:reindex`. Only including it in test runs the tests.

jywarren:
I think we /need/ `diabled:true` in development so that contributors can try their changes without needing solr, no? When does the reindex step need to run?

Second - i realized that the `rake test:solr` is passing even with solr off because it still respects the `disabled:true` setting for test mode in `sunspot.yml`. We need a way to temporarily make solr /not/ disabled, just for these tests. We could even manually rewrite the config file as a last resort, but there must be a way to set `disabled:false` temporarily without writing the config file, right?

jywarren:
I don't see a hint on how to re-enable Solr simply here: https://github.com/sunspot/sunspot/commit/b2dfd0e74d79c66915dafbdfff3c2ce1d364611b

jywarren:
Maybe we do have to manually write to the config file from the rake task... :-(

jywarren:
OK! I have it turning on solr dependence & launching solr during the `rake test:solr` rake task. I believe this will work, fingers crossed!

jywarren:
@carolineh101, @nickstaggs, @sandyvern, @ap916, @cpdavngr, @david-days, @ujithaperera, @RodCardenas, @sjones88, @ykl7, @mzishtiaq, @ryzokuken, @neliojrr, @rafael-acerqueira

Hello everyone! In the past two weeks, we merged in a set of changes which enables a powerful new search service in our codebase, based on Solr.

However, this has caused difficulties for people who are contributing code, because it requires additional install steps, among others Java. Please accept our apologies for the extra difficulty this has caused. 

Today we've just merged a solution (in this PR) -- and basic tests no longer require Solr, and so the install process is once again as easy as it was 2 weeks ago. 

To get the latest version of the master branch, so you can build your changes on this newer, easier version, follow these steps:

* `git add` and `git commit` your current work in your named feature branch (that's not master) 
* `git checkout master` - switch to your local copy of the master branch
* `git pull https://github.com/publiclab/plots2.git master` - get the new version into your local master branch
* `git checkout feature-branch` - switch back to your feature branch, here named `feature-branch` as an example
* `git rebase master` - detach your changes from the older master branch history, and replay (or 'rebase') them so they're relative to the new master branch

If you have trouble, or if you hadn't been developing in a named feature branch (always a good idea!), ping me here and I can help you out.

The above process is *always* a good idea before submitting your changes -- we really want any contributed code to be rebased onto the most recent available version of the code. 

Thanks for your hard work and patience!
Jeff

ryzokuken:
I bet this will be a lifesaver for a lot of people struggling with the dependencies. Kudos.

jywarren:
Hi, all - are folks having any trouble with this? Just a reminder to ping me if the process laid out in the my last comment gets you stuck (@carolineh101, @nickstaggs, @sandyvern, @ap916, @cpdavngr, @david-days, @ujithaperera, @RodCardenas, @sjones88, @ykl7, @mzishtiaq, @ryzokuken, @neliojrr, @rafael-acerqueira)

jywarren:
I've updated the "Contributing" page which summarizes these steps and offers other git assistance: https://publiclab.org/wiki/contributing-to-public-lab-software

sandyvern:
Can I make a suggestions as a 'newbie' to contributing? On step 4 of the "Sample Git Workflow" list...
> Susan develops the new feature in the add-map-tags feature branch.
For those of us following the directions step-by-step to make sure we have gotten it all right, it would be helpful to make mention of committing your (or in this case, Susan's) changes to her feature-branch before continuing on.

jywarren:
Oh, that's super helpful. I'll make the change now!

Done. New contributor input is extremely valuable for exactly this reason -- it's really easy to forget key steps which are not at all clear to others! Thanks.

sandyvern:
It only took me 3 or 4 times of getting an error to remember this step :-)

ykl7:
@jywarren this looks easy enough to do. Thanks

PR #864, opened 2016-10-02
Banning (spam) users drop the liked count on wiki pages; fixes #607

### LINKED ISSUE #607: Banning (spam) users doesn't drop the liked count on wiki pages
https://publiclab.org/barnraising had 14 likes with spammers, but I banned them all, so there should only be ~5 now. But maybe the cached_likes is not properly recalculated?

Like counts are listed in the upper right box -- to the right of the title, marked with a star icon.

`user.ban` is here: https://github.com/publiclab/plots2/blob/master/app/models/drupal_users.rb#L45

Perhaps this belongs in the admin controller? That would be here: https://github.com/publiclab/plots2/blob/master/app/controllers/admin_controller.rb#L87

Any time you ban a user, presumably you'd want to trigger a recalculation of likes on any node they've liked. Recalculating likes happens here: https://github.com/publiclab/plots2/blob/master/app/controllers/like_controller.rb#L35

rafael-acerqueira:
When I ban a user the like count is decrease and when I unban the user the like count is increased again

### DIFF
--- a/app/models/drupal_users.rb
+++ b/app/models/drupal_users.rb
@@ -63,6 +63,7 @@ class DrupalUsers < ActiveRecord::Base
 
   def ban
     self.status = 0
+    decrease_likes_banned
     self.save
     # user is logged out next time they access current_user in a controller; see application controller
     self
@@ -70,6 +71,7 @@ class DrupalUsers < ActiveRecord::Base
 
   def unban
     self.status = 1
+    increase_likes_unbanned
     self.save
     self
   end
@@ -198,4 +200,19 @@ class DrupalUsers < ActiveRecord::Base
     end
   end
 
+  private
+
+  def decrease_likes_banned
+    node_selections.each do |node|
+      node.drupal_node.cached_likes = node.drupal_node.cached_likes - 1
+      node.drupal_node.save!
+    end
+  end
+
+  def increase_likes_unbanned
+    node_selections.each do |node|
+      node.drupal_node.cached_likes = node.drupal_node.cached_likes + 1
+      node.drupal_node.save!
+    end
+  end
 end


jywarren:
Hello - i hope to review and merge in the next two days -- apologies, but I'm currently in a staff retreat and won't be able to today! Thank you very much for your contribution.

rafael-acerqueira:
Ok, no problem, don't worry. I'm here to contribute

jywarren:
This looks like it'd work -- and I can def. pull it in. But I wonder if we could do even better. Would you be willing to try something a little more compact and structurally well-organized as a followup? 

I'm thinking:

A. Write a test that:
  1. has a user like a node, and then check that the like count is increased
  2. ban the user, and check that the like count is decreased
B. Replace the like recalculations with more efficient code:
  1. Write a new method, perhaps under the DrupalNode model, like `node.toggle_like(user)` (and maybe directional ones like `node.like(user)` and `node.unlike(user)`), based on [this code](https://github.com/publiclab/plots2/blob/master/app/controllers/like_controller.rb#L35-L71) and use it both in the admin controller and in the likes controller.
  2. Write a method that checks if a user likes a node (this can be used in the `node.toggle_like` method above), such as `node.liked_by(user)` which returns `true` or `false`
  3. Write unit tests for `node.toggle_like`, `node.like`, `node.unlike`, and `node.liked_by(user)` 
C. Check that the test in A. still passes

This is a lot of steps, but the functionality is pretty simple and/or already written. I think that, done as outlined above, we could be sure not to break the currently-working system, and then all liking/banning code across the site would be much more effective. 

Another idea would be to write a method like `node.recalculate_likes` because can't we just write a query to count `node_selections` and save the sum? But this approach could be complimentary to the above.

jywarren:
And -- just to be clear, no obligation to take this on -- just curious if you're interested since you've just worked on some of this code. Test-driven development could make this a pretty straightforward development task. But I'm happy to repost these steps in a new `help-wanted` issue and see if anyone else wants to give it a try!

rafael-acerqueira:
I can try to do this. I'm here to learn too. How can I put "in progress" in this issue ?

jywarren:
Hi - actually, if you could copy-paste my steps into a new issue, we can organize/discuss there -- and I can mark it `in-progress` -- and merge this PR. I think that makes sense -- to do things in smaller, self-contained steps?

Thanks!

rafael-acerqueira:
Of course. I opened a issue

jywarren:
Thanks!

jywarren:
Rebased and merged, thanks!!!

PR #883, opened 2016-10-12
moderated wiki revisions dont appear wiki dashboard sidebar - closes #817

### LINKED ISSUE #817: moderated wiki revisions appear in Wiki dashboard sidebar

Wiki edits by `mariakatosvich` which I spammed still appear in the sidebar of the dashboard at https://publiclab.org/dashboard (https://publiclab.org/research if not logged in). 

These are generated by this query: https://github.com/publiclab/plots2/blob/master/app/controllers/home_controller.rb#L77

Notice on line 80, we specify `node.status = 1` -- but we don't specify `node_revisions.status = 1`, which we need to add as well. 

It would also be great to verify this with a test -- including a test fixture for a moderated spam node revision in this file: https://github.com/publiclab/plots2/blob/master/test/fixtures/node_revisions.yml#L49

which would have `status: 0` -- and a test in this file to check if it appears in the dashboard results: https://github.com/publiclab/plots2/blob/master/test/functional/home_controller_test.rb

rafael-acerqueira:

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -78,6 +78,7 @@ class HomeController < ApplicationController
                                 .order('timestamp DESC')
                                 .where('type = (?)', 'page')
                                 .where('node.status = 1')
+                                .where('node_revisions.status = 1')
                                 .where('timestamp - node.created > ?', 300) # don't report edits within 5 mins of page creation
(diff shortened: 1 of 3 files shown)


jywarren:
This looks perfect, thanks so much! Merging now.

Issue #958, opened 2016-11-06
Move comment input box to bottom of comments list on Questions pages


ryzokuken:
Hi, this is a `first-timers-only` issue. This means we've worked to make it more legible to folks who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**. 

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general [help wanted](https://github.com/publiclab/plots2/labels/help-wanted) issues. Thanks!

## Prerequisites

You don't really need to know anything in advance in order to tackle this issue. However, a little understanding about HTML and/or ERB (Embedded Ruby) would really help you not only crack this issue easily, but understand what's happening behind the scenes.

Get started with HTML:
1. http://www.w3schools.com/html/
2. https://developer.mozilla.org/en-US/docs/Learn/HTML
3. https://www.tutorialspoint.com/html/

Get started with ERB:
1. http://www.stuartellis.name/articles/erb/
2. https://www.tutorialspoint.com/ruby-on-rails/rails-introduction.htm

## The problem

Currently, the comment input box appears at the top of the comments list. (See [this](https://publiclab.org/questions/marfisistemidroni/11-03-2016/question-infragram-kit) question for reference) it needs to be moved to the bottom of the comments list.

## The Motivation

The Questions form had been designed with comments in ascending, instead of descending, time order, and we switched that back (to oldest-at-top) in a recent commit. But now the input itself is still at the top, but comments appear at the bottom when added.

Let's move the input to the bottom to match! This will also be more consistent across the site.

## Solution

https://github.com/publiclab/plots2/blob/master/app/views/questions/show.html.erb#L52-L70 is the comment form.

https://github.com/publiclab/plots2/blob/master/app/views/questions/show.html.erb#L72-74 render the comments partial.

All you need to do is to swap the positions of the two.

## Steps to Fix

- [ ] claim this issue with a comment here, below, and ask any clarifying questions you need
- [ ] set up a repository locally following the README instructions, and make sure that all tests pass
- [ ] try to fix the issue following the steps above, but even before you're done, you can:
- [ ] commit your changes and start a pull request (see [contributing to Public Lab software](https://publiclab.org/wiki/contributing-to-public-lab-software)) but mark it as "in progress" if you have questions or if you haven't finished
[...]

PR #961, opened 2016-11-07
Changed README.md to reflect optional dependencies

ashleypt:
While I was doing my first timer bug #958, I found myself installing java when i didnt need to for the bug. So I changed the documentation to accurately reflect optional dependencies.


jywarren:
I think this is great!! Merging now, and many thanks!

Issue #956, opened 2016-11-04
Deprecate Drupal legacy database structures/naming


jywarren:
This used to be a Drupal site (back in https://github.com/jywarren/plots days!)

This is a complex, multi-part issue which should be broken up into smaller issues, but just to provide an overview:

The legacy Drupal stuff is almost eliminated, but is confusing for new contributors -- mostly at this point it's:

## The word Drupal on some models

We could rename `DrupalNode` to `Node`, for example, and this line in the model would then be unnecessary:

https://github.com/publiclab/plots2/blob/master/app/models/drupal_node.rb#L20-L21

But this would need to be a find/replace throughout the codebase!

The filename here: `/app/models/drupal_node.rb` could then become `/app/models/node.rb`

So, in summary, we've completed:

* [x] `DrupalNode` => `Node`
* [x] `DrupalNodeRevision` => `Revision` (#1496)
* [x] `DrupalComment` => `Comment` (done by @ashleypt in #978)
* [x] `DrupalTag` => `Tag` (done by @ashleypt in #1178)
* [x] `DrupalNodeCommunityTag` => `NodeTag` (not UserTags as that'll be for tags upon users; #1495)
* [x] `DrupalNodeCounter` => `NodeCounter`
* [x] `drupal_profile_field.rb` - this and the below are used only [for storing bios](https://github.com/publiclab/plots2/blob/master/app/models/drupal_users.rb#L121-L135) which could be migrated to the `User` model as a text field. - moving to User.bio in #1497
* [x] `drupal_profile_value.rb`

## Remaining to refactor

There are a few others, but we should look into how heavily they're even used, as we could start a new project to begin simplifying/deprecating older models:

* [x] Code related to Answer model (not Drupal but we do want to deprecate it): https://github.com/publiclab/plots2/search?q=answer
* [x] `drupal_content_field_image_gallery.rb` - these could be merged into (added before) the body content field of the nodes they're attached to and this "gallery" feature finally dumped (broken out into https://github.com/publiclab/plots2/issues/4074)
* [x] `drupal_content_field_map_editor.rb`  - this and 2 below could be merged into the map node body field and not stored as separate records
    * [ ] `drupal_content_field_mapper.rb`
    * [ ] `drupal_content_type_map.rb` - https://github.com/publiclab/plots2/issues/4072
[...]

PR #978, opened 2016-11-10
[in-progress]Fix/change drupal comments to comment. Part of #956

ashleypt:
Currently, there is an issue with the "comments" method in node_shared.rb conflicting with its comments object, whereas previously it simply referred to self.drupal_comments, now it refers to self.comments and thus recurses. @aayushgupta1 and @jywarren we need to work together and figure out the best way to refactor the method name so it doesn't conflict with the comments object itself. (just removing the method causes all kinds of problems in tests with comment ordering.

Related: http://stackoverflow.com/a/5446209

Anyone curious see discussion in #956 

Thanks!

### DIFF
--- a/app/views/notes/_comments.html.erb
+++ b/app/views/notes/_comments.html.erb
@@ -2,11 +2,11 @@
 
   <div id="comments" class="col-md-10 comments">
 
-    <h3><span id="comment-count"><%= @node.drupal_comments.length %></span> <%= t('notes._comments.comments') %></h3>
+    <h3><span id="comment-count"><%= @node.comments.length %></span> <%= t('notes._comments.comments') %></h3>
  
     <div id="comments-container">
-      <% @node.drupal_comments.each do |comment| %>
-        <% if comment.cid == @node.drupal_comments.last.cid %><a id="last" name="last"></a><% end %>
+      <% @node.comments.order('timestamp ASC').each do |comment| %>
+        <% if comment.cid == @node.comments.first.cid %><a id="last" name="last"></a><% end %>
   
         <%= render :partial => "notes/comment", :locals => {:comment => comment} %>
   
--- a/test/unit/answer_test.rb
+++ b/test/unit/answer_test.rb
@@ -46,6 +46,6 @@ class AnswerTest < ActiveSupport::TestCase
 
   test "should list comments in descending order" do
     answer = answers(:one)
-    assert_equal answer.comments.first, DrupalComment.last
+    assert_equal answer.comments.last, Comment.last
   end
 end
--- a/test/unit/comment_test.rb
+++ b/test/unit/comment_test.rb
@@ -1,20 +1,20 @@
 require 'test_helper'
 
-class DrupalCommentTest < ActiveSupport::TestCase
+class CommentTest < ActiveSupport::TestCase
 
   test "should save comment" do
-    comment = DrupalComment.new
+    comment = Comment.new
     comment.comment = "My first thought is\n\nthat this is pretty good."
     assert comment.save
   end
 
   test "should not save comment without body" do
-    comment = DrupalComment.new
+    comment = Comment.new
     assert !comment.save, "Saved the comment without body text"
   end
 
   test "should scan callouts out of body" do
-    comment = DrupalComment.new({
+    comment = Comment.new({
       nid: node(:one).nid,
       uid: rusers(:bob).id
     })
@@ -25,7 +25,7 @@ class DrupalCommentTest < ActiveSupport::TestCase
   end
 
   test "should scan multiple callouts out of body" do
-    comment = DrupalComment.new({
+    comment = Comment.new({
       nid: node(:one).nid,
       uid: rusers(:bob).id
     })
@@ -36,7 +36,7 @@ class DrupalCommentTest < ActiveSupport::TestCase
   end
 
   test "should scan multiple space-separated callouts out of body" do
-    comment = DrupalComment.new({
+    comment = Comment.new({
       nid: node(:one).nid,
       uid: rusers(:bob).id
     })
@@ -47,7 +47,7 @@ class DrupalCommentTest < ActiveSupport::TestCase
   end
 
   test "should scan hashtags in comments and link them" do
-    comment = DrupalComment.new({
+    comment = Comment.new({
       nid: node(:one).nid,
       uid: rusers(:bob).id
     })
@@ -56,7 +56,7 @@ class DrupalCommentTest < ActiveSupport::TestCase
   end
 
   test "should ignore Headers as hashtags in markdown" do
-    comment = DrupalComment.new({
+    comment = Comment.new({
       nid: node(:one).nid,
       uid: rusers(:bob).id
     })
@@ -65,7 +65,7 @@ class DrupalCommentTest < ActiveSupport::TestCase
   end
 
   test "should ignore commas, exclamation, periods in hashtag" do
-    comment = DrupalComment.new({
+    comment = Comment.new({
       nid: node(:one).nid,
       uid: rusers(:bob).id
     })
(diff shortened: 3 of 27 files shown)


jywarren:
Hi, I left a longer comment with next steps at #956 -- thanks!

jywarren:
Hi, just checking if you've gotten stuck on this at all, or if I could help in any way? 

Thanks!

ashleypt:
Yes I did get a bit stuck and overwhelmed actually, thanks for asking :) I
really wasn't sure whether I was being lazy or just not familiar enough
with the codebase... then I had to go get batteries for my keyboard!

basically I changed the one instance where it called that method, after
eliminating that method, and a bunch of tests to do with the ordering of
the comments started failing - like my first issue all over again lol. I
think it's being called in more places than we realize. I'll re-open the PR
in a bit so you can see the tests that fail :)

ashleypt:
@jywarren I just pushed my work and it's testing it now. The problem seems to be with answers not showing up in the correct order. I looked in models/answers.rb and the answers_controller.rb and they have no mention of anything to do with comments. Are you sure that one instance in answers's show.html.erb is the only use of the comments method from node_shared? maybe it's being inadvertently called somewhere? :/

jywarren:
Well, this doesn't look too bad, and it makes some sense too -- since we removed:

```ruby
.order("timestamp #{direction}")
```

from [here](https://github.com/publiclab/plots2/pull/978/files#diff-a59c448107dbd2c805314c24ea66376cL12), the failing test which uses `.comments` should no longer expect that method to be sorted `ASC` by default. So we should tweak that test to match -- just add the `.order(...)` method, or potentially just change `.first` to `.last`, so it no longer expects the sorting to be done already.  

Make sense?

test/unit/answer_test.rb
```diff
@@ -46,6 +46,6 @@ class AnswerTest < ActiveSupport::TestCase
 
   test "should list comments in descending order" do
     answer = answers(:one)
-    assert_equal answer.comments.first, DrupalComment.last
+    assert_equal answer.comments.first, Comment.last
```
jywarren: Here, as I described in my longer comment, use `Comment.last` since we can no longer expect default ordering.

jywarren:
And thanks for your hard work, this is a complex one and you're doing really well on it. It's the kind of issue folks have shied away from in the past, honestly, and you're almost there!

ashleypt:
@jywarren I fixed it using the changes you suggested. I didn't catch that even though I stared at that test, thanks for pointing that out and all your support with my PR(s)!

It happens because comments.order timestamp ASC returns a collection that goes "backwards" - the first element is the most recent as the number is higher in that case. I guess the default comments method returns a descending list by default, so it will be the last comment.

I'm pretty tired so I can't think much more about it tonight, but I'll review it tomorrow. I'd also like to ask some advice about mastering development and which issue would be best to tackle next if it's not too much trouble... What would be the most convenient way for you for me to contact you?

The more time goes on the more I realize I have a personality that doesn't ask for help enough when getting stuck - I feel like I won't understand the problem if I don't solve it by myself, but while I may not understand it 100% as well that way I can gain a more rapid understanding of rails and the codebase by asking for help when I get stuck, which also means making contributions. 80% * 10 issues > 100% * 1 or 2 issues. I should change that as learning by doing (as opposed to boring courses that cover the same stuff again and again and are comprehensively instead of practically focused) working with an actual team and project on a real codebase where I can be mentored and recieve good code review in exchange for helping is one of my main motivations for contributing on github, and my other one is doing meaningful work and being valued :P

jywarren:
Also -- one suggestion I thought of was that for a deeper understanding of the application architecture, you might want to take a look at breaking up a complex feature proposal into smaller parts, which is something we could definitely use help with. Check out some of the ideas for ways to get involved beyond just fixing self-contained issues:

https://publiclab.org/notes/warren/11-08-2016/help-public-lab-s-software-grow-by-joining-a-supportive-team

Specifically, we were going to propose another role called "architects" or something, who are folks who can plan out new features, and likewise break them up into clean modular parts. We have a whole collection of both bugs and feature requests that really aren't broken up yet, and so lack a clear next step: 

https://github.com/publiclab/plots2/labels/break-me-up

If that's something you're interested in, I'm happy to help you choose one and make some suggestions. Or feel free to tackle another normal-sized issue and circle back to this suggestion later -- i just thought it might be of interest.

app/views/notes/_comments.html.erb
```diff
@@ -2,11 +2,11 @@
 
   <div id="comments" class="col-md-10 comments">
 
-    <h3><span id="comment-count"><%= @node.drupal_comments.length %></span> <%= t('notes._comments.comments') %></h3>
+    <h3><span id="comment-count"><%= @node.comments.length %></span> <%= t('notes._comments.comments') %></h3>
  
     <div id="comments-container">
-      <% @node.drupal_comments.each do |comment| %>
-        <% if comment.cid == @node.drupal_comments.last.cid %><a id="last" name="last"></a><% end %>
+      <% @node.comments.each do |comment| %>
```
jywarren: Here I guess we'd want to reverse the order with a `.order('timestamp ASC')` since that's no longer automatic.

app/views/notes/_comments.html.erb
```diff
@@ -2,11 +2,11 @@
...
+        <% if comment.cid == @node.comments.last.cid %><a id="last" name="last"></a><% end %>
```
jywarren: Likewise here, `.last` won't return the right `cid` due to the new ordering.

test/unit/comment_test.rb
```diff
@@ -136,28 +136,28 @@ class DrupalCommentTest < ActiveSupport::TestCase
...
-    comment = DrupalComment.new(comment: 'Test comment')
+    comment = Comment.new(comment: 'Test comment')
     comment.drupal_users = user
     comment.answer = answer
 
     assert comment.save
-    assert_equal user.drupal_comments.last, comment
-    assert_equal answer.drupal_comments.last, comment
-    assert_not_equal answer.node.drupal_comments.last, comment
+    assert_equal user.comments.last, comment
```
jywarren: I think this isn't an issue now because there's only one comment for the user `bob`, but it will be if the fixtures are ever expanded. Best make this `.first` or add the ordering to make it extra clear to future contributors.

jywarren:
OK we're really almost there. Just found a few spots where we have to re-add the `.order(...)` method. Then I'll merge!

test/unit/comment_test.rb
jywarren: and on the subsequent two lines as well!

jywarren:
Oof, i realized my earlier comment never posted successfully! Imagine I said this like about an hour ago:

Wow, all checks passed! 🎉 🚀 🎸 

I'll read through carefully but that's a really great sign.

I'm happy to look at potential next issues, starting with this list: https://github.com/publiclab/plots2/labels/fto-candidate and for some more challenging ones, https://github.com/publiclab/plots2/labels/help-wanted

I'd also say that given this particular tough nut of an issue, you're in a good position to offer some guidance and support to others tackling similar issues from #956. As you saw, it's a complex one!

I'd say that you're already digging into deeper issues than the typical contributor, and I can tell that you like to know not just that something works, but **why** -- which is super. Rails is a bit of a magical beast, and sometimes those explanations are not always as easy to find as in other frameworks -- but it's also super powerful and fast to write if you do know the conventions. I'm happy to try to find an issue which can give you some more insight into Rails architecture. 

Please feel free to email me at jeff@publiclab.org with other questions you'd prefer to ask offline, and once again, thank you so much for your contribution!

jywarren:
Just these last three changes left, @ashleypt - just wanted to be sure you saw. Eager to merge this in, thanks a million! 🎈

ashleypt:
@jywarren ah I see you minused those lines, not modified them. is the last one the only one that should be there? the one on line 144 failed with your suggestion

test/unit/comment_test.rb
jywarren: Hmm, 144 fails when its changed to `.first`? Hmm... ah but I see -- it's not a `node.comments` call, it's a `user.comments` call -- so... i guess we didn't change ordering there and `.last` is fine. Sorry!

jywarren:
Merged, awesome! Thanks a lot! 🎈

Issue #981, opened 2016-11-10
Map page showing error


ananyo2012:
This has been marked as a good candidate for becoming a [first-timers-only issue](http://www.firsttimersonly.com) like [these](https://github.com/publiclab/plots2/labels/first-timers-only), meaning that it's simple, self-contained, and with some extra formatting, could be a great entry point for a new contributor. If you're familiar enough with this code, please consider [reformatting or reposting it as a first-timers-only issue](https://publiclab.org/notes/warren/10-31-2016/create-a-welcoming-first-timers-only-issue-to-invite-new-software-contributors), and then ping @jywarren to get it labelled.

### What happened just before the problem occurred
When I try to open the page 2 in maps index page it shows an error

### Relevant URLs
https://publiclab.org/maps?page=2

### PublicLab.org username
ananyo2012
(to help reproduce the issue)

### Browser, version, and operating system
Google Chrome on Ubuntu 16.04 LTS

For bug reports, fill out the above template; for feature requests, you can delete the template.

To learn how to write really great issues, which increases the chances they'll be resolved, see:

https://publiclab.org/wiki/developers#Contributing+for+non-coders

jywarren:
Here's the error! (mixed up with some other errors in the logs) -- should be a pretty easy first timers issue by adding a test for `if node.map` before using `node.map.authorship`

```ruby
Started GET "/maps?page=2" for 184.186.1.179 at 2016-11-14 23:37:16 +0000
Processing by MapController#index as HTML
  Parameters: {"page"=>"2"}
Started GET "/notes/eustatic/09-06-2014/reference-bay-denesse-and-plaquemines-bayou-aerial-photo-history-2" for 98.185.50.69 at 2016-11-14 23:37:16 +0000
Processing by NotesController#show as HTML
  Parameters: {"author"=>"eustatic", "date"=>"09-06-2014", "id"=>"reference-bay-denesse-and-plaquemines-bayou-aerial-photo-history-2"}
  Rendered map/_maps.html.erb (1010.9ms)
  Rendered map/index.html.erb within layouts/application (1613.9ms)
Completed 500 Internal Server Error in 1615.3ms

ActionView::Template::Error (undefined method `authorship' for nil:NilClass):
    6:     <% end %>
    7:     <h4><a href="<%= node.path %>"><%= node.title %></a></h4>
    8:     <p style="color:#888;"><small>
    9:       by <a href="/profile/<%= node.map.authorship %>"><%= node.map.authorship %></a> 
[...]

Issue #754, opened 2016-08-30
"Browse other activities" next to "replicate this" prompt


jywarren:
This has been marked as a good candidate for becoming a [first-timers-only issue](http://www.firsttimersonly.com) like [these](https://github.com/publiclab/plots2/labels/first-timers-only), meaning that it's simple, self-contained, and with some extra formatting, could be a great entry point for a new contributor. If you're familiar enough with this code, please consider [reformatting or reposting it as a first-timers-only issue](https://publiclab.org/notes/warren/10-31-2016/create-a-welcoming-first-timers-only-issue-to-invite-new-software-contributors), and then ping @jywarren to get it labelled.

Or just solve it yourself!

****
A link next to the button would let people more easily horizontally browse activities: 

![screenshot 2016-08-30 at 2 13 06 pm](https://cloud.githubusercontent.com/assets/24359/18101418/a59511fa-6ebc-11e6-9c99-433e06aef840.png)

It could go on this line of code, though you should break it up unto multiple lines so it's easier to read:

https://github.com/publiclab/plots2/blob/master/app/views/notes/_responses.html.erb#L8

To determine where the link should lead to, we look for the `tagname`, and go to the page with a tag such as `activities:<tagname>`, so:

```ruby
DrupalTag.where(name: "activities:" + @node.power_tag('activity')).first.nodes.first.path
```

However, we must account for the possibility that:

* the DrupalTag doesn't exist and we can't run `.first` on an empty returned array
* the DrupalTag does exist but has no nodes (say, they've been deleted) and we can't call `.first` on an empty `.nodes` array

So we'll have to add some `if` statements to protect against that possibility. 

But once you have the URL, you can just add it in a link, like `<a href="/...">Browser other activities</a>`

500swapnil:
I would like to help with this issue. Can u explain further on how?

jywarren:
Maybe a good place to begin here would be to look over this test, which lays out what we expect to happen in that template:

https://github.com/publiclab/plots2/blob/master/test/integration/node_insert_extras_test.rb#L40-L53

Here, we're adding the tag `activity:test` for purposes of testing. So we'd then want this new code to also display a link (an `<a>` tag) pointed at a page which bears the tag `activities:test` -- that's the idea of this feature. 

You could add that expectation to the test and create a PR, which we should expect to fail, initially. 

But -- once it does, we can then begin writing the code to get that test to pass -- and we have a clear way to tell when it's working; to run the tests. Make sense?

[...]

PR #1019, opened 2016-11-24
Map page showing error #981

500swapnil:

Thanks!

### DIFF
--- a/app/views/map/_maps.html.erb
+++ b/app/views/map/_maps.html.erb
@@ -6,7 +6,10 @@
     <% end %>
     <h4><a href="<%= node.path %>"><%= node.title %></a></h4>
     <p style="color:#888;"><small>
-      by <a href="/profile/<%= node.map.authorship %>"><%= node.map.authorship %></a> 
+      by 
+      <% if node.map %>
+        <a href="/profile/<%= node.map.authorship %>"><%= node.map.authorship %></a> 
+      <% end %>
       <% unless !node.map || node.map.captured_on == "" %> <%= time_ago_in_words(node.map.captured_on) %> ago<% end %>
       | <a href="<%= node.path %>#comments"><i style="color:#888;" class="fa fa-comment-o"></i> <%= node.comment_count %></a>
       | <i style="color:#888;" class="fa fa-star-o"></i> <%= node.likes %>
--- a/app/views/map/_maps.html.erb
+++ b/app/views/map/_maps.html.erb
@@ -6,9 +6,8 @@
     <% end %>
     <h4><a href="<%= node.path %>"><%= node.title %></a></h4>
     <p style="color:#888;"><small>
-      by 
       <% if node.map %>
-        <a href="/profile/<%= node.map.authorship %>"><%= node.map.authorship %></a> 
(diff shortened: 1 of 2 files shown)


500swapnil:
How do I test locally, the changes I have made?

app/views/map/_maps.html.erb
```diff
@@ -6,7 +6,10 @@
     <% end %>
     <h4><a href="<%= node.path %>"><%= node.title %></a></h4>
     <p style="color:#888;"><small>
-      by <a href="/profile/<%= node.map.authorship %>"><%= node.map.authorship %></a> 
+      by
```
jywarren: Hi! It looks like this "by" should go inside the conditional `if` block, or if there's no author, it'll just say "by" and nothing after it, right? Would you mind moving "by" to the same line as the `<a>` tag 2 lines down?

500swapnil:
Is it ok now?

jywarren:
You could test your changes by making a fixture for a node of type "map" that has no `node.map` association -- that's actually defined on this line:

https://github.com/publiclab/plots2/blob/master/app/models/drupal_node.rb#L502

as: 

```ruby
def map
  self.drupal_content_type_map.first
end
```

I think we actually have only one fixture for maps, though we do test the show action here:

https://github.com/publiclab/plots2/blob/master/test/functional/map_controller_test.rb#L22-L28

The fixture is here: https://github.com/publiclab/plots2/blob/master/test/fixtures/node.yml#L135

So I believe your test could be a copy of the one for the map show action, but with a second map fixture that has no associated `node.drupal_content_type_map`

https://github.com/publiclab/plots2/blob/master/test/fixtures/content_type_map.yml

Or, since this is a pretty minor error, we could just leave it! I mean, really, why does this node not have a map?

jywarren:
OK, i see your changes, and they look great. Thanks! Did you want to try writing the extra test or just calling this one done? I think writing a test is good practice, but this seems like an obscure case that we don't really need a test for.

500swapnil:
I didnt quite understand how to add tests. And I meant that can I view the site in my own computer after making changes.

jywarren:
So, you can add tests by adding a new entry or method to one of the test files, like the one here: https://github.com/publiclab/plots2/blob/master/test/functional/map_controller_test.rb#L22-L28

There's a good overview here: http://guides.rubyonrails.org/testing.html

This lets you not only test on your own computer (see the Testing section of the README of this project) but also runs the tests automatically so we can see them in the pull request here -- that way we can all see the same output -- and also, all the tests are run each time someone adds new code, so you can be sure future changes don't break the work you've done here. Make sense?

Anyhow, I'm going to merge this, as I think a test is a little too much for this one case. But consider writing tests for your code in the future, as it's a more reliable and collaborative way to share your new changes, and the expectations they should fulfill! Thanks!

500swapnil:
Thank you! I will write tests next time. And I'm going to try testing the site locally now.
Any other issue I can help with?

jywarren:
Ah, i just responded to another comment of yours with some good thoughts on
testing as the first step. Thanks -- #754.

500swapnil:
Thank you I will have a look. And yes that sounds mighty nice :)

PR #1021, opened 2016-11-24
post to editor rename

### LINKED ISSUE #1014: Redirect /post to /editor unless it's an event post
We're pretty much ready with our new rich editor (at https://publiclab.org/editor) except for event posting, so we should set a redirect now in the editor_controller, unless someone specifically wants the old one. So let's rename the `post` action to `legacy`:

Then, make a new `post` method which checks if we definitely want the legacy version, and if not, redirects to the new editor. 

But since we don't yet have the events widget working in the new editor, we should also go to the legacy editor if it's going to be an event:

```ruby
def post
  if params[:legacy] || params[:template] == "event"
    flash.now[:notice] = "This is the legacy editor. For the new rich editor, <a href='/editor?#{request.env['QUERY_STRING']}'>click here</a>."
    legacy
  else
    flash.now[:notice] = "This is the new rich editor. For the legacy editor, <a href='/post?#{request.env['QUERY_STRING']}&legacy=true'>click here</a>."
    rich
  end
end
```

I think that'll work, though it'll likely cause some test errors which will have to be adjusted for the new rich editor page.

mridulnagpal:

Closes #1014 

Thanks!

### DIFF
--- a/app/controllers/editor_controller.rb
+++ b/app/controllers/editor_controller.rb
@@ -3,7 +3,7 @@ class EditorController < ApplicationController
   before_filter :require_user, :only => [:post, :rich]
 
   # main image via URL passed as GET param
-  def post
+  def legacy
     # /post/?i=http://myurl.com/image.jpg
     if params[:i]
       @image = Image.new({
@@ -13,6 +13,16 @@ class EditorController < ApplicationController
       flash[:error] = "The image could not be saved." unless @image.save!
     end
     redirect_to "/questions/new?#{request.env['QUERY_STRING']}" if params[:tags] && params[:tags].include?("question:")
+  end
+
+  def post
+    if params[:legacy] || params[:template] == "event"
+      flash.now[:notice] = "This is the legacy editor. For the new rich editor, <a href='/editor?#{request.env['QUERY_STRING']}'>click here</a>."
+      legacy
+    else
+      flash.now[:notice] = "This is the new rich editor. For the legacy editor, <a href='/post?#{request.env['QUERY_STRING']}&legacy=true'>click here</a>."
+      rich
+    end
   end
 
   def rich


jywarren:
OK, I see that there are a couple errors which we might expect:

https://travis-ci.org/publiclab/plots2/builds/178505020#L2658

This is because we're now changing the behavior of the `/post` action -- so in this test, the behavior is different here: 

https://github.com/publiclab/plots2/blob/master/test/functional/editor_controller_test.rb#L48

This is a little bit subtle -- we need to preserve the functionality of this line: https://github.com/publiclab/plots2/blob/master/app/controllers/editor_controller.rb#L15 which redirects question posts. Sorry I didn't anticipate this in my original description!

Also, I think it'd be best to add a new test for legacy vs. non-legacy requests to `/post` to confirm that those work properly. 

This test:

https://github.com/publiclab/plots2/blob/master/test/functional/editor_controller_test.rb#L14-L22

Should apply to the legacy form, so you might add to it to mark the request as such:

```ruby
    get :post,
        tags: 'one,two',
        legacy: true # <= here, we request the legacy version of the page
```

Then, we copy it and make a second /without/ the legacy parameter, but then the expectations for the HTML we'd encounter would be different; they'd be based on this page instead of the old post page:

https://github.com/publiclab/plots2/blob/master/app/views/editor/rich.html.erb

So we'd want to update those assertions to match, on the lines lower down in that test. Make sense?

ykl7:
@mridulnagpal if this is not being done at the moment, do you mind if i take a shot at it?

mridulnagpal:
@ykl7 Yeah sure, anyways I am stuck up with some work already :) . Glad to see you here

jywarren:
Oh great, thanks @ykl7 - and please base your work on this PR -- whether that means opening a new PR that references this one and uses its commits, or however. Thank you!

ryzokuken:

```
$ git fetch upstream
$ git checkout master
$ git merge upstream/master
$ git checkout <my-branch>
$ git rebase master
$ git push origin <my-branch>
```

Thanks for bearing with us,
Ujjwal

ykl7:
@ryzokuken Thanks for the heads-up. I haven't started on this yet since I got some unexpected work come up. Will keep in mind when I get to it

mridulnagpal:
@jywarren So you want the legacy editor on /post&legacy=true but what does it display on /post and once it is transferred to /editor there is no option to switch back to legacy editor. So what all specifications should it have (page wise).

jywarren:
Thanks -- 

* /post should show what's currently shown at /editor
* /post?legacy=true should show what's currently shown at /post (so you have a way out if the new editor doesn't work for you)
* /editor should redirect to /post

Make sense?

mridulnagpal:
#1091 Have a look please !

jywarren:
Also, it's fine (even preferable) to make additional changes to the same PR -- that way the conversation is threaded and not spread over multiple pages. It's fine this time, but next time please do your best to push to the same PR -- you can use `git push -f origin <feature-branch-name>`

Thanks!

PR #1022, opened 2016-11-24
helper maps

### LINKED ISSUE #1001: Separate map template out into a unique partial
Small Leaflet maps are currently shown in sidebars here (scroll to bottom, look at left side):

https://publiclab.org/search/lima

The code driving that is here:

https://github.com/publiclab/plots2/blob/master/app/views/sidebar/_related.html.erb#L80-L108

We should make a template helper like`render_map(lat, lon, items)` which accepts an array of items that have `.lat` and `.lon` methods, so we can use maps in more places. 

The helper could go in: https://github.com/publiclab/plots2/blob/master/app/helpers/application_helper.rb

Eventually we could use this to generate an inline tag that shows a map, just like they currently show a table of results: https://publiclab.org/wiki/power-tags#Inline+power+tags

mridulnagpal:

Closes #1001 
Thanks!

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -21,7 +21,7 @@ module ApplicationHelper
       ""
     end
   end
-  
+
   def locale_name_pairs
     I18n.available_locales.map do |locale|
       [I18n.t('language', locale: locale), locale]
@@ -121,3 +121,10 @@ module ApplicationHelper
   end
 
 end
+
+module render_map(lat, lon, items)
+  def .lat
+  end
+
+  def .lon
+  end


app/helpers/application_helper.rb
```diff
@@ -121,3 +121,10 @@ def insert_extras(body)
   end
 
 end
+
+module render_map(lat, lon, items)
```
jywarren: Actually this could go above the `end` on line 123 -- and just start with `def render_map(lat, lon, items)` instead of being a module. And then it'd have to render a partial template in a file at `/app/views/map/_map_helper.html.erb` -- containing these lines: https://github.com/publiclab/plots2/blob/master/app/views/sidebar/_related.html.erb#L80-L108

This is a little complex, because that actually contains two types of maps -- ones based on `@node.lat` type objects, and one on `@maps.first.lat` type objects. But if you replace the code in `_related.html.erb` with a call to your new helper function, you just pass those particular types in, so the top becomes:

```ruby
render_map(@node.lat, @node.lon, @maps)
```

and the bottom becomes:

```ruby
render_map(@maps.first.lat, @maps.first.lon, @maps)
```

Does that make sense? It'll take some more reworking to get right, but I'm happy to review.

ryzokuken:

```
$ git fetch upstream
$ git checkout master
$ git merge upstream/master
$ git checkout <my-branch>
$ git rebase master
$ git push origin <my-branch>
```

Thanks for bearing with us,
Ujjwal

jywarren:

mridulnagpal:
Does this look fine ?? In the application helper file
def render_map(@node.lat, @node.lon, @maps)
    render template : "/app/views/map/_map_helper.html.erb"
    end
end

  def render_map(@maps.first.lat, @maps.first.lon, @maps)
    render template : "/app/views/map/_map_helper.html.erb"
  end

jywarren:
Let's hold off on this one until your other PRs are resolved, as it'll help things be less complex. I think the next step, though, would be to push up a new commit so we can see exactly where your code is!

Thanks, and I appreciate your patience and persistence!

mridulnagpal:
#1094

PR #1023, opened 2016-11-25
Added unanswered questions, fixes #806

### LINKED ISSUE #806: Q&A Tab for unanswered questions

This will need to be broken up into multiple parts to become `first-timers-only` issues -- we've suggested 4 pieces below:

****

## The problem

I was just going through the Q&A feature and thought it might be a good idea to include a tab for **unanswered questions**. That way we don't lose them as new questions get asked and I think it could be a good place for people to start contributing. 

This is what it currently looks like at https://publiclab.org/questions:

screen shot 2016-11-18 at 11 34 18 am

## Solution

There are four parts to this:

## 1

We'd need to add a new controller method for `def unanswered`, like this one: https://github.com/publiclab/plots2/blob/master/app/controllers/questions_controller.rb#L44-L57

But where that one `.joins(:answers)`, we'd want to only query for questions which **have no answers** -- so that'd be:

```ruby
@questions.includes(:answers)
                .where( answers: { aid: nil } )
```

## 2

We'd also need to make a new title option like this one, but for "Unanswered questions": https://github.com/publiclab/plots2/blob/master/app/views/questions/index.html.erb#L6

And a new tab for "Unanswered questions": https://github.com/publiclab/plots2/blob/master/app/views/questions/index.html.erb#L40

## 3

We'd need a new route in `config/routes.rb` - a copy of this one but for `unanswered`: https://github.com/publiclab/plots2/blob/master/config/routes.rb#L243

## 4

Finally, we'd want to test the controller action, again, a copy of the one for the `answered` action, but where we also test that the first result `assigns(:questions).first` has `.answers.length == 0`:

https://github.com/publiclab/plots2/blob/master/test/functional/questions_controller_test.rb#L107-L113

saakuffo:
PR for [Issue **#806** Q&A Tab for unanswered questions](https://github.com/publiclab/plots2/issues/806)

1. Added the unanswered method in the questions controller. Mirrored the answered method but included the advice given (used includes instead of joins). Ordered ASC instead of DESC because changes are newer questions will get more natural traffic. That said, this is not necessarily intuitive.

2. Added the Unanswered link to the index page and added the route.

3.  Added the test for Unanswered by mirroring answered but included the advice given and did a functional test to check if a node had answered. **Note**, I need to explore more but not as familiar with this test suite; note sure what data was there to test against (aka how is the test 'site' seeded for questions.)

4. Rake test:all passes but not headless JS testing (this was done in C9, had some issues doing this locally.

------------------------

Thanks!

### DIFF
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -56,6 +56,18 @@ class QuestionsController < ApplicationController
     render :template => 'questions/index'
   end
 
+  def unanswered
+    @title = "Unanswered questions"
+    @questions = DrupalNode.questions.where(status: 1)
+    sort_question_by_tags
+    @questions = @questions.includes(:answers)
+                .where( answers: { id: nil } )
+                .order('answers.created_at ASC')
+                .group('node.nid')
+                .paginate(:page => params[:page], :per_page => 30)
+    render :template => 'questions/index'
+  end
+  
   def shortlink
     @node = DrupalNode.find params[:id]
     if @node.has_power_tag('question')
--- a/app/views/questions/index.html.erb
+++ b/app/views/questions/index.html.erb
@@ -6,6 +6,8 @@
   <h2>Recently Answered <small class="hidden-sm">Recently answered questions</small></h2>
   <% elsif params[:action] == 'popular' %>
   <h2>Popular Questions <small class="hidden-sm">Most viewed questions</small></h2>
+  <% elsif params[:action] == 'unanswered' %>
+  <h2>Unanswered Questions <small class="hidden-sm">Unanswered questions</small></h2>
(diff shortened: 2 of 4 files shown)


jywarren:
Hello, this looks great and thank you for the detailed notes! 

As for test data, it is stored in `/test/fixtures/` -- and can be referenced with `fixture_file_name(:fixture_name)` -- i.e. `node(:blog)` for the "node:" entry in `/test/fixtures/nodes.yml` -- make sense? 

Your test looks good, and since it passes, I'm very happy to merge this in right now! 

If you're interested in another challenge, please do check out https://github.com/publiclab/plots2/labels/help-wanted

Thanks a whole lot -- and curious, how did you hear of us? (We're working to be more welcoming to new contributors.)

PR #1049, opened 2016-12-02
Add ability to sort wiki index alphabetically (closes #193) 

### LINKED ISSUE #193: Add ability to sort wiki pages alphabetically

This would need to be broken into smaller parts before becoming a `first-timers-only` issue. 

****

This is a new feature request for https://publiclab.org/wiki. This is similar to what has been requested for /tags in https://github.com/publiclab/plots2/issues/179. 

The wiki index page is here: 

https://github.com/publiclab/plots2/blob/master/app/controllers/wiki_controller.rb#L261-L271

We'll be adding a new conditional to the Rails controller code to look for a `?order=alphabetic` parameter in the URL, and sorting accordingly if it exists.

Making it a GET parameter, like https://publiclab.org/wiki?order=alphabetic

The current ordering is done here: https://github.com/publiclab/plots2/blob/master/app/controllers/wiki_controller.rb#L266

We can detect this via `params[:order]` in the controller, but first we should remove the `.order()` that already exists, and add it to an `if/else` conditional. So if our GET param equals `alphabetic`, we sort alphabetically. If it doesn't, we sort by the existing method:

```ruby
                       .order("node_revisions.timestamp DESC")
```

Sorting alphabetically should be:

```ruby
                       .order("node_revisions.title DESC")
```

We could then test this with a similar test to this one, but where we specify `order: 'alphabetic'` in the GET request: https://github.com/publiclab/plots2/blob/master/test/functional/wiki_controller_test.rb#L30-L35

xssc:
Add ability for wiki index to be sorted alphabetically using the `order` get paramater

### DIFF
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -261,9 +261,15 @@ class WikiController < ApplicationController
   def index
     @title = I18n.t('wiki_controller.wiki')
 
+    if params[:order] == 'alphabetic'
+      order_string = "node_revisions.title ASC"
+    else
+      order_string = "node_revisions.timestamp DESC"
+    end
+
     @wikis = DrupalNode.includes(:drupal_node_revision, :drupal_node_counter)
                        .group('node_revisions.nid')
-                       .order("node_revisions.timestamp DESC")
+                       .order(order_string)
                        .where("node_revisions.status = 1 AND node.status = 1 AND (type = 'page' OR type = 'tool' OR type = 'place')")
                        .page(params[:page])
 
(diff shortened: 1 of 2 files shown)


jywarren:
Hi -- this looks perfect; merging now. If you're looking for a new challenge, please take a look at our `help-wanted` list: https://github.com/publiclab/plots2/labels/help-wanted

Thanks!

ananyo2012:
@xssc Thanks for your great work! Giving some good practices you should follow from next time. Write good commit messages for your work. Mention the issue no like `Fixes #193` Here #193 is the issue no. Use the imperative mood in commit messages. Here is a guide on writing good commit messages http://chris.beams.io/posts/git-commit/. Also a cool thing about mentioning the issue in the commit message or Pull request title auto closes the issue when merged!

xssc:
@ananyo2012 thanks for the tips!

PR #1055, opened 2016-12-03
Remove history graph for wildcard tag display (Fixes #878)

### LINKED ISSUE #878: History graph for wildcard tag display should not be displayed
Seen on https://publiclab.org/tag/question:*
The graph at the top of activity over time seems not to be populated: 

screen shot 2016-10-11 at 3 11 35 pm

****

@jywarren sez: This is because wildcards (`*`) are not real tags, so the graph should not really be shown at all. The following code (on this line: https://github.com/publiclab/plots2/blob/master/app/views/tag/show.html.erb#L25) is suppose to detect that and not show the graph, but is not working properly:

```ruby
  <% if @tags.first %>
```

We could:

1. Modify the conditional above to use the same logic as the controller  (https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L28), making it: `@tags.first && params[:id][-1..-1] != "*"` -- which ensures that the last character is not a wildcard.
2. Add a test that checks this page and ensures a graph is not shown:

The test could be based on this one: https://github.com/publiclab/plots2/blob/master/test/functional/tag_controller_test.rb#L85-L101

```ruby
  test "wildcard tag show" do

    get :show, id: 'question:*'

    assert :success
    assert_not_nil :tags
    assert :wildcard # if this fails, this line may need to be: assert assigns(:wildcard)

    assert_select '#note-graph', 0 # test that the graph is not shown
  end
```

xssc:
Fixed the issue of the graph showing on wildcard tag display (#878)

### DIFF
--- a/app/views/tag/show.html.erb
+++ b/app/views/tag/show.html.erb
@@ -25,7 +25,7 @@
     <p><a class="btn btn-primary" href="/wiki/<%= params[:id] %>"><span class="fa fa-plus fa-white"></span> Add one now</a></p>
   <% end %>
 
-  <% if @tags.first %>
+  <% if @tags.first && params[:id][-1..-1] != "*" %>
   <div id="note-graph" style="height:100px;"></div>
 
   <script type="text/javascript">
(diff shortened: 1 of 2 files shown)


ananyo2012:
@xssc Good work on this. @jywarren This can be merged.

jywarren:
Thank you!!!!

PR #1058, opened 2016-12-04
#1034: Add node.can_tag to check if given user can powertag a given tagname

### LINKED ISSUE #1034: new node.can_tag(tagname, user) check in DrupalNode model
## Problem

Both co-authorship (#761) and wiki locking (#397) features need a check in `node.add_tag(tagname, user)` to determine if the given user is allowed to make that kind of power tag -- now that we have some more powerful "power tags" -- https://publiclab.org/wiki/power-tags and both locking and co-authorship permissions will be tag-based. 

## Solution

I propose a new method called `node.can_tag(tagname, user)` which compares the tagname and determines if this user is allowed to apply that tag to this node. It'd go here:

https://github.com/publiclab/plots2/blob/master/app/models/drupal_node.rb

and can be based on the comparisons currently done in the tag_controller:

https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L114-L121

Once we create this **and unit-test it**, both co-authorship and wiki locking can use it -- and we can even add it to `node.add_tag()` and unit-test it. 

Also, controllers will be able to check for this and produce helpful error messages, more compactly than is done in the tag controller now:

https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L114-L121

This is also helpful because right now, tags can be added outside that controller (like on this line: https://github.com/publiclab/plots2/blob/master/app/controllers/notes_controller.rb#L141-L144) which bypass the tag_controller filters. 

So let's begin with just the `can_tag()` method and then break out the next steps!

ashleypt:

Thanks!

Added unit tests for node.can_tag method

Fixes #1034

### DIFF
--- a/app/models/drupal_node.rb
+++ b/app/models/drupal_node.rb
@@ -779,4 +779,18 @@ class DrupalNode < ActiveRecord::Base
               .where('term_data.name LIKE ?', "upgrade:#{tagname}")
   end
 
+  def can_tag(tagname, user)
+    if tagname[0..4] == "with:"
+      if self.author.uid != user.uid || User.find_by_username(tagname.split(':')[1]).nil? || tagname.split(':')[1] == user.username
+        return false
+      else
+        return true
+      end
+    elsif tagname[0..4] == "rsvp:" && user.username != tagname.split(":")[1]
+      return false
+    else
+      return true
+    end
+  end
+
 end
--- a/test/unit/drupal_node_tag_test.rb
+++ b/test/unit/drupal_node_tag_test.rb
@@ -119,4 +119,73 @@ class DrupalNodeTagTest < ActiveSupport::TestCase
     assert_nil node.tagged_lon
   end
 
+  test "can't powertag with: yourself" do
+    user = rusers(:bob)
+    node =  DrupalNode.new({
+      uid: user.id,
+      type: 'note',
+      title: 'My research note'
+    })
+    tagname = "with:#{user.username}"
+    assert_false node.can_tag(tagname, user)
+  end
+
+  test "can powertag with: another user" do
+    user = rusers(:bob)
+    jeff = rusers(:jeff)
+    node = DrupalNode.new({
+      uid: user.id,
+      type: 'note',
+      title: 'My research note'
+    })
+    tagname = "with:#{jeff.username}"
+    assert node.can_tag(tagname, user)
+  end
+
+  test "can't tag with: a nonexistent user" do
+    user = rusers(:bob)
+    node = DrupalNode.new({
+      uid: user.id,
+      type: 'note',
+      title: 'My research note'
+    })
+    tagname = "with:steven"
+    assert_false node.can_tag(tagname, user)
+  end
+
+  test "can't powertag if you're not author" do
+    user = rusers(:bob)
+    jeff = rusers(:jeff)
+    node = DrupalNode.new({
+      uid: jeff.id,
+      type: 'note',
+      title: 'My research note'
+    })
+    tagname = "with:#{jeff.username}"
+    assert_false node.can_tag(tagname, user)
+  end
+
+  test "can rsvp yourself" do
+    user = rusers(:bob)
+    node = DrupalNode.new({
+      uid: user.id,
+      type: 'note',
+      title: 'My research note'
+    })
+    tagname = "rsvp:#{user.username}"
+    assert node.can_tag(tagname, user)
+  end
+
+  test "can't rsvp someone else" do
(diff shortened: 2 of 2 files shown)


jywarren:
Hmm, error is:

```================================================================================

Failure: <false> is not true.

test_shows_things_tagged_with_child_tag(TagControllerTest)

test/functional/tag_controller_test.rb:257:in `block in <class:TagControllerTest>'

================================================================================
```

ashleypt:
yeah it's weird - I don't think my local tests gave that error

jywarren:
Hmm, the error is:

```ruby
    assert       assigns(:notes).first.has_tag_without_aliasing('spectrometer')
```

And it's for this action:

https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L16-L51

is it possible that you missed an instance of `drupal_node` or `DrupalNode`? You searched pretty thoroughly?

jywarren:
Oh oops, am thinking of the wrong issue, wait

jywarren:
Yeah that's so weird -- your code shouldn't touch that, right?

jywarren:
Gosh that's a stumper -- um, can you try to assert or `puts` more basic things about `assigns(:notes).first.has_tag_without_aliasing('spectrometer')` like `assigns(:notes).first.tags.collect(&:name)` to see what the values are? At least we should understand why it's failing, even if we're not sure how your code caused it...

Detective time, i guess -- 🕵️ / :female_detective:

ashleypt:
Alright, I'll try that next - it's absolutely bizarre. The only thing I can
think of is that since checking data on these things in ruby calls methods
(like that pesky original comments method in node_shared.rb for example),
maybe there's some weird method defined that I'm calling that instead of
just returning a value has some bizarre side effect somewhere else in the
code.

this is making me see why people learn functional programming!

jywarren:
or... somehow maybe `can_tag` is a reserved word or something... ??? but
that doesn't make sense either. And you're not saving any of the records
you're testing out in your unit tests, so it doesn't seem like they could
be affecting what's in the test db... i mean the only thing we can be sure
of is that your unit tests do run before the functional tests.

ryzokuken:
`rake test` passes on my computer (the failing test is a part of those)
`rake test:all` has again started throwing an error at the jasmine tests and idk why. I'd look into them.

jywarren:
> throwing an error at the jasmine tests

I don't think this is happening in Travis, if that's helpful?

Hmm, this is odd. I'll try running this myself in my local.

jywarren:
I'll also rebase it on top of the latest master...

jywarren:
It passed 100% clean for me too -- how strange. I'm going to rebase it now and see if somehow that's causing it; but I think the Travis tests run before squashing -- just on the `ashleypt:add-node-can-tag` branch itself... so strange!

jywarren:
Passed cleanly after rebase also. I'm going to merge this manually.

Done in a408340..96e5650 -- we can check if that passes here: https://travis-ci.org/publiclab/plots2/builds/181369881

Since 3 people confirmed it passed with no errors, I'm befuddled but confident enough to just move forward. Thanks!

jywarren:
Yikes, it failed again: https://travis-ci.org/publiclab/plots2/builds/181369881

What's going on? Reality uncertain.

jywarren:
re-running locally and I'll see if I can add some clarifying extra assertions

ashleypt:
this is the most interesting bug I've ever seen, it must have to do with
travis or the data being passed into it right?

jywarren:
I kind of want to google around for "failing in travis passing locally" or
something to see what could possibly be causing it. I mean, what are the
differences? I'm running tests in `mysql2` -- how about others?

jywarren:
Added:

```ruby

```

And now see this, which shows why it passes locally:

```ruby
=====================================
Failure:
  <[]> expected but was
  <[#<DrupalTag tid: 12, vid: 0, name: "spectrometer", description: nil, weight: 0, count: nil, parent: "spectrometry">]>.
```

I'll push this to a PR so we can see if/why it doesn't on TravisCI.

jywarren:
I guess merging before clearing this up was a mistake on my part, sorry! And the new assertion fails with:

```
Failure:

  <[]> expected but was

  <[#<DrupalTag tid: 13, vid: 0, name: "spectrometry", description: nil, weight: 0, count: 1, parent: "">]>.
```

jywarren:
So, that's different. How odd!

jywarren:
I'm going to continue chasing this error over at #1062.

PR #1062, opened 2016-12-05
testing inconsistencies tests

jywarren:
Followup to #1058

PR #1082, opened 2016-12-08
can_tag test expermentation

jywarren:
To try to isolate the exact cause after #1062.

### DIFF
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -299,6 +299,7 @@ class WikiController < ApplicationController
     render :template => "wiki/index"
   end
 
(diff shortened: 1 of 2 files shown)


jywarren:
Whoa! d92b5d4 passed!!! That's... different than what happened before. Trying last test.

jywarren:
Hooray. No idea why these were not passing before -- @ashleypt - the mystery is not solved, but at least it's bypassed. Maybe a cosmic ray was hitting the wrong bit over and over, and it was all random. Kind of unsettling!

ashleypt:
@jywarren that's good to know - but did the cosmic rays hit DHH or the test server?

PR #1090, opened 2016-12-12
"Browse other activities" next to "replicate this" prompt, fixes #754

### LINKED ISSUE #754: "Browse other activities" next to "replicate this" prompt

500swapnil:

Thanks!

### DIFF
--- a/app/views/notes/_responses.html.erb
+++ b/app/views/notes/_responses.html.erb
@@ -7,8 +7,10 @@
  
     <p id="<%= response_type %>s"><a class='btn btn-primary btn-lg' href="/post?tags=<%= response_type %>:<%= @node.id %><% if response_type == 'response' %>,hidden:response<% end %><%= ',' + (@tagnames.uniq.delete_if{|x| x.match(":") }).join(',') if @tagnames && @tagnames.length > 0 %>"><%= t('notes._responses.post_' + response_type) %></a>&nbsp;
       <% if @node.power_tag('activity') %>
-        <% if DrupalTag.where(name: @node.power_tag('activity')).first.nodes %>
-          <%= DrupalTag.where(name: @node.power_tag('activity')).first.nodes.first.path %>
+        <% if DrupalTag.where(name: @node.power_tag('activity')).first %>
+          <% if DrupalTag.where(name: @node.power_tag('activity')).first.nodes %>
+            <%= DrupalTag.where(name: @node.power_tag('activity')).first.nodes.first.path %>
+          <% end %> 
         <% end %>
       <% end %>
     </p>
--- a/app/views/notes/_responses.html.erb
+++ b/app/views/notes/_responses.html.erb
@@ -9,7 +9,7 @@
       <% if @node.power_tag('activity') %>
         <% if DrupalTag.where(name: @node.power_tag('activity')).first %>
           <% if DrupalTag.where(name: @node.power_tag('activity')).first.nodes %>
-            <a href="<%= DrupalTag.where(name: @node.power_tag('activity')).first.nodes.first.path %>">Browse other activities for "<%= @node.power_tag('activity') %>"</a>
+            <a id="other-activities" href="<%= DrupalTag.where(name: @node.power_tag('activity')).first.nodes.first.path %>">Browse other activities for "<%= @node.power_tag('activity') %>"</a>
           <% end %> 
         <% end %>
       <% end %>
--- a/app/views/notes/_responses.html.erb
+++ b/app/views/notes/_responses.html.erb
@@ -9,7 +9,7 @@
       <% if @node.power_tag('activity') %>
         <% if DrupalTag.where(name: @node.power_tag('activity')).first %>
           <% if DrupalTag.where(name: @node.power_tag('activity')).first.nodes %>
-            <%= DrupalTag.where(name: @node.power_tag('activity')).first.nodes.first.path %>
+            <a href="<%= DrupalTag.where(name: @node.power_tag('activity')).first.nodes.first.path %>">Browse other activities for "<%= @node.power_tag('activity') %>"</a>
           <% end %> 
         <% end %>
       <% end %>
--- a/app/views/notes/_responses.html.erb
+++ b/app/views/notes/_responses.html.erb
@@ -6,9 +6,12 @@
     <% @unpaginated = true %>
  
     <p id="<%= response_type %>s"><a class='btn btn-primary btn-lg' href="/post?tags=<%= response_type %>:<%= @node.id %><% if response_type == 'response' %>,hidden:response<% end %><%= ',' + (@tagnames.uniq.delete_if{|x| x.match(":") }).join(',') if @tagnames && @tagnames.length > 0 %>"><%= t('notes._responses.post_' + response_type) %></a>&nbsp;
-    <% if @node.power_tag('activity') %>
-    	<% if DrupalTag.where(name: @node.power_tag('activity')) %>
-    		<% @responses = DrupalTag.where(name: @node.power_tag('activity')).tag.nodes.first %><% end %><% end %></p>
+      <% if @node.power_tag('activity') %>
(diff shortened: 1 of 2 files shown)


app/views/notes/_responses.html.erb
```diff
@@ -5,7 +5,10 @@
     <% @responses = DrupalTag.find_nodes_by_type("#{response_type}:#{@node.id}", "note", 200) %>
     <% @unpaginated = true %>
  
-    <p id="<%= response_type %>s"><a class='btn btn-primary btn-lg' href="/post?tags=<%= response_type %>:<%= @node.id %><% if response_type == 'response' %>,hidden:response<% end %><%= ',' + (@tagnames.uniq.delete_if{|x| x.match(":") }).join(',') if @tagnames && @tagnames.length > 0 %>"><%= t('notes._responses.post_' + response_type) %></a></p>
+    <p id="<%= response_type %>s"><a class='btn btn-primary btn-lg' href="/post?tags=<%= response_type %>:<%= @node.id %><% if response_type == 'response' %>,hidden:response<% end %><%= ',' + (@tagnames.uniq.delete_if{|x| x.match(":") }).join(',') if @tagnames && @tagnames.length > 0 %>"><%= t('notes._responses.post_' + response_type) %></a>&nbsp;
+    <% if @node.power_tag('activity') %>
+    	<% if DrupalTag.where(name: @node.power_tag('activity')) %>
+    		<% @responses = DrupalTag.where(name: @node.power_tag('activity')).tag.nodes.first %><% end %><% end %></p>
```
jywarren: OK, - closer, great -- here, I don't think we need an instance variable for `@responses` -- but we could say:

```ruby
<% if @node.power_tag('activity') %>
  <% if DrupalTag.where(name: @node.power_tag('activity')).first.nodes %>
    <%= DrupalTag.where(name: @node.power_tag('activity')).first.nodes.first.path %>
  <% end %>
<% end %>
```

Please also put `end` statements and close `</p>` tags on their own line, with clean indentations (two spaces per indent) -- that makes it a lot easier to read!

See how this is a lot of logic to put into a template? If we imagined we'd use this logic more than just once, we might want to move some of it into the `DrupalTag` model, to simplify the code here. Then we could also unit test the code to ensure it doesn't break in the future. But it's your choice -- if it's just going to be used here, we're OK just doing a little bit of logic in the templates, as long as its readable and efficient.

jywarren: And see how we ask `@node.power_tag('activity')` three times? We could make that a bit more efficient too -- although it will probably cache the database call that fetches that info, so it's not too bad.

500swapnil: Okay it isnt used that much so we will see about adding it to the DrupalTag model. I will make these changes. And don't we need all this in 'Browse other activities' link and '<a>' tags?

jywarren:
OK - great - see how the tests ran, and some failed? This highlights exactly the IF checks we need -- i guess we forgot at least one. So the first test failure is:

```
Error: test_users_are_logged_out_and_alerted_when_banned,_and_notes_are_not_accessible(ModerateAndBanTest): ActionView::Template::Error: undefined method `nodes' for nil:NilClass
```

And note that these failed on the following line: `app/views/notes/_responses.html.erb:10`

So, we're calling `.nodes` on a `nil` -- that nil was the result of:

```ruby
DrupalTag.where(name: @node.power_tag('activity')).first
```

So i guess we have to add another check that the above call doesn't return `nil`. 

You can also write shorter checks by using this neat `.try()` method [in ActiveSupport](http://guides.rubyonrails.org/active_support_core_extensions.html#try). That can perhaps keep your code a bit more compact, if it's helpful.

Thanks!!!

jywarren:
This is a great example of how testing can help catch problems before we merge code!

500swapnil:
don't we need all this in 'Browse other activities' link and '' tags?

app/views/notes/_responses.html.erb
```diff
@@ -5,7 +5,15 @@
     <% @responses = DrupalTag.find_nodes_by_type("#{response_type}:#{@node.id}", "note", 200) %>
     <% @unpaginated = true %>
  
-    <p id="<%= response_type %>s"><a class='btn btn-primary btn-lg' href="/post?tags=<%= response_type %>:<%= @node.id %><% if response_type == 'response' %>,hidden:response<% end %><%= ',' + (@tagnames.uniq.delete_if{|x| x.match(":") }).join(',') if @tagnames && @tagnames.length > 0 %>"><%= t('notes._responses.post_' + response_type) %></a></p>
+    <p id="<%= response_type %>s"><a class='btn btn-primary btn-lg' href="/post?tags=<%= response_type %>:<%= @node.id %><% if response_type == 'response' %>,hidden:response<% end %><%= ',' + (@tagnames.uniq.delete_if{|x| x.match(":") }).join(',') if @tagnames && @tagnames.length > 0 %>"><%= t('notes._responses.post_' + response_type) %></a>&nbsp;
+      <% if @node.power_tag('activity') %>
+        <% if DrupalTag.where(name: @node.power_tag('activity')).first %>
+          <% if DrupalTag.where(name: @node.power_tag('activity')).first.nodes %>
+            <%= DrupalTag.where(name: @node.power_tag('activity')).first.nodes.first.path %>
```
jywarren: Not sure if this answers your question, but here, we now can make this appear in a link, like:

```ruby
<a href="<%= DrupalTag.where(name: @node.power_tag('activity')).first.nodes.first.path %>">Browse other activities for "<%= @node.power_tag('activity') %>"</a>
```

Make sense?

500swapnil: Got it! Yes it clears it up. Thanks!

500swapnil:
Now do we need to add a test for this link in **** ?

jywarren:
I think we could... You could add an extra assert_select to an existing
test on a node tagged with "activity:...." -- there should be one in
/test/functional/notes_controller_test.rb, though you might have to add an
id= property to the <a> tag so you can correctly select it.

I can help with this tomorrow!

500swapnil:
So I guess it would be
assert_select "a.other-activites" , true
where the id of '<a>' is 'other-activites'. Ok let me know where I can add this tomorrow. Thanks!

jywarren:
Yes, but # instead of . for ids -- periods are for class=

I can provide more input tomorrow!

jywarren:
So, if it's an id, you do `assert_select "a#other-activities"` (true is implied with only one param), and the test could go among these similar ones:

https://github.com/publiclab/plots2/blob/master/test/integration/node_insert_extras_test.rb#L47

500swapnil:
Is this ok?

500swapnil:
Is this ok?

jywarren:
Hi -- this looks perfect; merging now. If you're looking for a new challenge, please take a look at our `help-wanted` list: https://github.com/publiclab/plots2/labels/help-wanted

Thanks!

500swapnil:
Thanks!

ebarry:
Thanks so much @500swapnil !!!!! This is a really useful addition to helping everyone find the projects they are interested in.

500swapnil:
Happy to help! @ebarry

Issue #1040, opened 2016-11-30
incorrectly redirected to successful posting in new rich editor


jywarren:
### What happened just before the problem occurred

I posted a new note at

https://publiclab.org/editor

It published fine: https://publiclab.org/notes/warren/11-30-2016/how-many-people-to-do-an-activity

However, I saw this:

![screenshot 2016-11-30 at 4 56 16 pm](https://cloud.githubusercontent.com/assets/24359/20773080/24b60184-b71e-11e6-9532-bb100441758f.png)

I believe this is because it didn't recognize a successful response from this line: 

https://github.com/publiclab/plots2/blob/master/app/controllers/notes_controller.rb#L181

On this line, we should better recognize a response with zero errors, and go to the post:

https://github.com/publiclab/plots2/blob/master/app/views/editor/rich.html.erb#L213

Next step: go through this in development mode and find what a successful response looks like, and then test that the JavaScript code on line 213 above correctly identifies it as successful.

jywarren:
@ashleypt - curious if you're interested in this one -- it's at the intersection of the JavaScript editor library and the main Rails application, and shows some of the challenges of interfacing complex projects. Give a shout if you want to take a crack at it!

ashleypt:
@jywarren @ashleypt sure, I can take a look at it :)

ashleypt:
oops I accidentally hit close and comment D:

ashleypt:
I'll start off by console.logging the response object and the results of the condition in the erb. then I'll try passing in different json objects to see if I can replicate those errors

jywarren:
great!

jywarren:
Hi, @ashleypt - just checking if you've gotten stuck on this or needed any help. #1091 will make the new editor default, so I want to be sure we resolve this too before publishing that one. Thanks!

jywarren:
And if you don't think you'll be able to work on it soon, I'm happy to take this one and help you find another that'll be interesting. Thanks!

jywarren:
I went ahead and resolved this last night since we had a switchover to the rich editor in #1091 -- but I'm happy to help you find another issue, and there are a few in https://github.com/publiclab/PublicLab.Editor/ as well. Thanks!

PR #1091, opened 2016-12-12
Rich Editor as default at /post with legacy fallback option

mridulnagpal:

Thanks!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '2.3.1'
+ruby '2.1.2'
 gem 'rails', '~> 3.2.20'
 gem 'passenger'
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '2.1.2'
+ruby '2.3.1'
 gem 'rails', '~> 3.2.20'
 gem 'passenger'
 
@@ -42,7 +42,7 @@ group :test do
   gem 'test-unit'
   gem 'rake',  '~> 10.5.0'
   # gems to test RESTful API
-  gem 'rest-client' 
+  gem 'rest-client'
   gem 'rspec'
   gem 'json_expressions'
   gem 'timecop'
@@ -90,7 +90,7 @@ gem 'geocoder'
 gem "i18n-js", ">= 3.0.0.rc11"
 gem 'http_accept_language'
 
-# The default friendly_id version compatible with Rails 3 is v4.0 
+# The default friendly_id version compatible with Rails 3 is v4.0
 gem 'friendly_id'
 gem 'jbuilder'
 gem 'strong_parameters'
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '2.1.2'
+ruby '2.3.1'
 gem 'rails', '~> 3.2.20'
 gem 'passenger'
 
--- a/app/controllers/editor_controller.rb
+++ b/app/controllers/editor_controller.rb
@@ -29,7 +29,9 @@ class EditorController < ApplicationController
   end
 
   def post
-    if params[:legacy] || params[:template] == "event"
+    if params[:tags] && params[:tags].include?("question:")
+      redirect_to "/questions/new?#{request.env['QUERY_STRING']}"
+    elsif params[:legacy] || params[:template] == "event"
       legacy
     else
       rich
--- a/app/controllers/editor_controller.rb
+++ b/app/controllers/editor_controller.rb
@@ -17,12 +17,15 @@ class EditorController < ApplicationController
       node = DrupalNode.find(params[:n])
       params[:body] = node.body if node
     end
-    redirect_to "/questions/new?#{request.env['QUERY_STRING']}" if params[:tags] && params[:tags].include?("question:")
-    render template: 'editor/post'
+    if params[:tags] && params[:tags].include?("question:")
+      redirect_to "/questions/new?#{request.env['QUERY_STRING']}"
+    else
+      render template: 'editor/post'
+    end
   end
 
   def editor
-    redirect_to "/post"
+    redirect_to "/post?#{request.env['QUERY_STRING']}"
   end
 
   def post
--- a/app/controllers/editor_controller.rb
+++ b/app/controllers/editor_controller.rb
@@ -18,14 +18,12 @@ class EditorController < ApplicationController
       params[:body] = node.body if node
     end
     redirect_to "/questions/new?#{request.env['QUERY_STRING']}" if params[:tags] && params[:tags].include?("question:")
-    return "/editor/post"
+    render template: 'editor/post'
   end
-
 
   def editor
     redirect_to "/post"
   end
-
 
   def post
     if params[:legacy] || params[:template] == "event"
@@ -35,7 +33,6 @@ class EditorController < ApplicationController
       render "/editor/rich"
     end
   end
-
 
   def rich
     flash.now[:notice] = "This is the new rich editor. For the legacy editor, <a href='/post?#{request.env['QUERY_STRING']}&legacy=true'>click here</a>."
--- a/app/controllers/editor_controller.rb
+++ b/app/controllers/editor_controller.rb
@@ -3,8 +3,9 @@ class EditorController < ApplicationController
   before_filter :require_user, :only => [:post, :rich]
 
   # main image via URL passed as GET param
-  def post
+  def legacy
     # /post/?i=http://myurl.com/image.jpg
+    flash.now[:notice] = "This is the legacy editor. For the new rich editor, <a href='/editor'>click here</a>."
     if params[:i]
       @image = Image.new({
         :remote_url => params[:i],
@@ -17,9 +18,25 @@ class EditorController < ApplicationController
       params[:body] = node.body if node
     end
     redirect_to "/questions/new?#{request.env['QUERY_STRING']}" if params[:tags] && params[:tags].include?("question:")
+    return "/editor/post"
   end
 
+  def editor
+    redirect_to "/post"
+  end
+
+  def post
+
+  if params[:legacy] || params[:template] == "event"
+    legacy
+  else
+    rich
+    render "/editor/rich"
+  end
+end
+
   def rich
+    flash.now[:notice] = "This is the new rich editor. For the legacy editor, <a href='/post?#{request.env['QUERY_STRING']}&legacy=true'>click here</a>."
     if params[:main_image] && Image.find_by_id(params[:main_image])
       @main_image = Image.find_by_id(params[:main_image]).path
     end
--- a/app/controllers/editor_controller.rb
+++ b/app/controllers/editor_controller.rb
@@ -1,6 +1,6 @@
 class EditorController < ApplicationController
 
-  before_filter :require_user, :only => [:post, :rich]
+  before_filter :require_user, :only => [:post, :rich, :legacy, :editor]
 
   # main image via URL passed as GET param
   def legacy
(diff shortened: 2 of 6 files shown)


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '2.1.2'
+ruby '2.3.1'
```
jywarren: Hi, @mridulnagpal - although I'm not against upgrading this requirement, I don't think we should include it in this topically unrelated pull request. Would you mind opening a new issue/PR for this change, or at least removing it from your PR? I appreciate the thought, but one thing at a time, please!

app/controllers/editor_controller.rb
```diff
@@ -17,9 +18,25 @@ def post
...
+    return "/editor/post"
+  end
+
+  def editor
+    redirect_to "/post"
+  end
+
+  def post
+
+  if params[:legacy] || params[:template] == "event"
```
jywarren: Would you mind tweaking the indents here to be a bit more orderly? Thank you!

test/functional/editor_controller_test.rb
```diff
@@ -11,9 +11,9 @@ def setup
     assert_redirected_to '/login'
   end
 
-  test "should get post form" do
+  test "should get legacy form" do
     UserSession.create(rusers(:bob))
-    get :post,
+    get :legacy,
```
jywarren: I see what you're doing here, and it makes sense. But I think we need to change the tests, or add new tests, so that the new `post` action is as well tested as the previous one was. Does that make sense? So, for each one you're converting to a legacy test, we should add an equivalent test for the new `post` action. 

Is that something you think you could try? Not every bit of HTML will be the same, but we can look at how the new `/post` page renders, and adjust the assertions to match the new one. I'm happy to offer input/review as you go -- you can start with just one test, not the whole set -- thank you!

Gemfile
mridulnagpal: I had updated this for my version the final PR won't have this. Sorry! :)

mridulnagpal:
screenshot from 2016-12-13 13-11-40
screenshot from 2016-12-13 13-11-48
@jywarren The working seems fine now just have to update the tests. Please have a look.

app/controllers/editor_controller.rb
```diff
@@ -17,9 +18,27 @@ def post
       params[:body] = node.body if node
     end
     redirect_to "/questions/new?#{request.env['QUERY_STRING']}" if params[:tags] && params[:tags].include?("question:")
+    return "/editor/post"
+  end
+
+
```
jywarren: Can you remove these extra newlines? Thanks!

app/controllers/editor_controller.rb
```diff
@@ -17,9 +18,27 @@ def post
       params[:body] = node.body if node
     end
     redirect_to "/questions/new?#{request.env['QUERY_STRING']}" if params[:tags] && params[:tags].include?("question:")
+    return "/editor/post"
```
jywarren: Getting closer! I see the error:

`ActionView::MissingTemplate: Missing template editor/legacy, application/legacy with {:locale=>[:en], :formats=>[:html], :handlers=>[:erb, :builder, :coffee, :jbuilder]}. Searched in:`

I think you need to add a `render template: 'editor/post'` as the final line in this method, and `return "/editor/post"` doesn't seem necessary -- unless I'm missing something?

Thanks!

jywarren:
Almost there! I left a few requests. Thanks! 👍

mridulnagpal:
@jywarren Made some changes unable to solve that one test fail please review

jywarren:
Hi, the tests fail here: https://github.com/publiclab/plots2/blob/master/test/integration/login_flow_test.rb#L33

Try swapping lines 33 and 34, or just removing line 33 (since its implicit in 34) and we'll see where it actually did go?

mridulnagpal:
@jywarren had to get rid of both lines as now post is a 200 status i.e success because it is legacy which should redirect to questions but using legacy I get the error of Doublerender as we are redirecting and rendering a template at the same time

jywarren:
Ah - can you point me to where we are redirecting and rendering at the same time? I think we need to disambiguate the redirect to `/questions/new` form, but don't necessarily think we should remove those test lines -- where are we rendering? Are we perhaps just forgetting to redirect to `/questions/new`?

mridulnagpal:
@jywarren  We are rendering the post template whereas redirecting to questions one

jywarren:
Ah yes -- so that test was expecting to be redirected to the questions one, and I believe we've forgotten to deal with that scenario in the new code we've added.

app/controllers/editor_controller.rb
```diff
@@ -17,9 +18,24 @@ def post
       params[:body] = node.body if node
     end
     redirect_to "/questions/new?#{request.env['QUERY_STRING']}" if params[:tags] && params[:tags].include?("question:")
+    render template: 'editor/post'
+  end
+
+  def editor
+    redirect_to "/post"
```
jywarren: I think we also are losing the parameters when we redirect -- we should add `?#{request.env['QUERY_STRING']}` to this and any other new redirects we added here -- to preserve the URL parameters, see?

jywarren:
I left in-line comments - take a look, i think we found the issue. Thanks!

mridulnagpal:
@jywarren checked it for tests no fails on my local :)

jywarren:
Restarting the tests -- looks like an unrelated error.

mridulnagpal:
@jywarren Cheers!!

test/integration/login_flow_test.rb
```diff
@@ -28,10 +28,10 @@ class LoginFlowTest < ActionDispatch::IntegrationTest
       username: rusers(:jeff).username,
       password: 'secret'
     }
-    
+
     follow_redirect!
-    assert_response :redirect
-    assert_redirected_to  '/questions/new?tags=question%3Aquestion&template=question&title=What&redirect=question'
+    # assert_response :redirect
```
jywarren: OK - this is looking good, and we should be able to uncomment these again and they should work now! Great coding up above.

test/functional/editor_controller_test.rb
```diff
@@ -69,7 +77,7 @@ def setup
 
   test "should show title form input if title parameter present" do
```
jywarren: I'd like to mark this as "legacy" as well, and also can we please make copies of each test that is now of the "legacy" form, and adapt them to test the new form template? 

Then we'll be done!

mridulnagpal: I didn't get your point, please elaborate

jywarren: So, for each test which you've switched to test the "legacy" action, we should add the word "legacy" to the description. But second, we should really create new tests (they could be almost copies of the originals) to test the new template in each of the ways we've tested the legacy template. That way, both are tested and we are more assured that they're not broken as the code advances. Make sense? Thank you for your patience, we are very close!

jywarren:
The `post` action will now render the rich template, so yes, thanks!

mridulnagpal:
@jywarren Have a look

test/functional/editor_controller_test.rb
```diff
@@ -6,58 +6,97 @@ def setup
...
+  
+  test "newcomer should get post form" do
+    UserSession.create(rusers(:newcomer))
+    get :post
+    assert_response :success
+    assert_select "h1", "Share"
+    assert_select "p.ple-help", "Select an optional main image for your post."
+  end
 
   test "should redirect to login page while posting  question" do
```
jywarren: Can we add a check that requests to the new `:post` with `?template=question` **also** get redirected to "/legacy?redirect=question&tags=question%3Aquestion&template=question" ?

test/integration/login_flow_test.rb
jywarren: And just need these lines uncommented again -- thanks!

test/integration/login_flow_test.rb
```diff
@@ -30,8 +30,8 @@ class LoginFlowTest < ActionDispatch::IntegrationTest
     }
 
     follow_redirect!
-    # assert_response :redirect
-    # assert_redirected_to  '/questions/new?tags=question%3Aquestion&template=question&title=What&redirect=question'
+    assert_response :redirect
+    assert_redirected_to  '/questions/new?tags=question%3Aquestion&template=question&title=What&redirect=question'
     # assert_select "input#title" do
     # assert_select "[value=?]", "What"
```
jywarren: What about these -- do they not pass? Just checking, thanks!

mridulnagpal:
@jywarren This won't work as this is applicable for the post template but now our post action renders to the rich template I am adding other tests compatible for rich template

mridulnagpal:
I missed the fact that if it is redirecting it won't render anything hence no tests for assert_select

jywarren:
In an integration test you can do `follow_redirect!` and then select on the page you're redirected to. This is a special ability for integration tests!

mridulnagpal:
Cool then I would like to add some test for it as well just give me a minute

mridulnagpal:
@jywarren I don't understand why this isn't working I checked the link on my remote and there is an element with tag h4 and I tried for p as well but it says no p found

jywarren:
Maybe just before `assert_select "h4.visible-sm", "Ask a question"` you could use an `assert_template '...'` to ensure you're on the right template?

mridulnagpal:
It creates a user session and renders multiple templates

jywarren:
Yes, but won't we be able to tell that it's on the correct template -- the one that should contain the `h4` you're looking for? Adding that extra test should confirm or contradict this, so it's both a good test and a good debugging next-step. Thanks -- almost there!!!

mridulnagpal:
I checked in the browser using passenger start it works fine but it doesn't show the same on tests

mridulnagpal:
The page it is rendering to doesn't even have a div tag. There may be no
template in which there is no div tag

jywarren:
Yes, I think the best way to figure this out is to add `assert_template` and find what template it is actually looking at. And perhaps it's not looking at any template, and that's why we're not getting a match?

mridulnagpal:
It says rendering with user_sessions/new, layouts/_alerts, layouts/_header, layouts/_footer, layouts/application"

jywarren:
OK, great - so that means it's on the template at /app/views/user_sessions/new.html.erb -- so it's prompting us to log in. 

But that's odd because the test just before it asserts:      

`assert_redirected_to  '/questions/new?tags=question%3Aquestion&template=question&title=What&redirect=question'`

Can you push up your latest changes so I can see exactly what's being run, and what the error output is? Thanks!

mridulnagpal:
There are no changes actually

jywarren:
OK let's see how this runs.

mridulnagpal:
Yup still renders those templates

jywarren:
OK- I'm not sure, here -- perhaps `follow_redirect!` does not fetch the new templates. It may be a good idea to look at the documentation. It's odd because [this page](http://guides.rubyonrails.org/testing.html#helpers-available-for-integration-tests) seems to indicate that we can do that. 

`     assert_redirected_to  '/questions/new?tags=question%3Aquestion&template=question&title=What&redirect=question'
` should assure us that we are on the right page. 

In any case, I see that these were commented out beforehand, so let's just leave a comment by them, like:

```
# These don't pass, though they should according to http://guides.rubyonrails.org/testing.html#helpers-available-for-integration-tests
```

And just preserve the `assert_redirected_to` statement as "good enough" for now. 

Thank you for your patience, I think we are OK now. It'll be worth manually testing this once we publish it to the site, but it's also dependent on #1040 before we publish.

mridulnagpal:
Roger that!! Got to learn alot about the project by this issue. Thanks for the help :)

jywarren:
Super! Merged, and thank you again, this was a lot of work!

Issue #397, opened 2016-03-08
Wiki page locking


jywarren:
## Overview

For extremely cautious use, for around 10 or so "most visible" pages on the site and/or those related to the staff only (board, for example), we need some locking abilities. Pages like:

- https://publiclab.org/wiki/board (has been hit a lot)
- https://publiclab.org/about
- https://publiclab.org/getting-started

Admins (`user.role == "admin"`) would have ability to lock, and moderators to edit locked pages. We'd show a clear message indicating that it's locked and why, and a way to appeal this, probably by linking to now-draft https://publiclab.org/wiki/moderation. in an alert. 

**Update:** as of #1095, tag creation permissions are done and we now just have to disallow the `edit` and `update` actions on https://github.com/publiclab/plots2/blob/master/app/controllers/wiki_controller.rb -- See **Tasks**, below.

Goal is to keep locking to an absolute minimum -- only for the most trafficked and spam-hit pages. Wiki page edit notifications for anything else. (#396)

****

## Tasks

1. [x] Maybe display a lock icon, with Bootstrap tooltip explanation, where Edit button is. (#1099)
2. [x] Also, link to a listing of all locked pages, at `/tag/locked`, from an explanation page at https://publiclab.org/wiki/locked (that'll be done in #1099)
3. [x] And an entry in the power tags page to go to `/wiki/locked`: https://publiclab.org/wiki/power-tags
4. [x] Intercept requests to the `edit` page. Look to see if the "locked" tag exists, and if it does, redirects back to the wiki page with an error message, by adding just after this line: https://github.com/publiclab/plots2/blob/master/app/controllers/wiki_controller.rb#L74 as follows  (complete in #1114):

```ruby
if @node.has_power_tag('locked') && (current_user.role != "admin" && current_user.role != "moderator")
  flash.now[:warning] = "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
  redirect_to @node.path
end
```

5) [x] (closed in #1155) We'd want to add the same kind of catch to `def update` in the same controller (for when you click `save` on a wiki page), except that we'd want to put all of these lines in an `else` block, so they don't get run if there's a lock:

https://github.com/publiclab/plots2/blob/master/app/controllers/wiki_controller.rb#L137-L173

[...]

PR #1101, opened 2016-12-14
Replace all uses of wiki.slug in favor of wiki.path, fixes #736

### LINKED ISSUE #736: Replace all uses of wiki.slug in favor of wiki.path
We no longer use `node.slug` and it's redundant with the `node.path` field, so we should remove all instances from the codebase. Here they are:

```ruby
app/views/wiki/show.html.erb:  <div class="alert alert-warning"><%= raw t('wiki.show.revision_from', :time => @revision.created_at.to_s(:long)) %> <a href="/wiki/revisions/<%= @node.slug %>"><%= t('wiki.show.view_all_revisions') %></a></div>
app/views/wiki/show.html.erb:        <li><a href="/talk/<%= @node.slug %>"><i class="fa fa-comments-o"></i><span class="hidden-xs hidden-sm"> <%= t('wiki.show.talk') %></span></a></li>
app/views/wiki/show.html.erb:        <li><a href="/wiki/revisions/<%= @node.slug %>"><span class="hidden-xs"><%= @node.revisions.length %> </span><i class="fa fa-list"></i></a></li>
app/views/wiki/show.html.erb:    <% slug = @node.power_tag('parent') %>
app/views/wiki/show.html.erb:    <% parent = DrupalNode.find_by_slug(slug) %>
app/views/wiki/show.html.erb:          <p><a href="/tag/<%= @node.slug %>"><%= raw t('wiki.show.more_research', :tag => @node.slug) %> &raquo;</a></p>
app/views/wiki/edit.html.erb:      <li><a href="/wiki/<%= wiki.slug %>"><i class="fa fa-book"></i> <%= wiki.title %></a></li>
app/views/wiki/edit.html.erb:      <% if @node.path %><li><a href="/wiki/revisions/<%= @node.slug %>"><%= t('wiki.edit.revisions_for_this_page') %></a></li><% end %>
app/views/wiki/revisions.html.erb:      <li><a href="/wiki/<%= @node.slug %>"><%= t('wiki.revisions.view') %></a></li>
app/views/wiki/revisions.html.erb:      <li><a href="/wiki/edit/<%= @node.slug %>"><%= t('wiki.revisions.edit') %></a></li>
app/views/wiki/revisions.html.erb:      <li class="active"><a href="/wiki/revisions/<%= @node.slug %>">Revisions (<%= @node.revisions.length %>)</a></li>
app/views/talk/show.html.erb:<div><iframe src="https://pad.publiclab.org/p/<%= @node.slug[0,50] %>" style="width: 100%; height:800px;"><p><%= t('talk.show.try_direct_link', :url1 => 'https://pad.publiclab.org/p/'+@node.slug[0,50]) %></p></iframe></div>
app/models/drupal_node.rb:  attr_accessible :title, :uid, :status, :type, :vid, :cached_likes, :comment, :path, :slug
app/models/drupal_node.rb:  friendly_id :friendly_id_string, use: [:slugged, :history]
app/models/drupal_node.rb:    slug.blank? || title_changed?
app/models/drupal_node.rb:    self.slug[0..5] == 'place/'
app/models/drupal_node.rb:  def self.find_root_by_slug(title)
app/controllers/talk_controller.rb:    @node = DrupalNode.find_by_slug params[:id]
app/controllers/wiki_controller.rb:      @wikis = DrupalTag.find_pages(@node.slug,30) if @node.has_tag('chapter') || @node.has_tag('tabbed:wikis')
app/controllers/wiki_controller.rb:    @node = DrupalNode.find_root_by_slug(params[:id])
app/controllers/admin_controller.rb:        redirect_to "/wiki/revisions/<%= @node.slug %>" + '?_=' + Time.now.to_i.to_s
```

Note that each time, we can't just use `path` -- If a wiki has slug `hello-world`, it's supposed to have a path of `/wiki/hello-world`. So we should make a new method on `DrupalNode` like:

```ruby
def slug_from_path
  self.path.split('/').last
end
```

And then replace all of the above instances of `node.slug` with `node.slug_from_path`.

Also, `find_by_slug` and `find_root_by_slug` will have to be rewritten to search by `path` as well -- adding `/wiki/` back in, or potentially searching with a SQL wildcard like `%` in front of the `slug_from_path` result. I'm happy to offer ideas/help on that too, and it could become a separate issue for simplicity.

500swapnil:

Thanks!

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -339,3 +339,9 @@ DEPENDENCIES
   uglifier (>= 1.0.3)
   will_paginate (>= 3.0.6)
   will_paginate-bootstrap (>= 1.0.1)
+
+RUBY VERSION
+   ruby 2.1.2p95
+
+BUNDLED WITH
+   1.13.6
--- a/app/views/wiki/show.html.erb
+++ b/app/views/wiki/show.html.erb
@@ -26,7 +26,7 @@
   <h1 style="font-family: 'Junction Light';"><i class="fa fa-book"></i> <%= @revision.title %> <a onClick="$('#wiki-toolbar').toggle()" class="btn btn-link"><b class="caret"></b></a></h1>
 
 <% if @is_revision %>
-  <div class="alert alert-warning"><%= raw t('wiki.show.revision_from', :time => @revision.created_at.to_s(:long)) %> <a href="/wiki/revisions/<%= @node.slug_from_path %>"><%= t('wiki.show.view_all_revisions') %></a></div>
+  <div class="alert alert-warning"><%= raw t('wiki.show.revision_from', :time => @revision.created_at.to_s(:long)) %> <a href="/wiki/revisions/<%= @revision.node.slug_from_path %>"><%= t('wiki.show.view_all_revisions') %></a></div>
   
 <% end %>
 
@@ -44,8 +44,8 @@
         <li class="active"><a href="#"><span class="hidden-xs"><%= t('wiki.show.view') %></span><span class="visible-xs"><i class="fa fa-file-o"></i></span></a></li>
         <li><a href="<%= @node.edit_path %>?t=<%= Time.now.to_i %>"><i class="fa fa-pencil"></i><span class="hidden-xs hidden-sm"> <%= t('wiki.show.edit') %></span></a></li>
         <% if current_user && current_user.role == "admin" %><li><%= link_to "/wiki/delete/"+@node.id.to_s, :confirm => I18n.t('wiki.show.are_you_sure_delete', :path => @node.path) do %><i class="fa fa-trash"></i><span class="hidden-xs hidden-sm"> <%= t('wiki.show.delete') %></span><% end %></li><% end %>
-        <li><a href="/talk/<%= @node.slug_from_path %>"><i class="fa fa-comments-o"></i><span class="hidden-xs hidden-sm"> <%= t('wiki.show.talk') %></span></a></li>
-        <li><a href="/wiki/revisions/<%= @node.slug_from_path %>"><span class="hidden-xs"><%= @node.revisions.length %> </span><i class="fa fa-list"></i></a></li>
+        <li><a href="/talk/<%= @revision.node.slug_from_path %>"><i class="fa fa-comments-o"></i><span class="hidden-xs hidden-sm"> <%= t('wiki.show.talk') %></span></a></li>
+        <li><a href="/wiki/revisions/<%= @revision.node.slug_from_path %>"><span class="hidden-xs"><%= @node.revisions.length %> </span><i class="fa fa-list"></i></a></li>
       </ul>
     </div>
   </div>
@@ -76,7 +76,7 @@
       <% if @node.power_tags('tabbed').include?("notes") %>
         <div class="tab-pane" id="tab-notes">
           <%= render :partial => "notes/notes" %>
-          <p><a href="/tag/<%= @node.slug_from_path %>"><%= raw t('wiki.show.more_research', :tag => @node.slug_from_path) %> &raquo;</a></p>
+          <p><a href="/tag/<%= @revision.node.slug_from_path %>"><%= raw t('wiki.show.more_research', :tag => @revision.node.slug_from_path) %> &raquo;</a></p>
           <br />
         </div>
       <% end %>
--- a/app/views/wiki/show.html.erb
+++ b/app/views/wiki/show.html.erb
@@ -26,7 +26,7 @@
   <h1 style="font-family: 'Junction Light';"><i class="fa fa-book"></i> <%= @revision.title %> <a onClick="$('#wiki-toolbar').toggle()" class="btn btn-link"><b class="caret"></b></a></h1>
 
 <% if @is_revision %>
-  <div class="alert alert-warning"><%= raw t('wiki.show.revision_from', :time => @revision.created_at.to_s(:long)) %> <a href="/wiki/revisions/<%= @node.slug %>"><%= t('wiki.show.view_all_revisions') %></a></div>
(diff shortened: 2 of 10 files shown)


jywarren:
Aha - i see this is using some ERB templating code in the test:

```
Failure: Expected response to be a redirect to <"http://test.host/wiki/revisions/<%= @node.slug %>?_=1481746385"> but was a redirect to <"http://test.host/wiki/revisions/<%= @node.slug_from_path %>?_=1481746385">.
```

For that we ought to use normal Ruby string substitution, like: `"/wiki/revisions/" + @node.slug` and `"/wiki/revisions/" + @node.slug_from_path` -- so both in the test and also in the redirect itself:

https://github.com/publiclab/plots2/pull/1101/files#diff-6d0eec853a873b030e87f5a21aecdf63L150

and

https://github.com/publiclab/plots2/blob/master/test/functional/admin_controller_test.rb#L324

It looks like that code was already wrong, actually! It was not running the ERB replacements, just spitting out the whole thing as a string, see? So it's great that we caught that in our revisions here.

500swapnil:
Yeah that is great! I shall make the changes now!

500swapnil:
Done! So the @node.slug_from_path function cannot be accessed from a Ruby string and needs to be added separately. That was the problem?

jywarren:
Ah, looks like perhaps it should be `@revision.node.slug_from_path` since in that action we haven't defined `@node`. Fingers crossed that that'll get tests to pass!

500swapnil:
I defined the function slug_from_path in drupal_node.rb. So shouldnt it be @DrupalNode.slug_from_path?

jywarren:
DrupalNode is the classname -- so you can use it to make a new node like
DrupalNode.new. But once you make one, like:

node = DrupalNode.new

then you refer to it by the variable you've stored the instance in. In this
case, we didn't make a `node` or `@node` variable, but we can access the
node we want because `@revision` has a `.node` method which fetches it.

500swapnil:
Okay

jywarren:
Ah - sorry, looks like we need `@revision.drupal_node` not `.node` -- apologies!

500swapnil:
No problem. I had that doubt

500swapnil:
Same error

jywarren:
Checking now! Also, you can always ping `@publiclab/reviewers` for help if i'm not available -- no guarantee, but sometimes if I'm busy with other work and you really want to make progress, it's worth it. Thanks!

test/functional/admin_controller_test.rb
```diff
@@ -321,7 +321,7 @@ def teardown
     assert_equal 0, revision.status
     assert_equal 0, revision.author.status
     assert_not_equal node(:spam_targeted_page).latest.vid, revision.vid
-    assert_redirected_to "/wiki/revisions/<%= @node.slug %>" + '?_=' + Time.now.to_i.to_s
+    assert_redirected_to "/wiki/revisions/" + @revision.drupal_node.slug_from_path + '?_=' + Time.now.to_i.to_s
```
jywarren: OK -- while in the controller, we used an instance variable (starting with `@`) in the test, we are just using a local variable, so no `@` required. Check each changed line to ensure we're using what's available there -- either `@node` or `@revision` or `node` or `revision` -- and match the usage in that context.

app/views/wiki/show.html.erb
```diff
@@ -77,7 +76,7 @@
       <% if @node.power_tags('tabbed').include?("notes") %>
         <div class="tab-pane" id="tab-notes">
           <%= render :partial => "notes/notes" %>
-          <p><a href="/tag/<%= @node.slug %>"><%= raw t('wiki.show.more_research', :tag => @node.slug) %> &raquo;</a></p>
+          <p><a href="/tag/<%= @revision.drupal_node.slug_from_path %>"><%= raw t('wiki.show.more_research', :tag => @revision.drupal_node.slug_from_path) %> &raquo;</a></p>
```
jywarren: Yes -- see how here, we can indeed use `@node` -- it was only in some contexts where `@revision` was required, not all. Thanks!

500swapnil: I went through the errors. @node is fine for all except admin_controller.rb and admin_controller_test.rb.
What should I use in admin_controller.rb and admin_controller_test.rb??

500swapnil: Is there any way of testing a part of code and not the whole code locally? Some command instead of rake test:all .

jywarren: On this line in the admin controller test, `revision.drupal_node`:

https://github.com/publiclab/plots2/blob/master/test/functional/admin_controller_test.rb#L324

And on this line, `@revision.drupal_node`:

https://github.com/publiclab/plots2/blob/master/app/controllers/admin_controller.rb#L150

Yes, you can run just functional tests with `rake test:functionals`, units with `rake test:units`, integration with `rake test:integration`, and you can even run just one test file with:

`ruby -I test test/functional/notes_controller_test.rb` (for example)

I know it's time consuming to run them all!

500swapnil: Okay I will work on it now!

500swapnil:
Fine now?

jywarren:
I'll take a look but I have to head out now -- later tonight I hope! Thanks!

500swapnil:
Okay I will have a look at other issues till then!

Gemfile.lock
```diff
@@ -339,3 +339,9 @@ DEPENDENCIES
   uglifier (>= 1.0.3)
   will_paginate (>= 3.0.6)
   will_paginate-bootstrap (>= 1.0.1)
+
+RUBY VERSION
```
jywarren: I'd like to not include Gemfile or Gemfile.lock changes if not essential, but I'll leave these in this time. Thanks!

jywarren:
Looks great! merging now. Thanks for your careful work on this one!

500swapnil:
Thank you! Any idea on which issue I should try next?

jywarren:
Yes -- there are some sub-issues to this one that'd be great to have some assistance on! https://github.com/publiclab/plots2/issues/397

For example, the checks on the `edit` and `update` actions, and the corresponding tests!

PR #1114, opened 2016-12-16
Added test to block basic user edits, closes issue #397

500swapnil:

Thanks!

### DIFF
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -72,7 +72,7 @@ class WikiController < ApplicationController
     else
       @node = DrupalNode.find_wiki(params[:id])
     end
-    if @node.has_power_tag('locked') && (current_user.role != "admin" && current_user.role != "moderator")
+    if @node.has_tag('locked') && (current_user.role != "admin" && current_user.role != "moderator")
       flash[:warning] = "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
       redirect_to @node.path
     end
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -7,9 +7,9 @@ class WikiController < ApplicationController
   def subdomain
     url = "//#{request.host}/wiki/"
     case request.subdomain
-    when "new-york-city", 
-         "gulf-coast", 
-         "boston", 
+    when "new-york-city",
+         "gulf-coast",
+         "boston",
          "espana" then
       redirect_to url+request.subdomain
     when "nyc"
@@ -71,6 +71,10 @@ class WikiController < ApplicationController
       @node = DrupalNode.find_wiki(params[:lang]+"/"+params[:id])
     else
       @node = DrupalNode.find_wiki(params[:id])
+    end
+    if @node.has_power_tag('locked') && (current_user.role != "admin" && current_user.role != "moderator")
+      flash.now[:warning] = "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
+      redirect_to @node.path
     end
     if ((Time.now.to_i - @node.latest.timestamp) < 5.minutes.to_i) && @node.latest.author.uid != current_user.uid
       flash.now[:warning] = I18n.t('wiki_controller.someone_clicked_edit_5_minutes_ago')
@@ -146,7 +150,7 @@ class WikiController < ApplicationController
         # update vid (version id) of main image
         if @node.drupal_main_image && params[:main_image].nil?
           i = @node.drupal_main_image
-          i.vid = @revision.vid 
+          i.vid = @revision.vid
           i.save
         end
         @node.title = @revision.title
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -73,7 +73,7 @@ class WikiController < ApplicationController
       @node = DrupalNode.find_wiki(params[:id])
     end
     if @node.has_power_tag('locked') && (current_user.role != "admin" && current_user.role != "moderator")
-      flash.now[:warning] = "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
+      flash[:warning] = "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
       redirect_to @node.path
     end
     if ((Time.now.to_i - @node.latest.timestamp) < 5.minutes.to_i) && @node.latest.author.uid != current_user.uid
--- a/test/functional/wiki_controller_test.rb
+++ b/test/functional/wiki_controller_test.rb
@@ -139,7 +139,7 @@ class WikiControllerTest < ActionController::TestCase
       post :edit,
           id: 'organizers'
     end
-    assert_template "wiki/show"
+    assert_equal flash[:warning] , "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
     assert_redirected_to node.path
   end
 
--- a/test/functional/wiki_controller_test.rb
+++ b/test/functional/wiki_controller_test.rb
@@ -139,7 +139,6 @@ class WikiControllerTest < ActionController::TestCase
       post :edit,
           id: 'organizers'
     end
-    assert_redirected_to node.path
   end
 
 
--- a/test/functional/wiki_controller_test.rb
+++ b/test/functional/wiki_controller_test.rb
@@ -141,7 +141,6 @@ class WikiControllerTest < ActionController::TestCase
           uid:   rusers(:bob).id,
           title: ""
     end
-
     assert_template "wiki/show"
     assert_select ".alert", "flash.now[:warning] = 'This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it.'"
     assert_redirected_to wiki.path
--- a/test/functional/wiki_controller_test.rb
+++ b/test/functional/wiki_controller_test.rb
@@ -139,7 +139,6 @@ class WikiControllerTest < ActionController::TestCase
       post :edit,
           id: 'organizers'
     end
-    assert_equal flash[:warning] , "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
     assert_redirected_to node.path
   end
 
--- a/test/functional/wiki_controller_test.rb
+++ b/test/functional/wiki_controller_test.rb
@@ -141,9 +141,9 @@ class WikiControllerTest < ActionController::TestCase
           uid:   rusers(:bob).id,
           title: ""
     end
-    
+
     assert_template "wiki/show"
-    assert_select ".alert", "flash.now[:warning] = "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it.""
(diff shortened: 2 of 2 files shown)


500swapnil:
It failed the tests!

jywarren:
All right -- that's good, but check why it failed (click "Details" on the test check box) -- was it that it didn't show the right message?

` +    assert_select ".alert", "expected message"`

We should put a message there as outlined in #397 --

```ruby
  flash.now[:warning] = "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
```

Then, try to implement the feature itself -- as described in #397 --

> Look to see if the "locked" tag exists, and if it does, redirects back to the wiki page with an error message, by adding just after this line: https://github.com/publiclab/plots2/blob/master/app/controllers/wiki_controller.rb#L74 as follows:

```ruby
if @node.has_power_tag('locked') && (current_user.role != "admin" && current_user.role != "moderator")
  flash.now[:warning] = "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
  redirect_to @node.path
end
```

Then the test should pass (if it's written properly! I might've made errors...) because it'll match the right message. 

You could also enclose it in a block to make sure that the total # of `DrupalNodeRevisions` doesn't change, like:

```ruby
assert_difference 'DrupalNodeRevision.count', 0 do
  # put your POST call inside here
end
```

That's a nice type of test -- it'll ensure that within the block, the number of revisions cannot change, or the test will fail.

500swapnil:
I understood the working of the feature and the tests as well. I messed up the blocks. I will try to fix it now

500swapnil:
Still failing tests! Different error now

test/functional/wiki_controller_test.rb
```diff
@@ -132,11 +132,26 @@ def teardown
     assert_equal flash[:notice], "Edits saved."
   end
 
+  test "basic user blocked from updating a locked wiki page" do
+    node(:organizers).add_tag('locked', rusers(:admin)) # lock the page with a tag
+    # then try updating it
+    assert_difference 'DrupalNodeRevision.count', 0 do
+      post :update,
```
jywarren: Oh, note that your changes were to the `:edit` action, but your test is for the `:update` action. Both need to be changed and tested, but I think you're changing one but testing the other in this step.

500swapnil: Thank you! You are right! And the entire post block comes under assert_difference right?

test/functional/wiki_controller_test.rb
```diff
@@ -132,11 +132,26 @@ def teardown
     assert_equal flash[:notice], "Edits saved."
   end
 
+  test "basic user blocked from editing a locked wiki page" do
+    node(:organizers).add_tag('locked', rusers(:admin)) # lock the page with a tag
+    # then try editing it
+    assert_difference 'DrupalNodeRevision.count', 0 do
+      post :edit,
+          id:  node(:organizers).id,
+          uid:   rusers(:bob).id,
```
jywarren: These two lines shouldn't be needed in the edit action.

test/functional/wiki_controller_test.rb
```diff
@@ -108,21 +108,21 @@ def teardown
 
   test "viewing edit wiki page" do
 
-    get :edit, 
+    get :edit,
          id: 'organizers'
```
jywarren: Also look how the id for the edit action should be the page slug, not the numeric id! That should solve your last issue.

test/functional/wiki_controller_test.rb
500swapnil: So no id and uid? Will that work?

500swapnil:
@publiclab/reviewers Can anyone tell why I am getting this error?

500swapnil:
@ryzokuken Can you tell why I am getting this error and how to fix this?

ryzokuken:
I'll check the Travis log.

ryzokuken:
You failed the test you created yourself, as there were no elements in the page as `.alert` while the test requires one such element to be present. I suppose this is because you set `flash[:warning]` text but did not create the alert.

500swapnil:
Okay so I will remove the '.alert' test. But how do I test this?
  flash.now[:warning] = "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."

jywarren:
But wait, don't we want the test to identify the alert you've written? Why
isn't it detecting the notice?

500swapnil:
I dont know why its showing an error! How do I detect a 'flash.now[:warning]'?

app/controllers/wiki_controller.rb
```diff
@@ -72,6 +72,10 @@ def edit
     else
       @node = DrupalNode.find_wiki(params[:id])
     end
+    if @node.has_power_tag('locked') && (current_user.role != "admin" && current_user.role != "moderator")
+      flash.now[:warning] = "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
```
jywarren: Ah, i think the problem is here -- a `flash.now` runs **before** the redirect, but a `flash[]` runs after. If you remove the "now" it should work and pass the test!

500swapnil: Okay

test/functional/wiki_controller_test.rb
```diff
@@ -132,11 +132,24 @@ def teardown
...
 
+  test "basic user blocked from editing a locked wiki page" do
+    node(:organizers).add_tag('locked', rusers(:admin)) # lock the page with a tag
+    # then try editing it
+    assert_difference 'DrupalNodeRevision.count', 0 do
+      post :edit,
+          id: 'organizers'
+    end
+    assert_template "wiki/edit"
+    assert_select ".alert", "flash.now[:warning] = 'This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it.'"
```
500swapnil: Done it!

500swapnil:
The error is that there is no ".alert" message. We need to use a different method for detecting flash[:warning] any ideas? Can we use assert_equal?

jywarren:
Hmm, odd. You can also test the flash like this:

https://github.com/publiclab/plots2/pull/1114/files#diff-afcc4c4b1665a46e6c88e913be20b76dR132

But please add it above the existing flash test so we test both ways. Thanks!

500swapnil:
Okay so the statement would be like:
assert_equal flash[:warning] , "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
right?

jywarren:
Yes, that's right! Sorry for the slow response, we're still in a holiday here but I'll be more responsive now. Thanks!

jywarren:
Well, it's still failing. I think we have to figure out if all these conditions are actually passing in the test:

`if @node.has_power_tag('locked') && (current_user.role != "admin" && current_user.role != "moderator")`

Is there any way they wouldn't be? Could we echo out each one, like:

```
puts @node.has_power_tag('locked')
puts current_user.role != "admin"
puts current_user.role != "moderator"
```

in the controller code and see what we get?

app/controllers/wiki_controller.rb
```diff
@@ -72,6 +72,10 @@ def edit
     else
       @node = DrupalNode.find_wiki(params[:id])
     end
+    if @node.has_power_tag('locked') && (current_user.role != "admin" && current_user.role != "moderator")
```
jywarren: Ah! Here's the problem! `has_power_tag` looks for a tag with the given key, and any value, in the format `key:value` -- but the `locked` tag is not a `key:value` format tag. Here we should instead just use `@node.has_tag('locked')`. That should fix it!

jywarren:
@500swapnil -- the change i just proposed should work. But after this PR, you really should begin using named feature branches -- see [this guide](https://publiclab.org/wiki/contributing-to-public-lab-software#A+sample+git+workflow) for how to do this. That way your PRs won't come from your master branch, but from a branch named, for example, `edit-locking`.

500swapnil:
I made the required change and yes, I will work using branches after this PR!

jywarren:
Hmm, OK, can we assert template after a redirect? Maybe not -- I guess best remove that line.

jywarren:
See, because in functional tests we can only access the most recent request. Tests with a sequence of multiple requests really belong in integration tests, which have their own abilities and features. 

Thanks!!

jywarren:
And perhaps similarly the asserr_select can't be done without loading the redirect.

jywarren:
Hmm, it's only passing now because you removed all the assertions -- but we need to preserve some -- was the `assert flash` not working? Ah - i see the tests failed, but do check them by clicking the "X" link -- 961812c only failed due to a stalled test so I restarted it. You can tell this when it stalls at 1:30 instead of 7-8 minutes, and you can restart the tests manually by closing and re-opening the PR. It's a bit of a pain, but sometimes it gets stuck downloading remote dependencies. 

Let's see if 961812c passes, and if not, why!

jywarren:
OK -- got the test in 961812c to run -- https://travis-ci.org/publiclab/plots2/builds/187231729

```
Failure: expecting <"wiki/show"> but rendering with <"">
test_basic_user_blocked_from_editing_a_locked_wiki_page(WikiControllerTest)
test/functional/wiki_controller_test.rb:142:in `block in <class:WikiControllerTest>'
```

I believe it's catching on this line: https://github.com/publiclab/plots2/pull/1114/files#diff-afcc4c4b1665a46e6c88e913be20b76dR154

This makes sense -- and I think it means you can go back to the state you were in at 961812c -- because I believe this is showing that it's not rendering a template at that point -- it's redirecting since the basic user shouldn't be able to post an update -- the "organizers" page is locked. Does that help?

Thanks for your persistence on this -- the restarting-tests thing is annoying, but you're getting closer to completion here.

test/functional/wiki_controller_test.rb
```diff
@@ -132,11 +132,23 @@ def teardown
...
 
+  test "basic user blocked from editing a locked wiki page" do
+    node(:organizers).add_tag('locked', rusers(:admin)) # lock the page with a tag
+    # then try editing it
+    assert_difference 'DrupalNodeRevision.count', 0 do
+      post :edit,
+          id: 'organizers'
+    end
+    assert_equal flash[:warning] , "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can edit it."
+    assert_redirected_to node.path
```
jywarren: Ah - should be the last thing here -- this should be `node(:organizers).path` since we haven't made a variable called `node` in this test. Fingers crossed that this is the last thing!

500swapnil:
I made the required changes and yes it works now! Thank you. Can you explain in brief the syntax of this line? I am trying to understand the working of tests.
node(:organizers).add_tag('locked', rusers(:admin))

jywarren:
Great! Tests stalled but I restarted them; we'll see.

Yes, `node(:organizers).add_tag('locked', rusers(:admin))` is a bit obscure. Let's see:

`node(:organizers)` is accessing the test fixture data for nodes, in `/test/fixtures/node.yml`. It's fetching the "organizers" entry.

`node.add_tag('locked', rusers(:admin))` is using a method from `/app/models/drupal_node.rb` called `add_tag()` which requires a tagname and a user.

`rusers(:admin)` is accessing the `rusers.yml` fixture, so getting the test "admin" user. 

So that whole line is adding a tag to the test fixture data, but in such a way that it doesn't require a change to the fixture files themselves - it's just happening within the test itself. We might add a comment there to explain that line if you like, but it's not required. 

OK -- waiting for those tests!

jywarren:
Passed! Perfect. Merging! We actually also need similar tests and limits for the `update` action. They'd be virtually the same as the changes you've made here -- interested in giving it a try?

500swapnil:
Yeah I will try fixing that too! Thanks!

PR #1147, opened 2016-12-28
JSON renderer. Fixes #1144

### LINKED ISSUE #1144: add additional fields to JSON API calls for wiki pages
@devinbalkind requested as a followup to #1069, the following fields for wiki pages:

* thumbnail image link
* full path (i.e. hyperlinkable)
* first 100 character description
* tags

These should be pretty easy; the JSON is rendered here:

https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L54

We could break that up into multiple lines, something like this:

```ruby
      format.json  {
        nodes.each do |node|
          node.path = "https://" + request.host.to_s + node.path
          node['preview'] = node.body_preview
          node['image'] = node.main_image_url if node.main_image
          node['tags'] = node.tags.collect(&:name) if node.tags
        end
        render json: nodes
      }
```

This may need to be tested, with an expansion of this test: 

https://github.com/publiclab/plots2/blob/master/test/functional/tag_controller_test.rb#L150

Pk13055:
#1144 

Thanks!

### DIFF
--- a/test/functional/tag_controller_test.rb
+++ b/test/functional/tag_controller_test.rb
@@ -122,16 +122,17 @@ class TagControllerTest < ActionController::TestCase
 
   test "tag show" do
 
-    get :show, id: tags(:spectrometer)
+    get :show, id: tags(:spectrometer).name
 
     assert :success
     assert_not_nil :tags
 
     assert_equal tags(:spectrometer).parent, 'spectrometry'
     # iterate through results
+    assert assigns['notes'].length > 0
     assigns['notes'].each do |node|
       assert node.has_tag('spectrometry') # should return false
-      assert node.has_tag_without_aliasing('spectrometry') # should return false
+      assert_false node.has_tag_without_aliasing('spectrometry') # should return false
     end
 
     #assert_equal assigns['tags'].length, 1
@@ -140,7 +141,7 @@ class TagControllerTest < ActionController::TestCase
 
   test "tag show JSON" do
 
-    get :show, id: tags(:spectrometer), format: 'json'
+    get :show, id: tags(:spectrometer).name, format: 'json'
 
     assert :success
     assert_not_nil :tags
@@ -148,12 +149,12 @@ class TagControllerTest < ActionController::TestCase
     json = ActiveSupport::JSON.decode(@response.body)
 
     assert_not_nil json
-    assert_equal tags(:spectrometer).nodes.first.body_preview,
-    json.first[:preview]
-    assert_equal tags(:spectrometer).nodes.first.main_image_url,
-    json.first[:image] 
-    assert_equal tags(:spectrometer).nodes.first.tags.collect(&:name),
-    json.first[:tags]
+    assert assigns['notes'].length > 0
+    node = DrupalNode.find tags(:spectrometer).nodes.first.nid
+    assert_equal node.nid,                  json.first['nid']
+    assert_equal node.body_preview,         json.first['preview']
+    assert_equal node.main_image,           json.first['image'] # this won't check anything bc there is no main image
+    assert_equal node.tags.collect(&:name), json.first['tags']
   end
   
   test "wildcard tag show" do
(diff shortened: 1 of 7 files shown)


jywarren:
Hi! Please check out my suggestion for the tests here:

https://github.com/publiclab/plots2/issues/1144#issuecomment-269512845

A bit more specificity in the tests and we should be good!

test/functional/tag_controller_test.rb
```diff
@@ -148,9 +148,12 @@ def setup
...
 
     assert_not_nil json
-    assert :preview
-    assert :image
-      assert :tags
+    assert_equal tags(:spectrometer).notes.first.body_preview,
+    json.first[:preview]
+    assert_equal tags(:spectrometer).notes.first.main_image_url,
+    json.first[:image] 
+    assert_equal tags(:spectrometer).notes.first.tags,
```
jywarren: `tags(:spectrometer).notes.first.tags` will return an array of tags objects -- so we need to do `tags(:spectrometer).notes.first.tags.collect(&:name)` and it'll collect the tags' names. See if that works. 

Also - i noticed the tests ran in 1m 30 seconds, which means they stalled -- this happens sometimes -- they should take about 7-8 minutes. To restart them, you can close and reopen the pull request and they'll trigger again.

Pk13055:
It says there is no method `notes`:
`
Error: test_tag_show_JSON(TagControllerTest): NoMethodError: undefined method 'notes' for #<DrupalTag:0x005604ae44fcd8>
test/functional/tag_controller_test.rb:151:in block in <class:TagControllerTest>' 
`

jywarren:
Hmm, maybe it is "nodes"? But I believe that would return both wiki pages
and notes. Give it a try and if it returns more than we want, we can limit
by adding a .where(type: "note") after "nodes".

jywarren:
Indeed -- it's `tag.nodes` -- see here: https://github.com/publiclab/plots2/blob/master/app/models/drupal_tag.rb#L48

And now I see the error:

```
Error: test_tag_show_JSON(TagControllerTest): NoMethodError: undefined method `[]' for nil:NilClass
test/functional/tag_controller_test.rb:152:in `block in <class:TagControllerTest>'
```

I'll leave an inline suggestion.

test/functional/tag_controller_test.rb
```diff
@@ -148,7 +148,12 @@ def setup
     json = ActiveSupport::JSON.decode(@response.body)
 
     assert_not_nil json
-    # assert more things to check this
+    assert_equal tags(:spectrometer).nodes.first.body_preview,
+    json.first[:preview]
+    assert_equal tags(:spectrometer).nodes.first.main_image_url,
+    json.first[:image]
```
jywarren: OK - so we need to know how this is actually structured, so we can properly test it. I guess it's not just an array, so `.first` doesn't return anything. 

To debug, i added:

```ruby
puts json.inspect
``` 

on line 151. It shows an empty array. So `[].first` = `nil`. So... I added `puts @response.body.inspect` to see what happened.

...

It also returned "[]"! 

OK... I'm going to check that we're actually getting any nodes back at all.

jywarren: All right - there was a key error in the earlier test, and I managed to fix it -- unrelated to your work! Apologies! But it's great that we caught this. I'll push a new commit showing what happened and a fix.

jywarren:
So, there's actually a kind of obscure reason this is difficult -- Rails doesn't like it when we just make up new attributes for a model object, like a `node`. So it's showing warnings. But I think we can ignore them for now... there's a workaround, where we can make a new object -- a hash, or associative array -- for each item, and fill it with whatever we want, and render that to json. But let's just not worry about it for now -- it works as-is, and you've been very patient and persistent! 

Thank you! Uploading fixes shortly.

jywarren:
Argh - I tried a couple things and there was just too much going on to make the quick fix worthwhile -- so I did the workaround I was mentioning before -- I created a new array called `json` and filled it with everything from each node, and passed that new array of newly made hybrid objects to the JSON renderer. That worked! 

Sorry, this ended up being a more complex issue than I'd realized! I appreciate your putting time into it!

jywarren:
OK -- you can see the changes I made here: https://github.com/publiclab/plots2/pull/1147/commits/0626738c1a4fc889280dbea0779eab652ac227d1

Hope that makes some sense! Fingers X that it passes.

Pk13055:
@jywarren I took a look; correct me if I'm wrong but you created a JSON object which is asserted instead of the node directly? It's still failing the check though..

Pk13055:
@jywarren Changing the line to assert instead would solve the issue, but is that what we want?

jywarren:
Hmm, it is odd -- I think it's related to a very mysterious set of test failures that I shouldn't have messed with in this PR -- but I reverted that one line to how it was before, and it passes now. Looks good -- merging, and thank you for your patience!!!

Pk13055:
@jywarren Are you going to open an issue?

jywarren:
I believe I added a commit to your PR which resolved the issue. Thanks!

PR #1155, opened 2017-01-01
Added test to block normal user from updating wiki, fixes #397

### LINKED ISSUE #397: Wiki page locking

500swapnil:

Thanks!

### DIFF
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -143,7 +143,11 @@ class WikiController < ApplicationController
       title: params[:title],
       body:  params[:body]
     })
-    if @revision.valid?
+    if @node.has_tag('locked') && (current_user.role != "admin" && current_user.role != "moderator")
+      flash[:warning] = "This page is <a href='/wiki/power-tags#Locking'>locked</a>, and only <a href='/wiki/moderators'>moderators</a> can update it."
+      redirect_to @node.path
+    
+    elsif @revision.valid?
       ActiveRecord::Base.transaction do
         @revision.save
         @node.vid = @revision.vid
--- a/test/functional/wiki_controller_test.rb
+++ b/test/functional/wiki_controller_test.rb
@@ -143,7 +143,7 @@ class WikiControllerTest < ActionController::TestCase
     assert_redirected_to node(:organizers).path
   end
 
-  test "basic user blocked from update a locked wiki page" do
+  test "basic user blocked from updating a locked wiki page" do
     node(:organizers).add_tag('locked', rusers(:admin)) # lock the page with a tag
     # then try editing it
     assert_difference 'DrupalNodeRevision.count', 0 do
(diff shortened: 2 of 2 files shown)


jywarren:
Looks perfect. Great job!

PR #1174, opened 2017-01-06
Fixes #775: Separate display of power tags

### LINKED ISSUE #775: Separate display of power tags from normal tags

****

[Power tags](https://publiclab.org/wiki/power-tags) (in `key:value` format) and regular tags are mixed in display: 

https://publiclab.org/notes/liz/08-30-2016/check-out-these-activity-grids#tags

We should show the power tags separately, perhaps show them in purple or grey, and also link the words "Power tags" in the label to the above explanation of them. 

This may also be where we could show an expanded interface for adding powertags, as in #496 -- and should in any case link to power tags documentation at https://publiclab.org/wiki/power-tags

Tags are displayed in this template: https://github.com/publiclab/plots2/blob/master/app/views/tag/_tagging.html.erb#L5

`@node.community_tags`, used there, is defined here -- it's just an alias to an ActiveRecord relation: https://github.com/publiclab/plots2/blob/master/app/models/drupal_node.rb#L436

We should use the more specific `@node.power_tag_objects` and `@node.normal_tags` to separate these out ([see here for docs](https://github.com/publiclab/plots2/blob/master/app/models/drupal_node.rb#L385-L399)).

If we can think of a more compact way to display these, that'd be great, too!

gaeun:

Thanks!

- Separated display of power tags from normal tags
- Preview: 
screen shot 2017-01-06 at 8 50 00 pm

### DIFF
--- a/app/views/tag/_tagging.html.erb
+++ b/app/views/tag/_tagging.html.erb
@@ -1,17 +1,9 @@
 <%= render partial: 'tag/replication' %>
 
-<p>
-  <span id="tags">
-    <% @node.community_tags.each do |tag| %>
-      <span id="tag_<%= tag.tid %>" class="label label-primary" data-toggle="tooltip" data-html="true" title="created by <strong><%= tag.try(:author).try(:username) %></strong> <%= time_ago_in_words(Time.at(tag.date)) %> ago">
-        <a class="tag-name" href="<%= "/maps" if @node.type == "map" %>/tag/<%= tag.name %>"><%= tag.name %></a>
-        <% if current_user && (tag.uid == @node.uid || current_user.role == "admin" || current_user.role == "moderator") %>
-          <a data-confirm="This is a power tag (see https://publiclab.org/wiki/power-tags) -- and may drive a specific function on this page. Are you sure you want to delete it?" class="tag-delete" data-remote="true" href="/tag/delete/<%= @node.id %>/<%= tag.tid %>" data-tag-id="<%= tag.tid %>">x</a>
-        <% end %>
-      </span>
-    <% end %>
-  </span>
-</p>
+<span id="tags">
+<%= render partial: 'tag/tags', locals: { power_tag: true, label_name: 'label-default' } %>
+<%= render partial: 'tag/tags', locals: { power_tag: false, label_name: 'label-primary' } %>
+</span>
 
 <script>
 $(function () {
@@ -29,9 +21,7 @@ $(function () {
     </div>
     <script>
       jQuery(document).ready(function() {
-
         initTagForm(<%= @node.id %>, '#tagform');
-
       });
     </script>
   </div>
--- a/app/views/tag/_tags.html.erb
+++ b/app/views/tag/_tags.html.erb
@@ -1,0 +1,16 @@
+<p>
+  <% @node.community_tags.each do |tag| %>
+    <% if power_tag^tag.name.include?(':') %>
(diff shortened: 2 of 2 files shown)


app/views/tag/_tagging.html.erb
```diff
@@ -3,12 +3,21 @@
...
   <span id="tags">
     <% @node.community_tags.each do |tag| %>
-      <span id="tag_<%= tag.tid %>" class="label label-primary" data-toggle="tooltip" data-html="true" title="created by <strong><%= tag.try(:author).try(:username) %></strong> <%= time_ago_in_words(Time.at(tag.date)) %> ago">
-        <a class="tag-name" href="<%= "/maps" if @node.type == "map" %>/tag/<%= tag.name %>"><%= tag.name %></a>
-        <% if current_user && (tag.uid == @node.uid || current_user.role == "admin" || current_user.role == "moderator") %>
-          <a data-confirm="This is a power tag (see https://publiclab.org/wiki/power-tags) -- and may drive a specific function on this page. Are you sure you want to delete it?" class="tag-delete" data-remote="true" href="/tag/delete/<%= @node.id %>/<%= tag.tid %>" data-tag-id="<%= tag.tid %>">x</a>
-        <% end %>
-      </span>
+      <% if tag.name.include? ':' %>
+        <span id="tag_<%= tag.tid %>" class="label label-default" data-toggle="tooltip" data-html="true" title="created by <strong><%= tag.try(:author).try(:username) %></strong> <%= time_ago_in_words(Time.at(tag.date)) %> ago">
```
ananyo2012: @gaeun You don't have to put a `if-else` block for the entire block just to use a different class. Just putting a conditional operator in the class itself would do. Something like
```ruby
<span class="label <%= tag.name.include?(':') ? 'label-default' : 'label-primary' %>">
```

ananyo2012:
@jywarren Does it look good now? Do you want to have any extra tooltip text?

jywarren:
How about breaking them up onto two lines, would that be possible?

ananyo2012:
@jywarren You mean displaying power tags and normal tags in separate lines?

jywarren:
For now, just putting each in a separate <p> would be fine, yes!

gaeun:
@jywarren @ananyo2012 Done separating :)

gaeun:
@jywarren @ananyo2012 Separated the tags! Preview:

screen shot 2017-01-07 at 8 38 54 pm

ananyo2012:
@gaeun You are in the right track. But there is repetition of same code. Need to make it DRY. Put these lines in a separate partial. And then render this partial instead of writing over the lines again. Create a partial named suppose `tags` and include it like
```ruby
<%= render partial: 'tag/tags', locals: { labe_name: 'label-default' } %>
```
Keep the `label-default` or `label-primary` class in a local variable `label_name` suppose like I did above and put that variable in the partial like
```ruby
<span class="label <%= label_name %>">
```
You can learn more about partials [here](http://guides.rubyonrails.org/layouts_and_rendering.html#using-partials)

app/views/tag/_tags.html.erb
```diff
@@ -0,0 +1,18 @@
+<p>
+  <span id="tags">
```
ananyo2012: @gaeun Since id should be unique keep these outer span tags in the `tagging` partial surrounding the render tags.

gaeun: @ananyo2012 Fixed this

ananyo2012: @gaeun I think you didn't quite get it. `id` should be unique in a page. So wrap both the partial with a single `<span>` tags

gaeun: @ananyo2012 Ah, I see what you mean. I fixed it now.

ananyo2012:
Perfect this looks great! @jywarren This can be merged now!

jywarren:
Oh, super! Great job, both of you!

PR #1180, opened 2017-01-11
Counter fixes

### LINKED ISSUE #1166: page view count

It looks like the view count on the public lab wikis pages aren't working right now. Hard to tell if the same is going on for the Research Notes, I've seen some with really low counts that should be high (such as this one https://publiclab.org/notes/lenagd/12-20-2016/citizen-science-investigations-aka-common-legal-issues-when-using-community-sourced-data), and some that look ok (such as this one: https://publiclab.org/notes/sanchittechnogeek/01-03-2017/translating-project-files-using-transifex) maybe the count was just reset? Would be great to get them back if we can!

jywarren:
Fixes #1166 -- once it's thoroughly checked!

This will also require a followup PR which removes the `DrupalNodeCounter` model and drops its table in the database. 

This will need the following script run from a CSV of legacy data, in the rails console on production (but tested first on staging!):

```
CSV.foreach('counts.csv') do |row|
  DrupalNode.find(row[0]).update_attributes('legacy_views': row[1])
end
```


jywarren:
@ananyo2012 or @yukiisbored  - would you be able to look over my last commit here? I got it all working, I thought, but then added these two test additions in 664bbdd to confirm that `node.views` is incremented on wiki#show and nodes#show actions. But it doesn't seem to do so -- are the tests creating a unique impression? Am I doing it wrong?

@ashleypt - just a heads up, I figure out the test inconsistency issue we'd struggled with -- finally! 50727cb contains the fix, and it was because tests run so fast that the returned records had the same timestamp (they were created within a second of one another) and so the ordering wasn't deterministic; sometimes (essentially randomly) we got one **just** in the next timestamp second, and so the order would be reversed -- but only once in a while! I left a comment in the test, explaining it. Such a subtle issue, but certainly a comfort to have resolved!

ashleypt:
cool! glad to know it's finally fixed!

ananyo2012:
@jywarren I looked at the changes you made. You didn't extend impressionist over the `drupal_node` model. I think that might cause the problem. You need to make the `drupal_node` model impressionable and add the counter cache column as [explained here](https://github.com/charlotte-ruby/impressionist#adding-a-counter-cache).
For problems related to sort order in timestamps the method I use is assign a manual timestamp for each record like I did in the [comments.yml](https://github.com/publiclab/plots2/blob/master/test/fixtures/comments.yml#L8) fixture.

jywarren:
Hi, Ananyo, I thought I had on line 65 of drupal_node.rb... But did I do it
wrong?

ananyo2012:
Ok sorry I might have missed it. I think its because it's logging an impression with the same ip address. Can you try by creating some dummy data in a `impressions.yml` fixture [as done here](https://github.com/charlotte-ruby/impressionist/blob/master/tests/test_app/spec/fixtures/impressions.yml) with some ip other than `127.0.0.1`

jywarren:
sure, but do you know if we can set the IP address in the test? I'm trying
a few methods to do this now, but not seeing much success.

I realize there are probably already impressions made from earlier tests,
which is why this is not a unique one. There seems also not to be a way to
clear the impressions besides clearing the Impression table. I just pushed
a new commit which does that, but the tests still don't pass for some
reason.

jywarren:
Here's the exported `counts.csv`:

[counts.csv.zip](https://github.com/publiclab/plots2/files/712601/counts.csv.zip)

It's in format: `nid,totalcount,daycount,timestamp` -- and we only need the first two of those -- the script at top should work.

jywarren:
@ananyo2012 - I worked a while on this this morning but wasn't able to get `@node.views` to increment. I'm not sure what we should do -- if we don't use the counter cache, and just route `@node.views` to `@node.impressionist_count(:filter => :ip_address)`, we can skip the caching, but still get unique views based on IP address. Any other ideas?

jywarren:
I'm going to go ahead with this interim fix to not use counter caching. We can open an issue afterwards, and I'll leave the lines in there, commented!

jywarren:
https://github.com/publiclab/plots2/issues/1196 has docs on that persistent issue and workaround.

jywarren:
Whew, finally.

jywarren:
Completed! Script run and numbers recovered.

jywarren:
Completed recovery as well!

jywarren:
DrupalNodeCounter table is now gone! :-)

PR #1188, opened 2017-01-17
Embeddable activities grids for including on external sites, closes #830

### LINKED ISSUE #830: Embeddable activities grids for including on external sites

This could be broken up if it's too complex!

****

When showing PL projects on other sites, we could make it easier to display actionable lists of activities using an `<iframe>`. It could lead to:

`/embed/<tagname>` 

This would involve:

* a new route to the URL above in https://github.com/publiclab/plots2/blob/master/config/routes.rb, like `match 'embed/:tagname' => 'notes#embed`
* a new controller action for `def embed` in https://github.com/publiclab/plots2/blob/master/app/controllers/notes_controller.rb -- which contains: `render layout: false` 
* a new template at `/app/views/notes/embed.html.erb` with just an inline grid, so: 

```ruby
<%= insert_extras('[' + params[:tagname] + ']') %>
```

on this line: https://github.com/publiclab/plots2/blob/master/app/helpers/application_helper.rb#L31

Once this is all working, we could make a new issue for featuring the embeddable grid, perhaps in a dropdown somewhere like on the tag listing page...

ma8sa:

Thanks!


app/controllers/tag_controller.rb
```diff
@@ -279,4 +279,9 @@ def remove_all_tags
   def location
     render template: 'locations/_form'
   end
+
+  def embed
+      render layout: false
```
jywarren: Good start! Here, if we want to use the "gridsEmbed.html.erb" template, we'll have to specify it with `template: "gridsEmbed"` or name the action to match the template file: `def gridEmbed` (would be great for it to be singular, as it makes a bit more sense -- you'll have to change the template file too).

config/routes.rb
```diff
@@ -309,4 +309,6 @@
 
   match ':controller(/:action(/:id))(.:format)'
 
+  match 'embed/:tagname' => 'notes#embed
```
jywarren: and if you change it to `gridEmbed`, this'll have to change too, although perhaps we could make the URL something like `grid/:tagname` for simplicity?

app/views/tag/gridsEmbed.html.erb
```diff
@@ -0,0 +1 @@
+
```
jywarren: Great, here in this file, you can add this line:

```ruby
      <%= raw auto_link(insert_extras("[notes:%{params[:tagname]}]"), :sanitize => false) %>
```

See how I've put in the text `[notes:_______]` to be substituted in with the `params[:tagname]` variable which you set up in your routes? This will trigger the display of a table of tag related content. 

I know that line is a bit complex, but it's based on this line from the wiki show page: https://github.com/publiclab/plots2/blob/master/app/views/wiki/show.html.erb#L107

config/routes.rb
```diff
@@ -309,4 +309,6 @@
 
   match ':controller(/:action(/:id))(.:format)'
 
+  match 'gridsEmbed/:tagname' => 'notes#embed
```
jywarren: Ah - take a look, there's a missing end quote here. That's why the tests are failing. But this is looking good!

jywarren:
Hi, I'd like to add a functional test for this new view so we know it doesn't get broken by future work. It's pretty easy, would you be willing?

It'd go at the bottom of this file: 

https://github.com/publiclab/plots2/blob/master/test/functional/tag_controller_test.rb

And probably look like this -- something simple!

```ruby
  test "shows embeddable grid of tagged content" do
    get :gridsEmbed, tagname: 'spectrometer'

    assert_response :success
    assert_select "table" # ensure a table is shown
  end
```

jywarren:
I went ahead and added the test -- let's see how it does!

It also looks like you're working from your master branch - take a look at this guide for how to avoid using your own master branch, since that can make it harder to rewind and stay synced with the main repository: 

https://publiclab.org/wiki/contributing-to-public-lab-software#A+sample+git+workflow

jywarren:
Hey, it passed! Your code worked perfectly, and the test will ensure it continues to do so and nobody breaks it in the future. Thanks for your contribution! 

Merging now. If you're looking for a new challenge, please take a look at our `help-wanted` list: https://github.com/publiclab/plots2/labels/help-wanted

Thanks again! 🎈

PR #1231, opened 2017-02-02
event form test

jywarren:

Thanks!

### DIFF
--- a/app/views/editor/_event.html.erb
+++ b/app/views/editor/_event.html.erb
@@ -1,0 +1,14 @@
+<span style="margin-top:12px;margin-left:8px;" class="checkbox">
+  <label for="event-input">
+    <input tabindex="7" onChange="$('#event-info').toggle()" id="event-input" type="checkbox" name="event">
+    This is an event
+  </label>
+</span>
+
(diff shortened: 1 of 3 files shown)


jywarren:
Hi, @ananyo2012 -- do you think we could show something like "Tips" or "Suggestions" instead of "Warnings" to be a bit more encouraging?

jywarren:
Also, this now makes the issue fail if not every one is completed. I didn't realize that was the case. I think we should make them optional!

ananyo2012:
@jywarren Yes just change them from `warn` to `message`. messages are shown with the :book: icon instead of :warning: 
> Also, this now makes the issue fail if not every one is completed
What do you mean by this? The tests are failing due to test failures not due to danger.

jywarren:
OK -- did so here: #1233

I'll double check this, sorry -- I may have overlooked it. But does `warn` not make tests fail? Just checking. 

Thanks!

jywarren:
Great, thanks. Must have scrolled right past them!

ananyo2012:
No `warn` doesn't fail the build. there is another statement named `fail` to fail builds

PR #1243, opened 2017-02-06
[WIP] Section forms for updating subsections

### LINKED ISSUE #1005: Editing subsections of nodes
This is a planning issue, not yet complete, to eventually be broken up into sub-issues. 

### Idea

To make editing wikis easier, we're interested in making it possible to edit just one section; sections could be divided as block elements, so any header, paragraph, list, blockquote, or image would become a subsection. 

Subsections could have an "edit" link next to them, which opens an inline editor for just that subsection.

`DrupalNode` would have a "update_subsection" or similar method, which accepts strings A and B; A is what's replaced, B is what replaces it. A new revision is made with the post-swap version of the whole wiki's content. 

This architecture could be pretty robust, and might later lead to possible "suggestions" which affect just one block, and aren't yet "run". These could be stored as comments (so they're authored) or as some other new type -- and could apply to both notes and wikis. 

However, we'd need to think about authorship; can we efficiently show recent authors for a given subsection separately from for a whole wiki page? 

(Alternatively, could we think about per-line changes, rather than per-block changes, making it easier to split wiki pages up? But then list items would be one item per subsection, which doesn't sound right.)

Subsections would require:

* [x] Client-side `sections` method which splits latest revision body up by block element
* [x] `DrupalNode.update_subsection(old, new)` method
* [x] `DrupalNode.replace(old, new)` method which returns false if no replacements can be made
* [x] `replace` controller method in wiki_controller.rb
* [x] some view method for inserting "edit" links next to each block element on wiki pages
* [x] unit/functional tests for replace!
* [x] client-side tests for all that jazz
* [x] interface work to integrate new Rich editor inline

jywarren:

fixes #1005

### DIFF
--- a/app/assets/javascripts/methods.js
+++ b/app/assets/javascripts/methods.js
@@ -1,0 +1,8 @@
+jQuery(document).ready(function($) {
+
+  $('.topic-search').submit(function(e) {
+    e.preventDefault();
+    window.location = "/methods/" + $('.topic-search input').val();
+  });
+
+});
--- a/app/assets/javascripts/wikis.js
+++ b/app/assets/javascripts/wikis.js
@@ -1,13 +1,30 @@
+//= require wikis/processSections.js
+//= require wikis/replaceWithMarkdown.js
+
+/*
+* [edit]
+* respond to ajax WITHOUT flash[:notice]
+* respond with "failed" if replacement doesn't work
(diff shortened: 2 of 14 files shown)


jywarren:
Rebased and merged in #1246!

PR #1245, opened 2017-02-06
Emaill all tag folowwers

### LINKED ISSUE #1230: Email notify all followers of a tag if a new comment includes #tagname
Needs some fleshing out, but the comment notify code is here:

https://github.com/publiclab/plots2/blob/master/app/models/comment.rb#L97-L112

1. We'd want to scan for tags, which we already do when displaying, here: https://github.com/publiclab/plots2/blob/master/app/models/comment.rb#L43
2. We should make a method like `def notify_tag_followers`, where we collect those tags
3. we'd want to search for followers of that tag, which I believe is the `TagSelection` model
4. we'd want to de-duplicate as has been done elsewhere in the `notify` method, so nobody gets two emails
5. we'd want to write a test to confirm that the notifications get sent even if you're not otherwise related to this note (as an author, commenter, etc)

drfigs:
Fixes #1230

Thanks!


test/fixtures/comments.yml
```diff
@@ -95,3 +95,11 @@ question_callout:
   comment: Hey, @bob nice question
   timestamp: <%= Time.now.to_i + 11 %>
   thread: /03
+
+comment_tag:
+  uid: 1
+  nid: 8
+  status: 1
+  comment: 'comment with tag #ballom'
```
jywarren: Note that this says `#ballom` but the tag you're looking for below is `#ballon`.

test/functional/comment_controller_test.rb
```diff
@@ -106,7 +106,7 @@ def setup
          type: "question"
     comment.reload
     assert_equal new_comment_body, comment.comment
-    assert_redirected_to comment.answer.node.path(:question)
+    assert_redirected_to comment.answer.node.path(:question)+ "?_=" + Time.now.to_i.to_s
```
jywarren: These are good, thanks!

jywarren:
You had asked about:

> I created a test that makes sure we are calling the right mailer method CommentMailer.notify_tag_followers(user, tag)....on the test is missing to check the message body!
> I am having troubles in:
> 1. where should I call the method CommentMailer.notify_tag_followers?

I see how you're just [adding in tag followers within the `notify` method](https://github.com/publiclab/plots2/pull/1245/files#diff-3b4cc13a815e6a32db590e41f5d00d49R110), so perhaps that's fine, vs. an independent `notify_tag_followers` (which you could remove). Then you might merge your unit test code into the catch-all `notify` test, right?

> 2. where should I identify if there are tags on the comment and look for the followers? "comment create action on the controller, or inside the comment.notify method?"

I think there should be a `Comment` method like `comment.mentioned_tags` or `comment.tags` which you could unit test, does that make sense? 

Thank you and apologies for the slow response, i was traveling this week.

ryzokuken:
@drfigs Are you stuck working on this? Please feel free to ask for help. Thanks for contributing to Publiclab.

jywarren:
This has been solved in https://github.com/publiclab/plots2/issues/1330 -- thank you!

PR #1255, opened 2017-02-09
Fix Case Sensitivity issues

ryzokuken:

Thanks!

### DIFF
--- a/app/models/drupal_node.rb
+++ b/app/models/drupal_node.rb
@@ -41,8 +41,8 @@ class DrupalNode < ActiveRecord::Base
     updated_at.strftime('%B %Y')
   end
 
-  # FriendlyId is not being used; see 
-  # https://github.com/publiclab/plots2/pull/687 and 
+  # FriendlyId is not being used; see
+  # https://github.com/publiclab/plots2/pull/687 and
   # https://github.com/publiclab/plots2/pull/600
   extend FriendlyId
   friendly_id :friendly_id_string, use: [:slugged, :history]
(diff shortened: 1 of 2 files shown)


ryzokuken:
@jywarren Please look into this as soon as feasible.

jywarren:
I appreciate the modularity here, and I agree on using `find_by_username_case_insensitive` -- good idea -- but perhaps we can just use case insensitive for all environments, rather than switch on production vs. development? Or does that extra method not work in `mysql`?

It looks like `mysql` does have `lower()`: http://stackoverflow.com/questions/221774/is-there-a-mysql-command-to-convert-a-string-to-lowercase

ryzokuken:
@jywarren Cool, I just did that thinking 
> If it ain't broke, don't fix it
:P Anyway, moving to case insensitive for all environments.

jywarren:
Do you think it'd be a good idea to add a test to ensure two users can't get the same name with different capitalizations? A unit test? Or not necessary?

ryzokuken:
@jywarren Is it not desirable for capitalizations to make names unique (definitely not in this case, but otherwise) if that's the case, we could just remove all hassles by making usernames lowercase only. That would also make the URLs more legible (dunno why, capitals look strange in URLs)

jywarren:
I think that would have been easier to do at the beginning, but now we'd have to double check that there are no collisions, and also do some extra work to ensure that nothing would break if we shift everyone to lower-case... probably some testing, for starters. But for now, this is a good step -- if we do more, we should do it as a separate issue and PR. Thanks!

PR #1270, opened 2017-02-17
Update node_shared.rb

jywarren:


jywarren:
Coverage dropped by .008% due to longer method definition. I think that's OK!

PR #1284, opened 2017-02-21
location script rejigging

jywarren:


jywarren:
@mridulnagpal - note this change -- i noticed that the geocompletion code was running on every page, and I contained it with these changes. Just a heads up!

PR #1294, opened 2017-02-26
tests for redirect function

jywarren:
These tests fail, in order to demonstrate fixes we should make -- issue to be opened soon.


test/unit/drupal_node_tag_test.rb
```diff
@@ -181,4 +181,10 @@ class DrupalNodeTagTest < ActiveSupport::TestCase
     assert_equal I18n.t('drupal_node.only_admins_can_lock'), node(:blog).can_tag('locked', rusers(:bob), true)
   end
 
+  test "redirect tags to non-existent pages should not be accepted" do
```
jywarren: This was completed in #1309!

jywarren:
Introduced a fix, so 22b0a3c should fail and b81003e should resolve it.

PR #1317, opened 2017-03-04
Update Dangerfile

jywarren:
@ananyo2012 @ryzokuken what do you think of this? We may also want to explore why people keep checking in Gemfile.lock given that it's in the .gitignore file...?

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -8,6 +8,10 @@ end
 
 message "You have added multiple commits. It's helpful to [squash them](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) if the individual changes are small." if git.commits.size > 1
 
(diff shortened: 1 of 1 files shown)


ryzokuken:
I guess people add the Gemfile.lock because they're lazy and just do `git add -A` :stuck-out-tongue:

jywarren:
Hmm, is there any way to configure `.gitignore` to stop that? Or a way to automatically strip out the `Gemfile.lock` or something? Just trying to think creatively.

Do you like this change, though? I think it's a start.

ananyo2012:
@jywarren Yes this looks good. This would warn people who would did a `bundle update` . If there are minor changes to the gem version like the patch versions of the gem then that can be acceptable.

PR #1330, opened 2017-03-12
Fixes #1230 - Email notify all followers of a tag if a new comment includes #tagname 

### LINKED ISSUE #1230: Email notify all followers of a tag if a new comment includes #tagname

Khamba:

Thanks!

Hi,

I have worked on the issue #1230 and made some changes. I don't know why it shows that I have made 3 commits. The first two were to [sync my outdated fork with the main repository](https://help.github.com/articles/syncing-a-fork/).

[This](https://github.com/publiclab/plots2/commit/4a9c2ca6102c834e548fea73f4896601ee5bede7) is the main commit.

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -77,7 +77,7 @@ class Comment < ActiveRecord::Base
     User.find_all_by_username(usernames.map {|m| m[1] }).uniq
   end
 
-  def mentioned_tag_followers
+  def followers_of_mentioned_tags
     tagnames = self.comment.scan(Callouts.const_get(:HASHTAG))
     tagnames.map { |tagname| Tag.followers(tagname[1]) }.flatten.uniq
   end
@@ -91,7 +91,7 @@ class Comment < ActiveRecord::Base
 
   def notify_tag_followers(already_mailed_uids = [])
     # notify users who follow the tags mentioned in the comment
-    self.mentioned_tag_followers.each do |user|
+    self.followers_of_mentioned_tags.each do |user|
       CommentMailer.notify_tag_followers(self, user) if !already_mailed_uids.include?(user.uid)
     end
   end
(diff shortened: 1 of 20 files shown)


app/models/comment.rb
```diff
@@ -77,13 +77,25 @@ def mentioned_users
     User.find_all_by_username(usernames.map {|m| m[1] }).uniq
   end
 
+  def mentioned_tag_followers
```
jywarren: Here, is this followers of the tag who were /also/ mentioned? I think mentions should notify people regardless of whether they're following the tag, so this may need to be reworked and the method name adjusted to match. Thank you, though, your code is very readable!

Khamba: > Here, is this followers of the tag who were /also/ mentioned?

No, it is followers of the _tags_ that were mentioned. I can see now how it is be confusing. I wanted a similar method to `mentioned_usernames`. Will a comment on it suffice or should I change the method name?

> your code is very readable

Thank you very much :)

jywarren: Ha, i see, grammatically that's quite difficult, huh? What about `followers_of_mentioned_tags`? Thanks!

Khamba: I hear you. I will change the method name so that it is unambiguous.

jywarren:
thank you!!!!

app/models/comment.rb
Khamba: Made the changes. :+1:

jywarren:
Great, working on #1342 first, then I'll merge this. Super!

jywarren:
Done, thank you!!

PR #1335, opened 2017-03-14
add feature to sidebar, questions code overhaul

### LINKED ISSUE #897: Append keywords in questions tag filter as GET URL parameters
At https://publiclab.org/questions, when you add a "filter" of a given tag, the URL is not changed -- the added tag filters are stored in the session.

It's important that the keywords should be be appended to the url instead of being stored in the session. The routing should be more like:

`https://publiclab.org/questions/spectrometry`

and for multiple tags, perhaps:

`https://publiclab.org/questions/spectrometry,second-tag,another`

These would be fetched in the questions_controller via `params[:id].split(',')` -- here: https://github.com/publiclab/plots2/blob/master/app/controllers/questions_controller.rb. The `index`, `answered`, `popular`, `liked` action use this functionality presently.

The sorting is done using the `sort_questions_by_tag` method in the application controller, where it sources from the session: https://github.com/publiclab/plots2/blob/master/app/controllers/application_controller.rb#L190-L200. This method is called in all the places mentioned above.

The addition and deletion of tags from the session is done using the methods in tag_controller listed here: https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L239-L267. So you need to modify these methods too.

The tag sorting interface (that generates the sessions, and would generate the URLs in our new plan) shown on the sidebar, is here: https://github.com/publiclab/plots2/blob/master/app/views/tag/_sorting.html.erb

That itself might be moved into this directory for easier management: https://github.com/publiclab/plots2/blob/master/app/views/questions/

## Steps to solve
Summarizing the steps below
- [ ] Modify the tag methods mentioned [here](https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L239-L267) to append the keywords in url instead of session and fetch them accordingly
- [ ] Modify the `sort_questions_by_tag` [method](https://github.com/publiclab/plots2/blob/master/app/controllers/application_controller.rb#L190-L200) according to changes made and deprecate the use of session and use GET url parameters
- [ ] Modify the `index`, `answered`, `popular`, `liked`actions in questions controller accordingly
- [ ] Move the tag sorting template from [views/tag](https://github.com/publiclab/plots2/blob/master/app/views/tag) --> [views/questions](https://github.com/publiclab/plots2/blob/master/app/views/questions/
)

### LINKED ISSUE #1304: /questions page does not show any questions

I visited https://publiclab.org/questions/ and no questions are shown. 
Seems like when i click back and forth between "Recently asked" and "Recently answered", the yellow question blocks flicker for an instant before the page is just blank again. I wonder if this is happening for anyone else?

### Relevant URLs

https://publiclab.org/questions/

### PublicLab.org username

liz

Chrome

jywarren:
Fixes #1304, fixes #897 and closes #929 


jywarren:
Yay. Merging.

PR #1320, opened 2017-03-06
DrupalNode to Node and drupal_node to node

aayushgupta1:
Make sure these boxes are checked before your pull request is ready to be reviewed and merged. Thanks!

* [ ] all tests pass -- `rake test:all`
* [ ] code is in uniquely-named feature branch, and has been rebased on top of latest master (especially if you've been asked to make additional changes)
* [x] pull request is descriptively named with #number reference back to original issue

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/wiki/contributing-to-public-lab-software

We have a loose schedule of reviewing and pulling in changes every Tuesday and Friday, and publishing changes on Fridays. Please alert developers on plots-dev@googlegroups.com when your request is ready or if you need assistance.

Thanks!

PR #1341, opened 2017-03-15
Fixes #1337 - Paginate notes after ordering

### LINKED ISSUE #1337: Some recently edited methods pages don't sort to the top of /methods

****

### Problem

Some edits were made to https://publiclab.org/wiki/balloon-mapping, but it's still listed at the bottom of https://publiclab.org/methods

But that page is supposed to sort by last updated, descending, according to this code: https://github.com/publiclab/plots2/blob/master/app/controllers/notes_controller.rb#L14-L18

We can see that the latest revisions have reasonable timestamps, here: https://publiclab.org/wiki/revisions/balloon-mapping

Not sure why this is happening, but we should be able to narrow in on this error using a functional test, in this document:

https://github.com/publiclab/plots2/blob/master/test/functional/notes_controller_test.rb#L189-L201 (which can also use some indentation formatting help 😄 )

We ought to be able to test if `assigns(:notes)` is actually sorted by timestamp, maybe using Ruby's `sort_by` method.

amiedes:
Fixes #1337. Please read issue discussion.

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -14,8 +14,8 @@ class NotesController < ApplicationController
     @notes = DrupalNode.where(status: 1, type: ['page','tool'])
                        .includes(:drupal_node_revision, :tag)
                        .where('term_data.name = ?','tool')
-                       .page(params[:page])
                        .order("node_revisions.timestamp DESC")
+                       .page(params[:page])
     render :template => "notes/tools_places"
   end
 


jywarren:
Hi, thanks for your patience! Trying to give time for #1320 to complete before merging new stuff. Many thanks for your help here, this looks good!

jywarren:
I /believe/ this will merge properly but will double check afterward. Trying now!

jywarren:
Looks like one of the recent merges didn't turn out right. Filing a follow-up!

jywarren:
OK, will get it in #1335; i added one more commit there. 

Thank you very much! If you're looking for a new challenge, please take a look at our `help-wanted` list: https://github.com/publiclab/plots2/labels/help-wanted

🎈

PR #1349, opened 2017-03-19
Fix code that violate the Ruby Style Guide

### LINKED ISSUE #1348: Ruby style guide
There are many violations against the Ruby style guide. Is it OK to fix these violations using tools such as Rubocop? I can slowly complete this because there are over 10K violations.

siaw23:
Enforces changes that follow the Ruby Style Guide to make written code clearer
and easier to read. Closes #1348

### DIFF
--- a/app/views/dashboard/_node_meta.html.erb
+++ b/app/views/dashboard/_node_meta.html.erb
@@ -6,9 +6,8 @@
 <span class="">
   <% if params[:controller] == 'questions' %>
     | <a href="<%= node.path(:question) %>#answers"><i class="fa fa-comments-o" style="color: #DAA583;"></i> <%= node.answers.length %></a>
-  <% else %> 
-    | <a href="<%= node.path %>#comments"><i class="fa fa-comment-o"></i> <%= node.comment_count %></a>
   <% end %>
+    | <a href="<%= node.path %>#comments"><i class="fa fa-comment-o"></i> <%= node.comment_count %></a>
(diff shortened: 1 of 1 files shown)


jywarren:
Wow, this is really cool, and pretty big. Do you think we could try this just on controllers, to start with, or something a little smaller even? I just want us to clearly understand if merging this will cause rebase/merge difficulties for other contributors, and maybe tackling it in smaller pieces will be easier -- also, maybe easier to debug, since it looks like this failed tests? 

Thank you! Very much appreciated!

jywarren:
Hmm, it looks like a few syntax errors were introduced; they're [caught in the tests](https://travis-ci.org/publiclab/plots2/builds/212623972). For example:

```
/usr/local/lib/ruby/gems/2.1.0/gems/activesupport-3.2.22.4/lib/active_support/dependencies.rb:251:in `require': /app/app/controllers/subscription_controller.rb:137: syntax error, unexpected end-of-input, expecting keyword_end (SyntaxError)
```

siaw23:
@jywarren yes i saw the syntax errors. just came from a Ruby conference, exhausted, i'll check this and other stuff when i wake up. we can close this PR for now. the change is very big. i'll look for something smaller to fix for the start, gradually so you can track what i'm trying to do and suggest or provide feeback.

siaw23:
@jywarren can we close this PR I'd like to make a new one with a smaller change gradually. if that's ok. the changes in this commit are too much for a one time commit i think.

PR #1352, opened 2017-03-21
Improve code to match Ruby Style Guide

### LINKED ISSUE #1348: Ruby style guide

siaw23:
This make simple code improvements for a more readable code
Closes #1348 (part of)

All tests pass.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1,51 +1,50 @@
 class AdminController < ApplicationController
-
-  before_filter :require_user, :only => [:spam, :spam_revisions]
+  before_filter :require_user, only: [:spam, :spam_revisions]
 
   def promote_admin
     @user = User.find params[:id]
-    if !@user.nil?
-      if current_user && current_user.role == "admin"
+    unless @user.nil?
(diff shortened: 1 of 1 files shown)


jywarren:
This looks great; merging. Thanks! So, perhaps we could try /all controllers/? How do you suggest we approach this gradually? I'm very open to suggestions!

siaw23:
i'm going to fix all of the files in the controller tonight, run tests and make one big push for the controller, if it gets merged, i'll submit the fixes folder by folder... how about that?

jywarren:
I think that sounds reasonable. We should also link to the style guide from
the readme and perhaps we could integrate some automated feedback via
Danger as well, as long as it's friendly!

Thank you!!

siaw23:
yes, i'd like to include the README and gems for checking style guide after i have checked everything on my end. the reason is some tests failures are triggered and i wanted to fix them before so others can concentrate on contributing code instead of fixing test failures. so yeah, i'm on it... thanks @jywarren

PR #1353, opened 2017-03-21
Add support for A/B Testing on wiki pages (resolves #1339)

### LINKED ISSUE #1339: Enable A/B testing for wiki pages on PublicLab.org
## Problem

We're looking to test out some new page designs vs. existing pages in a simple A/B type way. The site uses Google Analytics, so we can see how people interact with the sites; we just need a way to direct equal proportions of people to one page or another. This could be 

## Proposed solution

We could do this with [power tags](https://publiclab.org/wiki/power-tags) -- similarly to the `redirect` tag -- but only affecting 50% of visitors. The tag could be formatted like:

`abtest:pagename` (which is obscure but this is a pretty obscure usage)

We could add a check similar to the redirect tag code: https://github.com/publiclab/plots2/blob/master/app/controllers/wiki_controller.rb#L29-L37

Using a random function to direct people to `/wiki/balloon-mapping` (for the example `abtest:balloon-mapping`

This may be hard to test for, but we should see the effects in Google Analytics. Then can use path analysis over there to assess how people act on each page.

ryzokuken:

Thanks!

### DIFF
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -39,11 +39,11 @@ class WikiController < ApplicationController
     if @node && @node.has_power_tag('abtest')
       if current_user == nil || (current_user.role != 'admin' && current_user.role != 'moderator')
         if Random.rand(2) == 0
-          redirect_to DrupalNode.find(@node.power_tag('redirect')).path, notice: "You have been redirected to this page randomly for testing purposes. We apologize for any inconvenience caused."
+          redirect_to DrupalNode.find(@node.power_tag('redirect')).path
         end
         return
       elsif (current_user.role == 'admin' || current_user.role == 'moderator')
-        flash.now[:warning] = "Only moderators and admins see this page, as it is redirected to #{DrupalNode.find(@node.power_tag('abtest')).title}.
+        flash.now[:warning] = "Only moderators and admins see this page, as it is redirected to #{DrupalNode.find(@node.power_tag('abtest')).title} roughly around 50% of the time.
         To remove this behavior, delete the tag beginning with 'abtest:'"
       end
     end
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -36,6 +36,18 @@ class WikiController < ApplicationController
       end
     end
 
+    if @node && @node.has_power_tag('abtest')
+      if current_user == nil || (current_user.role != 'admin' && current_user.role != 'moderator')
+        if Random.rand(2) == 0
+          redirect_to DrupalNode.find(@node.power_tag('redirect')).path
+        end
+        return
+      elsif (current_user.role == 'admin' || current_user.role == 'moderator')
+        flash.now[:warning] = "Only moderators and admins see this page, as it is redirected to #{DrupalNode.find(@node.power_tag('abtest')).title}.
+        To remove this behavior, delete the tag beginning with 'abtest:'"
+      end
+    end
+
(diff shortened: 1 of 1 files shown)


app/controllers/wiki_controller.rb
```diff
@@ -36,6 +36,18 @@ def show
       end
     end
 
+    if @node && @node.has_power_tag('abtest')
+      if current_user == nil || (current_user.role != 'admin' && current_user.role != 'moderator')
+        if Random.rand(2) == 0
```
jywarren: Will this output an integer? Maybe worth a manual test of some kind.

app/controllers/wiki_controller.rb
```diff
@@ -36,6 +36,18 @@ def show
...
+          redirect_to DrupalNode.find(@node.power_tag('redirect')).path
+        end
+        return
+      elsif (current_user.role == 'admin' || current_user.role == 'moderator')
+        flash.now[:warning] = "Only moderators and admins see this page, as it is redirected to #{DrupalNode.find(@node.power_tag('abtest')).title}.
```
jywarren: Perhaps here we might say "50% of the time" or something equivalent.

app/controllers/wiki_controller.rb
ryzokuken: Please look into this: https://repl.it/G7Ds/1

app/controllers/wiki_controller.rb
ryzokuken: Like... in a comment.

jywarren: Ah, I meant in the flashed message. Because it'll be confusing to folks if it only happens sometimes. Thanks!

app/controllers/wiki_controller.rb
jywarren: Ok perfect, thanks! Cool service, too.

app/controllers/wiki_controller.rb
ryzokuken: I will add the flash message. However, as the page is being redirected the exact next line, I have no idea if it'll show up. Will test it out manually.

ryzokuken:
@jywarren Added a notice with a different, more concise syntax. I guess we could use these more often. 

Reference: http://guides.rubyonrails.org/action_controller_overview.html#the-flash

app/controllers/wiki_controller.rb
```diff
@@ -36,6 +36,18 @@ def show
       end
     end
 
+    if @node && @node.has_power_tag('abtest')
+      if current_user == nil || (current_user.role != 'admin' && current_user.role != 'moderator')
+        if Random.rand(2) == 0
+          redirect_to DrupalNode.find(@node.power_tag('redirect')).path, notice: "You have been redirected to this page randomly for testing purposes. We apologize for any inconvenience caused."
```
jywarren: Ah, well, I don't think we need to tell people who are not mods/admins -- that would affect the test we're trying to run. But I do like the syntax, thanks! 

Could we remove the notice here, and then we're good to go? Thanks, and especially thanks for your helping people out while I was traveling!

ryzokuken: Don't we all try to do as much as we can? I have my exams around, but I will try my best to be engaged. Meanwhile, I will remove the notice.

ryzokuken:
@jywarren Removed the notice and modified the notice being sent to admins/mods. Seems okay now?

jywarren:
Awesome!!!!

PR #1357, opened 2017-03-26
[WIP] test for wiki pages with added parameter :node type fixes #1293

### LINKED ISSUE #1293: Redirected pages not listed in wildcard tag listing

****

### Background

Some pages have a tag like `redirect:pagename` where `pagename` is the name of another page which the viewer should be automatically redirected to. Documentation:

https://publiclab.org/wiki/power-tags (look for `redirect`)

This behavior is captured (at the controller level) in this test:

https://github.com/publiclab/plots2/blob/master/test/functional/wiki_controller_test.rb#L507-L516

### Problem

These redirected pages should be listed on this tag listing page for wiki pages, using the wildcard `redirect:*` -- but nothing is shown there:

https://publiclab.org/wiki/tag/redirect:*

There are definitely redirected pages, like this one:

https://publiclab.org/wiki/camera-triggering

### Next steps

We should write a functional test to ensure that such tags appear on that tag listing page. There is already one for notes, but the separate action to display wiki pages is not tested:

https://github.com/publiclab/plots2/blob/master/test/functional/tag_controller_test.rb#L160-L167

I believe it uses the same action, but changes the `params[:node_type]` parameter:

https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L23-L25

Once the test correctly describes this bug, changes to the tag controller will be able to resolve the problem.

StlMaris123:

Thanks!


jywarren:
Hi, @StlMaris123 - Is this still in progress, or shall I take a look? Thank you!

jywarren:
Looking at how this relates to #1293; I think we need to add an additional assertion -- to test the existence of (and non-zero length of) the `@wikis` variable. What we're worried about in #1293 is that the page does not return any wiki pages matching the wildcard. So if we add this to the test, we can confirm that it's not returning any wiki pages:

`assert :wikis`

also we could add:

`assert assigns(:wikis).length > 0`

Finally, we could look specifically to see if `assigns(:wikis)` (which gives us the actual variable object) contains a known wiki page which we'd expect to be listed. In this case, we're looking for a wiki page tagged with `question:_____` -- are we sure there is one? 

I do know that we have tagged the fixture `node(:method)` -- a "method page" (and a wiki) with the tag `activities:spectrometer` -- by looking here: https://github.com/publiclab/plots2/blob/master/test/fixtures/community_tags.yml#L115

So if we change the test to search for the wildcard `activities:*` we could then test to ensure that page shows up in the `@wikis` variable.

Sorry, I know that's a lot, but does it make sense? I'm happy to help you out on this. Thanks for your patience!

StlMaris123:
Hi @jywarren  I have modified the test. It is actually failing on this line:
`asserts assigns(:wikis).length > 0`
 meaning there is no any wiki pages

jywarren:
Ah, well then I think we've captured our big finally! So now we need to walk through the controller logic to find why it's not finding the node. Perhaps the wildcard logic is not feeding into the wiki listing code? Take a look and if you don't find anything I'm happy to help out!

Thanks!

StlMaris123:
Hi @jywarren,
I have used debugger to try and find out where the problem might be on the controller. On the browser i had the following params `http://0.0.0.0:3000/tag/question:*?node_type=wiki`.
The variable `nodes ` on line 32 returns an empty array. 
So i have gone through the methods being called on DrupalNode and had the following output: 
```ruby 
 DrupalNode.where(status: 1, type: node_type).count
```
7
```ruby
DrupalNode.where(status: 1, type: node_type).includes(:drupal_node_revision, :tag).count
```
7
```ruby
DrupalNode.where(status: 1, type: node_type).includes(:drupal_node_revision, :tag).where('term_data.name LIKE (?) OR term_data.parent LIKE (?)', params[:id][0..-2] + '%', params[:id][0..-2] + '%').count
```
0
Let me know if I am on track with the way i have approached the issue. :smile:

StlMaris123:
Hi Jeff,
 This issue has not been attended to for a while now.
Regards,
Stella

jywarren:
Hi, i'm sorry, i missed this! I'll look over it this weekend. Apologies!

jywarren:
Hi, so I think the `.where()` method may not be chained... but it's been a while and I'm rusty on it. Here are some places to read through on it:

* http://guides.rubyonrails.org/active_record_querying.html
* http://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html
* https://apidock.com/rails/ActiveRecord/QueryMethods/where

Perhaps this is important: http://guides.rubyonrails.org/active_record_querying.html#rewhere

But you might also want to just search for "chaining ActiveRecord where queries" ! Tell me what you find!

StlMaris123:
Just out of curiosity , why is this method being chained? 
`.where('term_data.name LIKE (?) OR term_data.parent LIKE (?)', params[:id][0..-2] + '%', params[:id][0..-2] + '%')` . It returns nothing.

jywarren:
There, I believe it's looking for matches to the search term, for example `balloon*` -- searching for anything starting with `balloon` -- the `*` wildcard represents an open-ended search like that. And we're also looking for anything with a parent tag that matches -- this is due to tag aliasing, where we want a search of `spectrometer` to also return things tagged `spectrometry` -- so one is the "parent" of the other. 

One thing is that there are probably not any wiki pages tagged with `question:*` - since that's a tag we've typically used with "notes" and not "wikis" -- so can you check for what tags your wiki pages have (add some if necessary) and then search for a tag you can definitely expect them to have? Say, tag one with "balloon-mapping" and then do a search with `http://0.0.0.0:3000/tag/balloon*?node_type=wiki`?

jywarren:
Ah, i also have seen elsewhere in the code that chaining `where` methods is OK -- as long as it's not the same conditions. So sorry if that was a mistake and wasn't really the problem!

jywarren:
Hi, Stella - wow, big additions here. But I wonder if you are reverting some changes during your rebase. See how there are changes to many many files, possibly because in your merge conflicts you chose the older versions? 

Luckily your commit is preserved, so I think you could rebase it on top of the latest master branch: https://github.com/publiclab/plots2/pull/1357/commits/9100d105d146844851f7f2c022f43586fb94e80b

I think what you'd do is:

1. backup your work with `git branch bkp-temporary`
2. switch to your master branch with `git checkout master`
3. get the latest master with `git pull https://github.com/publiclab/plots2 master`
4. open a new branch with `git checkout -b rebase-wiki`
5. try to "cherry pick" your last (intact) commit with `git cherry-pick 9100d105d146844851f7f2c022f43586fb94e80b` (i think this is right - you may want to read about cherry-picking in the Git docs)
6. try opening a new pull request with that branch? Let's see what happens!

And no worries, if this doesn't quite work I can help untangle things. Git can get a little out of control sometimes. :-)

StlMaris123:
Hello @jywarren , the issue is that i pulled the branch. I had to forked the repository again after I changed the machine. Could this cause the change in many files?

StlMaris123:
All changes are in the new pull request.

PR #1363, opened 2017-04-04
[WIP] Update search by tag api test

### LINKED ISSUE #1332: API calls for tags returning "null" -- need better test to diagnose

https://publiclab.org/api/srch/tags?srchString=ball returned:

```
{"srchParams":{"srchString":"ball","seq":null,"showCount":null,"pageNum":null}}
```

https://github.com/publiclab/plots2/wiki/API covers this usage, as does https://publiclab.org/api/swagger_doc.json

I'm not sure what's going on here, but we do test this out here:

https://github.com/publiclab/plots2/blob/master/test/functional/restful_search_test.rb#L89

We should tighten up this test to actually check that the returned JSON contains a valid tagname, so for example `spec` as a search term should return several tagnames from our fixtures: https://github.com/publiclab/plots2/blob/master/test/fixtures/tags.yml

Once this test is created and failing (capturing the bug), we can investigate why, and fix it! 

Later, this could be adapted into https://github.com/publiclab/plots2/blob/master/app/assets/javascripts/tagging.js#L67 and the older suggested tags query format can be deprecated.

ffiebig:
fixes #1332.

@jywarren this is what I've tried

### DIFF
--- a/test/functional/restful_search_test.rb
+++ b/test/functional/restful_search_test.rb
@@ -92,16 +92,19 @@ class RestfulSearchTest < ActiveSupport::TestCase
     # Expected tag pattern
     pattern = {
       items: [{
-        docId: 8,
+        docId: 13,
         docType: 'tag',
-        docUrl: '/notes/admin/03-14-2017/blog-post',
-        docTitle: 'Blog Post',
+        docUrl: '/notes/jeff/'+Time.now.strftime("%m-%d-%Y")+'/blog-post',
+        docTitle: 'Blog post',
         docSummary: '',
         docScore: 0
       }],
-      srchString: {
-        srchString: 'blog'
-      }.ignore_extra_keys!
+      srchParams: {
+        srchString: 'blog',
+        seq: nil,
+        showCount: nil,
+        pageNum: nil
+      }
     }
 
     matcher = JsonExpressions::Matcher.new(pattern)
--- a/test/functional/restful_search_test.rb
+++ b/test/functional/restful_search_test.rb
@@ -87,17 +87,22 @@ class RestfulSearchTest < ActiveSupport::TestCase
 
 
   test 'returns results from search tags functionality' do
-    get '/api/srch/tags?srchString=lon:24.484315929497463&seq=7'
+    get '/api/srch/tags?srchString=blog'
     assert last_response.ok?
     # Expected tag pattern
     pattern = {
-      # Need more/better understanding and data for the test database, so ignoring null results for now
-      # items: Array,
-      srchParams: {
-        srchString: 'lon:24.484315929497463',
-        seq: 7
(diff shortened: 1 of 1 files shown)


test/functional/restful_search_test.rb
```diff
@@ -87,17 +87,22 @@ def app
...
-        seq: 7
+      items: [{
+        docId: 8,
+        docType: 'tag',
+        docUrl: '/notes/admin/03-14-2017/blog-post',
+        docTitle: 'Blog Post',
+        docSummary: '',
+        docScore: 0
+      }],
+      srchString: {
```
ffiebig: There's a typo here, it was meant to be srchParams

test/functional/restful_search_test.rb
```diff
@@ -87,17 +87,22 @@ def app
...
+        srchString: 'blog'
       }.ignore_extra_keys!
-    }.ignore_extra_keys!
+    }
 
     matcher = JsonExpressions::Matcher.new(pattern)
```
jywarren: Cool, so can `JsonExpressions::Matcher` do a comparison of two JSON objects?

jywarren: So then before the "assert" statement on the next line, we could run `puts last_response.body` to see what's output. If it does look like it should match, then we can manually try it out in the rails console with `rails console` and try some variations. Does that help?

ffiebig:
@jywarren Just pushed a new commit, please review it when you get the chance. Thanks ! :bowtie:

jywarren:
Hi, I'm not sure, but can you help me figure out exactly what this test is doing that the previous one wasn't? This is a subtle bug and I'm not sure what's going on, so if you have more insight than I that's great! We appreciate your help!

ffiebig:
Previous test was ignoring extra keys, it was only checking `srchParams`. This one, by searching for `blog`, checks that the correct tag is returned. The tag is coming from `fixtures/node.yml`. It should fail if it returns null.

To be honest I'm a bit lost, when I do ` get '/api/srch/tags?srchString=foo'` by running the test it searches within the fixtures on `fixtures/node.yml` and there should be no results for `ball` (right?). On #1332  you mentioned that searching for `ball` was returning null, isn't that the correct response?

You mentioned that it is a subtle bug, am I not seeing it?

jywarren:
Hi, apologies for my slow reply. 

So, I'd thought that `ball` should return results for `balloon-mapping`, but in fact it does not: you need `balloon-mapping` in its entirety:

https://publiclab.org/api/srch/tags?srchString=balloon-mapping

So, it looks like this is a non-bug! However, your test adjustment does improve that test, so thank you, and I'll merge that in now. Much appreciated, and apologies for the confusion!

PR #1367, opened 2017-04-04
two separate fixes

jywarren:


jywarren:
Oh, that's dumb, of course it might be a redirect sometimes; we can't test that.

PR #1366, opened 2017-04-04
rework API to use Solr when available

jywarren:
Needs now to be done on top of (after) #1386 -- could redo this manually, actually.

@icarito - check this change out -- it would make API node searches draw upon Solr when available, and ActiveRecord when not. If you like it, and #1176 is ready, you can merge this into the `fix_solr` branch.

****

### Jeff's remaining to-do list:

- [x] `Node.search` (fulltext) inside of `SearchService` and test it (on `SearchService` unit test and JSON API output, & https://publiclab.org/search/dynamic/)

PR #1176, opened 2017-01-06
In-progress Solr containerization and reindexing work

jywarren:
@icarito and I are working on this branch to resolve Solr issues through better containerization; working on this in #1112. 

This will show a pass/fail based on any commits pushed to this branch: 

http://staging.laboratoriopublico.org/job/Plots-Solr/

http://branch1.laboratoriopublico.org/

Look for success here:

http://branch1.laboratoriopublico.org/searches/test

PR #1386, opened 2017-04-27
re-enable solr-specific tests, next Solr steps

jywarren:
* [ ] figure out why Coveralls isn't running on here, and get it running so we can see if coverage drops - maybe by reopening this from `publiclab/plots2`?
* [ ] re-enable test.rake "solr-specific" tests in such a way as to not confuse Coveralls and maintain 86% coverage -- this may be due to those extra tests running in a subshell...
  * possibly with http://stackoverflow.com/questions/18894060/rails-and-minitest-add-additional-folder
  * possibly using alternative shell command runner: http://stackoverflow.com/questions/2232/calling-shell-commands-from-ruby#2400 
* [ ] fix tests on `test/solr/search_record_test.rb:50` to ensure unique results are actually returned
  * was this bc it was not actually indexing? So, maybe if we get Solr indexing properly, this will pass. 
* [ ] document why 23-hour reindex is needed, or fix the toggle to make this unnecessary
* [ ] re-enable solr on production

Then, the rest of this:

* [ ] get `branch1` building from this PR
* [ ] reindexing and then looking for if a search for "balloon" works on this page: http://branch1.laboratoriopublico.org/searches/new
* [ ] Once it does, we want to try turning off Solr and ensuring the site doesn't break (as the new toggle code should allow for).
* [ ] then, we can try merging in #1366, and seeing if http://branch1.laboratoriopublico.org/search/dynamic returns results for `balloon` with Solr on AND off
* [ ] then we have to do the remainder of #1366's listed to-dos

### DIFF
--- a/lib/tasks/test.rake
+++ b/lib/tasks/test.rake
@@ -6,8 +6,8 @@ namespace :test do
     Coveralls::RakeTask.new
     puts "Running Rails tests"
     Rake::Task["test"].execute
-    #puts "Running Solr-dependent tests"
-    #Rake::Task["test:solr"].execute
+    puts "Running Solr-dependent tests"
+    Rake::Task["test:solr"].execute
     puts "Running jasmine tests headlessly"
     Rake::Task["spec:javascript"].execute
     Rake::Task["coveralls:push"].execute


jywarren:
Why aren't coveralls checks running on this?

jywarren:
Maybe we have to close this and open it from `publiclab/plots2` to get Coveralls to run? @ryzokuken @ananyo2012 -- do either of you know how to get Coveralls running on this PR?

ryzokuken:
@jywarren Sorry, I have no idea. Will look it up though.

jywarren:
Ah, i just enabled it in Coveralls by adding the `jywarren/plots2` repo.

jywarren:
Hmm ok, so i tried to enable Coveralls but it looks like I have to input a unique token, and we've already done that for the `publiclab` repository. So the easiest way will be to open this from another branch on the `publiclab` repository, perhaps? 

But has something gone wrong with the Coveralls integration? I had thought it was run on every single PR, but I don't see it anywhere now? Maybe we did something odd while working on this in #1176?

icarito:
Okay then I'll move this PR into a branch of plots2 repo to get coveralls.

icarito:
this is very wierd, now I do see coveralls in this PR. Still, moving to https://github.com/publiclab/plots2/pull/1409 for easier teamwork.

jywarren:
I think Coveralls may have been having issues yesterday, as i had one which
stalled for almost 40 minutes. I ultimately just went ahead and merged it.

PR #1403, opened 2017-05-09
initial tag-filtered maps display

jywarren:
Ability to filter an inline map by tag, using `[map:tag:__tagname__:lat:lon]`


jywarren:
@mridulnagpal - this also relates to your work! There are a variety of different types of map we hope to have, and your module will be adding the tags which drive this system, while this code actually displays a map with all the location-tagged data. Make sense?

PR #1407, opened 2017-05-09
Added test/functional/search_api_test.rb

### LINKED ISSUE #1395: Add new functional tests to each method in Search API
Hi, this is a [first-timers-only issue](https://github.com/publiclab/plots2/projects/2). This means we've worked to make it more legible to folks who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**. 

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general [help wanted](https://github.com/publiclab/plots2/labels/help-wanted) issues. Thanks!

## The problem

> Tests help us ensure that the site behaves as we expect it to -- and that as new features are added, they don't break existing functions. We're working to get to near 100% "coverage" -- which would mean all our code is covered by tests.

We have good tests written for the typeahead API, here: https://github.com/publiclab/plots2/blob/master/test/functional/typeahead_api_test.rb

This tests each method in: https://github.com/publiclab/plots2/blob/master/app/services/typeahead_service.rb

But we should have similar ones for the search API, for the methods in: https://github.com/publiclab/plots2/blob/master/app/services/search_service.rb

## Solution

Let's create a new file at:

`test/functional/search_api_test.rb` -- containing, as a start:

```js
require 'test_helper'

class SearchApiTest < ActiveSupport::TestCase
  include Rack::Test::Methods

  def app
    Rails.application
  end

  test 'search notes functionality' do
    get '/api/srch/notes?srchString=Blog'
    assert last_response.ok?

    # Expected search pattern
    pattern = {
      srchParams: {
        srchString: 'Blog',
        seq: false,
      }.ignore_extra_keys!
    }.ignore_extra_keys!

    matcher = JsonExpressions::Matcher.new(pattern)

    assert matcher =~ JSON.parse(last_response.body)

    # check for actual Blog note, too
    # assert_equal ___, node(:blog)

  end

end
```

You can try running this with `rake test` or (just to run the functional tests only) `rake test:functionals` -- hopefully it will pass! But you can also just open a pull request (see below) and our automated test system will run it, and you'll be able to see the output there. 

If you can help with this, please open a pull request with this test added, and let's see how it goes. We may need to tweak the `srchString` parameter, and (as noted in the snippet above) add an extra assertion to ensure the results are right. 

After this, we can move on to adding additional tests for the other methods in the API.

## Steps to Fix

- [ ] claim this issue with a comment here, below, and ask any clarifying questions you need
- [ ] set up a repository locally following the README instructions, and make sure that all tests pass
- [ ] try to fix the issue following the steps above, but even before you're done, you can:
- [ ] commit your changes and start a pull request (see [contributing to Public Lab software](https://publiclab.org/wiki/contributing-to-public-lab-software)) but mark it as "in progress" if you have questions or if you haven't finished
- [ ] alert someone via the developers list (see below) to have your pull request merged. This may take a few extra steps depending on when you do it, but we'll help you out!

Please email the developers list (see https://publiclab.org/wiki/developers) or go to [the chatroom](http://publiclab.org/chat) if you have questions, and take a look at [our first-timers landing page for more information](https://github.com/publiclab/plots2/projects/2)!

orcutt989:
Was told per the chat to generate a pull request even though the test does not pass.

per @jywarren "this is a good moment to post a PR so we can all look at the output of the Travis automated tests!"

Fixes #1395

### DIFF
--- a/test/functional/search_api_test.rb
+++ b/test/functional/search_api_test.rb
@@ -13,10 +13,6 @@ class SearchApiTest < ActiveSupport::TestCase
 
     # Expected search pattern
     pattern = {
-      items: [{
-        docUrl: "/notes/jeff/05-10-2017/blog-post",
-        docTitle: "Blog post"
-      }.ignore_extra_keys!],
       srchParams: {
         srchString: 'Blog',
         seq: false,
@@ -24,8 +20,14 @@ class SearchApiTest < ActiveSupport::TestCase
     }.ignore_extra_keys!
 
     matcher = JsonExpressions::Matcher.new(pattern)
+
+    json = JSON.parse(last_response.body)
+
+    assert_equal "/notes/jeff/05-10-2017/blog-post", json['items'][0]['docUrl']
+    assert_equal "Blog post",                        json['items'][0]['docTitle']
+    assert_equal 13,                                 json['items'][0]['docId']
     
-    assert matcher =~ JSON.parse(last_response.body)
+    assert matcher =~ json
 
   end
 
--- a/test/functional/search_api_test.rb
+++ b/test/functional/search_api_test.rb
@@ -15,7 +15,7 @@ class SearchApiTest < ActiveSupport::TestCase
     pattern = {
       srchParams: {
         srchString: 'Blog',
-        seq: false,
+        seq: nil,
       }.ignore_extra_keys!
     }.ignore_extra_keys!
 
@@ -23,9 +23,9 @@ class SearchApiTest < ActiveSupport::TestCase
 
     json = JSON.parse(last_response.body)
 
-    assert_equal "/notes/jeff/05-10-2017/blog-post", json['items'][0]['docUrl']
-    assert_equal "Blog post",                        json['items'][0]['docTitle']
-    assert_equal 13,                                 json['items'][0]['docId']
+    assert_equal node(:blog).path, json['items'][0]['docUrl']
+    assert_equal "Blog post",       json['items'][0]['docTitle']
+    assert_equal 13,                json['items'][0]['docId']
     
     assert matcher =~ json
 
--- a/app/views/grids/_notes.html.erb
+++ b/app/views/grids/_notes.html.erb
@@ -63,6 +63,8 @@
 <% end %>
 </table>
 
+<a class="grid-embed" title="Embed this table on another site."><i class="fa fa-code" style="color:#aaa;"></i></a>
+
 <% if type == "questions" %>
   <p><a href='/post?tags=question:<%= tagname %>,<%= tagname %>&template=question&title=How%20do%20I...&redirect=question' class='btn btn-primary add-activity'>Ask a question</a> &nbsp;or <a href='/subscribe/tag/question:<%= tagname %>'>help answer future questions<span class='hidden-sm hidden-xs'> on this topic</span></a></p>
 <% elsif type == "activity" %>
@@ -76,6 +78,18 @@
   $(".<%= className %>-<%= randomSeed %> .show-all a").click(function() { 
     $(".<%= className %>-<%= randomSeed %> tr.hide").toggleClass("hide");
   });
+  $(".grid-embed").click(function() {
+    prompt('Use this HTML code to embed this table on another site.', '<iframe border=0 width="100%" height="500px" src="//publiclab.org/embed/grid/&?t=<% if type != "notes" %><%= type %>:<% end %><%= tagname %>"></iframe>')
+  });
   setupGridSorters(".<%= className %>-<%= randomSeed %>");
 })()
 </script>
+<style>
+  a.grid-embed {
+    float:right;
+    padding:8px;
+  }
+  a.grid-embed:hover {
+    color: #888;
+  }
+</style>
--- a/app/views/map/_leaflet.html.erb
+++ b/app/views/map/_leaflet.html.erb
@@ -8,6 +8,6 @@
   L.tileLayer("//a.tiles.mapbox.com/v3/jywarren.map-lmrwb2em/{z}/{x}/{y}.png").addTo(map);
 
   <% items.each do |item| %>
-    L.marker([<%= item.lat %>, <%= item.lon %>]).addTo(map).bindPopup("<a href='/<%= item.path %>'><%= item.title %></a>");
+    L.marker([<%= item.lat %>, <%= item.lon %>]).addTo(map).bindPopup("<a href='<%= item.path %>'><%= item.title %></a>");
   <% end %>
 </script>
--- a/app/views/map/_leaflet.html.erb
+++ b/app/views/map/_leaflet.html.erb
@@ -4,7 +4,7 @@
 </style>
 <div class="leaflet-map" id="map<%= unique_id %>"></div>
 <script>
-  var map = L.map('map<%= unique_id %>').setView([<%= lat %>,<%= lon %>], 10);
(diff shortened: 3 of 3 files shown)


jywarren:
Hi, i see this! Thanks, will check it out and get back to you tomorrow.

jywarren:
Note the "checks" running below -- if you click on Details in the Travis line, you can see the test output as it runs.

test/functional/search_api_test.rb
```diff
@@ -0,0 +1,31 @@
...
+    pattern = {
+      srchParams: {
+        srchString: 'Blog',
+        seq: false,
+      }.ignore_extra_keys!
+    }.ignore_extra_keys!
+
+    matcher = JsonExpressions::Matcher.new(pattern)
+
+    assert matcher =~ JSON.parse(last_response.body)
```
jywarren: OK, so this comparison didn't work -- we've probably missed some of the content on lines 16-18. Could be best to do `puts JSON.parse(last_response.body)` to see what comes out, and we can work from there!

jywarren:
Oh, actually it needs to go on its own line, like:

```
puts JSON.parse(last_response.body)
assert matcher =~ JSON.parse(last_response.body)
```

jywarren:
OK -- so now, we have to debug a little. `JSON.parse(last_response.body)` is returning:

```js
{"items"=>[{"docId"=>13, "docType"=>"file", "docUrl"=>"/notes/jeff/05-10-2017/blog-post", "docTitle"=>"Blog post", "docSummary"=>nil, "docScore"=>0}], "srchParams"=>{"srchString"=>"Blog", "seq"=>nil, "showCount"=>nil, "pageNum"=>nil}}
```

So we have to think about how to craft a "matcher" that will return true if we compare it to this using the `~=` operator. I'm not that familiar with it, but @david-days wrote this code, I wonder if he's able to chip in? 

But in any case, the matcher is based on this code:

```ruby
 pattern = {
      srchParams: {
        srchString: 'Blog',
        seq: false,
      }.ignore_extra_keys!
    }.ignore_extra_keys!
```

For one, `seq` is not false, it's `nil`. That's probably a good first thing to try, since it'll "ignore extra keys" so perhaps the rest doesn't matter? But then we should add some extra stuff to the matcher to get a closer match. But not before first adding a commit to change `seq` and see if that works!! One step at a time. 

Also, would you mind reverting the changes to your Gemfile.lock? We'll have to do this to merge this PR.

orcutt989:
@jywarren how do I revert Gemfile.lock only?

Also where do you see the output for ```JSON.parse(last_response.body)``` ?

jywarren:
If you click Details in the Travis check, you can see the full log.

Re: reverting just one file, honestly i have to google this one each time
-- 
http://stackoverflow.com/questions/2733873/reverting-a-single-file-to-a-previous-version-in-git
:-)

orcutt989:
@jywarren You know - I did find that page, but I was still a big confused.  Don't I have to do a push afterwards? How does gitub know I reverted the file?

jywarren:
hmm, well, i had thought that it would affect your /next/ commit, but
perhaps it makes a commit itself? you can run `git log` to see if it's made
a new commit and `git status` to check if it's staged something for your
next commit.

orcutt989:
@jywarren Thanks. After bumbling around a bit I was able to get those changes committed. I used the hash from your last commit of GemFile.lock to revert it. I think its ok now. Also changed seq to nil instead of false.

test/functional/search_api_test.rb
```diff
@@ -0,0 +1,36 @@
...
+      }.ignore_extra_keys!
+    }.ignore_extra_keys!
+
+    matcher = JsonExpressions::Matcher.new(pattern)
+
+    puts JSON.parse(last_response.body)
+    assert matcher =~ JSON.parse(last_response.body)
+    
+
+    # check for actual Blog note, too
```
jywarren: OK, great -- this looks super and it's great to see the test passing now. Can you remove these comments, and add a few more items to the above search pattern, based on the output that the `puts` generated? (i copied this into an earlier comment, so you can work from there). 

Because at the moment, we don't actually test anything about the returned results -- see how the pattern is just a confirmation of the /request/ we made, not the returned search?

test/functional/search_api_test.rb
```diff
@@ -0,0 +1,36 @@
...
+    pattern = {
+      srchParams: {
+        srchString: 'Blog',
+        seq: nil,
+      }.ignore_extra_keys!
+    }.ignore_extra_keys!
+
+    matcher = JsonExpressions::Matcher.new(pattern)
+
+    puts JSON.parse(last_response.body)
```
jywarren: Also, we can now remove this debugging line again, now that we've captured the output and know what it's supposed to look like. Thanks!

jywarren:
This is going great. Just a couple more requests and we'll be able to merge it. Thanks a lot!!

jywarren:
Hi, @orcutt989 - just checking - can you remove the `puts` line now?

Also, see how the returned results has things like:

`"docUrl"=>"/notes/jeff/05-10-2017/blog-post", "docTitle"=>"Blog post"`

That means we can expand the pattern, so for example:

```js
 pattern = {
      items: [{
        docUrl: "/notes/jeff/05-10-2017/blog-post"
      }],
      srchParams: {
        srchString: 'Blog',
        seq: false,
      }.ignore_extra_keys!
    }.ignore_extra_keys!
```

And others from [the response we captured above](https://github.com/publiclab/plots2/pull/1407#issuecomment-300613087). Does that make sense? Then we'll be testing for a more specific response and if that ever fails in the future, the test will catch it.

orcutt989:
@jywarren My bad. Just removed the puts. What else would you like me to do?

jywarren:
Why don't we try expanding the pattern to include at least the URL or the title, so we can check the returned result. Then this should be good to go! Thanks for your persistence and good work!

orcutt989:
I added both the title and the URL as items in the pattern. Does that work?

orcutt989:
I thnk I got it. I am not sure about the indentation.

orcutt989:
How about now?

jywarren:
Hmm that's odd. This might take some manual trial and error on the rails console to really figure out. I'll see if I can give it a try tomorrow. Sorry it looks a little more complex than it originally seemed!

jywarren:
OK, let's see if this works...

jywarren:
Oh, you know what? the error may be the date in the URL... i think it's generated ad hoc at the time the tests are run!

So you could revert back to your version, and just test the title and say the `docId` but not the path, since the path includes the date. I wonder if that'd work??

jywarren:
On thinking this over, while the matcher is OK, it's not a great test because it doesn't actually return what's not matching. So while I'm going to leave it in, let's also do the path and title matching using a direct comparison, so we know what the bad data is if it fails. 

Here, this should pass. It's pretty close to what you had, and just uses the `node(:blog)` fixture reference instead of hard-coding the path. It passes locally -- let's see how it does here!

jywarren:
Great, it passed! Does my change make sense? 

Sorry this last step was such a tough one; your initial commit was almost all there, and we just got stuck on these last few lines. Your contribution's much appreciated though, and thanks for sticking with it! Sometimes these testing scenarios turn out a little more complex than it first appears. 

Merging now -- we'd love to help you with another if you want to look through https://github.com/publiclab/plots2/labels/fto-candidate or even https://github.com/publiclab/plots2/labels/help-wanted -- we can add more info to any of those to help you get started.

Thanks again!

PR #1420, opened 2017-05-18
typeahead search reworking

jywarren:
Hi, @david-days - finally got to do some more tweaking to the typeahead search and adjusted a few things; the code is very readable, thanks!

I did notice we don't seem to be using the separate `Doc`-based typeahead code, and I commented it out. We get both types (Tags, documents) via the same JSON API, and that seems fine (i did some extending of the API to include TagSource for both types). So we could adjust the naming to acknowledge the use of one type, or we could rewire things... I guess easier to just use less redundant code?

### DIFF
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -1,7 +1,7 @@
 require 'grape'
 require 'grape-entity'
 
-module Search
+module Srch
   class Search < Grape::API
     # Endpoint definitions
     resource :srch do
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -1,7 +1,7 @@
 require 'grape'
 require 'grape-entity'
 
-module Srch
+module Search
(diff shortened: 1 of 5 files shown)


jywarren:
The other thing I might like to try is to use the Bootstrap `typeahead()` code, which is less code for us, and gets us the ability to arrow up and down in the results list:

https://github.com/publiclab/plots2/blob/master/app/assets/javascripts/tagging.js#L63-L75

https://github.com/bassjobsen/Bootstrap-3-Typeahead

I think this wouldn't be too hard a changeover, and may give it a try.

jywarren:
Hmm, oh i think i see that the tag code here is not for "searching tags from the navbar input" but for "searching tags from a tag input" -- a usage we haven't connected up yet. I guess I'm inclined to work with the above-referenced `tagging.js` code for that. Looking into this a bit further.

PR #1424, opened 2017-05-19
user stats query optimization

jywarren:


jywarren:
Hmm, we'll see what happened here.

PR #1423, opened 2017-05-19
Upgrade mysql

icarito:
This is a first step of a two step-process to upgrade to MySql 5.7 in production.
Hopefully will help with https://github.com/jywarren/web/issues/94

Make sure these boxes are checked before your pull request is ready to be reviewed and merged. Thanks!

* [ ] all tests pass -- `rake test:all`
* [x] code is in uniquely-named feature branch, and has been rebased on top of latest master (especially if you've been asked to make additional changes)
* [x] pull request is descriptively named with #number reference back to original issue

PR #1384, opened 2017-04-26
Include custom configuration for mysql container

icarito:
Make sure these boxes are checked before your pull request is ready to be reviewed and merged. Thanks!

* [x] all tests pass -- `rake test:all`
* [x] code is in uniquely-named feature branch, and has been rebased on top of latest master (especially if you've been asked to make additional changes)
* [ ] pull request is descriptively named with #number reference back to original issue
Reference to internal sysadmin tracker: https://github.com/jywarren/web/issues/91

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/wiki/contributing-to-public-lab-software

We have a loose schedule of reviewing and pulling in changes every Tuesday and Friday, and publishing changes on Fridays. Please alert developers on plots-dev@googlegroups.com when your request is ready or if you need assistance.

Thanks!

PR #1428, opened 2017-05-22
Mysql upgrades

icarito:
This is to sync the container setup for the database with the code repository.
Note the mysql configuration and docker-compose are already in production.
This includes the bump in mysql version.
This PR supersedes #1384 and #1423, which I will close.

Additionally this PR comments out Solr and Web containers from docker-compose to make it on par with our production use pattern.


jywarren:
Looks good -- did you want me to merge this to stable so we can observe it? Also -- remember we wanted to have Jenkins set up to build from a `testing` branch that can be force-pushed to in order to try out features without "committing" (in the non-Git sense) to them? That is, in order to test things currently, we have to merge them to the `stable` branch, which is essentially deciding to adopt the commits. If we wanted to try something out but not make a final decision, we could force push it to this hypothetical `testing` branch and see how it builds, then make a decision about whether to merge it into `master`. 

Is that something you'd like to try here, or should we try to do that later and just assume we can make a commitment to this if it passes tests?

icarito:
Yes, except I seem to have mixed up a couple of commits, I'll fix this
PR and expect we can merge today, as this is already in production. (I
did it manually to look over the database conversion process).

Currently I did my testing directly on Tycho, without the web hook
auto-deploy. But this does not scale to other developers. I agree this
is needed, I've removed the /lowpriority/ tag on issue
https://github.com/jywarren/web/issues/80 ;-)

icarito:
Okay now it should be ready for merge. Thanks!

icarito:
Ah. We do need to uncomment docker-compose.yml to have tests run properly. I'll look at how to just run a single container in production without this hack to comment out the other containers.

jywarren:
OK - well, i'd like to merge new things to stable and publish to production tonight, if possible (as i'll be in meetings tomorrow) so give me a heads up if this becomes ready to merge in! Although I guess it's not stopping anything?

Will I now have to rewind production to merge in from stable?

icarito:
Hi,
The merge as I foresee it would be:
- Following file already exist untracked in production: `config/my.cnf`. Should be no conflict they are the same.
- Changes in `docker-compose.yml` really are only configuration for our production system, but they are making tests fail. 
 
One way to keep it simple would be to setup a separate `docker-compose-production.yml` file (untracked in git) to deal with tweaks in production. I'll research if there's a way to avoid to always remember to type 
`RAILS_ENV=production docker-compose -f docker-compose-production.yml ...` every single time, perhaps with a wrapper?

icarito:
> One way to keep it simple would be to setup a separate docker-compose-production.yml file (untracked in git) to deal with tweaks in production. I'll research if there's a way to avoid to always remember to type RAILS_ENV=production docker-compose -f docker-compose-production.yml ... every single time, perhaps with a wrapper?

Of course, Makefile should do it ;-)

icarito:
Also, to clarify, none of these changes are needed for today. The database is already running version 5.6 and mysql configuration files to be merged previously existed outside git anyway.

jywarren:
Hi, @icarito - does this have an update since we've upgraded? Just checking!

jywarren:
After today's changes, does that mean we're ready for this?

jywarren:
Just a note that i am currently backing up and checking out docker-compose.yml before git pull when i publish on production.

jywarren:
Is this still a change we'd like to merge in? Thanks!

ryzokuken:
@icarito are you still working on this? What is the current status?

Issue #1440, opened 2017-05-30
Adding a Comment button for subsections in wiki pages


aspriya:
### Feature 
This regarding the Long term project `Wiki Discussion`. A wiki page can broken down into subsections and can be edit per subsection basis. As the project describes, a wiki should be more interactive. Users having the ability to comment on section basis of is one step for acheiving that.

### Relevant URLs
A detailed description of the procet can be found in following links
https://github.com/publiclab/plots2/issues/1438
https://publiclab.org/notes/Ashan/03-13-2017/wiki-discussion

### Methodology
When a logged in user pass a query string paremeter `raw=true` for a wiki page, the rendering of that page is switched from `RDiscount` to `publiclab/inline-markdown-editor`. Doing so `inline-markdown-editor` brakes the wiki page into subsections on the basis two new line characters (`\n\n`). `Inline-markdown-editor` library is coded so that it adds an edit button for every on of these broken subections. To add a comment button, in `wiki.js` we have add a new property called `extraButtons` for the object we are sending to the `inline-markdown-editor` module. Value of that property should also be a object which has a property to identify the button type (So this will be something like `fa-comment`). This 'fa-comment` property should be assigned with a setup function to setup the comment functionality. Creating that function is another task, So with this feature is only to add a comment button for every commentable subsection.

Issue #1438, opened 2017-05-29
Planning issue for Wiki Discussion GSoC project


aspriya:
### Why
Wiki pages are web pages that anyone can create or edit. PublicLab use them to collect information, documentation and instructions on projects. At present users can create wikis, edit them and associate tags with them. But users can not add comments to this, can not make suggestions, can not tag a user for a special part in the wiki. Wikis can made more interactive by implementing these features. For this I will use the publiclab/inline-markdown-editor.

### Phase 1

- [ ] Add a comment button aside of the pencil button (edit button)
- [ ] Building a inline form for commenting
- [ ] Wirte tests to check the functionlity of comment and building forms attributes.
- [ ] Add view suggestions and view Comments Buttons

### Phase 2

- [ ] Make a table to store comments for a perticuler wiki page
- [ ] Make the comment form submittable
- [ ] Write tests to check the Commenting functionality
- [ ] List suggestions when view suggestions button is clicked.
- [ ] create two functions to approve and discard suggestions by a valid user.
- [ ] Write tests to check the functionality of approval fucntion and discarding function.
- [ ] List comments when view Comments button is clicked.
- [ ] Create a function to remove comments by admin or moderators.
- [ ] Check the functionality of appearing of comments by tests.

Will keep updating this thread as and when things get clearer. Please feel free to provide feedback and suggestions.

jywarren:
I'm thinking some rearranging could make sense so that we don't have to merge in a non-functional comment button before the commenting functionality is ready.

So, what's the first step in getting a commenting form to appear? Let's think through where we'd want to make the form ([template here](https://github.com/publiclab/plots2/blob/master/app/views/comments/_form.html.erb)) appear.

Even before that, maybe we want to create a functional test of adding a comment to a wiki page. This could be a good project to break out and make a checklist out of. Comment controller is functionally tested here: https://github.com/publiclab/plots2/blob/master/test/functional/comment_controller_test.rb#L20-L30

Is there anything stopping us from making a comment attached to a wiki page instead of a note? Could we test that out in the comment controller functional test, as a first step? 

(The above is copied in [from here](https://github.com/publiclab/plots2/pull/1441#discussion_r119249749), but could be used to edit/revise the checklist above.) 

Also - thanks for posting this checklist, it looks great!!! Very helpful.

jywarren:
[...]

PR #1441, opened 2017-05-30
Wiki discussion-Adding a comment button for subsections in wiki pages #1440

aspriya:

Thanks!


app/assets/javascripts/wikis.js
```diff
@@ -8,7 +8,8 @@ function setupWiki(node_id, raw, logged_in) {
       selector: '#content-raw-markdown',
       wysiwyg: true,
       preProcessor: preProcessMarkdown,
-      postProcessor: postProcessContent
+      postProcessor: postProcessContent,
+      extraButtons: {"fa-comment": true}
```
jywarren: Ah, this is looking good, but is it perhaps not ready until we actually have a function that this button will do? I think a good place to start will be to integrate this into an example in https://github.com/publiclab/inline-markdown-editor, showing that when you click the button, something runs (it could even be a simple `alert(some_output_here)` that the user can see working in the test. 

Then, we'd add to this PR a function that would run to actually show a comment form. Until then, I think we don't want to actually merge this PR in, but it's great to see it here since it will definitely be an important step once the comment form actually works. 

So, what's the first step in getting a commenting form to appear? Let's think through where we'd want to make the form (shown here: https://github.com/publiclab/plots2/blob/master/app/views/comments/_form.html.erb) appear. 

Even before that, maybe we want to create a functional test of adding a comment to a wiki page. This could be a good project to break out and make a checklist out of. Comment controller is functionally tested here: https://github.com/publiclab/plots2/blob/master/test/functional/comment_controller_test.rb#L20-L30

Is there anything stopping us from making a comment attached to a wiki page instead? Could we test that out in the comment controller functional test, as a first step?

jywarren: I see that maybe my comments are more appropriate on #1438 -- great to see you've posted that! Shall I copy this over? I'll copy over the last 3 paragraphs.

jywarren:
I'm going to close this for now but let's re-open when it comes back up in the checklist, ok?

PR #1459, opened 2017-06-13
List Users Following tags

StlMaris123:
#1421 

Thanks!


StlMaris123:
I have done something on:
`Node has X tags`
`(revised) Get all users who are following the new tag`

Please shed some light on 
`For each of these following users, none of the existing tags is being followed by the user`
`the user has not already gotten an email for that node creation or any of its tag creations`

StlMaris123:
What could be the issue causing Travis_ci to fail?

jywarren:
hi @publiclab/soc - can anyone look through the logs here and help @StlMaris123 debug? Thanks!

ryzokuken:
@StlMaris123 There's a pending migration. Maybe you could try migrating the database locally by running `rake db:migrate`? That  *should* solve the issue.

Great work, by the way.

jywarren:
Ah, if there's a pending migration you need to update the schema file, i think?

db/migrate/20170612162455_add_email_sent_at_to_rusers.rb
```diff
@@ -0,0 +1,5 @@
+class AddEmailSentAtToRusers < ActiveRecord::Migration
```
jywarren: But I think we want to separate out this migration into its own PR, right? because this PR would be just for the model method `followers_who_dont...` -- right? If you can move this migration to its own branch, then remove it from this PR then things should pass. Good catch, @ryzokuken!

ryzokuken: Glad to be of help

test/models/user_test.rb
```diff
@@ -0,0 +1,21 @@
+require 'test_helper'
+class UserTest < ActiveSupport::TestCase
```
jywarren: Also, here, I don't see an explicit test of `followers_who_dont_follow_tags` -- but is this still in progress? Great to see the test shaping up here!

ryzokuken:
@StlMaris123 again, look's like you need to migrate.

StlMaris123:
Splitting this into two PR's:
- return a list of users who follow this tag, but don't follow the list of given tag
- A check that user has not received any email

PR #1466, opened 2017-06-15
Solr tweaks

jywarren:

Thanks!


jywarren:
Again back to https://github.com/sunspot/sunspot#manually-adjusting-solr-parameters

Based on console testing, it could also work to remove `&defType=edismax`.

jywarren:
OK, here i'm stuck -- bc based on my testing on the Solr console on staging, we need to set `params[:qf] = ""` -- but that seems to break the tests using embedded Solr. 

I can try some other variations, like `params[:qf] = "''"` -- empty quotes -- but maybe we should instead try removing the `edismax` part...

jywarren:
Trying `.delete(:defType)`... odd but who knows

jywarren:
Should be able to test this with http://staging.laboratoriopublico.org/searches/test/test?q=stk500_recv

jywarren:
Unfortunate... @icarito -- any inspirations? We know the exact Solr query we want, we just can't get Sunspot to create it. 

Alternatively, we aren't sure why the default Sunspot-generated query (with `qf=title_text+body_text+comment_text` and `defType=edismax`) doesn't work -- removing either of those portions of the query gets us full results.

jywarren:
Added more detail to [the sunspot issue](https://github.com/sunspot/sunspot/issues/863) and opened one on Stack Overflow: https://stackoverflow.com/questions/44617017/difficulty-getting-more-than-exact-field-match-results-on-sunspot-solr-with-qf

jywarren:
Trying to set `params[:defType]` to an empty string, which on rails console testing resulted in:

```
irb(main):011:0> s = Node.search { fulltext('balloon');adjust_solr_params { |p| p[:defType] = '' } }.results
  SOLR Request (109.0ms)  [ path=select parameters={fq: ["type:Node"], q: "balloon", fl: "* score", qf: "title_text body_text comments_text", defType: "", start: 0, r
ows: 30} ]
```

And full results...

jywarren:
OK, another variation which worked on external Solr but did not pass embedded solr tests:

```ruby
Failure:
  <[]> expected to be != to
  <[]>.
test_should_get_search_test_action(SearchesControllerTest)
test/solr/searches_controller_test.rb:23:in `block in <class:SearchesControllerTest>'
```

Sebastian -- the alternative here is that we could simply **drop support of the embedded gem `sunspot_solr` engine** (which is [for development only anyways]()) and spin up the containerized Solr in Travis. 

What would that take? Changes to:

* uncomment the solr container in docker-compose.yml (for purposes of Travis only)
* repoint `/config/sunspot.yml` to the container (just the port number, i guess)
* would the [solr-specific test.rake lines](https://github.com/publiclab/plots2/blob/master/lib/tasks/test.rake#L33) require tweaks so that the commands to start/stop the solr engine would go to the containerized solr, not the embedded solr? 

This seems like a way forward even if we don't figure out what's going on. I suspect the embedded solr is just an older version and is not responding to the same queries in the same ways. Maybe it's not using `edismax` queries, for example.

jywarren:
Actually i like this plan a lot. It's closer to how we'll be deploying anyways.

jywarren:
Attempting these steps now -- not 100% sure of the `sunspot.yml` or `.travis.yml` or `test.rake` changes, but let's see how it does!

jywarren:
Seems close -- solr container starts up, but it seems to run /after/ `bower install` for some reason, though i tried adding 40 seconds wait. The last thing that runs is WeBrick starting Rails, but i never see `rake db:migrate` run, or tests.

jywarren:
Hmm, so what is going on here? Can you push up a copy without logging on, not a force push? What's the specific error?

jywarren:
```

solr_1  | 17863 INFO  (qtp1112280004-13) [   x:default] o.a.s.c.S.Request [default] webapp=/solr path=/select params={q=test&defType=edismax&qf=title_text+body_text+comments_text&fl=*+score&start=0&fq=type:Node&rows=30&wt=ruby} status=400 QTime=26 
solr_1  | 17890 ERROR (qtp1112280004-14) [   x:default] o.a.s.c.SolrCore org.apache.solr.common.SolrException: undefined field type
solr_1  | 	at org.apache.solr.schema.IndexSchema.getDynamicFieldType(IndexSchema.java:1298)
solr_1  | 	at org.apache.solr.schema.IndexSchema$SolrQueryAnalyzer.getWrappedAnalyzer(IndexSchema.java:444)
solr_1  | 	at org.apache.lucene.analysis.DelegatingAnalyzerWrapper$DelegatingReuseStrategy.getReusableComponents(DelegatingAnalyzerWrapper.java:74)
```

Hmm, this is why you were talking about the schema again. Is there an ordering issue? Or do we need to run the reindex? I don't know... why does this happen here and not in the container you already made?

jywarren:
Are we sure that Solr is: a) up and running and b) able to see mySQL/Rails by the time we run `rake db:setup`?

icarito:
Okay I have the following to report on this:

Embedded Solr is version 5.3.1.
The container version of Solr is 5.3.2, selected because it was the closest version available.

Solr admin of the embedded Solr engine is available in Tycho at http://tycho.media.mit.edu:8982/solr/#/
Solr admin from production is available but firewalled (access with `ssh pad.publiclab.org -p 22 -nN -L8983:localhost:8983`, then at http://localhost:8983/solr/#/).

From here I can find that there's a `schema.xml` which seems to belong to Sunspot and is setting up types for Solr. I think we can add the `configsets` directory to git, and mount it as a volume within the container. I'll try that.

I found a very good reference about running Solr in a container: https://www.norconex.com/use-solr-5-with-docker/

We're close!

icarito:
We're so close! https://github.com/docker/compose/issues/3593 gives a solution for copying the config files _after_ the core has been created. I'll try it out.

icarito:
`docker cp` can only copy single files! :-/
Looking for better options configuring the container at https://hub.docker.com/_/solr/

icarito:
Hi @jywarren I managed to get the Solr engine working with the Sunspot configset! Now an apparently unrelated test failed, I'd appreciate a look! Thanks!

icarito:
Actual error is:
```

===============================================================================
Failure:
  <["test", "awesome", "spectrometer", "activity:spectrometer", "hashtagshere"]> expected to include
  <"hashtags">.
test_should_tag_hashtags_in_headers(RevisionsTest)
test/unit/revision_test.rb:123:in `block in <class:RevisionsTest>'
===============================================================================
```

jywarren:
Uhhhhh exciting! but that's a real mystery... it's not failing on anything else... is there any possible way this could relate to our changes? Gotta think about this in the morning.

jywarren:
I can't find any instance of `hashtagshere` anywhere in the codebase. I pushed up a rebased version of this PR.

jywarren:
Ok, for what it's worth, i think the `hashtagshere` line may come from these two lines: https://github.com/publiclab/plots2/blob/master/test/fixtures/node_revisions.yml#L117-L118

```
body: |+
    # Heading 1 is about #hashtags
    Here is the message
```

I think it may have to do with the `|+` operator in Yaml, which i'm not familiar with.

jywarren:
I removed the `+` -- going to sleep now!

jywarren:
Oh no! now we have the same error as we had! How is that the case -- can we see if it's actually running them on the containerized solr, or if the containerized solr for some reason resulted in the same error as we'd only previously seen on the embedded solr?

icarito:
I'll try to issue a reindex, perhaps the mock data isn't being indexed automatically?
I don't remember, did the embedded solr fail with this same error? I'll also review if the parameters need to be adjusted. Still, progress!

jywarren:
I thought it did, yes, almost exactly this error. You're right, we should
check to see that the index is occurring.

jywarren:
Yeah, indeed i don't see a line where solr indexes! That must be the missing part!!!

jywarren:
hmm, i don't see what that last merge commit did... isn't it just the same as the last change I made?

jywarren:
Oh sorry i had to refresh the page.

icarito:
I was expecting that doing db:setup would properly index the mock data, but added a reindex step now. Also moved to output logs after that step, hoping to learn what's going on.

icarito:
Ok, so the reindex appears to be working fine. I'll try to reproduce in a local environment.

jywarren:
Hi, Sebastian -- how's this going?

icarito:
I don't why, I thought we had solved it. Likely needs a last push, will review it now. Sorry!

icarito:
I've merged this into `unstable` branch so that we can use jenkins to track this.
This made me notice that in our current setup, staging and production are sharing the same pad.publiclab.org Solr index! At one point I stopped the Solr container in pad.publiclab.org and experienced a 500 error in publiclab.org/dashboard! So I think we need to doublecheck our fallbacks.

jywarren:
Yikes! So staging should have its own index, right? Is that then in
sunspot.yml?

icarito:
Yes! I've closed the firewall at pad.publiclab.org to only accept connections from _publiclab.org_ (dropping requests from _tycho.media.mit.edu_). That should ensure isolation from staging. I'm going to set SOLR_URL in Jenkins, it should override the config in `sunspot.yml`.

icarito:
In fact, having firewalled tycho from pad.publiclab.org should affect staging and allow to debug the fallback. Checking it...

icarito:
I'm not seeing the staging server fail after restricting access to pad.publiclab.org. I'll check it further tonight.

icarito:
My fault, I had put the wrong address in SOLR_URL env variable in jenkins config. It appears to be working fine after changing it.

icarito:
I think staging is working except without access to Solr. I'll have a closer look tonight.

icarito:
Okay I was trying to reindex tonight and got this:
```
icarito@tycho:/srv/plots_testing/plots$ RAILS_ENV=production SOLR_URL=http://solr:8983/solr/ docker-compose exec web rake DISABLE_SOLR_CHECK=1 sunspot:solr:rei
ndex
rake aborted!
Errno::ETIMEDOUT: Connection timed out - connect(2) for "pad.publiclab.org" port 8983
(eval):2:in `post'
Tasks: TOP => sunspot:solr:reindex => sunspot:reindex
(See full trace by running task with --trace)
icarito@tycho:/srv/plots_testing/plots$
```
This is because the firewall is set to DROP and I guess the Solr-check is timing out instead of immediately failing, taking a long time. I'll make a patch to enable Solr container in `docker-compose.yml` and also try setting the firewall to REJECT, to see if the Solr-check will work better in that case.

icarito:
@jywarren sincerely I've spent a lot of time on this one and I can't tell why it's failing. I would appreciate your help.

jywarren:
Ok we can look at this tomorrow! Thanks for spending time on this. 

My first thought is can we view and reproduce this query directly on solr console on staging.

jywarren:
http://staging.laboratoriopublico.org/searches/test?q=spectrometer works

icarito:
Today Jeff and I worked on this, configuring the staging environment to run from the _test_ RAILS_ENV, and discovered that the indexer was indexing the _seed_ data instead of the _test_ data. We're attempting to reindex _after_ the original tests have been run, to see if it will include the test data.

icarito:
YESS!

jywarren:
HOORAAYYYYYYYY

lets get this on staging fast!

jywarren:
Ready to merge, right? if so, can you hit merge and i'll push to staging?

icarito:
Yes, please!

icarito:
Note that staging will be broken until I restore the staging database from a dump. I can do it later today.

icarito:
Staging is back with a reindexed solr container, all lights green.

PR #1467, opened 2017-06-15
removing code from search_service.rb

jywarren:
Just checking @david-days -- this was marked for deprecation so taking it out to simplify?


jywarren:
I've read through this and am pretty sure this will be OK and no code touches this. We can double check on Staging!

david-days:
@jywarren--yeah, this is a good call--the deprecated code was a copy of the original controller code, but I didn't want to make too many changes at once.

Late to the party again!

:-)

PR #1471, opened 2017-06-17
.gitignore Gemfile.lock

jywarren:
@icarito and @ananyo2012 -- i know we've been through this a bit, but I wanted to try adding this since almost every contributor is getting stuck on the Gemfile.lock changes. And when it is necessary, we can always do `git add Gemfile.lock` -- right? or maybe an `-f`?

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ data
 .DS_Store
 .idea
 .ruby-version
+Gemfile.lock
 coverage*
 development.sqlite
 test.sqlite


ananyo2012:
Well commiting Gemfile.lock is a convention so people can use the same gem versions throughout each repo. I am not sure why some gems keep updating. Have to list out them and see through. Till then you can merge this one. But we should re add Gemfile.lock after this is solved.

ryzokuken:
The gemfile.lock changes only if bundler tried to installed the specified version but could not. In which case, we *should* commit the changes so that the next person does not face any problems.

I think the issue is that the gemfile is outdated, and therefore changes for every new contributor who installs the gems anew. Once we regenerate the file and commit those changes, chances are that there won't be any more gemfile.lock issues.

@ananyo2012 what do you say?

ananyo2012:
Yes thats what I think too. We should perhaps commit the changes in Gemfile.lock.

jywarren:
Will there be any variations due to different processor architecture?

jywarren:
OK - so that makes sense, but this will happen any time any gem is updated, which could be quite often. Note that I'm not saying we stop tracking Gemfile.lock, we'd leave it in the repository. You'd just have to manually force-add it to update it. So what do you think of this step -- the .gitignore -- and we can check in a new version from the production server (so that we're sure that it runs on there)?

jywarren:
So we would set a schedule -- say, every 3 months? To update Gemfile.lock and ensure it runs on production.

jywarren:
Counter case -- anytime someone /does/ submit a Gemfile.lock, and tests pass, we could decide "that's good enough" and just merge it. Gemfile.lock changes will still happen sometimes due to the `bundled by version ___` messages, though.

ryzokuken:
@jywarren

**EDIT:** I am not sure, but the Gemfile.lock will not be updated every time a new gem version is available, but everytime **our** version of the gem goes *unavailable*. Will need sources to double up that one though.

jywarren:
I think it's updated any time anyone runs `bundle install` -- to the latest versions of permissible gems.

jywarren:
Having thought this through, I'm going to make a quarterly schedule and start doing this on a schedule, so i'll merge this. Thanks for your thoughts, all!

ananyo2012:
Oh I saw you already did that in #1474

PR #1472, opened 2017-06-18
Followers not following given tags fixes#1421

StlMaris123:

Thanks!

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -201,4 +201,11 @@ class Tag < ActiveRecord::Base
         .order('node_revisions.timestamp DESC')
         .limit(limit)
   end
+
+  def followers_who_dont_follow_tags(tags)
+    tag_followers = self.subscriptions.map {|user| user.user_id}
+    following_given_tags = tags.map { |tag| tag.subscriptions  }.flatten
+    following_given_tags_ids = following_given_tags.map { |userid| userid.user_id }
+    tag_followers.reject { |userid| following_given_tags_ids.include? userid }
+  end
 end
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -204,8 +204,12 @@ class Tag < ActiveRecord::Base
 
   def followers_who_dont_follow_tags(tags)
     tag_followers = self.subscriptions.map {|user| user.user_id}
+    users = User.find_all_by_id([tag_followers])
+    p tag_followers_names = users.map { |user| user.username }
     following_given_tags = tags.map { |tag| tag.subscriptions   }.flatten
     following_given_tags_ids = following_given_tags.map { |userid| userid.user_id  }
-    tag_followers.reject { |userid| following_given_tags_ids.include? userid  }
+    foll_users = User.find_all_by_id([following_given_tags_ids])
+    following_given_tags_names = foll_users.map { |user| user.username }
+    tag_followers_names.reject { |userid| following_given_tags_names.include? userid  }
   end
 end
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -204,8 +204,8 @@ class Tag < ActiveRecord::Base
 
   def followers_who_dont_follow_tags(tags)
     tag_followers = self.subscriptions.map {|user| user.user_id}
-    following_given_tags = tags.map { |tag| tag.subscriptions  }.flatten
-    following_given_tags_ids = following_given_tags.map { |userid| userid.user_id }
-    tag_followers.reject { |userid| following_given_tags_ids.include? userid }
+    following_given_tags = tags.map { |tag| tag.subscriptions   }.flatten
+    following_given_tags_ids = following_given_tags.map { |userid| userid.user_id  }
+    tag_followers.reject { |userid| following_given_tags_ids.include? userid  }
   end
 end
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -205,6 +205,6 @@ class Tag < ActiveRecord::Base
   def followers_who_dont_follow_tags(tags)
     tag_followers = User.find(self.subscriptions.collect(&:user_id))
     following_given_tags = User.find(tags.collect { |tag| tag.subscriptions.collect(&:user_id)}).flatten
-    tag_followers.reject { |userid| following_given_tags.include? userid  }
+    tag_followers.reject { |user| following_given_tags.include? user  }
   end
 end
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -203,13 +203,8 @@ class Tag < ActiveRecord::Base
   end
 
   def followers_who_dont_follow_tags(tags)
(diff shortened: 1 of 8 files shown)


StlMaris123:
The above method gets users who are following this tag but not the given tags
1. Get the users following this tag -done
2. Get the users following the given tags -done
3. Reject the users in step 2 and are included in step 1 -done

test/models/tag_test.rb
```diff
@@ -0,0 +1,8 @@
+require 'test_helper'
+class TagTest < ActiveSupport::TestCase
+  test "it returns users following this tag but not given tags" do
+    tag_followers = [1,4,5]
```
jywarren: Looking cool, but where do these numbers come from? I think we should try to show, in the test, that these users are following the tag -- assert that in a sequence of assertions here to help demonstrate the test more clearly. 

Also, rather than using the uids directly, best use `ruser(:Bob)` and so forth -- use them via the standard syntax for fixtures. Make sense?

StlMaris123:
yes @jywarren.  working o  it

jywarren:
Hi, StellaMaris - re: our discussion today, this is a perfectly-sized PR -- it does one thing and tests it. Once we figure it out and merge it, the next one could be in #1421, the one labelled ` A new method in called notify_tag_added ` -- make sense?

StlMaris123:
@jywarren That is okay, I had been stuck on the test for the past two days but now am on track. Hopefully I will be done by today then I move to `notify tag added`

jywarren:
Super. Note there's a method called followers in the tag model already,
that you can use or model off of.

https://github.com/publiclab/plots2/blob/master/app/models/tag.rb#L135

StlMaris123:
@jywarren , check on the updates and give feedback. I realised that when a user subscribes to the tag its actually the `TagSelection` table which used. I have used this to create tag followings for the tests.  I feel confident and ready for the next one :smiley:

app/models/tag.rb
```diff
@@ -201,4 +201,11 @@ def self.find_research_notes(tagnames, limit = 10)
         .order('node_revisions.timestamp DESC')
         .limit(limit)
   end
+
+  def followers_who_dont_follow_tags(tags)
+    tag_followers = self.subscriptions.map {|user| user.user_id}
```
jywarren: Why not just use `self.followers`? https://github.com/publiclab/plots2/blob/master/app/models/tag.rb#L135 -- maybe there's a good reason that I'm missing?

test/fixtures/tags.yml
```diff
@@ -63,3 +63,5 @@ replications:
 activities-spectrometer:
   tid: 16
   name: activities:spectrometer
+
```
jywarren: Some extra space here? Could you remove it?

test/models/tag_test.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  test " returns users following this tags but not given tags" do
+    # users following tag are 1,4,5
+    # users following tag1 are 1,4,6
+    # users following tag2 are 3,7
+    tag = tags(:test)
+    tag1 = tags(:awesome)
+    tag2 = tags(:spam)
+    given_tags = [tag1, tag2]
+    assert_equal [5], tag.followers_who_dont_follow_tags(given_tags)
```
jywarren: Here, I'd like to be able to do `assert_equal` with a full user. Something like: 

```ruby
assert_equal [rusers(:bob)], tag.followers_who_dont_follow_tags(given_tags)
```

jywarren:
@icarito @ujithaperera how's this look?

app/models/tag.rb
StlMaris123: Hi, @jywarren that did not work, from my understanding tags have following through tag_selection

StlMaris123:
@jywarren @ujithaperera @icarito I have updated the tests to use `usernames` instead of `user_ids`. The method  `followers_not_following_given_tags ` has however become more verbose as I tried to get the usernames. Please have a look and tell me what you think. Thank you.

app/models/tag.rb
jywarren: There's a function for collecting all the properties from an array of objects, `.collect(&:id)` will return an array of all `id`s from an array of users. Could help to make it more readable!

But I guess if you're looking for followers, you can do: `self.subscriptions.collect(&:user) -- see: https://github.com/publiclab/plots2/blob/master/app/models/tag_selection.rb

Although maybe you need to refactor that to use `rusers`!

app/models/tag.rb
```diff
@@ -201,4 +201,15 @@ def self.find_research_notes(tagnames, limit = 10)
...
+    users = User.find_all_by_id([tag_followers])
```
jywarren: Also here you can just do `User.find(uids)` -- or if you don't want to refactor DrupalUsers above, you can do `User.find(self.subscriptions.collect(&:user_id))` and it'll run on the whole array of ids.

I think `collect()` can make this code a lot more readable. It's hard to see what's going on! But now that you have a test, you can mess with it and always know if your code is correctly fulfilling the test :-)

test/models/tag_test.rb
```diff
@@ -0,0 +1,27 @@
...
+    given_tags = [tag1, tag2]
+    assert_equal ['palpatine'], tag.followers_who_dont_follow_tags(given_tags)
+  end
+
+  test 'returns all users in this tag if none is following the given tags' do
+    tag = tags(:spam)
+    tag2 = tags(:test)
+    tag1 = tags(:awesome)
+    given_tags = [tag1, tag2]
+    assert_equal ['newcomer', 'spammer'], tag.followers_who_dont_follow_tags(given_tags).sort
```
jywarren: Also here, you'll want to test these against actual full users, not usernames. You can `assert_equal [rusers(:newcomer), rusers(:spammer)], tag.followers_who_dont_follow_tags(given_tags).sort` -- it's generally better to pass around the whole object and not just one field!

test/unit/comment_test.rb
```diff
@@ -53,7 +53,7 @@ class CommentTest < ActiveSupport::TestCase
     comment = Comment.new(nid: node(:one).nid,
                           uid: rusers(:bob).id)
     comment.comment = 'Hey, #everything, #awesome followers.'
-    assert_equal comment.followers_of_mentioned_tags.length, 2
+    assert_equal comment.followers_of_mentioned_tags.length, 3
     assert comment.followers_of_mentioned_tags.collect(&:id).include?(rusers(:bob).id)
```
jywarren: Here, couldn't you just do `assert comment.followers_of_mentioned_tags.include?(rusers(:bob))` ?

StlMaris123:
Refactoring really helped :smile: , in `app/models/tag.rb` I was left with only 3 lines of code which is more readable.

test/unit/comment_test.rb
StlMaris123: That one has already been included in line 57, so I just updated the third `ruser`

PR #1473, opened 2017-06-19
Update wikis.js to connect inline questions back to origin wiki page

jywarren:

Thanks!

### DIFF
--- a/app/assets/javascripts/wikis.js
+++ b/app/assets/javascripts/wikis.js
@@ -12,7 +12,7 @@ function setupWiki(node_id, title, raw, logged_in) {
       wysiwyg: true,
       preProcessor: preProcessMarkdown,
       postProcessor: postProcessContent,
-      extraButtons: {"fa-question": questionForm}
+      extraButtons: { "fa-question": questionForm }
     });
     $('#content').hide();
   } else {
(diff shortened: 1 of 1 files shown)


jywarren:
@aspriya - just a note here -- i made a slight change as a followup, using a power tag (which we haven't done much with yet, but just for consistency) to try to link the question back in to the wiki page via `node_id`, and also moved the `questionForm` function into the `setupWiki` function so that it doesn't exist globally and also so it gets access to the `node_id` parameter. Make sense?

app/assets/javascripts/wikis.js
```diff
@@ -12,14 +12,20 @@ function setupWiki(node_id, title, raw, logged_in) {
...
     $('#content').hide();
   } else {
     $('#content').html(shortCodePrompt($('#content')[0], { submitUrl: '/wiki/replace/' + node_id }));
     postProcessContent();
     addDeepLinks($('#content'));
   }
+
+  function questionForm(qbutton, uniqueId) {
+    wiki_title = wiki_title.replace(/ /g, "-");
+    qbutton.attr('href','/questions/new?tags=response:' + node_id + '&question%3A' + wiki_title + ', ' + wiki_title + '-' + uniqueId+', a-wiki-question&template=question&redirect=question').attr('target', '_blank');
```
aspriya: @jywarren, I think here `&` symbol in `&question%3A` should be changed into a `,` symbol. Because with this change other tags are not setting correctly

aspriya:
Ah yes, that makes sense. Any I think i found a small bug here:
https://github.com/publiclab/plots2/pull/1473/files#diff-5964ca08334d80fd5d91e87d38b7150eR26

jywarren:
Can you open a new PR with the fix?

jywarren:
Oops you're right!

aspriya:
Sure will do :+1:

PR #1509, opened 2017-07-07
Update notes_controller.rb for direct images

jywarren:

should improve dashboard of SpectralWorkbench.org


icarito:
The reason this fails in staging is that the database has inconsistencies:
- Some `rusers` don't have a corresponding `users` record.
- Some users have strange username characters (which currently confuses ActiveRecord?)
I'll make a pull request that hopefully works around the first issue. Not sure how to tackle the second.

icarito:
Btw I had to enter `mysql` and manually drop / create the half-migrated fields, in order to attempt the migration again. `rollback` appears to start rolling back from the last successful migration.

PR #1512, opened 2017-07-08
Issue 1337

### LINKED ISSUE #1337: Some recently edited methods pages don't sort to the top of /methods

scosbet:
Fixes #1337 but missing test (see issue description for explanation of why) and format of query could potentially be improved, any guidance with that would be appreciated!

One test is failing but it's also failing on master so not related to my change

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -364,6 +364,3 @@ DEPENDENCIES
   uglifier (>= 1.0.3)
   will_paginate (>= 3.0.6)
   will_paginate-bootstrap (>= 1.0.1)
-
-BUNDLED WITH
-   1.14.6
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -364,3 +364,6 @@ DEPENDENCIES
   uglifier (>= 1.0.3)
   will_paginate (>= 3.0.6)
   will_paginate-bootstrap (>= 1.0.1)
+
+BUNDLED WITH
+   1.14.6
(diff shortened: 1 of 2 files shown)


ryzokuken:
The failing test issue was fixed, I suppose, and your PR does not show a failure anymore, so cheers to that.

Off-topic, but #1337 sounds like a pretty rad issue number :stuck_out_tongue:. Good work, and thanks for contributing.

jywarren:
This looks fantastic. 

Merging now. If you're looking for a new challenge, please take a look at our `help-wanted` list: https://github.com/publiclab/plots2/labels/help-wanted

Thanks!

PR #1513, opened 2017-07-08
[API] Commenting through a token

ryzokuken:
@jywarren take a look. The function does not do anything yet, but a laid out a possible flow about how it could work. Maybe you had something else in mind? I'd be happy to incorporate your suggestions.

```ruby
def create_by_token
    # Parse the user's token from the headers

    # Use logic similar to the above create controller but with the user that
    # the token belongs to instead of the `current_user`

    # In order to avoid looking up the entire database for a user having a
    # matching token, we could instead ask the user to pass username/id in the
    # params along with the node id and comment body.
    # This would also mean that if someone tries to bruteforce a token, they
    # would to do so for each user individually.

    # If the username/id and token do not match, send an error in the same
    # format as the request (JSON/XML)
  end
```

### DIFF
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -48,6 +48,9 @@ class CommentController < ApplicationController
 
     if @user && @user.token == @token
       begin
+        # The create_comment is a function that has been defined inside the
+        # CommentHelper module inside app/helpers/comment_helper.rb and can be
+        # used in here because the module was `include`d right at the beginning
         @comment = create_comment(@node, @user, @body)
         respond_to do |format|
           format.all { render :nothing => true, :status => :created }
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -36,6 +36,22 @@ class CommentController < ApplicationController
     end
   end
 
+  def create_by_token
+    # Parse the user's token from the headers
+
+    # Use logic similar to the above create controller but with the user that
+    # the token belongs to instead of the `current_user`
+
+    # In order to avoid looking up the entire database for a user having a
+    # matching token, we could instead ask the user to pass username/id in the
+    # params along with the node id and comment body.
+    # This would also mean that if someone tries to bruteforce a token, they
+    # would to do so for each user individually.
+
+    # If the username/id and token do not match, send an error in the same
+    # format as the request (JSON/XML)
+  end
+
   # create answer comments
   def answer_create
     @answer_id = params[:aid]
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -49,32 +49,17 @@ class CommentController < ApplicationController
     if @user && @user.token == @token
       begin
         @comment = create_comment(@node, @user, @body)
-        msg = {
-          status: :created,
-          message: 'Created'
-        }
         respond_to do |format|
-          format.xml { render xml: msg.to_xml }
-          format.json { render json: msg.to_json }
+          format.all { render :nothing => true, :status => :created }
         end
       rescue CommentError
-        msg = {
-          status: :bad_request,
-          message: 'Bad Request'
-        }
         respond_to do |format|
-          format.xml { render xml: msg.to_xml }
-          format.json { render json: msg.to_json }
+          format.all { render :nothing => true, :status => :bad_request }
         end
       end
     else
-      msg = {
-        status: :unauthorized,
-        message: 'Unauthorized'
-      }
       respond_to do |format|
-        format.xml { render xml: msg.to_xml }
-        format.json { render json: msg.to_json }
+        format.all { render :nothing => true, :status => :unauthorized }
       end
     end
   end
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -47,12 +47,12 @@ class CommentController < ApplicationController
       # TODO: Do the usual commenting stuff
     else
       msg = {
-        :status => :unauthorized
-        :message => "Unauthorized"
+        status: :unauthorized
+        message: "Unauthorized"
       }
       respond_to do |format|
-        format.xml { render :xml => msg.to_xml }
-        format.json { render :json => msg.to_json }
+        format.xml { render xml: msg.to_xml }
+        format.json { render json: msg.to_json }
       end
     end
   end
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -42,9 +42,9 @@ class CommentController < ApplicationController
 
   def create_by_token
     @node = Node.find params[:id]
-    @user = Users.find_by_username params[:username]
-    @body = pararms[:body]
-    @token = request.headers["token"]
+    @user = User.find_by_username params[:username]
+    @body = params[:body]
+    @token = request.headers['Token']
 
     if @user && @user.token == @token
       begin
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -47,7 +47,7 @@ class CommentController < ApplicationController
       # TODO: Do the usual commenting stuff
     else
       msg = {
-        status: :unauthorized
+        status: :unauthorized,
         message: "Unauthorized"
       }
       respond_to do |format|
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -44,14 +44,14 @@ class CommentController < ApplicationController
     @node = Node.find params[:id]
     @user = User.find_by_username params[:username]
     @body = params[:body]
-    @token = request.headers['Token']
+    @token = request.env['rack.session']['token']
 
     if @user && @user.token == @token
       begin
         @comment = create_comment(@node, @user, @body)
         msg = {
           status: :created,
-          message: "Created"
+          message: 'Created'
         }
         respond_to do |format|
           format.xml { render xml: msg.to_xml }
@@ -60,7 +60,7 @@ class CommentController < ApplicationController
       rescue CommentError
         msg = {
           status: :bad_request,
-          message: "Bad Request"
+          message: 'Bad Request'
         }
         respond_to do |format|
           format.xml { render xml: msg.to_xml }
@@ -70,7 +70,7 @@ class CommentController < ApplicationController
     else
       msg = {
         status: :unauthorized,
-        message: "Unauthorized"
+        message: 'Unauthorized'
       }
       respond_to do |format|
         format.xml { render xml: msg.to_xml }
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -52,14 +52,32 @@ class CommentController < ApplicationController
   end
 
   def create_by_token
-    @username = params[:username]
-    @node_id = params[:id]
+    @node = Node.find params[:id]
+    @user = Users.find_by_username params[:username]
     @body = pararms[:body]
     @token = request.headers["token"]
 
-    @user = Users.find_by_username @username
     if @user && @user.token == @token
-      # TODO: Do the usual commenting stuff
+      begin
+        @comment = create_comment(@node, @user, @body)
+        msg = {
+          status: :created,
+          message: "Created"
+        }
+        respond_to do |format|
+          format.xml { render xml: msg.to_xml }
+          format.json { render json: msg.to_json }
+        end
+      rescue CommentError
+        msg = {
+          status: :bad_request,
+          message: "Bad Request"
+        }
+        respond_to do |format|
+          format.xml { render xml: msg.to_xml }
+          format.json { render json: msg.to_json }
+        end
+      end
     else
       msg = {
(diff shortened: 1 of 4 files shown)


jywarren:
Sounds good -- plus a test; i added a token to the `bob` user fixture for testing. We'd want to handle errors and rejections as well, so that's a second and third test. 

If you're creating a whole new method instead of shoehorning it into the existing one, maybe this is best to put into a new API file after all? Your call.

I like the idea of the username, but lookup by token isn't that bad, esp. if we only search users that have `status != 0`. Your call, i do think it's nice to have few requirements so writing api calls is not a big deal, but i'm agnostic.

ryzokuken:
@jywarren I'm more concerned by the security problem here.

`SecureRandom.uuid` gives us 16 bytes of entropy, which means that if there's only one "right" token (as in the case of a single user, we have 1/2^16 (~0.00001525878) chance of it being the right one (which in itself is awfully low).

However, if the API accepts any token and looks up the user, the possibility will increase to n/2^16 where n is the number of users with `status != 0`. Considering a scenario (not very unlikely) where we have 10,000 users, this goes up to 0.1525878 or 15%. That's too bad, don't you think?

ryzokuken:
P.S. I added a new function so that we could point the `comment_by_token` function to a more appropriate URL inside `config/routes.rb`. (Possibly something like `/api/comment`. It would be a good idea IMO to group all the API routes together with an `/api/` prefix. What do you say?

jywarren:
Ah ok, good point. Sounds good!

jywarren:
On both counts.

On Jul 8, 2017 2:28 PM, "Jeffrey Warren" <jeff@unterbahn.com> wrote:

Ah ok, good point. Sounds good!

ryzokuken:
@jywarren if I have reference to user `DrupalUser`, how do I retrieve their token? Do I have to add a method for this in `app/models/drupal_users.rb`?

ryzokuken:
Oh, wait. I could do `@user.user.token`. Thanks.

jywarren:
But please try to use user instead, if possible. We want to deprecate
DrupalUsers soon. Either try to switch over entirely or you can add an
adapter method to DrupalUsers aliasing token, so you can do
drupal_user.token, see?

Thanks!

ryzokuken:
@jywarren I will. Was using `DrupalUsers` for the `find_by_name` function.

jywarren:
Ah, so, User.find_by_username should work!

ryzokuken:
Awesome. Will use that instead.

ryzokuken:
@jywarren how does this look? Should I wrap the commenting logic inside the `create` function into a private function and then call it from both of the functions?

ryzokuken:
Reference: https://web.archive.org/web/20130909062027/http://peepcode.com/blog/2011/rip-ruby-hash-rocket-syntax

ryzokuken:
@jywarren It finally works now, I just forgot the comma inside the hash literal. Could you please guide me regarding the last question I asked so that I could continue working on this?

jywarren:
Sure, a shared private function sounds great.

ryzokuken:
@jywarren started to work on the private function, it is mostly on the lines of the regular function. Hopefully, if its well tested, it will be super easy for me to make just the right function. The problem is, I had trouble understanding this part:

```ruby
respond_with do |format|
        if params[:type] && params[:type] == 'question'
          @answer_id = 0
          format.js
        else
          format.html do
            if request.xhr?
              render partial: 'notes/comment', locals: { comment: @comment }
            else
              flash[:notice] = 'Comment posted.'
              redirect_to @node.path + '#last' # to last comment
            end
          end
        end
```

I would start digging into it, but if you get the time, could you explain what exactly is achieves? I will then add some comments, I guess. That would be great for later contributors.

ryzokuken:
@jywarren Take a look now. If you like it, I would start writing unit tests for it. Hopefully, we could get it merged by today.

app/controllers/comment_controller.rb
```diff
@@ -1,3 +1,16 @@
+class CommentError < ArgumentError
+end
+
+def create_comment(node, user, body)
```
ananyo2012: This type of logic should go in helpers. Keep the controller code as thin as possible. Only keep action related methods

ryzokuken: Okay, If @jywarren approves the function, I would move it to the corresponding helper class.

jywarren:
format.js lets you create a response to a request with format = js, so ending in ".js" in the uRL. Chris is also a way to specify that it's an asynchronous xhr request, most likely from a JavaScript ajax function and not an html form request. So that code is helping respond to different types of requests but giving essentially the same information.

jywarren:
I guess some are rendering html in response and others are redirecting. It's not very tidy or compact code... @ananyo2012 is familiar with this stuff... Is there any more standard or compact way to do this? 

Otherwise looks good, go ahead!

app/controllers/comment_controller.rb
jywarren: Maybe look at how we've done error responses in other parts of the code, and compare also with best practices from modern Rails guides?

ryzokuken:
@jywarren Okay, I will move the class and function definition to the corresponding helper function and then move on with writing the tests. I will use the other tests as reference. By the way, do we have API-specific tests for any controllers? Tests that work with say -- JSON data instead of HTML?

jywarren:
Some I think... Search for xhr in the functional tests.

ryzokuken:
@jywarren How does this look? I think I should make integration tests instead.

jywarren:
Hmm, integration tests are the most expensive, and there isn't any persistent state we need here, I'd prefer not. 

Are you going to wire up create_by_token to use the helper too?

ryzokuken:
`create_by_token` and `create` both use the same function from the helper. I decided to use an integration test because I was thnking about making the tests somewhere on the lines of:

```ruby
test "posts comment successfully" do
  get '/api/comments',
     params: {
      ...
    }.
    {
      "token": ...
    }
end
```

shouldn't it be in an integration test?

jywarren:
You can prepare a post object and send it in a functional test, can't you?
Integration tests are just so heavy...

ryzokuken:
@jywarren Okay, I will do that inside the functional tests. There's one problem: The `/api/` endpoint is mounted to Swagger, I think. Can I add my own non-swagger endpoint to it?

app/controllers/comment_controller.rb
```diff
@@ -30,12 +34,51 @@ def create
...
+  def create_by_token
+    @node = Node.find params[:id]
+    @user = Users.find_by_username params[:username]
+    @body = pararms[:body]
+    @token = request.headers["token"]
+
+    if @user && @user.token == @token
+      begin
+        @comment = create_comment(@node, @user, @body)
+        msg = {
```
jywarren: I guess this just has a /lot/ of repeated code here -- the only unique line in 52-59 vs 61-68 vs 71-78 is `:created` and `"Created"`, you know? If you're going through the trouble to a whole new class for CommentError, i just feel that it's not justified given how little this code is reducing repetition. Also, isn't it a bit odd to need to `begin/rescue` in here? 

I'm trying to find a guide to best-practices in Rails controller error handling, to help DRY out this code. But I'm not finding much that's not based on model validation error handling...

ryzokuken:
@jywarren wrote a basic test. I have no idea why, but the test fails because the response is not `:success` but 406. Will be going to bed for today, but if you can, please look into this and post a solution if possible. I will try to figure something out myself first thing in the morning, and as soon as I do, I will add the next two tests and hopefully this will be ready to merge.

app/controllers/comment_controller.rb
ryzokuken: I get your point. I used the `raise -> rescue` workflow here because I call the `@comment.save` inside the function but wanted the calling function to know that something went wrong and do something sensible, and decided that throwing an error would be the best possible way to achieve this. If you could suggest a better way, I would be more than happy to use it instead, though.

jywarren:
Re the `/api/` endpoint... if you're doing such a thorough rewrite, do you think it's worth using the same URL before we actually try to integrate this with teh Grape/swagger stuff? 

If so, i think you can set a more specific (narrower) path higher in the routes.rb file and it'll catch ones that fit the pattern `/api/comment/____`, for example. The rest should miss that, and default to Swagger.

ryzokuken:
@jywarren got it, will definitely add it higher up.

jywarren:
Maybe you could pass back the output of `.save` and then you'd have the
built-in ActiveRecord exceptions to work with?

ryzokuken:
@jywarren I failed to figure out the reason I was getting an HTTP `406` status. I have made a stack overflow question. Here's the link: https://stackoverflow.com/questions/45005178/rails-responds-with-status-code-406

ryzokuken:
@jywarren figured out a solution :tada: https://stackoverflow.com/questions/45005178/rails-responds-with-status-code-406/45013651#45013651

Will push my changes now and will accept the answer after 2 days.

jywarren:
Ah one more thing -- did you want to display tokens to logged-in users on their profiles, and to admins for all profiles? You could make it show only during a `prompt('Copy your token', token)` sort of thing if you wanted.

ryzokuken:
@jywarren Yes. Showing the token only when the user explicitly asks for it sounds perfect.

ryzokuken:
@jywarren I was curious. When can `@comment.save` fail?

ryzokuken:
@jywarren figured it out. Check out the three tests I added. Do we need more?

app/controllers/comment_controller.rb
```diff
@@ -30,12 +34,36 @@ def create
...
       flash[:error] = 'The comment could not be saved.'
       render text: 'failure'
     end
   end
 
+  def create_by_token
+    @node = Node.find params[:id]
+    @user = User.find_by_username params[:username]
+    @body = params[:body]
+    @token = request.env['rack.session']['token']
```
jywarren: What are the advantages of this over passing the token in a POST req?

app/controllers/comment_controller.rb
```diff
@@ -30,12 +34,36 @@ def create
...
+
+    if @user && @user.token == @token
+      begin
+        @comment = create_comment(@node, @user, @body)
```
jywarren: Maybe add a comment here telling where to find this method? I think for newcomers the Rails organization can be a bit arcane and although it adds a line, this could be really helpful for some.

config/routes.rb
```diff
@@ -3,13 +3,15 @@
   mount JasmineRails::Engine => '/specs' if defined?(JasmineRails)
   mount JasmineFixtureServer => '/spec/javascripts/fixtures' if defined?(Jasmine::Jquery::Rails::Engine)
 
+  # Manually written API functions
+  post '/bot/comment/id.:format', to: 'comment#create_by_token'
```
jywarren: i'd prefer this to be more general -- anyone developing an API use -- could we do `/comment/create/token/:id.:format` maybe?

app/controllers/comment_controller.rb
ryzokuken: Sure, will do.

config/routes.rb
ryzokuken: Definitely. This was a placeholder anyway, because I realized that there would be a few issues with the `/api` endpoint because it's already being used.

app/controllers/comment_controller.rb
ryzokuken: I will look around for the best practices regarding tokens and come back to you. AFAIK, there's an important security benefit of passing the auth token inside the header, but I totally forgot what it was. :sweat_smile:

ryzokuken: https://www.codeschool.com/blog/2014/02/03/token-based-authentication-rails/ says that there's an inbuilt Rails function for dealing Authorization tokens "the right way".

It goes like:
```ruby
authenticate_or_request_with_http_token do |token, options|
  # authenticate user...
end
```

should we use it instead?

jywarren:
let's not - your implementation is clean and readable enough that we
needn't worry. I hadn't seen that before and some Rails conventions are
more widely adopted than others. Sometimes they're obscure enough to
actually make using them harder :-P

ryzokuken:
@jywarren Agreed. I will push a commit for the other two things you pointed out, and hopefully, we'd probably get this show on the road today.

ryzokuken:
@jywarren made the suggested changes. Take a look.

ryzokuken:
@jywarren merge this whenever you're ready, I see you have approved the changes but I do not have access to the repo so I cannot do it myself. :sweat_smile:

ananyo2012:
Great work guys!

ryzokuken:
Thanks, @ananyo2012. Hopefully, this will clear the way for a few basic publiclab-related behaviors, especially ones that involve simple commenting.

ananyo2012:
Yes you can move on with the https://github.com/publiclab/plotsbot/issues/13 after this is merged.

jywarren:
Merging! Thanks -- can you add to https://github.com/publiclab/plots2/wiki/API with a sample query?

PR #1511, opened 2017-07-08
Update 20170704165711_add_user_bio_and_token.rb (fix for #1510)

icarito:
Make sure these boxes are checked before your pull request is ready to be reviewed and merged. Thanks!

* [ ] all tests pass -- `rake test:all`
* [ ] code is in uniquely-named feature branch, and has been rebased on top of latest master (especially if you've been asked to make additional changes)
* [ ] pull request is descriptively named with #number reference back to original issue

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/wiki/contributing-to-public-lab-software

We have a loose schedule of reviewing and pulling in changes every Tuesday and Friday, and publishing changes on Fridays. Please alert developers on plots-dev@googlegroups.com when your request is ready or if you need assistance.

Thanks!

PR #1514, opened 2017-07-09
Excluding non-UTF-8 names from migration

jywarren:
Follow-up fix to #1511, attempting to deal with this error on staging server:

```
ActiveRecord::StatementInvalid: Mysql2::Error: Illegal mix of collations (latin1_swedish_ci,IMPLICIT) and (utf8_general_ci,COERCIBLE) for operation '=': SELECT  `rusers`.* FROM `rusers`  WHERE `rusers`.`username` = 'トリーバーチ クラッチ' LIMIT 1
```

### DIFF
--- a/db/migrate/20170704165711_add_user_bio_and_token.rb
+++ b/db/migrate/20170704165711_add_user_bio_and_token.rb
@@ -7,12 +7,14 @@ class AddUserBioAndToken < ActiveRecord::Migration
 
     # copy bios into new fields for non-spam users
     DrupalUsers.where('status != 0').each do |u|
-      user = u.user
-      if user and defined? :u.status then
-        user.status = u.status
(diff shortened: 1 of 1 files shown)


jywarren:
@icarito - check this out. It's a bit brute force, but honestly i think such usernames not in UTF-8 in the first place will break most controller actions anyways, so I think excluding them is unlikely to make any active or real users unhappy at this point. 

(Ref: a LOT of info on encodings here: http://yehudakatz.com/2010/05/05/ruby-1-9-encodings-a-primer-and-the-solution-for-rails/)

jywarren:
If it passes, i'll merge to stable and we'll see.

jywarren:
Darn, same thing. Opening an issue as this is becoming a blocking concern.

jywarren:
https://github.com/publiclab/plots2/issues/1515

PR #1517, opened 2017-07-09
Update 20170704165711_add_user_bio_and_token.rb

jywarren:

Thanks!

### DIFF
--- a/db/migrate/20170704165711_add_user_bio_and_token.rb
+++ b/db/migrate/20170704165711_add_user_bio_and_token.rb
@@ -3,8 +3,14 @@ class AddUserBioAndToken < ActiveRecord::Migration
     add_column :rusers, :bio, :text, limit: 2147483647
     add_column :rusers, :token, :string
     add_column :rusers, :status, :integer, default: 0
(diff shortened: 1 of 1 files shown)


ananyo2012:
@jywarren Modifying migrations is not a good idea as it breaks the backward compatibility of the already migrated databases in some cases.

jywarren:
Yes, very much agreed. However this migration broke staging but not the
Travis tests, and so could not be modified with a subsequent migration. It
was caused by data that existed only in the production/staging databases
and this caused the flurry of urgent activity over the weekend to resolve
the issue as fast as possible.

Regrettable but that's why we have staging to catch this sort of thing. I
think occasionally encodings-related obscure errors will crop up and cause
an issue like this.

Once we have a testing instance with a full replication of our production
server, we should be able to push migration code to that to confirm that it
will pass on the "real" data before committing to it in staging (which at
the moment is the only time we run the full real database through our code
before publishing to the live site).

So as good as it is that staging caught this before going live, our testing
branch will make this even tighter, so thanks in advance, @icarito!!

PR #1536, opened 2017-07-13
Make Solr test available in route

jywarren:
Merging stable changes into master -- not ideal but preserves master as canonical


jywarren:
Note @icarito

Issue #1463, opened 2017-06-15
Planning issue for Search integration


jywarren:
* [x] #1466 -- done!

Start hooking up Solr Node.search to various things, one at a time:

1. Solr Typeahead? 
    * [ ] Add Node.search with toggle to: https://github.com/publiclab/plots2/blob/master/app/services/typeahead_service.rb#L56 (doing in #1537)
1. Start stripping back different search interfaces from: https://github.com/publiclab/plots2/blob/master/config/routes.rb#L143-L152
    * [x] Remove https://github.com/publiclab/plots2/blob/master/app/controllers/questions_search_controller.rb
    * [x] Remove almost all methods from https://github.com/publiclab/plots2/blob/master/app/controllers/searches_controller.rb#L69-L79 (keep dynamic for now)
    * [x] rewire normal search methods with search (form) and results methods/views
    * [x] rewire these to use SearchService - #1470
    * [ ] rewire SearchService to use Solr `Node.search`
        * [ ] Use fallbacks using toggle: https://github.com/publiclab/plots2/blob/master/lib/solr_toggle/solr_toggle.rb
    * [ ] Use UI from https://publiclab.org/search/dynamic/  (this uses JSON API)
1. Consolidate:
    * [x] Does https://github.com/publiclab/plots2/blob/master/app/models/search.rb relate to SearchService, or could it? -- attempting removal in #1469
    * [x] remove typeahead code from https://github.com/publiclab/plots2/blob/master/app/services/search_service.rb#L64-L93
    * [ ] Functional test https://github.com/publiclab/plots2/blob/master/app/api/srch/search.rb and https://github.com/publiclab/plots2/blob/master/app/api/srch/typeahead.rb
    * [ ] Unit test https://github.com/publiclab/plots2/blob/master/app/services/search_service.rb and https://github.com/publiclab/plots2/blob/master/app/services/typeahead_service.rb
    * [x] Get rid of: https://github.com/publiclab/plots2/blob/master/app/models/search_record.rb
1. search tags, profiles, comments? Add new search fields: 
    * [ ] Comments: https://github.com/publiclab/plots2/blob/master/app/models/node.rb#L32
        * Add to `Node.search`:    `fields(:title, :body, :comments)`
    * [ ] Begin indexing tags: https://github.com/publiclab/plots2/blob/master/app/models/node.rb#L32
        * `text :tags do / tags.map(&:name) / end`
        * and add: `fields(:title, :body, :tags)`
[...]

PR #1537, opened 2017-07-13
Solr typeahead

### LINKED ISSUE #1654: Automated tests fail due to Solr being enabled in test mode -- should disable it
Unfortunately, this issue has always occurred for me on my machine. While some tests do successfully run & pass, many others simply timeout. 

### What's the problem?
Oftentimes, I see `Net::OpenTimeout: execution expired` come up in the test output. Based on my understanding, something is trying to open up a HTTP connection when I run the tests locally, but cannot appear to do so.

### Relevant URLs
1. When I run `rake test TEST=test/functional/comment_controller_test.rb`, I get the following [output for comment_controller_test.rb](https://gist.github.com/fuentesj/cf0d3075198705989961fc9af9249457)

2. When I run `rake test:functionals`, I get the following [output for functional tests](https://gist.github.com/fuentesj/3de86190b1e1d1e5286c26338e8a9a78): Note how in some cases you'll see the following: 
```
SocketError: Failed to open TCP connection to solr:8983 (getaddrinfo: nodename nor servname provided, or not known)
```
According to `config/sunspot.yml`, I believe solr should only listen on port `8983` in a production environment, and solr should listen only on port `8981` in a test environment. Could some tests be misconfigured? I might have messed up while setting up my local dev environment.

### Additional Information
1. I've noticed that many tests appear to fail at `lib/authlogic_openid/authlogic_openid/acts_as_authentic.rb:74`. Here's an excerpt containing that bit of code: 
```
def save(perform_validation = true, &block)
  return false if perform_validation && block_given? && authenticate_with_openid? && !authenticate_with_openid
  result = super
  yield(result) if block_given?
  result
end
```

OS: MacOS Sierra 10.12.5
Ruby: ruby 2.3.5p376 (2017-09-14 revision 59905) [x86_64-darwin16]
Rails: Rails 3.2.22.4

jywarren:
May need a toggle, however, as in #1463

Fixes #1654


jywarren:
Ok we need to follow
https://github.com/sunspot/sunspot/blob/master/README.md

jywarren:
@icarito - not sure, i added a new field here, do we need to update the schema when we do that?

jywarren:
May still need a toggle to switch between solr and non solr, and tests for this service.

jywarren:
OK -- so now, this is working (probably, once my last commit runs) but it takes a timeout to determine if Solr is working... which is too long. 

Here, we only need to know if a Solr service is available and we can toggle between -- but we need to know /fast/. Ideas, @icarito ?

jywarren:
Here's the toggle area:

https://github.com/publiclab/plots2/pull/1537/files#diff-1bc1c213d7a40db44441cd43fd8cd96cR35

It depends on an attempt to run a `Node.search` in the `SolrToggle` library we built in `/lib/` -- but that has to time out before determining which way to go. That'll take forever and also occupy a zillion threads if we put it in production.

icarito:
Is there a possibility to just instead of making a pre-query, try the _actual query_ and fallback _only if_ it fails?

icarito:
> The big thing (...) is trying to find a faster way than a timeout to determine if Solr is present (from in a rails app). I think that may be a sunspot question and could be searched on Stack overflow and/or posted on sunspot's issue tracker.

Looking for ideas.

jywarren:
OK, no longer seeing the timeout, but resolving a failure and some errors. Added a check to ensure solr is not available when the test failure occurs.

jywarren:
Our plan is to simply monitor publiclab.org/searches/test instead of having a fallback on production.

icarito:
We could monitor pad.publiclab.org solr service directly too.

Note also there are two kinds of failures: either Solr is (a) not
listening or (b) not responding. Both will likely fail differently on
the rails side.

jywarren:
Well, `solrAvailable` is evaluating to true, but not returning the blog nodes. Can we reproduce this in an interactive environment to see why?

icarito:
Maybe it's a good opportunity to push to unstable branch and see the
magic happen? ;-)

jywarren:
Yes, lets. We should be able to use `/searches/test?q=blog` and other queries to ensure Solr is actually working. for now, i'll also output a blank solr query to the log just to see.

jywarren:
https://github.com/publiclab/plots2/pull/1537/files#diff-25ca85bb0284da55ad87e7c39dfe8247R9 -- one more try. just have to get that test passing!

jywarren:
You know, it's weird but the unit tests are running with solr available, it seems. I guess I thought we were still testing with Solr off, except for the `/test/solr` tests. 

We aren't changing anything about SolrAvailable or the solr toggle here, nor are we touching `sunspot.yml`. Not sure why the toggle should show "on" here.

icarito:
You mean on sunspot.yml for test environment? I'm not sure about how it's supposed to run. I only see it run on travis, I guess this makes a difference running locally?

jywarren:
OK, I think we may be trying to run a Solr-dependent test in a non-Solr area of our tests. That would be pretty silly. I'm going to try moving `/test/unit/typeahead_service_test.rb` into `/test/solr/typeahead_service_test.rb`.

jywarren:
Before merging this, we should consider having a "hidden toggle" so we can selectively test this on production for a while.

app/services/typeahead_service.rb
```diff
@@ -6,98 +6,90 @@
...
   end
 
-  ## search for node title only
-  ## FIXme with solr
-  def find_notes(input, limit = 5)
-    Node.limit(limit)
-        .order('nid DESC')
-        .where('type = "note" AND node.status = 1 AND title LIKE ?', '%' + input + '%')
+  def notes(input, limit = 5)
+    if solrAvailable
```
jywarren: Can we access the `params` object in a service, here? If so, we could check for something like `?solr=true` here.

app/services/typeahead_service.rb
```diff
@@ -6,98 +6,90 @@
...
+      search = Node.search do
+        fulltext input
+        with :status, 1
+        with :type, "note"
+        order_by :updated_at, :desc
+        paginate page: 1, per_page: limit
+      end
+      search.results
+    else 
+      Node.limit(limit)
```
jywarren: Here's where we could/should turn Solr off and add a test without.

jywarren:
OK, i re-inserted the non-solr unit test so we can ensure it fails properly and doesn't break, but what I think is happening is that embedded Solr (that we're not using anymore) actually is on (we can check the gems) but is not indexed. So the `solrAvailable` check returns true, failing the test. 

I'm going to try disabling the `sunspot_solr` gem since we run our real Solr tests on a container anyways, and maybe that'll fix this.

jywarren:
Hmm, well, `solrAvailable` doesn't seem to be responding correctly -- either Solr is not actually off, or the toggle is not responding. It's based on `Node.search`.

jywarren:
@icarito -- so here, we no longer have disabled Solr for regular tests, so our Solr toggle doesn't work. I tried removing the embedded solr gem, but I think it's essentially just using the containerized Solr in tests, so I'm tempted to just remove the non-Solr test. This reduces the meaning of our toggle, but... as you said, we should just be working to keep Solr online, rather than getting hung up on this toggle situation. What do you think?

app/services/typeahead_service.rb
jywarren: I would guess it can't access `params`. But we can test this out in unstable so i guess that'd be OK.

jywarren:
:-)))) @icarito this is ready, i guess, to be tested on unstable!!

jywarren:
Checklist:

* [ ] turn Solr off in testing
* [ ] re-add the normal unit test that confirms Solr is off via toggle

solve anything remaining...

jywarren:
75cdc6f was published to https://unstable.publiclab.org -- but on there, this query fails when typing into the search bar: http://unstable.publiclab.org/api/typeahead/all?srchString=summer

jywarren:
Error was: 

```
web_1   | Started GET "/api/typeahead/all?srchString=summe" for 181.66.127.141 at 2017-10-06 16:17:05 +0000
web_1   | 
web_1   | NoMethodError (undefined method `name' for #<User:0x0055c7ef449ca8>):
web_1   |   app/services/typeahead_service.rb:100:in `block in search_profiles'
web_1   |   app/services/typeahead_service.rb:96:in `search_profiles'
web_1   |   app/services/typeahead_service.rb:76:in `search_all'
web_1   |   app/api/srch/typeahead.rb:24:in `block (2 levels) in <class:Typeahead>'
```

jywarren:
OK great, i've introduced a new test to try to capture the error we just saw, which was mistaking `user` for `username` -- and i have the fix too but won't yet submit it.

jywarren:
lots of small issues! added status to rusers fixtures

jywarren:
After re-disabling Solr in test mode, i had to re-introduce the system which turns it on again for the Solr-specific tests.

jywarren:
But this seems to be working on unstable; despite some JavaScript bugs, i see: 

http://unstable.publiclab.org/api/typeahead/all?srchString=about returns:

`{"items":null,"srchParams":{"srchString":"about","seq":null,"showCount":null,"pageNum":null}}`

Versus https://publiclab.org/api/typeahead/all?srchString=about --

```
{"items":[{"tagId":14975,"tagVal":"What can the contributor data tell us about the community around PL?","tagType":"file","tagSource":"/notes/mollydb/09-30-2017/what-can-the-contributor-data-tell-us-about-the-community-around-pl"},{"tagId":14891,"tagVal":"About Hydrogen Sulfide Research Paper:  \"Materializing Exposure\"","tagType":"file","tagSource":"/notes/Sara/09-13-2017/about-hydrogen-sulfide-research-paper-materializing-exposure"},{"tagId":14580,"tagVal":"Why does the first elevation reading show up about order of magnitude too large?","tagType":"file","tagSource":"/notes/liz/06-24-2017/why-does-the-first-elevation-reading-show-up-about-order-of-magnitude-too-large"},{"tagId":14101,"tagVal":"Hi, about fido alarm output's","tagType":"file","tagSource":"/notes/Roey/04-08-2017/hi-about-fido-alarm-output-s"},{"tagId":14041,"tagVal":"Thinking about local level advocacy","tagType":"file","tagSource":"/notes/stevie/03-20-2017/thinking-about-local-level-advocacy"},{"tagId":13246,"tagVal":"Questions About Gain Correction and their use in Plot Data","tagType":"file","tagSource":"/wiki/questions-about-gain-correction-and-their-use-in-plot-data"},{"tagId":9987,"tagVal":"Summary of discussion about Sensor Journalism","tagType":"file","tagSource":"/wiki/summary-of-discussion-about-sensor-journalism"},{"tagId":9411,"tagVal":"Sample curriculum:  All about light","tagType":"file","tagSource":"/wiki/sample-curriculum-all-about-light"},{"tagId":4,"tagVal":"About Public Lab","tagType":"file","tagSource":"/about"},{"tagId":0,"tagVal":"miikeabout117","tagType":"user","tagSource":"/profile/miikeabout117"},{"tagId":0,"tagVal":"Paarkertheaboutbe1999","tagType":"user","tagSource":"/profile/Paarkertheaboutbe1999"},{"tagId":0,"tagVal":"allaboutimports","tagType":"user","tagSource":"/profile/allaboutimports"},{"tagId":0,"tagVal":"caustaboutir1974","tagType":"user","tagSource":"/profile/caustaboutir1974"},{"tagId":0,"tagVal":"allaboutblockeddrains","tagType":"user","tagSource":"/profile/allaboutblockeddrains"},{"tagId":0,"tagVal":"parent:about","tagType":"tag","tagSource":null},{"tagId":0,"tagVal":"about","tagType":"tag","tagSource":null},{"tagId":14975,"tagVal":"What can the contributor data tell us about the community around PL?","tagType":"question-circle","tagSource":"/notes/mollydb/09-30-2017/what-can-the-contributor-data-tell-us-about-the-community-around-pl"},{"tagId":14975,"tagVal":"What can the contributor data tell us about the community around PL?","tagType":"question-circle","tagSource":"/notes/mollydb/09-30-2017/what-can-the-contributor-data-tell-us-about-the-community-around-pl"},{"tagId":14580,"tagVal":"Why does the first elevation reading show up about order of magnitude too large?","tagType":"question-circle","tagSource":"/notes/liz/06-24-2017/why-does-the-first-elevation-reading-show-up-about-order-of-magnitude-too-large"},{"tagId":14101,"tagVal":"Hi, about fido alarm output's","tagType":"question-circle","tagSource":"/notes/Roey/04-08-2017/hi-about-fido-alarm-output-s"},{"tagId":13721,"tagVal":"Query about Infragram DIY Filter Pack ( blue \u0026 red filter)","tagType":"question-circle","tagSource":"/notes/suman/11-22-2016/query-about-infragram-diy-filter-pack-blue-red-filter"}],"srchParams":{"srchString":"about","seq":null,"showCount":null,"pageNum":null}}
```

(so, perhaps it's not indexed. But at least it does return something without erroring. 

@icarito, would it be possible to reindex Solr on unstable?

jywarren:
Well, there we are. This looks to be ready to go if we can:

* [ ] resolve the JavaScript issues re: bower dependencies on `unstable`
* [ ] reindex unstable's Solr container so we can test the results it returns

icarito:
Reindexing unstable now...

icarito:
Okay so after reindexing, http://unstable.publiclab.org/api/typeahead/all?srchString=about still shows zero results.

jywarren:
I've added another test, for that specific query, just to be sure.

jywarren:
And we're 100% sure the Solr setup between production and unstable is the same? So I'm not off to be looking in the code to be sure it's wired up correctly?

icarito:
We could have a difference. It would be a good idea to rebuild and reindex both cores. Production core never failed since we created it but we've rebuilt the staging core when debugging the issues we had reindexing.

icarito:
I've resolved the asset loading paths. Having moved around to better organize stable/unstable instances, nginx was looking for assets in the wrong directory. That bit is fixed now.

jywarren:
Cool, thanks. Can you try rebuilding the unstable core and I will get this
last test fixed?thank you!

icarito:
Will do, created task https://github.com/jywarren/web/issues/105

jywarren:
Ok, this looks good. So next step is to rebuild both cores just to be sure?

icarito:
The idea would be to drop both cores and recreate + reindex them.
The process for dropping the cores if I recall is to stop and remove the solr container, then remove the solr volume, then recreate the solr container (it should create the core automatically) and then reindex. I'll try to do it this morning, sorry for the delay.

jywarren:
OK, just for comparison, this does return results:

http://unstable.publiclab.org/searches/test?q=about

This does not: 

http://unstable.publiclab.org/api/typeahead/all?srchString=about

So I think this is a coding issue?

icarito:
I'm in the middle of reindexing unstable.

icarito:
Oh actually reindex was just done for unstable. So the first is with SQL and the second with Solr? I don't know really.

icarito:
Just a not from logs, Solr **is** being queried:
```
solr_1  | 1002747 INFO  (qtp1112280004-19) [   x:default] o.a.s.c.S.Request [default] webapp=/solr path=/select params={q=farmhack&defType=edismax&qf=title_text+body_text+comments_text&fl=*+score&start=0&fq=type:Node&fq=status_s:1&fq=type_s:note&sort=updated_at_s+desc&rows=5&wt=ruby} hits=17 status=0 QTime=4 
solr_1  | 1007618 INFO  (qtp1112280004-15) [   x:default] o.a.s.c.S.Request [default] webapp=/solr path=/select params={q=test&defType=edismax&qf=title_text+body_text+comments_text&fl=*+score&start=0&fq=type:Node&rows=30&wt=ruby} hits=1960 status=0 QTime=1 
solr_1  | 1007627 INFO  (qtp1112280004-19) [   x:default] o.a.s.c.S.Request [default] webapp=/solr path=/select params={q=ica&defType=edismax&qf=title_text+body_text+comments_text&fl=*+score&start=0&fq=type:Node&fq=status_s:1&fq=type_s:note&sort=updated_at_s+desc&rows=5&wt=ruby} hits=2 status=0 QTime=5 
solr_1  | 1007874 INFO  (qtp1112280004-18) [   x:default] o.a.s.c.S.Request [default] webapp=/solr path=/select params={q=test&defType=edismax&qf=title_text+body_text+comments_text&fl=*+score&start=0&fq=type:Node&rows=30&wt=ruby} hits=1960 status=0 QTime=1 
solr_1  | 1007883 INFO  (qtp1112280004-13) [   x:default] o.a.s.c.S.Request [default] webapp=/solr path=/select params={q=icar&defType=edismax&qf=title_text+body_text+comments_text&fl=*+score&start=0&fq=type:Node&fq=status_s:1&fq=type_s:note&sort=updated_at_s+desc&rows=5&wt=ruby} hits=0 status=0 QTime=5 
solr_1  | 1008102 INFO  (qtp1112280004-17) [   x:default] o.a.s.c.S.Request [default] webapp=/solr path=/select params={q=test&defType=edismax&qf=title_text+body_text+comments_text&fl=*+score&start=0&fq=type:Node&rows=30&wt=ruby} hits=1960 status=0 QTime=0 
solr_1  | 1008108 INFO  (qtp1112280004-15) [   x:default] o.a.s.c.S.Request [default] webapp=/solr path=/select params={q=icari&defType=edismax&qf=title_text+body_text+comments_text&fl=*+score&start=0&fq=type:Node&fq=status_s:1&fq=type_s:note&sort=updated_at_s+desc&rows=5&wt=ruby} hits=0 status=0 QTime=3 
```

jywarren:
OK, just noting also that this change only uses Solr for note search, not for wikis, for example. But this should still return results. Compare these two lines for note search for the `/searches/test` query above, and the `typeahead` search:

Test: https://github.com/publiclab/plots2/blob/04c7e22c2f2f31074d43176eafc149815115393f/app/controllers/searches_controller.rb#L9

Typeahead: https://github.com/publiclab/plots2/blob/04c7e22c2f2f31074d43176eafc149815115393f/app/services/typeahead_service.rb#L37

Are we having trouble based on `pageNum` or `seq` being `null`?

`{"items":null,"srchParams":{"srchString":"farmhack","seq":null,"showCount":null,"pageNum":null}}`

jywarren:
I've introduced a change to disable the extra fields on the Solr query. I also added a test specifically for note (not wiki) content from fixtures.

app/services/typeahead_service.rb
```diff
@@ -6,118 +6,110 @@
...
-  ## FIXme with solr
-  def find_notes(input, limit = 5)
-    Node.limit(limit)
-        .order('nid DESC')
-        .where('type = "note" AND node.status = 1 AND title LIKE ?', '%' + input + '%')
+  def notes(input, limit = 5)
+    if solrAvailable
+      search = Node.search do
+        fulltext input
+        #with :status, 1
```
jywarren: we can re-enable these if they're not related.

jywarren:
They should actually both be for Solr, and I'm wondering if the extra parameters makes it return nothing... we'll see -- if this passes, could you push it to unstable? Thanks!

icarito:
I see what you mean! It might very well be, here's the logged query for both cases. I'll push to unstable when it passes.
```
solr_1  | 1549882 INFO  (qtp1112280004-15) [   x:default] o.a.s.c.S.Request [default] webapp=/solr path=/select params={q=about&defType=edismax&qf=title_text+body_text+comments_text&fl=*+score&start=0&fq=type:Node&fq=status_s:1&fq=type_s:note&sort=updated_at_s+desc&rows=5&wt=ruby} hits=2003 status=0 QTime=3 

solr_1  | 1582962 INFO  (qtp1112280004-19) [   x:default] o.a.s.c.S.Request [default] webapp=/solr path=/select params={q=test&defType=edismax&qf=title_text+body_text+comments_text&fl=*+score&start=0&fq=type:Node&rows=30&wt=ruby} hits=1960 status=0 QTime=1
```

jywarren:
This is odd, and I'm restarting the build -- it looks to be stuck on `rake db:setup` -- may just be an intermittent Travis issue?

jywarren:
Promising! It returned 2 results instead of 3 for a couple tests. I think that means the Solr query is working now; I adjusted the test assertion to match. Next I can re-enable some of the extra Solr parameters to see which caused the issue... hopefully there are, for example, pagination defaults we can lean on.

icarito:
Pushed to unstable!

jywarren:
Ug! Unstable still doesn't show results. @icarito, could you paste in the most recent solr queries from the logs and we can try to see why no results are returned?

jywarren:
Huh, interesting! One where query works but not the other?

jywarren:
OK, so the only line that makes the tests fail is `with :type, "note"` -- that's odd. 

Oh, we don't index that! 

https://github.com/publiclab/plots2/blob/master/app/models/node.rb#L24-L39

Changing that to see.

jywarren:
Oh wait, that's weird - we do index that, at least in this branch. Never mind, I'm not sure what's going on here...

jywarren:
@icarito -- finished this; pushing it to unstable too.

icarito:
Already! Wow. Will reindex production then in preparation for deployment.

jywarren:
testing it in unstable now. If this passes, can you merge it and push it to stable? Thanks!

jywarren:
It works in unstable: 
screenshot 2017-11-16 at 2 37 15 pm

jywarren:
🍔 🔨 🍬 🔥 🎆

icarito:
Reindexing production now in advance of deploying this.

icarito:
Reindexing stable too.

icarito:
Deployed to production!

icarito:
https://publiclab.org/api/typeahead/all?srchString=precocious no results, not sure why

PR #1551, opened 2017-07-26
attempt to fix token and bio

jywarren:
#1549 fixes

* [ ] all tests pass -- `rake test:all`
* [x] code is in uniquely-named feature branch, and has been rebased on top of latest master (especially if you've been asked to make additional changes)
* [x] pull request is descriptively named with #number reference back to original issue

PR #1552, opened 2017-07-27
Migration to fix bio, token, status. Already ran in staging.

icarito:
Token was generated, but bio is not correctly pulled.
Will need another migration for that.

Note that this is very similar to https://github.com/publiclab/plots2/pull/1551#pullrequestreview-52529835 but it already ran (I was doing it already when 1551 was issued). Process too six hours and a half and *did not* copy over the Bio - will attempt that with executing a SQL query in the migration.

### DIFF
--- a/db/migrate/20170726174757_recover_bios_tokens_statuses.rb
+++ b/db/migrate/20170726174757_recover_bios_tokens_statuses.rb
@@ -1,0 +1,27 @@
+
+# will only work in Ruby 1.9+:
+class String
+   def multibyte?
+     chars.count < bytes.count
+   end
+end
+
+class RecoverBiosTokensStatuses < ActiveRecord::Migration
+  def up
+      DrupalUsers.find(:all).each do |u|
+        unless u.name.multibyte?
+          user = u.user
+          if user
+            user.status = u.status unless u.nil?
+            user.bio = DrupalProfileValue.find_by_uid(user.id, conditions: { fid: 7 }) || ''
+            user.token = SecureRandom.uuid
+            user.save({})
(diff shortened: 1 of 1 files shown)


db/migrate/20170726174757_recover_bios_tokens_statuses.rb
```diff
@@ -0,0 +1,27 @@
...
+end
+
+class RecoverBiosTokensStatuses < ActiveRecord::Migration
+  def up
+      DrupalUsers.find(:all).each do |u|
+        unless u.name.multibyte?
+          user = u.user
+          if user
+            user.status = u.status unless u.nil?
+            user.bio = DrupalProfileValue.find_by_uid(user.id, conditions: { fid: 7 }) || ''
```
icarito: user.id ? :-( now everyone has numeric bios! :-p That explains it. This took so long that I'm tempted to use `execute` to run a SQL query directly.

db/migrate/20170726174757_recover_bios_tokens_statuses.rb
```diff
@@ -0,0 +1,27 @@
...
+   end
+end
+
+class RecoverBiosTokensStatuses < ActiveRecord::Migration
+  def up
+      DrupalUsers.find(:all).each do |u|
+        unless u.name.multibyte?
+          user = u.user
+          if user
+            user.status = u.status unless u.nil?
```
icarito: I managed to _not_ filter by status, so that explains why it took 6.5 hours.

jywarren:
moving to #1553

PR #1557, opened 2017-07-31
Modify unit test to make sure token isn't nil

ryzokuken:

Thanks!

@jywarren I'm sorry I pushed to a branch on publiclab/plots2 instead of ryzokuken/plots2 by mistake.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -39,10 +39,8 @@ class User < ActiveRecord::Base
   validates_with UniqueUsernameValidator, on: :create
   validates_format_of :username, with: /^[A-Za-z\d_\-]+$/
 
-  before_create do
-    create_drupal_user
-    set_token
-  end
+  before_create :create_drupal_user
+  before_save :set_token
   after_destroy :destroy_drupal_user
 
   def create_drupal_user
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -78,7 +78,7 @@ class User < ActiveRecord::Base
   end
 
   def set_token
-    self.token = SecureRandom.uuid if !self.token.nil?
+    self.token = SecureRandom.uuid if self.token.nil?
   end
 
   # this is ridiculous. We need to store uid in this model.
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -39,7 +39,10 @@ class User < ActiveRecord::Base
   validates_with UniqueUsernameValidator, on: :create
   validates_format_of :username, with: /^[A-Za-z\d_\-]+$/
 
-  before_create :create_drupal_user
+  before_create do
+    create_drupal_user
+    set_token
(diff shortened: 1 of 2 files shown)


ryzokuken:
@jywarren Locally, I get all the authlogic tests failing.

ryzokuken:
@jywarren Perfect, the error is:

```
Failure: <nil> expected to not be nil.

test_user_creation(UserTest)

test/unit/user_test.rb:18:in `block in <class:UserTest>'
```

Now in order to make the tests pass, all we need to do is to make sure the token is set properly.

Maybe we could set the token inside `create_drupal_user` method?

jywarren:
Actually could we do it outside that method since we may eventually deprecate that once we move away from `DrupalUsers` completely? Is there another `before_save` method or could we add a new one?

ryzokuken:
@jywarren I will make it inside a function called `set_token` add it to to `:before_create`. We can also use this function to later implement a "Reset Token" feature if we would like.

ryzokuken:
@jywarren It must be working perfectly I think, but because `User.create` isn't being invoked in this instance (it will be in the normal workflow), Our method inside `before_create` isn't being triggered. I would change it to `before_save` instead, and it should work. However, I think that might be counterproductive because then it would be called everytime the user is updated, making the token reset a side-effect of an update of data.

ryzokuken:
@jywarren It worked. (YAY!)

Maybe we should implement the "Reset Token" functionality too so that all users with their tokens set to `nil` can use that to set it instead?

jywarren:
I think you are right that before_save will cause a token reset on any update -- can you try another event like after_create or something? Why isn't before_create triggered? It should be in the users_controller create method. 

Alternatively you could have the create token only occur if the token is nil.

jywarren:
Sorry for my slow response - I'm still not all caught up and am in meetings all day for this week. But if you can resolve this last issue I'll try to publish this tonight. Worst case I can generate you a token randomly.

ryzokuken:
I think before_create isn't being called because we aren't calling the
create method in the first place, we are creating it manually by calling
User.new and user.save 😅

You're right, a good options seems to be to add a condition inside the
before_save method or to change the token right inside User.new. I will add
the condition for now, but will make appropriate changes if you request.

jywarren:
Let's do the conditional; that keeps the logic inside the model.

ryzokuken:
@jywarren Added the condition; Take a look.

jywarren:
:-)

jywarren:
Ok, published to the live site! Have fun! 🎈

ryzokuken:
@jywarren should I delete the user and then remake one, or trigger the save
method by changing the bio or something instead?

jywarren:
Change your bio! Good to know if that works.

PR #1558, opened 2017-08-01
example powertag dropdown [WIP]

### LINKED ISSUE #496: dropdown menu for "power tags"
### Problem

There are an increasing number of powertags which can be used to modify or add features to a post or a page. But these must be manually entered using the documentation at https://publiclab.org/wiki/power-tags

Tag form shown here:

screenshot 2016-04-12 at 4 13 02 pm

### Proposed fix

Provide a low-profile dropdown menu (so as not to confuse newcomers, but be available for more experienced folks) which auto-tags or prompts for input in order to generate appropriate power tags. For example, if you select `response`, it'd prompt you in a dialog box (or a [Bootstrap popover](http://getbootstrap.com/javascript/#popovers), perhaps?) for the ID of the post you're responding to -- and even offer a search function to find that post. 

This last "autocomplete" addition could be a little bit like the Github interface which helps you link to other issues, as shown here:

screenshot 2016-04-12 at 4 11 10 pm

Code for the tagging form is here: https://github.com/publiclab/plots2/blob/master/app/views/tag/_tagging.html.erb#L11

The tagging form JavaScript might also be moved into its own file at `/app/assets/javascripts/tagging.js`.

shelbyfirebaugh:
Beginning of a solution: fixes #496

We just added the Bootstrap example code for a dropdown menu in a new file called _advanced_tagging.html.erb.

We added it just in the format of a normal HTML page, since we don't know the correct syntax to use for .erb files. It also is not clickable at the moment, as we also don't know how to actually call the powertag functions. Is Active Record Querying the solution that we should be looking at?

Any suggestions on how we should proceed would be greatly appreciated!


shelbyfirebaugh:
Also, this is what we got when running the tests. Should we proceed with `rake:db migrate` ?

screen shot 2017-08-01 at 13 16 01

jywarren:
Looks great! Yes, try rake db:migrate - and can you remove the npm debug file that's included?

Are you able to start the application with `passenger start` on your local machine?

Thanks!

jywarren:
And you should only need `rake test`, not the "all" option. Hope that helps!

shelbyfirebaugh:
I ran `rake db:migrate`:

screen shot 2017-08-01 at 13 52 53

Then I ran `rake test` and got these errors:

screen shot 2017-08-01 at 13 51 18

Then I ran `passenger start`. It says I have an error, but that path that it specifies isn't a file that I edited. It also still says I'm not running mysql, if that matters.

screen shot 2017-08-01 at 13 48 24

jywarren:
you probs need to run `bower install` to get bootstrap

jywarren:
1. lets strip out the <html> start and end tags, so just lines 17-27
2.

jywarren:
Forgot to mention -- you can view the tagging partial on any note or wiki page, and I'll paste in a link to the relevant lines of code -- sorry should've done that already!

shelbyfirebaugh:
okay, so yesterday we met with our coach Katie and were trying to make it so that the changes that i have made would show up on my local version of the app. so my file `_advanced_tagging.html.erb` is rendered as a partial inside of the file `_tagging.html.erb`. but after lots of digging, Katie said that she couldn't find a place where the `_tagging` partial is actually rendered, and thus `_advanced_tagging` as well. if you know that there is somewhere that it's being rendered, could you please point us to that path?

aside from that issue, we are unsure of how to approach the issue of defining the parameters to make the code only visible when you add something to the end of the url. Katie wanted to know where we would need to define the route for that, or if there is some other way that we should address this issue

jywarren:
So for example, on wiki pages like `/wiki/_____` (you should be able to see a list of all at `/wiki`), you'll see the tagging interface below the main body content. But only if you're logged in!

That partial include is here: https://github.com/publiclab/plots2/blob/master/app/views/wiki/show.html.erb#L84

shelbyfirebaugh:
wow okay we were definitely just on the wrong page, it works now! but could you just explain a little bit about why typing `?potter=harry` at the end works even though we didn't define that route? unless that's a route that you already have that you use for these cases?

jywarren:
Wow, cool! Does this work, can you see it on your local install? Is it ready to publish? This one is pretty straightforward, but sometimes with UI related work, it's helpful to attach a screenshot so I or another reviewer can get a look without having to pull the whole thing first. 

But if this looks good I'll merge it in! Fantastic! And we can check it off the list at #1560.

shelbyfirebaugh:
yes, it does work on my local version! at least the drop down shows up and it's clickable, i still want to test it with a post with actual stuff in it to make sure that the tag function is being called and makes it ~ fancy ~

i'm at work at the moment but when i get home i'll test that and post screenshots on the PR

thanks!

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

jywarren:
Awesome -- I think this is all good though, since it'll be hidden. I'm going to merge!

jywarren:
🎈 🎉  Thanks! And onward to #1560 :-)

jywarren:
Great -- this is live now -- https://publiclab.org/wiki/sandbox?potter=harry

It's not exactly as I'd imagined -- there was a way I'd seen in Bootstrap for the button to be kind of attached to the input form, on the same line. Like these: http://getbootstrap.com/components/#input-groups-buttons-dropdowns

But the potter hidden thingy is working perfectly! https://publiclab.org/wiki/sandbox

Kudos -- you should share this out with your demo in my request for 1-line updates! @lilliankorinek @shelbyfirebaugh great work.

PR #1562, opened 2017-08-02
Fix commenting

ryzokuken:

Thanks!

### DIFF
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -44,7 +44,7 @@ class CommentController < ApplicationController
     @node = Node.find params[:id]
     @user = User.find_by_username params[:username]
     @body = params[:body]
-    @token = request.env['rack.session']['token']
+    @token = request.headers["HTTP_TOKEN"]
 
     if @user && @user.token == @token
       begin
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,7 +4,7 @@ Plots2::Application.routes.draw do
   mount JasmineFixtureServer => '/spec/javascripts/fixtures' if defined?(Jasmine::Jquery::Rails::Engine)
 
   # Manually written API functions
-  post '/comment/create/token/id.:format', to: 'comment#create_by_token'
+  post 'comment/create/token/:id.:format', to: 'comment#create_by_token'
 
   get 'searches/test' => 'searches#test'
   #Search RESTful endpoints
--- a/test/functional/comment_controller_test.rb
+++ b/test/functional/comment_controller_test.rb
@@ -259,7 +259,7 @@ class CommentControllerTest < ActionController::TestCase
     }
 
     @headers = {
-      'token' => 'invalid-token'
+      'HTTP_TOKEN' => 'invalid-token'
     }
 
     post :create_by_token, @params, @headers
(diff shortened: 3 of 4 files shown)


ryzokuken:
@jywarren @ebarry this is how commenting looked locally 🎉 

screenshot from 2017-08-03 01-21-53

ryzokuken:
@jywarren This will most likely fail, as the tests were written keeping the faulty code in mind. I will fix those tomorrow if that's not a problem, Thanks.

ananyo2012:
@ryzokuken This looks great! Did you use curl for making the post request?

ryzokuken:
No, I was using Firefox's http requester add-on. You could also use Chrome
if you're using Chrome.

ananyo2012:
Well I mostly Use Postman plugin for making post requests.

ryzokuken:
@jywarren @ananyo2012 There's an issue. Until now, I had been adding stuff to the session in the request, and had been reading from the session itself. In order to fix this, we could make an integration test instead, as I had been confused between http://api.rubyonrails.org/v3.2/classes/ActionController/TestCase/Behavior.html#method-i-post and http://api.rubyonrails.org/v3.2/classes/ActionDispatch/Integration/RequestHelpers.html#method-i-post. What do you guys say?

jywarren:
Oh huh... just curious, why do you use the session?

ryzokuken:
I was confused between the two, and set the session instead of the header.

ananyo2012:
Well we can simply use `params[:token]` and send a simple key-value pair for the token via a post request. It doesn't necessarily need to be in the header. Though the convention for token based authentication is sending the token in an `Authorization` header.

app/controllers/comment_controller.rb
```diff
@@ -44,7 +44,7 @@ def create_by_token
     @node = Node.find params[:id]
     @user = User.find_by_username params[:username]
     @body = params[:body]
-    @token = request.env['rack.session']['token']
+    @token = request.headers["HTTP_TOKEN"]
```
ananyo2012: @ryzokuken Can you check this with `@request.headers["HTTP_TOKEN"]` and check if the tests pass.

ryzokuken:
Exactly. I don't think sending the token as a parameter is safe at all.

ryzokuken:
It's being read perfectly, as comments work. The tests are configured
improperly. There's no way to set headers inside controller tests, you can
only set headers in request inside integration tests.

ananyo2012:
So what is assigned to the `@token` during tests ? Is it null ?

ryzokuken:
@ananyo2012 yes.

jywarren:
Ah, i see. Well, yes, if you are confident that the header is the appropriate place (and worth the extra complexity for users of the API; consider how difficult it is to set a header from any particular scripting tool), then go ahead and move this to be an integration test. Thanks!

ryzokuken:
@jywarren sure, thanks! I will submit a PR that moves the tests to an integration test instead today.

ryzokuken:
@jywarren @ananyo2012 yay! This looks ready from my side.

ryzokuken:
@jywarren @icarito Let me know when these changed are pushed to staging and stable, then I could proceed with using these.

jywarren:
It is live now on PublicLab.org!

ryzokuken:
@jywarren @ebarry @ananyo2012 demonstrated this on the live site, it works!

jywarren:
This is so super!

PR #1571, opened 2017-08-08
Implementing Inline commenting #1438

aspriya:
Make sure these boxes are checked before your pull request is ready to be reviewed and merged. Thanks!

* [x] all tests pass -- `rake test:all`
* [x] code is in uniquely-named feature branch, and has been rebased on top of latest master (especially if you've been asked to make additional changes)
* [x] pull request is descriptively named with #number reference back to original issue

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/wiki/contributing-to-public-lab-software

We have a loose schedule of reviewing and pulling in changes every Tuesday and Friday, and publishing changes on Fridays. Please alert developers on plots-dev@googlegroups.com when your request is ready or if you need assistance.

Thanks!

PR #1593, opened 2017-08-30
Wiki Inline-Commenting #1438

aspriya:

Picking up from where https://github.com/publiclab/plots2/pull/1571 left off.


jywarren:
What's the error here? Can you copy it in? @ananyo2012 I wonder if danger
is responding to test failures but not errors? Or if an error happens early
enough maybe danger doesn't even run? If the latter is the case there's not
much we can do, but good to know.

jywarren:
Ooh, cool - but see the warning about new schema file!

jywarren:
I wanted to note that I think /all/ rich wiki updating is a bit slow. We'll look into this, but I believe we should:

* monitor unstable for performance (@icarito)
* add a "spinner" icon as with the Post button on `/post` (this also happens in regular comments, i think) so people know it's doing **something**
* fix the schema file as advised by Dangerbot above

then let's test one more time -- i'd like to work a bit more on how the comments are displayed too; perhaps we can show `3 comments` by the comment icons, instead of the full display shown in your screenshot:

image

Then that full display could be toggled to show.

app/assets/javascripts/buildSectionCommentForm.js
```diff
@@ -0,0 +1,70 @@
+function buildSectionCommentForm(uniqueId, title, node_id, subsection_string) {
+  var formHtml = "<form style='display:none;' id='inline-comment-"+uniqueId+"' data-remote='true' class='well' action='/comment/create_inline_comment/"+node_id+"'>";
+  formHtml += "<h4 style='margin-top:0; padding-top:0;'>"+title+"</h4>";
```
jywarren: Here I think you can do:

```js
var html = "<div>\
<p>Stuff</p>\
</div>";
```

app/assets/javascripts/showInlineComment.js
```diff
@@ -0,0 +1,36 @@
...
+  var iComment = "<div id='inline-comment-"+cid+"' style='padding-top: 10px; padding-left: 10px;'>";
+  iComment += "<div class='inline' style='vertical-align: top;'>";
+  if(photo_file_name){
+    iComment += "<img class='img-circle' src='"+author_photo_path+"' />";
+  }else{
+    iComment += "<div class='img-circle'></div>";
+  }
+  iComment += "</div>";
+
+  iComment += "<div class='inline' id='comment-content'>";
```
jywarren: same here?

app/assets/javascripts/wikis.js
```diff
@@ -15,7 +14,8 @@ function setupWiki(node_id, title, raw, logged_in) {
       preProcessor: preProcessMarkdown,
       postProcessor: postProcessContent,
       extraButtons: {
-        "fa-question": questionForm
+        "fa-question": questionForm,
+        "fa-comment": setupCommentFunction
```
jywarren: Can we make this:

```ruby
<% if params[:inlineCommenting] %>"fa-comment": setupCommentFunction<% end %>
```
? That way we can test this more on production without committing to it just yet. And we can publish the general Rich Wikis inline features without also having to publish inline commenting which may take a little longer?

db/schema.rb.example
```diff
@@ -49,6 +49,7 @@ ActiveRecord::Schema.define(:version => 20170824172336) do
     t.string  "mail",      :limit => 64
     t.string  "homepage"
     t.integer "aid",                             :default => 0,  :null => false
+    t.string  "reference"
```
jywarren: Here I believe you have to update the timestamp at the top of the file to match the latest migration version -- that may be what's stopping you!

icarito:
A note that not the whole site seems slow on unstable.publiclab.org - I
see no reason for slowness in the logs.

I will continue to investigate.

db/schema.rb.example
aspriya: Oh thanx. Corrected this

app/assets/javascripts/buildSectionCommentForm.js
jywarren: Actually it looks like that syntax only works in node.js or maybe ES6; ignore!

app/assets/javascripts/wikis.js
jywarren: Hi, @aspriya - i think if you can add this flag we can merge and see how this is working. Thanks!

jywarren:
I made the "hidden" flag in javascript so this will only work when this is at the end of the URL: `?inlineComments=true`

Now all that's left is:

* [ ] add a "spinner" icon as with the Post button on /post (this also happens in regular comments, i think) so people know it's doing something

Very close!

jywarren:
Hi, @Aspriya - what do you think about these changes? Hope you're well!

ryzokuken:
@jywarren @aspriya hasn't replied to this thread for a very long time. Is this mergeable, or should we close this for now?

jywarren:
do you want to download the code and try it out? I worried there were
things we overlooked in this and it needed a bit more work.

ryzokuken:
@jywarren I could definitely give this a try!

SidharthBansal:

Issue #511, opened 2016-04-24
Draft text to welcome newcomers to the dashboard


jywarren:
This has been marked as a good candidate for becoming a [first-timers-only issue](http://www.firsttimersonly.com) like [these](https://github.com/publiclab/plots2/labels/first-timers-only), meaning that it's simple, self-contained, and with some extra formatting, could be a great entry point for a new contributor. If you're familiar enough with this code, please consider [reformatting or reposting it as a first-timers-only issue](https://publiclab.org/notes/warren/10-31-2016/create-a-welcoming-first-timers-only-issue-to-invite-new-software-contributors), and then ping @jywarren to get it labelled.

****

We need something like (but in HTML):

```
## Hello, welcome to Public Lab!

This is your dashboard, where recent updates from other community members show up. It's helpful to keep up to date with new progress, news, and recent developments.

But to get started at Public Lab, try:

* asking a question (/questions)
* exploring projects (/methods)
```

This would go in the Dashboard template, probably just above the Activity feed, in a box, perhaps with a light grey 1px border and rounded corners:

https://github.com/publiclab/plots2/blob/master/app/views/dashboard/dashboard.html.erb#L26

If you take this on, please share a screenshot of your work so we can see how it looks!

mayank-mittal:
So do i only have to create a container which displays this message and add it to the dashboard.html.erb? if it is can i take this one.

jywarren:
We'd have to also check if the user is new -- i.e. have they signed up in the past week, for example? And put that in a conditional. Something like:

```ruby
<% if Time.now - current_user.created_at > 1.week %>
  <!-- container here -->
<% end %>
```

I *think* that'd work, but you probably need to test it out! 

Thanks a lot!

dristybutola:
Can i work on this issue?
Secondly is the dashboard to be shown to users newer than a week?
thanks.

ryzokuken:
Sure, @dristybutola.

Yes and no. The dashboard is visible to everyone, but the welcome text would only be visible if the user is newer than a week.

For any queries, feel free to ask here or on our Gitter channel.

jywarren:
Hi, are either @mayank-mittal or @dristybutola still interested in trying this one out? We'd love to help if you've gotten stuck.

dristybutola:
Hi @jywarren , I am going to start work on this :), can I?

jywarren:
Yes, please do, if you're able -- much appreciated!

dristybutola:
[...]

PR #1642, opened 2017-09-25
issue #511 resolved - draft text to welcome newcomers to dashboard added using modal

Gauravano:
Hi, I have tried to solve issue #511 by adding a modal as the content is long so I tried this. The modal will open if the user is just a week old for the website.It can be closed by crossing the window or by clicking out of modal. Here;s the preview of modal added.
welcome_text

If you want me to handle this issue using normal notification like this, kindly respond.Thank You.
check1


jywarren:
Hi, I see what you're saying; do you think we could make the WELCOME header just bold, but on the same paragraph as the rest? That should compact it. 

Also, I think `asking a question` and `exploring projects` could just be links to their respective URLs, rather than on bullet points. What do you think? I guess I prefer to compact this instead of to use a modal. 

Thanks!!

jywarren:
Hi, just checking if you'd had a chance to look at my suggestions? This is very close, thank you!

Gauravano:
hi, @jywarren, your suggestions are very useful and I have tried to work on them. Kindly check the screenshot and tell for more optimisations.Thank You.
update

jywarren:
That looks really good. Can you put the two links at the bottom in buttons, so for example:

```html
<p>
  <a class="btn btn-sm btn-default" href="...">Asking a question</a> or 
  <a class="btn btn-sm btn-default" href="...">Exploring projects</a>
</p>
```

And could you push a new commit to this branch with your new code? Super and thanks for making the additional changes!

app/views/dashboard/dashboard.html.erb
```diff
@@ -1,3 +1,32 @@
+<% if Time.now - current_user.created_at < 1.week %>
```
jywarren: Also, you'll probably need to add an `if current_user && ` before this to only attempt it if you're logged in. That's what's causing the failure in the tests!

Gauravano:
check2pull
Buttons like this or any change required? Also, thanks for clearing about the existence of current_user

jywarren:
That looks perfect! Thanks!

app/assets/javascripts/dashboard.js
```diff
@@ -1,3 +1,25 @@
+$(document).ready(function() {
```
jywarren: Hi, is the modal code here still needed? It seems like the alert box approach may not require the modal code. Thanks!!

Gauravano: Yes, this code is redundant now .Thanks for pointing and I will take care of these types of mistakes for future contributions.

Gauravano:
Please help me to figure out why checks are failing as I have only removed redundant code in my last commit. Thank you.

jywarren:
Ok, this was confusing to me too, but i see in the logs:

> `raise_if_we_cannot_find_the_commit': Commit 3e4f22c7 doesn't exist. Are you running `danger local/pr` against the correct repository? Also this usually happens when you rebase/reset and force-pushed. (RuntimeError)

On this line - https://travis-ci.org/publiclab/plots2/builds/286597109?utm_source=github_status&utm_medium=notification

I'm going to try restarting the build. You can do this yourself by closing and reopening the PR. Sometimes this happens if two pushes are made in quick succession... Not sure why exactly here but it doesn't seem to be related to your code. 

Sorry this was a bit frustrating!

jywarren:
Looks like it worked! I'd like to ask @steviepubliclab and @ebarry to take a final look at the text on Monday before merging in. But your work looks great! Thanks so much!

jywarren:
And just to be clear @ebarry @steviepubliclab  -- this message is shown to everyone who's signed up less than 1 week ago. If there are any text tweaks you'd like to do now, please ping in! 

This is just a starting point. We can always add/edit/expand later, in addition to using this as an example for future first-time coders of how we can add more triggers for different time periods.

ebarry:
+1 great!

jywarren:
Super -- merging now, thank you!!!

steviepubliclab:
Looks good to me too!

Gauravano:
Thank you @jywarren @ebarry @steviepubliclab

PR #1647, opened 2017-09-26
new users by date of posting

jywarren:
This'll end up at https://publiclab.org/people?new=true and https://publiclab.org/people?new=true&format=json (we can improve routing later!)


jywarren:
I see `Error: test_list_new_users_in_json(UsersControllerTest): ActiveSupport::JSON::Encoding::CircularReferenceError: object references itself`

PR #1653, opened 2017-09-27
Adapt Dangerbot to catch errors too

jywarren:
@ananyo2012 - i'm trying to adapt the Dangerfile to report errors as well as failures! Hope this works. I need to introduce a failure to test this :-)

Based on this sample error output: https://gist.github.com/ananyo2012/5f74fadf8a15b76dfc669c1e39337d44#file-test-admincontrollertest-xml

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -40,6 +40,8 @@ junit.failures.collect(&:nodes).flatten.each do |failure|
   end
 end
 
+message junit.inspect
+
 junit.errors.collect(&:nodes).flatten.each do |error|
   error.nodes.each do |f|
     match = f.match(/(test[a-z_\/]+.rb):([0-9]+)/)
--- a/Dangerfile
+++ b/Dangerfile
@@ -40,8 +40,6 @@ junit.failures.collect(&:nodes).flatten.each do |failure|
   end
 end
 
-message junit.inspect
-
 junit.errors.collect(&:nodes).flatten.each do |error|
   error.nodes.each do |f|
     match = f.match(/(test[a-z_\/]+.rb):([0-9]+)/)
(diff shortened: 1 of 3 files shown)


jywarren:
Once this passes, i'll introduce an error in https://github.com/publiclab/plots2/blob/error-dangerbot/app/controllers/notes_controller.rb

jywarren:
`/usr/local/bundle/gems/activesupport-3.2.22.4/lib/active_support/dependencies.rb:251:in `require': /app/app/controllers/notes_controller.rb:335: syntax error, unexpected keyword_end, expecting end-of-input (SyntaxError)`

So, just a syntax error doesn't get picked up by Danger because the whole thing fails to run. I don't know if we'd be able to prevent that. 

But an error in the tests themselves might trigger Danger.

jywarren:
OK, this was clever, but the error actually causes no output xml from Danger, so we can't analyze it: `tail: cannot open ‘./test/reports/TEST*.xml’ for reading: No such file or directory`

jywarren:
Tried introducing an error in the test, instead of in the app code.

jywarren:
Hmm, this doesn't seem to work... out of ideas here!

jywarren:
Hmm, that last seems to have run tests properly and found an error at:

`Error: test_should_not_save_answer_without_content(AnswerTest): NameError: uninitialized constant AnswerTest::TAnswer
test/unit/answer_test.rb:5:in `block in <class:AnswerTest>'`

but then, i see:

```

/home/travis/.rvm/gems/ruby-2.4.1/gems/octokit-4.7.0/lib/octokit/response/raise_error.rb:16:in `on_complete': PATCH https://api.github.com/repos/publiclab/plots2/issues/comments/332656015: 403 - Issue not viewable by PublicLabBot // See: https://developer.github.com/v3/issues/comments/#edit-a-comment (Octokit::Forbidden)
```

Weird. Does that mean the bot couldn't update the comment it left??

jywarren:
AHA - it somehow thought my citing of its error message meant it had to edit my comment, not its own. It must use a text match to find its own comment. Changed mine so it doesn't get confused.

jywarren:
It worked!!! 

The error is long but good: https://gist.github.com/jywarren/1e7e88f7ca6868cc8e3dfd3a441f8002

jywarren:
OK, notable that errors in the code itself are still not caught -- only errors in the testing code are parseable by Dangerbot. Well, then wrapping this up for merge in!

PR #1660, opened 2017-10-02
Add node revisers to list of email recipients on new comments (1589)

### LINKED ISSUE #1589: Notifications for comments on wiki pages - do they go to the right people?
Looking at @aspriya's new interface for making comments on wiki pages, it's not clear who comments notify. I guess they should notify all authors for that page, and all commenters, no?

https://publiclab.org/wiki/new-concepts-for-oshw-in-indoor-gardening/comments

screen shot 2017-08-25 at 11 22 14 am

Let's **add an italic line above the form indicating who will receive it**:

`Comments will notify all authors of this page, and all commenters, by email.`

## Code

The controller code for that is here:

https://github.com/publiclab/plots2/blob/master/app/controllers/comment_controller.rb#L21

Then it uses `create_comment` here: https://github.com/publiclab/plots2/blob/master/app/helpers/comment_helper.rb

Then `comment.notify` here: https://github.com/publiclab/plots2/blob/master/app/models/comment.rb#L106

It looks like it only notifies the page's original author, and all commenters. 

## Solution

We need to modify this so that if there are any /additional authors/ -- that is, authors of revisions attached to this node -- they get notified too. The change should go inside the `  def notify(current_user)` block in comment.rb -- and be sure we're not notifying revision authors twice in the extra checks below. 

Finally, we should write some tests to ensure this all works. That probably means making sure we have fixture data for a wiki page with revisions by two different authors:

https://github.com/publiclab/plots2/blob/master/test/fixtures/node_revisions.yml

And could be modeled on some of the test in this file: https://github.com/publiclab/plots2/blob/master/test/unit/comment_mailer_test.rb

maxtedford:
Resolves #1589

## Summary

As part of the notification process for comments on wiki pages, it's important that any authors of revisions to the same wiki are also notified of new comments.

## Implementation

This work accomplishes that goal by making a few updates to the `#notify` method within `comment.rb`:

1. Add four methods to the `CommentsShared` concern located at `app/models/concerns/comments_shared.rb`:
* `parent_commenter_uids` (collects all the uids for the parent node's comments)
* `parent_liker_uids` (collects all the uids for the parent node's likers)
* `parent_revisers_uids` (collects all the uids for the parent node's revisers, minus the original author)
* `uids_to_notify` (a summation of the three above, uniqued to avoid duplication)

2. Refactor the guts of the alread- tested `#notify` method on the `Comment` model as follows:

FROM:
```ruby
already = mentioned_users.collect(&:uid) + [parent.uid]

uids = []
# notify other commenters, and likers, but not those already @called out
(parent.comments.collect(&:uid) + parent.likers.collect(&:uid)).uniq.each do |u|
  uids << u unless already.include?(u)
end
```

TO:
```ruby
# notify other commenters, revisers, and likers, but not those already @called out
already = mentioned_users.collect(&:uid) + [parent.uid]
uids = uids_to_notify - already
```

That list of uids is then passed on to the already-tested `notify_users` method, which finds all users matching those in the list, and sends them an email via the `CommentMailer`.

## Additional Notes

There's an `#answer_comment_notify` method within the same `comment.rb`, which seems to function very similarly to the `#notify` method I've refactored. I considered mirroring my refactor there, but believed it to be outside the scope of this particular ticket, so decided against it. However, if the code I've produced here is amenable to you, I'd be happy to take a stab at refactoring that similarly, and perhaps even DRYing the two methods up via another method export into the shared concern. Perhaps something like `notify(current_user, answer: false)`, that makes use of an options hash as the second argument which indicates node type and helps determine which users to notify. Or perhaps a single-responsibility PORO whose job is simply to collect the relevant uids is better. I may be off base there, given my very remedial exposure to the codebase at this point, but just thought I'd throw those ideas out there.

Thanks!

### DIFF
--- a/test/fixtures/node_revisions.yml
+++ b/test/fixtures/node_revisions.yml
@@ -18,6 +18,20 @@ about_rev_2:
   title: About
   body: All about Public Lab, AGAIN
 
+about_rev_3:
+  nid: 2
+  vid: 3
+  uid: 1
+  title: About
+  body: All about Public Lab, ONCE AGAIN
+
+about_rev_3:
+  nid: 2
+  vid: 3
+  uid: 1
+  title: About
+  body: All about Public Lab, ONCE MORE
+
 three:
   nid: 3
   uid: 3
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -110,12 +110,9 @@ class Comment < ActiveRecord::Base
 
     notify_callout_users
 
+    # notify other commenters, revisers, and likers, but not those already @called out
     already = mentioned_users.collect(&:uid) + [parent.uid]
-    uids = []
-    # notify other commenters, and likers, but not those already @called out
-    (parent.comments.collect(&:uid) + parent.likers.collect(&:uid)).uniq.each do |u|
-      uids << u unless already.include?(u)
-    end
+    uids = uids_to_notify - already
 
     notify_users(uids, current_user)
     notify_tag_followers(already + uids)
--- a/app/models/concerns/comments_shared.rb
+++ b/app/models/concerns/comments_shared.rb
@@ -12,4 +12,20 @@ module CommentsShared
   def author
     DrupalUsers.find_by_uid uid
   end
+
+  def parent_commenter_uids
+    parent.comments.collect(&:uid)
+  end
+
+  def parent_liker_uids
+    parent.likers.collect(&:uid)
+  end
+
+  def parent_reviser_uids
+    (parent.revision.collect(&:uid).uniq - [parent.author.uid])
+  end
+
+  def uids_to_notify
+    (parent_commenter_uids + parent_liker_uids + parent_reviser_uids).uniq
+  end
 end
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -182,7 +182,7 @@ class Node < ActiveRecord::Base
   def answered
     self.answers && self.answers.length > 0
   end
-  
+
   # users who like this node
   def likers
     node_selections
--- a/config/sunspot.yml
+++ b/config/sunspot.yml
@@ -21,4 +21,4 @@ test:
     read_timeout: 120
     log_level: WARNING
     path: "/solr/default"
-  disabled: true
+  disabled: false
--- a/config/sunspot.yml
+++ b/config/sunspot.yml
@@ -21,4 +21,4 @@ test:
     read_timeout: 120
     log_level: WARNING
     path: "/solr/default"
-  disabled: false
+  disabled: true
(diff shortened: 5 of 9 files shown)


maxtedford:
FYI @jywarren, it looks like adding the `disable: true` flag in the solr config may have broken the build (I've taken care of the errors from the bot above already). I'm happy to drop it from this PR if you'd like, just let me know what you prefer :)

jywarren:
Ah, do you think you could leave the sunpsot.yml change as it was for the purposes of this PR? I think we do have to fix that but it may be separate and a bit more involved. Thanks!

maxtedford:
Sure thing, no problem at all. Reverted that commit 👍

maxtedford:
I noticed that the build failed after I reverted the solr commit. I tried to reproduce that failure locally, I'm not getting that test failing. I did another rebase against the upstream master just to make sure, and still nothing. Is it possible it's an intermittent failure or something?

test/fixtures/node_revisions.yml
```diff
@@ -114,7 +128,7 @@ hashtag_in_header:
   uid: 1
   title: grabs hashtags in headers
   body: |+
-    # Heading 1 is about #hashtags 
+    # Heading 1 is about #hashtags
```
jywarren: I believe this change may be triggering the bug. Tough one, and really it shouldn't be -- but the regex driving the hashtag splicing is not as broad as we'd like: that's here: https://github.com/publiclab/plots2/blob/master/config/initializers/constants.rb#L3

If we try to fix this, I'd strongly prefer opening a new issue, and re-introducing this error in the test in a separate PR, so as to respect the one-issue-per-PR convention. But I do think it's one we'd really love to fix! 

Thanks!

maxtedford: Oh wow, okay. I think that was my text editor auto-squashing anything it considers whitespace on file save. Let me add that trailing space back in, re-push, and see what happens.

If you'd like, I'd be happy to tackle the regex change as well.

jywarren:
Yeah, obscure issue! Thanks, and yes that'd be awesome. I think the regex
also choked on line starts, so perhaps we need a few good tests for that.
There's some good test text in the publiclab/PublicLab.Editor project
specs, actually.

jywarren:
This looks great, is it ready to go now? Thanks so much!

maxtedford:
Yep, if you're cool with it, it's ready to go 😄

jywarren:
Awesome and many thanks!

maxtedford:
@jywarren anytime! I'm a fan of your project. Is it cool with you if I claim a few more issues from the list and run with them?

jywarren:
Yes please, that would be wonderful! There are a number that need breaking
up into smaller projects as well, if you're interested! The "break-me-up"
tag. Thanks!

PR #1708, opened 2017-10-21
[wip] begin notes rss feeds solves #1421

StlMaris123:

Thanks!


StlMaris123:
@jywarren @ujithaperera I have opened the PR so that we can have a conversation on what is needed. Please share details of what should be included Thank you.

jywarren:
Awesome! I hope to take a look today, thanks and good to hear from you
again! Hope you're well.

jywarren:
Cool! So this is to be an RSS feed of what a particular user is following, is that right? Sorry just refreshing myself. 

For starters do you think we could reuse the RSS template from the tag views, and just pass it a distinct set of nodes? This could happen in the subscription controller. Did you get a look at how the tag controller or user controller queued up the notes for other RSS feeds?

jywarren:
This template: https://github.com/publiclab/plots2/blob/master/app/views/tag/rss.rss.builder

Look under `def rss` in this file: https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb

app/models/user.rb
```diff
@@ -278,7 +278,7 @@ def questions
     Node.questions.where(status: 1, uid: id)
   end
 
-  def content_followed_in_past_period(time_period)
+  def self.content_followed_in_past_period(time_period)
```
jywarren: Ah, here, actualy perhaps you want to skip the `.self` -- so you can run this for a specific user. I'll elaborate in the controller code.

app/controllers/subscription_controller.rb
```diff
@@ -109,6 +109,13 @@ def delete
       end
     end
   end
+  
+  def notes
+    User.content_followed_in_past_period(time_period)
```
jywarren: So here, would you do this /only/ for the current user? 

So then, `current_user.content_followed...`

To very specifically define what you're trying to do here, maybe a functional test is in order? And a unit test for `content_followed...` could help here too. But you're off to a great start!!!

jywarren:
OK, so i think you do need to remove `self.` from the model, and then in SubscriptionsController, you can do:

```ruby
def notes
  if current_user
    current_user.content_followed_in_past_period(time_period)
  else
    redirect_to '/'
  end
end
```

Does this make sense? You could also do, instead, something like this for any user:

```ruby
def notes
  user = User.find(params[:id])
  @content = user.content_followed_in_past_period(time_period)
end
```

StlMaris123:
Thank you @jywarren that solved the problem

jywarren:
Oh, looks like wer'e seeing some merge issues come through here. Do you think you could try to rebase your last six commits on top of master, rather than merging in from master? Sorry you're having a bit of trouble here, @StlMaris123 -- if you like I can try to resolve the merge errors and get a clean copy to work from! 

There's some guidance on rebasing and just selecting a few commits here: https://help.github.com/articles/using-git-rebase-on-the-command-line/ Hope that helps!

StlMaris123:
I will try doing it myself so that i too can learn. Thank you

jywarren:
:-) we're here if you run into trouble! Save a copy of your branch just in
case.

jywarren:
How's this going, @StlMaris123 -- any way we can help out?

StlMaris123:
Hi @jywarren I need some help in this issue then I can get going

jywarren:
Aha! There we go. What I did was to a) rebase everything over latest master (https://www.atlassian.com/git/tutorials/merging-vs-rebasing) but also b) i skipped the rss feed template and simplified things to return just JSON. This is looking good! It should pass; fingers X!

jywarren:
Sorry this took me a while, @StlMaris123 !! Glad to have you back -- how's the welcoming of new contributors going? Can we help out at all?

jywarren:
Hooray! I'm merging this now -- great work, and happy to have this fixed!

Issue #1702, opened 2017-10-19
Publiclab.org top menu bar: change "Store" and "Donate" to buttons


ebarry:
Hello! This is the latest update to the top menu bar. Part of https://github.com/publiclab/plots2/issues/1621

Margie (Fundraising Working Group) requested that Donate and Store appear differently from the dropdown menus, specifically as warm-colored buttons. 
We should think about how to approach this. Some points: 

* Donate and Store links in top menu bar are not dropdown menu items, instead, they both leave the publiclab.org website and go to monetary transaction areas. 
* Margie requests that Donate in particular be a warm color like orange. Maybe not red because too alarming. 
* Suggested that Donate and Store buttons should have a complementary color scheme. Perhaps a terracotta and sunflower. 
* Margie provided "Some examples of DONATE buttons from nonprofits kind of like us or with similar funders (#1 would be very nonintrusive). 1. https://www.techandciviclife.org/ 2. https://www.revealnews.org/ 3. https://ecodistricts.org/ 4. http://www.carpediemwest.org/"

ianroberts131:
If this is still available, I can take a look over the weekend.

jywarren:
Hi, I appreciate the offer of help but we need a little more time to think
through the color scheme before we're ready. If you'd like to submit a pull
request for the buttons themselves, and we can tweak the colors before
merging it, that would be helpful, but I want to be clear this may get
delayed a bit more than the usual issue due to the undecided colors. Thanks
so much though!

ianroberts131:
Yep that's fine. I can submit a pull request, and then we can iterate on the color scheme until it's final.

ryzokuken:
@jywarren I think we're using the standard color scheme that comes with bootstrap v3. If that's the case, we would just use something like the `danger` scheme to convey the idea.

Also, now that we're on this, I would also suggest to switch to bootstrap v4. It's out of alpha and looks very promising, plus it will improve the look and feel of the website quite a bit. We also won't have to take too many design decision for the same as the look of the components currently being used (eg: `navbar`) has been improved quite a bit as well.

jywarren:
@ryzokuken - what would it take to switch to Bootstrap 4? could you make an
issue with a checklist so we can start preparing for it and possibly
breaking it up into smaller pieces? Would anything immediately break, or is
there a guidance doc with a provided upgrade checklist as some projects
provide?

Thanks!

ryzokuken:
[...]

PR #1710, opened 2017-10-21
Convert Donate and Store links to buttons - solves #1702

ianroberts131:
This is a first attempt on #1702. There may be a few iterations on this to get the style/color right.

I put all of my css in `style.css` - please let me know if I should create a separate css file for the navbar or if there is a more appropriate file to place this code.

Thanks!


Gemfile.lock
```diff
@@ -370,6 +370,3 @@ DEPENDENCIES
   uglifier (>= 1.0.3)
   will_paginate (>= 3.0.6)
   will_paginate-bootstrap (>= 1.0.1)
-
```
jywarren: Hi, this is looking good -- do you think you could revert this file as it's unrelated to your changes? Sorry, this happens sometimes. Thanks!

jywarren:
Ah cool, thank you! Do you think you could upload a screenshot so we can take a look, and mention `ebarry` when you do? Thanks!!

ianroberts131:
Here is a gif of the buttons, showing how each changes on hover.

plots-2-buttons

Gemfile.lock
```diff
@@ -372,4 +372,4 @@ DEPENDENCIES
   will_paginate-bootstrap (>= 1.0.1)
 
 BUNDLED WITH
-   1.14.6
```
jywarren: Hmm, looks like it's still registering some kind of change. Maybe the best way would be to download the raw version from the master publiclab branch and just copy it over your local version, and re-add it in a new commit. Sorry, this is a pain!

jywarren:
Hi, Ian - thanks, this is looking really good! 

I wonder if we could save some lines of code by using existing Bootstrap style for the button -- see how the button on this line is defined by the Bootstrap library using the classnames `btn btn-sm btn-primary` to make it blue? 

https://github.com/publiclab/plots2/pull/1710/files#diff-faf6482c7d306c24c4e64dcdf191f7f4R124

If you look on this page there are some options for button colors; could you try refactoring to use the native Bootstrap styles instead of custom CSS (although the styles you created do look great!) --

https://getbootstrap.com/docs/3.3/css/#buttons-options

I think we could use the red and orange styles there, or perhaps better, the green and yellow styles? Green for Donate perhaps, and I wonder if we could put that one along the right side of the menubar just next to the button I linked to above (the Join button).

For the store, I think it may be worth seeing what the `btn-default` style looks like. If it shows a filled-in white button, maybe that's too much visual punch, but if it shows a white-outlined button on the black menubar, that might work. Could you give that a try?

Thanks, we really appreciate your work on this and are excited to accept these changes! 🎈

ianroberts131:
I'm working through this but wanted to run a few things by you before updating the pull request. 

First, outline options for Bootstrap aren't available until v4, and I think we're using v3 so I won't be able to get the same effect without some custom css. If I use the built in bootstrap classes, I can still do something similar to what was done before with the outlines, but with fewer lines of code (still a little custom css though).

As for the colors, the green really isn't coming out very well, so if we go the bootstrap color route I think warning (orange), primary (blue), and then default (grey) are the best options.

If I go this route, I'm able to trim the custom css to the following:

```
.nav-button-donate, .nav-button-store {
  display: inline-block;
  height: 35px;
  background-color: transparent;
  margin-left: 1em;
  text-transform: uppercase;
}

.nav-button-donate:hover a, .nav-button-store:hover a {
  color: black;
}

.nav-button-donate a {
  color: #eea236;
  text-decoration: none;
  font-weight: 600;
}

.nav-button-store a {
  color: #ccc;
  text-decoration: none;
  font-weight: 600;
}
```

And the final result is in the gif below.

plots-2-buttons-v2

You also mentioned possibly moving the donate button over to the right hand side next to the 'Join' button. Currently, we are just hiding the store/donate buttons when the screen size gets too small to fit it (using the class `hidden-xs hidden-sm hidden-md`). Do you want to possibly make it so that the donate button is always visible? I think it may be worthwhile to split that into another issue (which I'm happy to tackle of course).

Let me know your thoughts.

jywarren:
Hi, sorry i took a bit to reply. I think you make a good point that moving the buttons is a bit complex due to screen size change, so let's just explore that in a follow-up issue; good call. 

I think this change looks good! Have you had a chance to try changing the window width a bit to see how it flows at different widths? Sometimes there can be a gotcha near a transition point. 

Thanks so much for your careful work on this!

ianroberts131:
I tested what happens when I change the window width, and the buttons hide in the same manner as they do currently as just links. There don't seem to be any layout issues when the window size is changed.

jywarren:
Cool! I resolved conflicts -- letting tests run again.

jywarren:
OK, to test this out live, i've pushed this up to http://unstable.publiclab.org, so we can try it out one more time. just being extra careful since this affects every page on the site! But if it looks good we can merge in and close this up! I saw the other issue you created -- thanks so much!!!

jywarren:
The `unstable` build will take some minutes. You can watch at http://jenkins.laboratoriopublico.org/ !

jywarren:
OK - having played with it a bit, I'd like to ask for one more change -- apologies! I think the orange is just a bit too loud for every page, and I'd like to ask if we could change it to `#3355ee` -- like this:

screenshot 2017-10-26 at 11 39 18 pm

jywarren:
I appreciate your hard work on this! Doing design work alongside coding is not an easy combo.

jywarren:
And note that you'll probably have to `git pull https://github.com/ianroberts131/plots2 create-store-and-donate-buttons` to get the resolved conflicts I added! 

Thanks!!!

ianroberts131:
I think I finally got that to work....apologies for the unnecessary commits.

jywarren:
OK, awesome, no problem. I pushed it to unstable again so we can take a look. Thanks!

jywarren:
Wow, it looks great! Thanks so much, this is super. I'm going to merge now. 

If you're looking for a new challenge, please take a look at our `help-wanted` list: https://github.com/publiclab/plots2/labels/help-wanted

Thanks for your help!!!

Issue #1641, opened 2017-09-25
Performance issue optimize query


icarito:
On 25/09/17 11:38, Jeffrey Warren wrote:
> Hi! I believe this has to do with this query:
>
> https://github.com/publiclab/plots2/blob/master/app/controllers/admin_controller.rb#L51
>
> And could be solved by refactoring and removing status != 1 users. 

So two actions required (or one of either?):
  - [ ] refactoring (adding index to model?)
  - [x] remove users with status not 1 (confirm?)

icarito:
> removing status != 1 users

You mean doing something like DROP FROM rusers WHERE status!=1 ?
Or would a migration be preferable?

jywarren:
nono, i just mean adding a filter like `.where(status: [1,4])` -- though this would limit the search (used by admins) to verified accounts. But that should be OK. 

### Steps to resolve:

```ruby
        @users = User.where(email: params[:address])
                 .where(status: [1,4])
```

jywarren:
This has been marked as a good candidate for becoming a [first-timers-only issue](http://www.firsttimersonly.com) like [these](https://github.com/publiclab/plots2/labels/first-timers-only), meaning that it's simple, self-contained, and with some extra formatting, could be a great entry point for a new contributor. If you're familiar enough with this code, please consider [reformatting or reposting it as a first-timers-only issue](https://publiclab.org/notes/warren/10-31-2016/create-a-welcoming-first-timers-only-issue-to-invite-new-software-contributors), and then ping @jywarren to get it labelled. Or, if this is not your first time, try to solve it yourself!

****

icarito:
Here's the top queries for Monday

QUERY TIME - BY MEAN | HITS | SUM | MEAN | STDDEV | MIN | MAX | 95 %TILE
-- | -- | -- | -- | -- | -- | -- | --
publiclab@172.18.0.1: SELECT term_data.* FROM term_data INNER JOIN community_tags ON community_tags.tid = term_data.tid INNER JOIN community_tags node_tags_term_data_join ON node_tags_term_data_join.tid = term_data.tid INNER JOIN node ON node.nid = node_tags_term_data_join.nid WHERE (node.status = :int) GROUP BY name ORDER BY count DESC LIMIT :int OFFSET :int | 26 | 14m46s | 34.07s | 7.69s | 25.83s | 1m01s | 25.76s-1m02s
[...]

PR #1738, opened 2017-10-27
Optimization for front page queries

jywarren:
re: #1641

For docs, see: `1.6 Low-Level Caching`:

http://guides.rubyonrails.org/caching_with_rails.html

We may have to enable `memcached` -- but it may already be on. If it's not, the cache will not expire, so it's OK to push to production to find out, if we want.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -9,7 +9,9 @@ class HomeController < ApplicationController
   # caches_action :index, :cache_path => { :last => Node.find(:last).updated_at.to_i }
 
   def home
-    self.activity
+    Rails.cache.fetch("front-activity", expires_in: 10.minutes) do
+      self.activity
+    end
     @title = I18n.t('home_controller.science_community')
     if current_user
       redirect_to '/dashboard'
@@ -34,7 +36,9 @@ class HomeController < ApplicationController
   # route for seeing the front page even if you are logged in
   def front
     @title = I18n.t('home_controller.environmental_investigation')
-    self.activity
+    Rails.cache.fetch("front-activity", expires_in: 10.minutes) do
+      self.activity
+    end
     render template: 'home/home'
   end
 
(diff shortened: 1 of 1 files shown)


jywarren:
OK, we can't do `Rails.cache.fetch` except to return a specific value. So we need to refactor accordingly -- doing so now!

jywarren:
Refactored!

icarito:
Thanks for looking into this! When this passes I'll deploy it and hopefully things will be back to normal...

jywarren:
Hmm, it doesn't like us putting arbitrary records in that kind of cache. https://github.com/stripe/stripe-ruby/issues/90

Trying something different...

jywarren:
Hmm, this is saying we would need to stringify this and then reconstitute it. That's expensive and seems wrong-headed. I think we have to figure out a different way to do this; maybe caching the sub-requests, rather than the whole thing?

jywarren:
OK @icarito -- took longer than I thought but this looks good now, whew...

icarito:
Thanks! I'll be testing and deploying and we'll see.

icarito:
Shall I do squash and merge then, right?

jywarren:
Yes please!

icarito:
Okay, jenkins deployed to staging and we're facing 500 error. Log says:
```
Started GET "/" for 190.42.85.127 at 2017-10-27 23:22:46 +0000                                                                                
Processing by HomeController#home as HTML                                                                                                     
Completed 500 Internal Server Error in 355.7ms                                                                                                
                                                                                                                                              
NoMethodError (undefined method `group' for #<Array:0x000000060a89c0>):                                                                       
  app/controllers/home_controller.rb:114:in `activity'                                                                                        
  app/controllers/home_controller.rb:13:in `block in home'                                                                                    
  app/controllers/home_controller.rb:12:in `home'                                                                                             

```

jywarren:
Ufff. I'll look into it tonight or very early tomorrow... apologies
Sebastian!

jywarren:
Found the issue and am pushing to master and merging to production. Oversight on my part. see http://jenkins.laboratoriopublico.org

jywarren:
OK, we're good on stable.

jywarren:
publishing to production.

Issue #1713, opened 2017-10-22
Email notifications for questions, etc., Could be better for UX


pdhixenbaugh:
The template for alerting users about new answers makes it difficult for the user to see at a glance if the information is relevant.

Improving this would help users return to the PL website after seeing the emails with relevant information at a glance in their inbox.

It might be useful to look at all the email templates and compare them to examples from other way platforms to see if they can be improved to provide a better experience for users.

A better template might be something like,

~~~
"[PL]Q: What are some lessons learned...? An answer has been posted on Public Lab"

Hi! <User> responded: <first few words of answer>. Do NOT reply to this email; click this link to see the answer:
https://publiclab.org/questions/pdhixenbaugh/03-14-2017/what-are-some-lessons-learned-using-the-riffle-especially-underwater#a497
~~~

Compared to the older version:
~~~
"[PublicLab] New answer to Question: What are some lessons learned using the Riffle? (Especially Underwater)

Hi! A new answer has been posted for your question 'What are some lessons learned using the Riffle? (Especially Underwater)'. Do NOT reply to this email; click this link to see the answer:
https://publiclab.org/questions/pdhixenbaugh/03-14-2017/what-are-some-lessons-learned-using-the-riffle-especially-underwater#a497

Here is the answer that anuj2017 wrote:

(First few words of answer)..... Lorem ipsem"
~~~

The only problem I can foresee with this is spammers taking advantage of this vector to get eyeballs on their message. To minimize that problem it might be prudent to hold answers in moderation for new users.

What do you all think?

pdhixenbaugh:
![screenshot_20171022-110743](https://user-images.githubusercontent.com/16969821/31864992-226b3026-b71c-11e7-897b-77fe38149107.png)

Here is the information-poor current version, and how it shows up in the inbox

jywarren:
I like this -- the templates are available in this folder: https://github.com/publiclab/plots2/tree/master/app/views/answer_mailer

There are a few of them, but perhaps we can start with the one to the original author, and work from there: 

https://github.com/publiclab/plots2/blob/master/app/views/answer_mailer/notify_question_author.html.erb

This may disrupt peoples' filters a little bit but I think it's worth it for the easier readability. Thanks for the great suggestion -- i'm marking this as a first-timers-only candidate, and hopefully someone will submit a fix based on your before/after changes above!

Gauravano:
@jywarren I would like to take this issue 

A few doubts:-
[...]

PR #1763, opened 2017-11-07
Notify_author_question template changed

Gauravano:
solves #1713 
I have changed the UI according to the issue discussion 
@jywarren  kindly suggest if any changes required. Thank You.

### DIFF
--- a/app/views/answer_mailer/notify_question_author.html.erb
+++ b/app/views/answer_mailer/notify_question_author.html.erb
@@ -1,5 +1,5 @@
 Hi! <a href='https://<%= ActionMailer::Base.default_url_options[:host] %>/profile/<%= @answer.author.name %>'><%= @answer.author.name %></a> responded :
-<p><%= @answer.content.truncate(30) %></p>
+<hr /><p><%= @answer.content %></p><hr />
 
 Do NOT reply to this email; click this link to see the answer:
 
--- a/app/views/answer_mailer/notify_question_author.html.erb
+++ b/app/views/answer_mailer/notify_question_author.html.erb
@@ -1,23 +1,18 @@
-Hi! A new answer has been posted for your question '<a href='https://<%= ActionMailer::Base.default_url_options[:host] %><%= @answer.node.path(:question) %>'><%= @answer.node.title %></a>'. Do NOT reply to this email; click this link to see the answer:
+Hi! <a href='https://<%= ActionMailer::Base.default_url_options[:host] %>/profile/<%= @answer.author.name %>'><%= @answer.author.name %></a> responded :
+<p><%= @answer.content.truncate(30) %></p>
 
-<p>https://<%= ActionMailer::Base.default_url_options[:host] %><%= @answer.node.path(:question) %>#a<%= @answer.id %></p>
+Do NOT reply to this email; click this link to see the answer:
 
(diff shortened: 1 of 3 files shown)


app/views/answer_mailer/notify_question_author.html.erb
```diff
@@ -1,23 +1,18 @@
...
+<p><%= @answer.content.truncate(30) %></p>
 
-<p>https://<%= ActionMailer::Base.default_url_options[:host] %><%= @answer.node.path(:question) %>#a<%= @answer.id %></p>
+Do NOT reply to this email; click this link to see the answer:
 
-<p><b>Here is the answer that <a href="https://<%= ActionMailer::Base.default_url_options[:host] %>/profile/<%= @answer.author.name %>"><%= @answer.author.name %></a> wrote:</b></p>
-
-<hr /> 
-
-<p><%= raw auto_link(RDiscount.new(@answer.body_email(ActionMailer::Base.default_url_options[:host])).to_html) %></p>
```
jywarren: I think we're not quite ready to remove the full text from the email -- maybe we could revert this and think that through as a follow-up? I'd like to keep the entire text in for now. I'm not as worried about spammers, and think it's really great for a "quick look" at a new post. Thanks!

app/views/answer_mailer/notify_question_author.html.erb
```diff
@@ -1,23 +1,18 @@
-Hi! A new answer has been posted for your question '<a href='https://<%= ActionMailer::Base.default_url_options[:host] %><%= @answer.node.path(:question) %>'><%= @answer.node.title %></a>'. Do NOT reply to this email; click this link to see the answer:
+Hi! <a href='https://<%= ActionMailer::Base.default_url_options[:host] %>/profile/<%= @answer.author.name %>'><%= @answer.author.name %></a> responded :
```
jywarren: Cool!

Gauravano:
hi @jywarren, I have implemented changes as requested by you, sir. Now the whole answer would be visible but only initial part of question would be visible on the subject of mail sent to users. 

pull2_att2

pull2_attt2

jywarren:
Great! @pdhixenbaugh - what do you think about this? The screenshot here shows the new subject line. 

@Gauravano - I think @pdhixenbaugh also noted we could change the prefix from [PublicLab] to [PL] -- what do you think of that? 

Sorry for my slow response, we just had the Barnraising (our annual conference) and then I've been traveling since so I'm just catching up now! Your work on this is much appreciated :-)

jywarren:
Oh, actually it was `[PLab]` and not `[PL]`. I think that makes sense!

Gauravano:
First of all, no problem about delay @jywarren sir.
As you suggested I have changed [PublicLab] to [PLab], also I want to highlight that apart from the subject, content part of mail is also different.Here's the updated template, kindly review.Thank you.

screenshot from 2017-11-14 16-31-24

jywarren:
Awesome, this looks great. Just going to wait one more day to see if @pdhixenbaugh has a chance to chime in, thanks!!!

pdhixenbaugh:
Looks beautiful!

jywarren:
Awesome, thanks to everybody for pitching in on this one. We're making solid progress!

Issue #1751, opened 2017-11-01
@usernames are not linked when they begin a bulleted line


ebarry:
### Please describe the problem

In a research note, i noticed two usernames were not linked, and so i checked and indeed they are still active accounts

### Please show us where to look

https://publiclab.org/notes/Shannon/07-06-2016/public-lab-code-of-conduct
-- scroll down into the Refinements section, where neither /@klie/ nor /@kanarinka/ are linked

### What's your PublicLab.org username?

liz

### Browser, version, and operating system

> Many bugs are related to these -- please help us track it down and reproduce what you're seeing!

****

## Thank you!

Your help makes Public Lab better! We *deeply* appreciate your helping refine and improve this site. 

To learn how to write really great issues, which increases the chances they'll be resolved, see:

https://publiclab.org/wiki/developers#Contributing+for+non-coders

jywarren:
Great catch! Can you add a sample text that's not working and someone
hopefully can write a test for it, and develop a fix?

Here's the line of code we use to find usernames starting with a @ -

https://github.com/publiclab/plots2/blob/master/config/initializers/constants.rb#L2

http://rubular.com is a great site for debugging regular expressions like
this.

ebarry:
Ah -- left out the important observation, that this is happening after an `asterisk space` combo to make a bullet:

Sample text: 

* @kanarinka
* @klie

SidharthBansal:
I think FINDER = /(^|\s)\@([\w-]+)/ should be replaced by FINDER = /[\u2022,\u2023,\u25E6,\u2043,\u2219]?(^|\s)\@([\w-]+)/  as bullet can be of five types 
    Bullet (•)
    Triangular Bullet (‣)
    White Bullet (◦)
    Hyphen Bullet (⁃)
    Bullet Operator (∙)
So, I have used 5 regex coding with it. 
I have made a pull request #1764

SidharthBansal:
Sorry, I checked it but it should be /(^|\s)?@([\w-]+)/ 
This is my first time.
It is working 
![image](https://user-images.githubusercontent.com/20972099/32632938-8642e400-c5cb-11e7-9001-f870fde3a927.png)
I have made a pull request #1765. Tell me if I am wrong anywhere.

jywarren:
Awesome! I'll be going through new submissions tomorrow morning; sorry have
been a bit behind with an event we were running! Thanks so much for your
contribution!

PR #1765, opened 2017-11-09
@user made work after bullets #1751

SidharthBansal:


jywarren:
Hi @SidharthBansal -- are you stuck on this? I really recommend using Rubular.com to help debug; having a great set of test strings and running them through Rubular can let you iterate quickly on getting a regular expression working quickly. If you get stuck, please share a Rubular.com link with us here and we can help out! Thank you!!!

SidharthBansal:
First of all, I found that there are 2 errors instead of one.
First error is if there is a new line then the regular expression is not working
Second: If it is after a non-word character it is not working. May be bullet or '-' etc. Eg. Checkout my profile -@sidharth will not work
So, I tried the following:
I used a '^' for new line and \W for non word character.So, it should be   FINDER = /(^|\W)\@([\w-]+)/ instead of   FINDER = /(^|\s)\@([\w-]+)/ .
The rubular link is http://rubular.com/r/dKqYq2ZKbe
The tests are passing in rubular. 
Also I ran the server again after having changes, output is:
image
The earlier commit was not passing because I think there was a clash between email addresses and user tags. But now, I guess there isn't.
Thanks @jywarren

SidharthBansal:
First of all, there are 2 errors instead of one.
First error is if there is a new line
Second: If there is a non-word character before @user_name. For example: Bullet point and a username or Checkout of profile at -@user or there is a space before the name
So I did the following
I replaced the space by non word character ie \s by \W
So finder should be 
  FINDER = /(^|\W)\@([\w-]+)/
It is passing test cases on rubular - http://rubular.com/r/dKqYq2ZKbe
Also 
image
Thanks @jywarren

SidharthBansal:
I dont know why travis test cases are failing .
It is showing ================================================================================
Failure:
  <"REPLACE REPLACE `@StlMaris123`"> expected but was
  <"REPLACE REPLACE  REPLACE`">.

jywarren:
The test string is here: https://github.com/publiclab/plots2/blob/43dcbcd8601e8a4fa23bef9eda7030b8e60a61e7/test/unit/constants_test.rb#L6

See how in this version it matches all 3 instances of 'StlMaris123': http://rubular.com/r/fh9glRXmpQ

It's not supposed to match the one with ``` ticks -- so I think if you can get that string to match only the first two instances of 'StlMaris123' and not the one in back-ticks, you'll be good!

Thanks!!!

jywarren:
Moving to #1775 !

PR #1771, opened 2017-11-12
tests and adjustments for user record creation/update

### LINKED ISSUE #1769: Error on adding tag related to tag subscriber email notification
https://github.com/publiclab/plots2/blob/348334366f6ccf499a80498e5e06667417f8d764/app/mailers/subscription_mailer.rb#L44

```
Sent mail to AAAAAAAAAA (47.6ms)
Completed 500 Internal Server Error in 413.7ms

Net::SMTPSyntaxError (501 <AAAAAAAAAA>: recipient address must contain a domain
):
  app/mailers/subscription_mailer.rb:44:in `block in notify_tag_added'
  app/mailers/subscription_mailer.rb:41:in `each'
  app/mailers/subscription_mailer.rb:41:in `notify_tag_added'
  app/models/node.rb:677:in `block in add_tag'
  app/models/node.rb:661:in `add_tag'
  app/controllers/tag_controller.rb:167:in `block in create'
  app/controllers/tag_controller.rb:161:in `each'
  app/controllers/tag_controller.rb:161:in `create'
```

### LINKED ISSUE #1770: Email address not recognised when registering
Hi - I'm having a problem registering my email address for a new account. 

I went to the register page - https://publiclab.org/register, filled in the information and received this error message "Email should look like an email address." I am using a .today domain name for my email address which I believe might be the problem. 

Chrome, Version 61.0.3163.100, Mac OS. Also the same result on Chromium on Ubuntu. 

Thank you! 
~Gabi

jywarren:
fixes #1769 and fixes #1770

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -9,17 +9,15 @@ class UsersController < ApplicationController
   end
 
   def create
-    # craft a publiclaboratory OpenID URI around the PL username given:
-    params[:user][:openid_identifier] = "https://old.publiclab.org/people/"+params[:user][:openid_identifier]+"/identity" if params[:user] && params[:user][:openid_identifier]
-    using_recaptcha = !params[:spamaway] && Rails.env == "production"
-    @spamaway = Spamaway.new(params[:spamaway]) unless using_recaptcha
     @user = User.new(params[:user])
-    if ((@spamaway && @spamaway.valid?) || (using_recaptcha && recaptcha = verify_recaptcha(model: @user))) && @user.save({})
+    using_recaptcha = !params[:spamaway] && Rails.env == "production"
+    recaptcha = verify_recaptcha(model: @user) if using_recaptcha
+    @spamaway = Spamaway.new(params[:spamaway]) unless using_recaptcha
+    if ((@spamaway && @spamaway.valid?) || recaptcha) && @user.save({})
       if current_user.crypted_password.nil? # the user has not created a pwd in the new site
         flash[:warning] = I18n.t('users_controller.account_migrated_create_new_password')
         redirect_to "/profile/edit"
       else
-        @user.update_attribute(:bio, params[:drupal_user][:bio])
         @user.add_to_lists(['publiclaboratory'])
         flash[:notice] = I18n.t('users_controller.registration_successful').html_safe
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -38,13 +39,13 @@ class User < ActiveRecord::Base
...
   validates_with UniqueUsernameValidator, on: :create
   validates_format_of :username, with: /^[A-Za-z\d_\-]+$/
+  validates_format_of :email, with: /@/
 
   before_create :create_drupal_user
   before_save :set_token
   after_destroy :destroy_drupal_user
 
   def create_drupal_user
-    self.bio = "" # needed to set a default bio value of ""
```
sagarpreet-chadha: I guess we should add **self.bio=""** here or in **user_test.rb** when we create a new user ☺️ .

app/controllers/users_controller.rb
```diff
@@ -10,6 +10,7 @@ def new
 
   def create
     @user = User.new(params[:user])
+    @user.bio ||= ''
```
sagarpreet-chadha: During Testing of **'user creation'**  : https://github.com/publiclab/plots2/blob/348334366f6ccf499a80498e5e06667417f8d764/test/unit/user_test.rb#L4 
we are creating user over there , so i guess this **@user.bio ||= ' '**  statement here will not run (for testing).
So i guess we should either give **bio: 'something'** attribute here :
https://github.com/publiclab/plots2/blob/348334366f6ccf499a80498e5e06667417f8d764/test/unit/user_test.rb#L5
OR
maybe here (because this function is called before the user is created) : 
https://github.com/publiclab/plots2/blob/348334366f6ccf499a80498e5e06667417f8d764/app/models/user.rb#L47

What do you think sir ? :smile:
Thank you !

jywarren: Oh thank you for your help -- i've been traveling and working on events recently and have been slow on responding to pull requests, but will do my best to catch up!

jywarren: `user.rb#47` was causing me trouble in tests, as I've rearranged the ordering a bit here. But I think you're right we have to do it in the model. Let me try again here, thanks for the assistance!

app/models/user.rb
jywarren: I tried changing it to self.bio ||= "" instead of moving it to the controller. Let's see how that does!!!

jywarren: Because it had been overwriting the given bio in tests.

jywarren:
Awesome! Merging now.

PR #1818, opened 2017-12-01
upgrade to solr 5.5 container

jywarren:
This doesn't yet work, but is something we may want to debug and merge in later.


jywarren:
This actually seems OK -- @icarito shall we push this to unstable to test it out?

jywarren:
Closing as we're switching to mysql full text search in #1913

Issue #232, opened 2015-02-21
forced password reset


jywarren:
@btbonval sez:

> It'd be nice to have a feature for admins which allows them to forcibly change a user's password to some kind of internally generated nonsense, which is emailed to the user. The admin need never see the generated password.
> If an account is compromised, this would be a simple fix so long as the user's email address is a trusted out of band communication channel.

> This should actually be fairly simple to implement. While trying to reset a password by hand, I found that a user's hashed password and salt may be changed using the `password` attribute. The `reset_key` attribute allows the user to reset his or her password by typing in the generated key; this should probably be done using the pre-existing functions.

> For a quick preview, it might look like this:

``` ruby
if current_user && current_user.role != 'moderator' && current_user.role != 'admin':
  flash[:error] = 'YOU SHALL NOT PASS!'
else:
  user = User.find_by_whatever(something distinct)
  user.password = random_generator()
  user.reset_key = key_generator()
  user.save!
  email_key_to_user()
```

> I can't test this personally due to #166

> It turns out no accounts were (as yet) compromised. This ticket is still useful should an account be compromised at a later date, but there is less urgency.

btbonval:
Not sure if I meant this originally, but I see this as being a pretty simple process

1. change the user's password to random gibberish using JS or Ruby and don't report that password to anyone. Effectively the user account is disabled from logins until the password is changed.
2. Issue a password reset key using the conventional mechanism. The only difference is that the user didn't request it by clicking a button on the UI, otherwise the same internal mechanisms should work.

jywarren:
If only admins can do this, I think it should be OK to use something like `(rand * 100000000000).to_i` for the password and key generators... or is there a stronger way to generate these? 

I'd like to break this up into smaller pieces:

* random pwd and key generating functions
* a method contained in user.rb: https://github.com/publiclab/plots2/blob/master/app/models/user.rb  called `def force_reset_key` or something
* a unit test

Marking it `break-me-up` so we can give people next steps and get this done.

Kevin-Kawai:
@jywarren Hi I've been looking around issues trying to find something I can take part in. Is this issue still live? I think I generally understand the task and what to do. If it's still something that needs to get done I'd be super excited to give it a go!

jywarren:
Yes, this would be great -- thank you, we'd love your help!

Take a look at the admin_controller to see the functions we have now. I
think this is a remix of password reset key and password gibberish-izing as
should be reflected in that controller -- thanks!!!

[...]

PR #1834, opened 2017-12-08
admin forced password reset

Kevin-Kawai:

This is the branch for the issue #232. It allows an admin to start the password reset procedure for a user through their profile page. 

Thanks!


config/routes.rb
```diff
@@ -38,11 +38,11 @@
   get 'ioby' => "legacy#ioby"
 
   get 'login' => "user_sessions#new",      :as => :login
-  delete 'logout' => "user_sessions#destroy", :as => :logout
+  get 'logout' => "user_sessions#destroy", :as => :logout
   post 'register' => 'users#create'
   get 'reset' => 'users#reset'
   post 'reset' => 'users#reset'
-  post 'reset/key/:key' => 'users#reset'
+  get 'reset/key/:key' => 'users#reset'
```
Kevin-Kawai: I had to change this verb and the above to get in order to get them to work on my local machine. Since they aren't really the parts I had to work on I assume they are correct? And I will revert them back after I have finished.

app/controllers/admin_controller.rb
```diff
@@ -43,6 +43,21 @@ def demote_basic
...
+    if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
+      user = User.find(params[:id])
+      if user
+        key = user.generate_reset_key
+        user.save({})
+        # send key to user email
+        PasswordResetMailer.reset_notify(user, key) unless user.nil? # respond the same to both successes and failures; security
+      end
+      flash[:notice] = I18n.t('users_controller.password_reset_email')
+      redirect_to "/login"
```
Kevin-Kawai: I think the page should not redirect to the log in page since it's the admin that's starting the password reset

config/routes.rb
```diff
@@ -222,6 +222,7 @@
   put 'moderate/spam/:id' => 'admin#mark_spam'
   put 'moderate/publish/:id' => 'admin#publish'
   put 'admin/promote/moderator/:id' => 'admin#promote_moderator'
+  get 'resetuserpassword/:id' => 'admin#reset_user_password'
```
Kevin-Kawai: I think this name might not be that great and another path name would be good.

Kevin-Kawai:
@jywarren I have created a PR for #232 this is still a work in progress but if you have a chance to take a look and let me know if I'm on the right track that would be great! Thanks!

config/routes.rb
jywarren: I think you fixed a latent bug related to the Rails 4 upgrade-- i've spent much of the day chasing a few of these which we missed when doing the `routes.rb` refactoring. Thank you!

app/controllers/admin_controller.rb
jywarren: Hmm, can you detect that it's not a typical user and redirect to... the profile page?

config/routes.rb
jywarren: maybe... `/user/reset/password` ?

jywarren:
Very cool and apologies for the routes.rb conflicts -- as i mentioned i've been fixing lots of things in that file. Please try to preserve your changes and rebase them into the latest version. I'm happy to help with this of course! Many thanks!

jywarren:
This is looking good -- actually we don't want moderators to be able to force reset -- sorry if that was wrong in the request. Is this otherwise good to go? Awesome!

Kevin-Kawai:
oh sorry that last commit might be misleading. I meant that an Admin can reset a moderator's password if they need to. Or do we not want that to be a possibility? 

I think for the most part it is finished. I'm just thinking about changing the flash message that pops up when the Admin resets the password because right now it's the same message as if a user reset it themselves (something like "You should receive an email.....") and maybe "@User's email has been reset. They should receive instructions...."

Kevin-Kawai:
@jywarren I just changed the message so I think the whole PR is good to go! Let me know if there is anything you need me to change

app/controllers/admin_controller.rb
```diff
@@ -43,6 +43,22 @@ def demote_basic
     redirect_to '/profile/' + @user.username + '?_=' + Time.now.to_i.to_s
   end
 
+  def reset_user_password
+    if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
```
jywarren: Hi, doesn't this mean that moderators would be able to reset passwords? If we could just change this to admins only, that'd be great. Thanks!!!

Kevin-Kawai: I apologize, you are correct. This would allow moderator's to reset a user's password. I will fix this after work and push the changes.

Kevin-Kawai:
@jywarren Sorry about that last bit of confusion. I fixed the controller so the moderator doesn't have the ability to reset passwords. Everything else should be good to go! Let me know if there's anything else.

jywarren:
Super, thanks so much!!!

🎉

PR #1838, opened 2017-12-08
Expand Tag.tagged_nodes_by_author() (Resolves #1833)

### LINKED ISSUE #1833: Expand Tag.tagged_nodes_by_author() method to accept "*" tagname wildcards
This line shows the new method added by @SidharthBansal -- 

```ruby
Tag.tagged_nodes_by_author(tagname, user_id)
```
https://github.com/publiclab/plots2/blob/e72cb97ce12f81f68b5195c00e5fb6a22ee08f83/app/models/tag.rb#L227-L232

We want to be able to match "wildcard" tagnames like any tag starting with `activity:` -- often denoted with an asterisk (`*`) -- like in this query: https://publiclab.org/tag/activity:*

This isn't such a tough query -- there's an example here:

https://github.com/publiclab/plots2/blob/e72cb97ce12f81f68b5195c00e5fb6a22ee08f83/app/controllers/tag_controller.rb#L60-L77

See how we detect a wildcard and then switch to an alternative query? This could potentially be more compact, but it works.

Any help with this would be much appreciated!

Once we do this, we can move on to listing all activities a user has done on their profile, by using the query:

```ruby
Tag.tagged_nodes_by_author('replication:*', @profile_user.id)
```

But not yet! Let's do this first and then we'll have a good basis to work from. Thanks!!!

SidharthBansal:
In progress
Updated tag.rb
Tests remaining
Fixes #1833

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -239,10 +239,4 @@ class Tag < ActiveRecord::Base
     end
   end
 
- def self.tagged_nodes_by_author(tagname, user_id)
-    Node.where('term_data.name = ?', tagname)
-        .includes(:node_tag, :tag)
-        .references(:term_data)
-        .where('node.uid = ?', user_id)
- end
 end
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -28,7 +28,7 @@ class Tag < ActiveRecord::Base
   validates :name, presence: :true
   validates :name, format: { with: /\A[\w\.:-]*\z/, message: 'can only include letters, numbers, and dashes' }
   # validates :name, :uniqueness => { case_sensitive: false  }
-                                                            
+
   def id
     tid
   end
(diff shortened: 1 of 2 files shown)


jywarren:
This is looking good -- can you demonstrate that it works using a test? You should be able to get, for example, `balloon-mapping` to return even if you search for `balloon*`. Make sense? Thanks!

SidharthBansal:
I have added tests.
Thanks for help @jywarren  sir

jywarren:
Awesome! This looks perfect, thank you!!

jywarren:
🎉 super -- would you be interested in adding a new page, similar to the "tag" page like https://publiclab.org/tag/hydrogen-sulfide, but for just one author's work, so:

https://publiclab.org/tag/hydrogen-sulfide/author/warren ?

This'd require:

1. a `config/routes.rb` entry
2. a re-use of the `/app/views/tag/show.html.erb` template with a custom title
3. a new action in `/app/controllers/tag_controller.rb` for, maybe `def show_for_author` ? which uses your new method here
4. a functional test in `/test/functionals/tag_controller_test.rb` to show that the new `show_for_author` action works

This would be a bigger project, but would be pretty awesome to see. I'd be happy to help you through it, maybe starting with creating a new issue describing it?

jywarren:
https://github.com/publiclab/plots2/issues/1854 !! :-)

SidharthBansal:
Yeah I am happy to work at it

jywarren:
:-) super! So glad to have your help.

PR #1861, opened 2017-12-11
Add performance tests via rails-perftest (stuck!)

### LINKED ISSUE #1801: Add performance tests for leading pages, like / and /dashboard
https://github.com/publiclab/plots2/blob/8e6ec5a29a7ebb4c873cf5b953468addc1df6547/test/performance/browsing_test.rb#L10

This is documented at https://rubygems.org/gems/rails-perftest (as of Rails 4, which we're moving to shortly)

@icarito - just a note!

bellvat:
* [x] all tests pass -- `rake test:all` **Please see "Issue" below**

I have made the following changes -- fixes #1801 :
1. Added perftest gem
2. Added ruby prof gem
3. Uncommented code in browsing_test.rb, and changed syntax of line 9 to: 
 `  test "browsers" do `
4. I have not make changes to the Rakefile yet. 

**Issue:** When I ran `$ bundle exec rake test:benchmark`, which should give us the performance statistics, I kept getting this error:
screen shot 2017-12-08 at 3 39 52 pm

1. I tried to debug this by changing my ruby version, as I read in some posts that it might be that Rails 4.1.6 may not be compatible with Ruby 2.4. I am not as familiar with how rbenv works, and although I was able to install a different version of ruby, I was not able to get rails to work with that version.
2. The same error message came up when I tried running `rails s' to try and start the server

Any guidance would be appreciated! 

Thanks for your help!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -121,6 +121,7 @@ gem 'jbuilder'
 # Pin mustermann to Ruby 2.1 compatible
 gem 'mustermann' , '~> 0.4'
 
+<<<<<<< 51c9abdf9d37eef37439e845a1e1fe59b569e717
 #Gem for assertions used in testing
 gem 'rails-dom-testing'
 #OAuth Based login
@@ -139,3 +140,6 @@ gem 'sanitize'
 gem 'rails-perftest'
 gem 'ruby-prof', group: :test
 gem 'ruby-prof', :github => 'ruby-prof'
+=======
+gem 'ruby-prof','~> 0.15.9', group: :test
+>>>>>>> Added ruby-prof 0.15.9 and uncommented browsing test
--- a/Gemfile
+++ b/Gemfile
@@ -131,3 +131,7 @@ gem 'omniauth-github', '~> 1.1', '>= 1.1.2'
 gem 'activerecord-tableless'
 gem 'figaro'
 gem 'sanitize'
+
+#Performance testing
+gem 'rails-perftest'
+gem 'ruby-prof', group: :test
--- a/Gemfile
+++ b/Gemfile
@@ -121,7 +121,6 @@ gem 'jbuilder'
 # Pin mustermann to Ruby 2.1 compatible
 gem 'mustermann' , '~> 0.4'
 
-<<<<<<< 51c9abdf9d37eef37439e845a1e1fe59b569e717
 #Gem for assertions used in testing
 gem 'rails-dom-testing'
 #OAuth Based login
@@ -140,6 +139,3 @@ gem 'sanitize'
 gem 'rails-perftest'
 gem 'ruby-prof', group: :test
 gem 'ruby-prof', :github => 'ruby-prof'
-=======
-gem 'ruby-prof','~> 0.15.9', group: :test
->>>>>>> Added ruby-prof 0.15.9 and uncommented browsing test
--- a/Gemfile
+++ b/Gemfile
@@ -51,7 +51,7 @@ end
 group :test, :development do
   gem 'test-unit'
   gem 'rails-perftest'
-  gem 'minitest-reporters', '~> 1.1.19'
+  gem 'minitest-reporters', '~> 1.2.0.beta2'
   gem 'rake',  '~> 10.5.0'
   # gems to test RESTful API
   gem 'rest-client'
--- a/Gemfile
+++ b/Gemfile
@@ -121,7 +121,6 @@ gem 'jbuilder'
 # Pin mustermann to Ruby 2.1 compatible
 gem 'mustermann' , '~> 0.4'
 
-<<<<<<< 3326f593e11ec51c6529431588ea4d7e705f1a4b
 #Gem for assertions used in testing
 gem 'rails-dom-testing'
 #OAuth Based login
@@ -138,6 +137,5 @@ gem 'sanitize'
 
 #Performance testing
 gem 'rails-perftest'
-=======
->>>>>>> Moved gem perftest to test block in gemfile
 gem 'ruby-prof', group: :test
+gem 'ruby-prof', :github => 'ruby-prof'
--- a/Gemfile
+++ b/Gemfile
@@ -63,6 +63,9 @@ group :test, :development do
   gem 'coveralls', require: false
   gem 'ci_reporter_test_unit'
   gem 'openssl', '~> 2.0.0.beta.1'
+
+  # Performance testing
+  gem 'rails-perftest'
 end
 
 # run with `bundle install --without production` to exclude these
@@ -118,6 +121,7 @@ gem 'jbuilder'
 # Pin mustermann to Ruby 2.1 compatible
 gem 'mustermann' , '~> 0.4'
 
+<<<<<<< 3326f593e11ec51c6529431588ea4d7e705f1a4b
 #Gem for assertions used in testing
 gem 'rails-dom-testing'
 #OAuth Based login
@@ -134,4 +138,6 @@ gem 'sanitize'
 
 #Performance testing
 gem 'rails-perftest'
+=======
+>>>>>>> Moved gem perftest to test block in gemfile
 gem 'ruby-prof', group: :test
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,6 @@ end
 # run with `bundle install --without test` to exclude these
 group :test, :development do
   gem 'test-unit'
-  gem 'rails-perftest'
   gem 'minitest-reporters', '~> 1.2.0.beta2'
   gem 'rake',  '~> 10.5.0'
   # gems to test RESTful API
@@ -66,6 +65,7 @@ group :test, :development do
 
   # Performance testing
   gem 'rails-perftest'
+  gem 'ruby-prof','~> 0.15.9', group: :test
 end
 
 # run with `bundle install --without production` to exclude these
@@ -134,8 +134,3 @@ gem 'omniauth-github', '~> 1.1', '>= 1.1.2'
 gem 'activerecord-tableless'
 gem 'figaro'
 gem 'sanitize'
-
-#Performance testing
-gem 'rails-perftest'
-gem 'ruby-prof', group: :test
-gem 'ruby-prof', :github => 'ruby-prof'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,7 @@
 GIT
   remote: git://github.com/ruby-prof/ruby-prof.git
   revision: d6867cf1133bc6602058166a6c1d3301b9b0529c
+  branch: master
   specs:
     ruby-prof (0.16.2)
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,3 +1,9 @@
+GIT
+  remote: git://github.com/ruby-prof/ruby-prof.git
+  revision: d6867cf1133bc6602058166a6c1d3301b9b0529c
+  specs:
+    ruby-prof (0.16.2)
+
 GEM
   remote: https://rubygems.org/
   specs:
@@ -346,7 +352,6 @@ GEM
       ruby-progressbar (~> 1.7)
       unicode-display_width (~> 1.0, >= 1.0.1)
     ruby-openid (2.7.0)
-    ruby-prof (0.16.2)
     ruby-progressbar (1.9.0)
     sanitize (4.6.4)
       crass (~> 1.0.2)
@@ -483,11 +488,15 @@ DEPENDENCIES
   rspec
   rubocop (~> 0.52.1)
   ruby-openid
+<<<<<<< 59f260e3120e184c9376835dda5879118e6ba4d8
 <<<<<<< d5759ea4a5ffe7a59cf03091d29a1e22bccce10c
   sanitize
 =======
   ruby-prof
 >>>>>>> Added rails-perftest and ruby-prof gems, uncommented lines 9-11 in borwsing_test.rb
+=======
+  ruby-prof!
+>>>>>>> Updated gemfile to pull ruby prof from Github
   sass-rails (~> 4.0.0)
   scrypt (~> 3)
   sqlite3
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -197,13 +197,8 @@ GEM
     mimemagic (0.3.2)
     mini_mime (1.0.0)
     mini_portile2 (2.3.0)
-<<<<<<< 4c5fa537f3fa9c9e7ae4a42ff622065dbeb88711
-    minitest (5.11.3)
-    minitest-reporters (1.1.19)
-=======
     minitest (5.11.1)
     minitest-reporters (1.2.0.beta2)
->>>>>>> Update Gemfile.lock
       ansi
       builder
       minitest (>= 5.0)
@@ -252,6 +247,7 @@ GEM
     omniauth-twitter (1.4.0)
       omniauth-oauth (~> 1.1)
       rack
+    openssl (2.0.8)
     options (2.3.2)
     paperclip (5.2.1)
       activemodel (>= 4.2.0)
@@ -467,6 +463,7 @@ DEPENDENCIES
   omniauth-github (~> 1.1, >= 1.1.2)
   omniauth-google-oauth2
   omniauth-twitter
+  openssl (~> 2.0.0.beta.1)
   paperclip (~> 5.2.0)
   passenger
   php-serialize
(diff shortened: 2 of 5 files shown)


jywarren:
OK, I heard from another contributor (here: https://github.com/publiclab/plots2/issues/1829#issuecomment-350531276) that switching to Ruby 2.3.3 works. We should file an issue to make the app compatible with Ruby 2.4, but for now I think that should resolve your local issue? Are you using RVM or Rbenv, and can you try installing Ruby 2.3.3 for now?

jywarren:
Well cool ! I see your tests do run -- but we see an error:

```
BrowsingTest#test_browsers (29 ms warmup)
        process_time: 120 ms
/usr/local/bundle/gems/ruby-prof-0.16.2/lib/ruby-prof/printers/call_tree_printer.rb:46:in `print': wrong number of arguments (given 2, expected 0..1) (ArgumentError)
```

I think we need `puts` instead of `print`? Let me check the code.

Great work on this in any case!

Gemfile
```diff
@@ -106,3 +106,7 @@ gem 'jbuilder'
 
 # Pin mustermann to Ruby 2.1 compatible
 gem 'mustermann' , '~> 0.4'
+
+#Performance testing
+gem 'rails-perftest'
```
jywarren: Could we move this inside the "test" section of the gemfile? Thank you!

jywarren:
Aha! Same issue here: https://github.com/ruby-prof/ruby-prof/issues/203

jywarren:
Hmm, potentially we need to create a directory for the results? 

> It crashes because call_tree_printer needs a directory. Better error message will given when using latest master. (https://github.com/ruby-prof/ruby-prof/issues/215#issuecomment-306163269)

Gemfile
bellvat: Yes, will move it!

jywarren:
@ryzokuken -- any ideas why we're getting this error? We'd love some help if you have any insights. Thanks!

Gemfile
bellvat: To clarify, by 'test' section, are you referring to inserting it below 'group :test, :development do', but outside of that that block in the gem file?

jywarren: It can go inside that `group :test, :development do` block, thanks!

jywarren:
I left a comment at the `rails-prof` project, but it's odd that I don't see this when running it locally... anyhow, hope we can get some input from them.

ryzokuken:
@jywarren sure, I will take a look.

ryzokuken:
> I tried to debug this by changing my ruby version, as I read in some posts that it might be that Rails 4.1.6 may not be compatible with Ruby 2.4. I am not as familiar with how rbenv works, and although I was able to install a different version of ruby, I was not able to get rails to work with that version.

I believe this error appears because @bellvat seems to be using ruby 2.4. This is a known issue with Rails 4.1 and Ruby 2.4 (https://stackoverflow.com/questions/41504106/ruby-2-4-and-rails-4-stack-level-too-deep-systemstackerror). Infact, Rails 4.1 does not even support Ruby 2.4.

@bellvat please try using Ruby 2.3 (which is what we all use). That'll solve this issue for now. As a long term solution, we'd be migrating to Rails 4.2 soon (Rails 4.2.8 fully supports Ruby 2.4).

test/performance/browsing_test.rb
```diff
@@ -1,12 +1,12 @@
...
+require 'test_helper'
+require 'rails/performance_test_help'
 
-#class BrowsingTest < ActionDispatch::PerformanceTest
-  # Refer to the documentation for all available options
-  # self.profile_options = { :runs => 5, :metrics => [:wall_time, :memory]
-  #                          :output => 'tmp/performance', :formats => [:flat] }
+class BrowsingTest < ActionDispatch::PerformanceTest
+   #Refer to the documentation for all available options
+   #self.profile_options = { :runs => 5, :metrics => [:wall_time, :memory]
```
ryzokuken: Should these still be commented out?

ryzokuken:
@jywarren As of the error that says

```
BrowsingTest#test_browsers (29 ms warmup)
        process_time: 120 ms
/usr/local/bundle/gems/ruby-prof-0.16.2/lib/ruby-prof/printers/call_tree_printer.rb:46:in `print': wrong number of arguments (given 2, expected 0..1) (ArgumentError)
```

it has been fixed in https://github.com/ruby-prof/ruby-prof/commit/837edf447914663d0100e5599ab8c44f81f0eae0. The issue is, the latest version of this gem is 0.16.2, which was released on August 25, 2016. On the other hand, This particular commit was merged on October 19th, 2016. To get over the issue, we should use the gem from GitHub rather than fetching it from RubyGems. http://bundler.io/v1.16/guides/git.html seems to be a good resource regarding this. Basically, we need to change the line to:

```ruby
gem 'ruby-prof', :github => 'ruby-prof'
```

And hopefully that will solve all our problems.

bellvat:
Hi @ryzokuken 

Thank you for taking a look at the issues and advising on possible solutions.

I updated this PR for ruby-prof line in the gemfile to include `:github => 'ruby-prof'` according to the comment above.
Unfortunately, it looks like the error remains.

Not sure what could be the cause..

test/performance/browsing_test.rb
bellvat: I believe tests run with a default setups. If we would like to configure the test runs, we can uncomment the section and specify our own values.

ryzokuken:
A few things: Firstly, the gem still has to be grouped under the `test` group as it had been previously. I didn't mention that in my comment and I'm sorry for the trouble.

Also, after making the changes to the Gemfile, you'd have to run `bundle update` to actually bring those changes in effect. Again, I'm not sure why I missed that as well.

Please run `bundle update` and then try it and let me know if it works. Thanks for your patience.

jywarren:
Thanks both of you! After bundle update, you'll need to re add Gemfile.lock
-- fingers crossed that it works!

Gemfile.lock
```diff
@@ -315,11 +315,11 @@ GEM
     thread_safe (0.3.6)
     tilt (1.4.1)
     timecop (0.9.1)
-    tins (1.15.1)
+    tins (1.16.3)
     tool (0.2.3)
     tzinfo (1.2.4)
       thread_safe (~> 0.1)
-    uglifier (3.2.0)
+    uglifier (4.0.2)
```
ryzokuken: @jywarren any idea why it was bumped up by a major version? Wouldn't `4.0.2` be inconsistent with `3.2.0`?

jywarren:
Was version specified in the Gemfile?

ryzokuken:
@bellvat could you try adding `:branch => 'master'` at the end as well? I hope that'll help.

ryzokuken:
@jywarren not even close to 4

https://github.com/publiclab/plots2/blob/d5da035c98f64d5ad0f51ea463e6a27c802dda67/Gemfile#L32

bellvat:
@ryzokuken Sure. 
Just to confirm, I would add `:branch => 'master'` to the end of the ruby-prof line in the Gemfile, then run `bundle update`, correct?

bellvat:
@jywarren and @ryzokuken 
I am going to try to install an earlier version of ruby-prof to see if it will help. If not, we can revert to an earlier commit.

ryzokuken:
No, I don't think that'll help. Which ruby version are you using BTW?

bellvat:
@ryzokuken  
Got it. Will revert back. Ruby version 2.3.4.

ryzokuken:
Yeah, it might help. If it doesn't also try mentioning the `master` branch in your Gemfile and run `bundle update` everytime you make a change in the Gemfile.

jywarren:
For uglifier, `>=` is anything equal or higher, while `~>` is anything within a `0.0.x` bugfix/patch level change. So if you don't want it to go up to 4, we'd want to change it to `~>3.0.0` (i'm not sure if there's a way to specify a `0.x` level change but not a major version # change)

Do you think `uglifier` is the problem? If so, shall we limit it to v3? I still see the error:

`/usr/local/bundle/bundler/gems/ruby-prof-d6867cf1133b/lib/ruby-prof/printers/call_tree_printer.rb:46:in `print': wrong number of arguments (given 2, expected 0..1) (ArgumentError)`

And we seem to be on `ruby-prof` version `0.16.2`

jywarren:
I'm sorry this one has turned out to be so much trouble, thanks so much for your persistence! 

I think there's a typo in this last test, let me see.

jywarren:
It's a weird error;

`ERROR: Service 'web' failed to build: failed to process "\"This": unexpected end of statement while looking for matching double-quote`

I restarted the test to see if it was a fluke. Fingers X!

ryzokuken:
@jywarren Did we specify the master branch yet? `v0.16.2` doesn't work for us.

jywarren:
I think so - i see branch => master

sorry, what version should we specify?

bellvat:
Hi @jywarren  and @ryzokuken,

I made the following changes in the latest commit:
1. I reverted the branch, so currently, versions for existing gems match masters.
2. I am using Ruby version 2.3.4
3. I found that when running `rake test:all` on the Master branch, I see that 577 tests pass, but 7/7 solr tests are erroneous- please see the following: 
screen shot 2017-12-12 at 9 20 12 am
screen shot 2017-12-12 at 9 20 31 am
screen shot 2017-12-12 at 9 20 47 am
screen shot 2017-12-12 at 9 21 16 am

4. When I ran `rake test:all` with ruby-prof version 0.16.2 on the Performance-test branch, I got the following message: 
screen shot 2017-12-12 at 9 48 07 am

5. When I switched to ruby-prof version 0.15.9 on the Performance-test branch, `rake test:benchmark` passes, as follows:
screen shot 2017-12-12 at 9 13 13 am

6. Then when I run `rake test:all` on the Performance-test branch with ruby-prof(0.15.9), I see that 577 tests pass, and get the same 7/7 solr test errors like in the Master branch. Please see below:
screen shot 2017-12-12 at 9 16 14 am
screen shot 2017-12-12 at 9 07 57 am
screen shot 2017-12-12 at 9 08 22 am
screen shot 2017-12-12 at 9 09 05 am

It might be something about my environment that is causing this?
Would it be worthwhile for one of you to try adding the ruby-prof gem and uncommenting the browser_test.rb on your local to give it a try?

jywarren:
Hmm, I wonder if the new error `: Service 'web' failed to build: failed to process "\"This": unexpected end of statement while looking for matching double-quote` is now blocking, and the original version bug is actually resolved now. I'm wondering where we could possibly have missed a double quote...

jywarren:
Literally the only double quote in your pull requests is `test "browsers" do` -- it sounds wild, but maybe changing that to single quotes could have some effect?

jywarren:
OK I think @ryzokuken solved the unrelated issue -- and i'm rerunning the tests now to see if they pass. I think they should now run on top of the change in #1881 -- thanks everyone for your help and patience!

jywarren:
Hmm, maybe we have to push some new commit here. I still see the old Dockerfile:

https://github.com/bellvat/plots2/blob/c6b28d54097ca6dc6d3aae88af27fc2a99b903fa/Dockerfile#L7

Let me try pushing something insignificant.

jywarren:
I went ahead and rebased all these changes on top of the latest, just to be sure.

ryzokuken:
@jywarren It works! All we now need to do is find a way to make the reporter parse this test. I'm taking a look at https://github.com/ruby-prof/ruby-prof#printers

ryzokuken:
@jywarren Can we just exclude it from report generation? AFAIK, Rails Prof can generate it's own reports in multiple formats.

jywarren:
For now, sure. I think it'd be great to have Dangerbot report the effects on performance of different PRs -- but we can make a follow-up issue for that!

jywarren:
How might it be disabled?

jywarren:
Hi, @ryzokuken - did you have an idea of how to disable report generation
for performance tests? Your help is MUCH appreciated, thank you!

jywarren:
Happy new year, all! @ryzokuken did you have an idea of how we could disable report parsing, and just output the rails-perf native report? The error is currently when junit tries to parse the report here:

`/usr/local/bundle/gems/minitest-reporters-1.1.19/lib/minitest/reporters/junit_reporter.rb:114:in `+': nil can't be coerced into Fixnum (TypeError)`

Or we could try running this locally and see what format the report is in, and see why there's a `nil` in there...

jywarren:
Interestingly it's failing on the "time" line in this function of the `junit_reporter.rb` file:

```ruby

      def analyze_suite(tests)
        result = Hash.new(0)
        tests.each do |test|
          result[:"#{result(test)}_count"] += 1
          result[:assertion_count] += test.assertions
          result[:test_count] += 1
          result[:time] += test.time
        end
        result
      end
```

jywarren:
Well, this is good!  I get the same error when running this locally.

jywarren:
Some clues in the `ruby-prof` docs say we need to add a Rack adapter to `config.ru` --

```
if Rails.env.profile?
  use Rack::RubyProf, :path => '/temp/profile'
end
```

ugggg but it'd be `Rails.env.test?`

jywarren:
Ooh, a lot of clues here: http://ccaloha.cc/blog/2015/07/14/howto-performance-test-in-rails-4/

and here: http://railscasts.com/episodes/411-performance-testing

jywarren:
I'm a bit stuck, to be honest. Performance testing docs in Rails 3.2: http://guides.rubyonrails.org/v3.2/testing.html#rake-tasks-for-running-your-tests are missing from Rails 4.1: http://guides.rubyonrails.org/v4.1/testing.html

I guess we need to get junit to ignore rails-perftest. I'm going to use some of the options that were commented out in the test...

jywarren:
No luck! The error is... odd. Especially since I didn't get it when we first started! I think it may have to do with us now being on Rails 4.1, but not sure.

jywarren:
Trying to echo out the reports with `cat test/reports/*` -- see if we can see inside a bit!

jywarren:
That was instructive -- i believe it's not generating a report at all! @ryzokuken what do you think?

ryzokuken:
@jywarren let me check.

jywarren:
@publiclab/reviewers -- this is a real tough one. If anyone can give it a try to see if we can move this forward, much appreciated! I'm totally stuck!

ryzokuken:
@jywarren me too, tried a couple of times.

jywarren:
OK, i'm going to try this again! Fingers X -- pushing a newly rebased version up!

jywarren:
OK, a clarification -- this seems to be happening when /generating/ the junit reports, but before reading and parsing them in the Dangerfile for report-back in the PR.

jywarren:
I disabled the offending line of minitest-reporters -- 113 -- and got it to run. THe result:

```
(xenial)warren@localhost:~/sites/plots2$ less test/reports/
BrowsingTest#test_browsers_memory.csv          BrowsingTest#test_browsers_wall_time.csv       TEST-BrowsingTest.xml
BrowsingTest#test_browsers_memory_flat.txt     BrowsingTest#test_browsers_wall_time_flat.txt  
(xenial)warren@localhost:~/sites/plots2$ less test/reports/TEST-BrowsingTest.xml 
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="BrowsingTest" skipped="0" failures="0" errors="0" tests="1" assertions="0" time="0">
  <testcase name="test_browsers" classname="BrowsingTest" assertions="0" time="">
  </testcase>
</testsuite>
```

jywarren:
OK, I think we're going to have to give up on this for now, unless the folks over at https://github.com/kern/minitest-reporters/issues/251 are able to help us out at all; 

For anyone trying to solve this, the tests are running, but the XML JUnit reports produced by `minitest-reporters` are failing to get generated due to line 114 of this file:

https://github.com/kern/minitest-reporters/blob/9d3f5777c1a45e576a0c6844fa1f612831768595/lib/minitest/reporters/junit_reporter.rb#L114

@bellvat I'm so sorry we weren't able to make it on this! I hope it wasn't a terrible experience for your first contribution and I'd love to help you choose another over at https://publiclab.github.io/community-toolbox/#r=all if you're still interested. Many thanks for working with us on this one that turned out to be unexpectedly awful!

bellvat:
@jywarren No problem! I wish I could have helped out more with the reporting issues. Thanks for really trying to work this out!

ryzokuken:
@jywarren keep this open, I'll keep trying to fix this somehow as my understanding of our test reporting improves.

jywarren:
Thanks, all!

jywarren:
Whoa, they just adopted our solution in `minitest-reporters` -- look! https://github.com/kern/minitest-reporters/issues/251

`minitest-reporters v1.2.0.beta2` might now work for us!

jywarren:
I just tried updating minitest-reporters to `v1.2.0.beta2` to see if that passes.

jywarren:
Ooh! A new error!

`/usr/local/bundle/gems/minitest-reporters-1.2.0.beta2/lib/minitest/reporters/junit_reporter.rb:56:in `method': undefined method `test_newcomer_should_get_legacy_form' for class `Minitest::Result' (NameError)`

jywarren:
That's odd. It's for this test, of all things: 

https://github.com/publiclab/plots2/blob/5ffd7c6b3f9d2a3a53946a3433970ce1590ceb15/test/functional/editor_controller_test.rb#L71

I can't see why that test buried in the middle of our tests should have anything to do with performance tests. I'll try running this again...?

jywarren:
Unfortunately, [we've learned](https://github.com/publiclab/plots2/pull/2726#issuecomment-394833053) via @siaw23 that Rails performance testing was mostly for Rails 3 and has been supported less and less. I think he's right that we should close this and pursue something with #2660 -- Skylake -  Thanks @siaw23 !!!

And thanks @bellvat for your help with this and I'm sorry we never got this attempt running, but it looks like it was none of our faults!

PR #1872, opened 2017-12-11
Added new page for tagged content by a single author with '/tag/:tagname/author/:authorname'

### LINKED ISSUE #1854: Add a new page for tagged content by a single author
Let's add a new page, similar to the "tag" page like https://publiclab.org/tag/hydrogen-sulfide, but for just one author's work, so:

https://publiclab.org/tag/hydrogen-sulfide/author/warren ?

This'd require:

1. a `config/routes.rb` entry
2. a re-use of the `/app/views/tag/show.html.erb` template with a custom title
3. a new action in `/app/controllers/tag_controller.rb` for, maybe `def show_for_author` ? which uses the new `Tag.tagged_nodes_by_author()` method here
4. a functional test in `/test/functionals/tag_controller_test.rb` to show that the new `show_for_author` action works

This would be a bigger project, but would be pretty awesome to see. 

I'd be happy to help anyone through it -- @SidharthBansal ?

SidharthBansal:
fixes #1854

This'll allow:

http://publiclab.org/tag/replication:*/author/warren would show all replications by author `warren`!

Progress: 
- [x] a config/routes.rb entry
- [x] a re-use of the /app/views/tag/show.html.erb template with a custom title
- [x] a new action in /app/controllers/tag_controller.rb for, maybe def show_for_author ? which uses the new Tag.tagged_nodes_by_author() method here
- [x] a functional test in /test/functionals/tag_controller_test.rb to show that the new show_for_author action works


jywarren:
This old code tried to do this! 

https://github.com/publiclab/plots2/blob/8069e3024aa8e2518fce3cca83a2424b9df9ae2f/app/controllers/notes_controller.rb#L264-L272

Let's update this or replace it with your new code. Great!

SidharthBansal:
@jywarren  I have made a commit can you please view it. At my localhost:3000 it is working properly.
All previous tests are passing. 
Please tell me the required changes if any.
Thanks

SidharthBansal:
@jwarren I am little confused about this commit. What are we trying to build?
I guess we want to have a page like tag/show for the tags with authors, as you wrote in the comment. In that page, we have buttons for the research notes, questions, wiki pages, maps, and contributors. So, what will they link to in this navbar? I checked the routes file but I could not find any route linking to questions tagged with the tag name and author name and same for other fields. Are we suppose to create them later on? Also, I tried to understand what wiki leads are but can't understand. Can you please help me in understanding them..? Curious to know about them.
I have created the notes view which is supposed to be wrong if we need to use the tag/show. Otherwise, it is working fine with the notes view.
Also, I found out that the contributor's page on the navbar links to another page which is completely different from the tag/show view. So, does it needs to be changed?
One more thing, maybe not important. But I felt a little odd to have space at the end of the search of publiclab.org/tag/balloon*. I think there should be either 28 search results or 32 search results or say 20. So, that there is no extra space. I know by default will_paginate has 30 records. We can change it. Also this is happening not only on this but throughout the publiclab.org.

app/controllers/tag_controller.rb
```diff
@@ -110,6 +109,19 @@ def show
     end
   end
 
+
+  def show_for_author
+    if params[:id][-1..-1] == '*'
+      @wildcard = true
+    end
+    @tagname = params[:id]
+    @user = DrupalUser.find_by(name: params[:author])
```
jywarren: Can we just search `User.find_by` here, not `DrupalUser` which we're trying to reduce usage of (and eliminate)? Thanks!

app/models/tag.rb
```diff
@@ -239,7 +239,6 @@ def self.trending(limit = 5 , start_date = DateTime.now - 1.month , end_date = D
   #select nodes by tagname and user_id
   def self.tagged_nodes_by_author(tagname, user_id)
     if tagname[-1..-1] == '*'
-      @wildcard = true
```
jywarren: Ah, we use this to disable some things that don't currently work with wildcard searches... so maybe preserve this unless we're sure that the whole page works identically for normal tags and wildcard queries?

app/views/tag/show_for_author.html.erb
```diff
@@ -0,0 +1,43 @@
+<% if params[:controller] == "tag" || params[:controller] == "search" %>
```
jywarren: Was there not an efficient way to use the same `show.html.erb` template, and adapt it using one of these `if/else` statements for displaying posts by a single author?

jywarren: Perhaps anything that doesn't currently work with just one author, like the other tabs, could be disabled by checking the `params[:action]` to see if this is an instance of an author-specific tag show page and not just a normal one?

test/functional/tag_controller_test.rb
```diff
@@ -185,6 +185,19 @@ def setup
     assert_select '#wiki-content', 1
   end
 
+  test 'show note with author and tagname without wildcard' do
```
jywarren: These look great. Awesome job on the tests.

jywarren:
Sorry, just getting to this after the backlog over the new year. I made some requests above. 

The bigger goal here is to make it possible to list completed activities on [peoples' profiles](https://publiclab.org/profile/warren). 

When you complete an activity and post a response, it's tagged `replication:______` and so my hope was to develop a query we can use to display any post that is:

1. by a given author
2. tagged with `replication:______` -- a wildcard search

Then, we can:

1. display the count of such posts on peoples' profiles and
2. have a consistent page and route to view a list of such posts (this route could have other uses too!)

Does this make sense?

> understand what wiki leads are

Do you mean wiki likes? Sorry, what section of code?

Regarding pagination of 30, i agree! We can change it to 24 if you want to open a new PR.

SidharthBansal:
This 
image
on https://publiclab.org/tag/balloon

app/views/tag/show_for_author.html.erb
SidharthBansal: I will try to change it

jywarren:
The idea of the tag page wiki button is that if there's a wiki page whose URL matches the tagname, like /wiki/gsoc for tag `gsoc`, then we show the first paragraph and the lead image of that page on the tag page, as a kind of "intro". If it doesn't exist, then we prompt people to create such a wiki page to provide a nice intro for the page. 

> but it was then redirecting the login page eventhough I was logged in.

Are you sure? i just tried this for /wiki/abcdef and it worked for me -- presented me a blank wiki page form. 

It looks like your `_header.html.erb` is out of sync due to other changes we've since merged. You can probably rebase all your changes over the latest master -- see https://publiclab.org/wiki/developers for some guidance! -- or you can use the GitHub interface button to `Resolve conflicts`. I'm happy to help if you get in trouble! Thanks!

SidharthBansal:
Ok, got it. I checked it again after logging in and out. It worked. It is not displaying the login page. Thanks...
I will change the view soon
Regarding conflicts --yeah I will resolve all of them. I know how to resolve them.

SidharthBansal:
image
image
image
image
I think it is working fine. As per your suggestion I have disabled the buttons which are not needed in this with a tooltip.
I have used the google translator for the locale files. I don't know Dutch. Hope it works.

jywarren:
I would leave blank the translations you don't know for sure. Also, were
you able to consolidate the template with the existing tag show template
then? Cool!

jywarren:
Indeed you did! This looks great; is it ready? Want to throw in a functional test to confirm this at least loads successfully in tests? This will protect your code from future issues too. 

Awesome!

config/locales/views/tag/show/de.yml
```diff
@@ -18,3 +18,4 @@ de:
       no_results_found: "Keine Ergebnisse gefunden; versuchen, für '<b>%{tag}</b>' Suche"
       try_advanced_search: "Oder versuchen Sie eine <a href='%{url1}'>erweiterte Suche</a>"
       contributors: "Mitwirkende"
+      tagged_by: "Notities getagd met \"%{tags}\" door <a href='%{url1}'>%{user_name}</a>"
```
jywarren: Let's remove this, and let it default to English, ok? Thanks!

jywarren:
Just one more tweak to leave the German translation blank, and we'll be good to merge!

jywarren:
Thanks, this is exciting!

jywarren:
Amazing. This is a great implementation. 🎉 🎈

jywarren:
I hope we'll deploy it today and then we can check http://publiclab.org/tag/replication:*/author/warren -- and then make a link to it from the profile page. Thank you!!!

SidharthBansal:
@jywarren Just thinking, it will be better not to display "has no lead wiki " for power tags. Because even though Gsoc wiki page exists, searching /tag/Gsoc* gives us that it leads to no wiki page. I think, if a person is searching for power tag, then he/ she should get a message that "Enter a tag instead of power tag to get related wiki pages" or we can remove the lead wiki section when there is a power tag type search.
One ambiguity with power tag search is a user may assume that if no g* page exists this implies gsoc cant be a wiki page. 
Another way can be for power tags it will display all wiki pages related to that tag. But this would be ugly.

jywarren:
I think i agree -- we could could truncate the * and just lead to `/wiki/replication` for `replication*` -- and then be sure we make explanatory pages for those sorts of things. If you'd like to open a new issues for that, go for it! 

Same with the link from profile page. I think we could show a count and link with a fourth column where it displays questions, answers, and comments. What do you think?

SidharthBansal:
I think truncating is not a better option. Reason --  if a lazy person who knows a little code writes ``replica`` instead of ``replication`` because he knows replication is derived from the ``replica`` this will again end up in a chaos.
One more thing, power tags eg ``*`` itself by the code(regex) means all tags. I think, we should optimise the condition of power tags too, so the user will enter first part fully.

Fastie:
This is really great. There have been many times when I have wanted to use something like this. I added some links to my profile with this type of search. Thanks much.   
https://publiclab.org/profile/cfastie

Fastie:
The link to the author's profile at the author's tag result page is broken, e.g., "Notes on ndvi by **cfastie**" on this page:  https://publiclab.org/tag/ndvi/author/cfastie

SidharthBansal:
@fastie thanks, I will change it

PR #1888, opened 2017-12-13
displayed redirected wiki page

### LINKED ISSUE #1773: Display redirected wiki page at top of /tag/____ pages if redirect exists
Hi, this is a [first-timers-only issue](https://github.com/publiclab/plots2/projects/2). This means we've worked to make it more legible to folks who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**. 

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general [help wanted](https://github.com/publiclab/plots2/labels/help-wanted) issues. Thanks!

### Problem

Tag pages, like https://publiclab.org/tag/h2s-copper-pipe where the tag doesn't have a corresponding wiki page prompt you to make a new page. 

But some pages just have a non-matching wiki page -- this one has for example, the matching:

https://publiclab.org/wiki/hydrogen-sulfide-copper-pipe

### Solution

So we should detect if there's a redirect tag on https://publiclab.org/wiki/h2s-copper-pipe leading to another page, and display /that/ page's lead image and first paragraph, instead of the original (which is empty). 

We can do this in the controller here:

https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L45

Basically after this line, we can see if @wiki exists, and if not, check if it has a redirect, so:

```ruby
@wiki = Node.find(@wiki.power_tag('redirect'))  if @wiki.has_powertag('redirect') # use a redirected wiki page if it exists
```

That'll check if there's a redirect tag, then if there is one, it'll assign the Node with the correct new node id to `@wiki`. Make sense? I'm happy to help here!

## Steps to Fix

- [ ] claim this issue with a comment here, below, and ask any clarifying questions you need
- [ ] set up a repository locally following the README instructions, and make sure that all tests pass
- [ ] try to fix the issue following the steps above, but even before you're done, you can:
- [ ] commit your changes and start a pull request (see [contributing to Public Lab software](https://publiclab.org/wiki/contributing-to-public-lab-software)) but mark it as "in progress" if you have questions or if you haven't finished
- [ ] alert someone via the developers list (see below) to have your pull request merged. This may take a few extra steps depending on when you do it, but we'll help you out!

Please email the developers list (see https://publiclab.org/wiki/developers) or go to [the chatroom](http://publiclab.org/chat) if you have questions, and take a look at [our first-timers landing page for more information](https://github.com/publiclab/plots2/projects/2)!

namangupta01:
Fixes #1773


jywarren:
I suspect the errors here are because this came before our big Rails 4.1 upgrade, perhaps? In any case i'm seeing if we can just rerun it on the latest version. Let's see how this goes!

jywarren:
Hmm, i'm going to try manually rebasing this... let's see...

jywarren:
OK, let's see how this does. Fingers X!

app/controllers/tag_controller.rb
```diff
@@ -43,6 +43,8 @@ def index
   def show
     # try for a matching /wiki/_TAGNAME_ or /_TAGNAME_
     @wiki = Node.where(path: "/wiki/#{params[:id]}").try(:first) || Node.where(path: "/#{params[:id]}").try(:first)
+    @wiki = Node.find(@wiki.power_tag('redirect'))  if @wiki.has_powertag('redirect') # use a redirected wiki page if it exists
```
jywarren: OK, so maybe occasionally there is no @wiki variable and we can't run `.has_powertag` on `nil` -- so can we add:

`if @wiki && @wiki.has_powertag('redirect')` to catch this?

jywarren: oof, that one almost stumped me!

jywarren:
Oof, the correct line is `has_power_tag`, not `has_powertag` -- my bad! This should fix the issue (though TravisCI is stalling currently so we won't see until that gets fixed!)

PR #1894, opened 2017-12-15
Redirects to login if not logged in(Fixes#1893)

SidharthBansal:


app/views/like/_like.html.erb
```diff
@@ -9,9 +9,19 @@
         <% end %>
     <li class="btn btn-default btn-sm" data-html="true" rel="popover" data-content="<%= "No tags" if tagnames.nil? || tagnames.length == 0 %><% if tagnames %><% tagnames.each do |tagname| %><p style='margin-bottom:3px;'><a href='/subscribe/tag/<%= tagname %>' class='btn btn-default btn-mini'><%= tagname %></a></p><% end %><% end %><hr /><a class='btn btn-sm'><i class='fa fa-user'></i> <%= node.author.name %></a>" title="Follow by tag or author" data-placement="left"><i class="fa fa-eye"></i><span class="hidden-xs hidden-sm"> Follow</span></li>
 
-    <li rel="tooltip" title="Helpful? Like it and get updates!" class="btn btn-default btn-sm btn-like" node-id="<%= node.id %>" id="like-button-<%= node.id %>"><span id="like-star-<%= node.id %>" class="fa fa-star<% if current_user && !node.liked_by(current_user.uid) %>-o<% end %>"></span> <span id="like-count-<%= node.id %>"><%= node.likers.length %> </span></li>
+  <li rel="tooltip" title="Helpful? Like it and get updates!" class="btn btn-default btn-sm btn-like" node-id="<%= node.id %>" id="like-button-<%= node.id %>">
+    <% if !current_user %>
+      <a href = '/login' >
```
jywarren: Can we set a `return_to` value here? It'll get picked up by the controller after being relayed here: https://github.com/publiclab/plots2/blob/master/app/views/user_sessions/new.html.erb#L24

jywarren: Also, happy new year!

SidharthBansal: Happy New year!!!

SidharthBansal: Yeah that would be better for user experience to redirect back to the notes after sign up. Will change it

SidharthBansal:
@jywarren  now it is redirecting back to the post. Please review it.
Thanks

jywarren:
Awesome, thanks!!!

PR #1895, opened 2017-12-15
Typo in TESTING.md

SidharthBansal:
There is a typing error in TESTING.md
"Stack Overflow sites" should be there instead of "Stack Overflow cites"


ryzokuken:
I don't think it's a typo 😅
screen_shot_2017-12-16_at_2_21_05_am

SidharthBansal:
OK thanks

PR #1910, opened 2018-01-03
added a time bound version of contributor_count

### LINKED ISSUE #1891: make a time-bounded version of tag.contributorcount ?
Hi everyone!

In order to show how many unique contributors have been active in a certain period of time, perhaps we could make add a time-bounded version of the tag.contributorcount function

Contributor counts are already looking great on tag pages! Check out how the number count is shown right on the tag page https://publiclab.org/tag/spectrometer and it's clickable leading to a page that shows individuals: https://publiclab.org/contributors/spectrometer 

### Next steps after this

Adding contributor count by time period to publiclab.org/stats (in discussion over at ) https://github.com/publiclab/plots2/pull/1672

ViditChitkara:
closes #1891

Thanks!


ViditChitkara:
@jywarren Could you please have a look at this? I have added a screenshot in the description!
The contributor count is now shown on the stats page.
Thanks and Happy new Year!!

jywarren:
Hello - this looks good in that it's separated out, but I'd prefer if we could keep this code as closely parallel to the `contributor_count` in the Tag model that is /not/ time-bounded, here:

https://github.com/publiclab/plots2/blob/c20f81e35ec73c284d1b893fcd71eeba60e547dd/app/models/tag.rb#L55

Do you think we could move this to the User model, perhaps? So, `User.contributor_count_for(start..end)`? That way we can add a unit test in https://github.com/publiclab/plots2/blob/master/test/unit/user_test.rb

And let's put a description in a comment just before the method -- so:

```ruby
# all uses who've posted a node, comment, or answer in the given period
def self.contributor_count_for(start, end)

end
```

Also, I'd like it if we could consolidate all the code for the method within the method, so maybe `uniq_users` could be unpacked -- i like the modularity, but let's try to keep things really readable within the method itself. 

Thanks so much, this is looking really good! just a few more changes and we'll be able to merge it in. Your help is much appreciated!

jywarren:
And happy new year!!

ViditChitkara:
Yes, perhaps it is better to move the time bound method to User model. I was reluctant to keep it in Tag model since this is more of a user's property rather than tag's. Thanks for your help I will do the required changes.

ViditChitkara:
@jywarren I have made the changes, please see if it is good to go. I'll write the unit tests in the meanwhile.

app/models/user.rb
```diff
@@ -321,4 +321,14 @@ def map_openid_registration(registration)
   def self.find_by_username_case_insensitive(username)
     User.where('lower(username) = ?', username.downcase).first
   end
+
+  def self.contributor_count_for(start_time,end_time)
+    notes = Node.where(type: 'note', status: 1, created: start_time.to_i..end_time.to_i).pluck(:uid)
+    answers = Answer.where(created_at: start_time..end_time).pluck(:uid)
+    questions = Node.questions.where(status: 1, created: start_time.to_i..end_time.to_i).pluck(:uid)
+    comments = Comment.where(timestamp: start_time.to_i..end_time.to_i).pluck(:uid)
```
jywarren: This looks great but I think we're missing one more type -- revisions -- could we add: `Revision.where...` the same way? Thanks!

ViditChitkara: Sure! Will add that

ViditChitkara: Jeff, I have made the changes.

test/unit/user_test.rb
```diff
@@ -110,7 +110,8 @@ class UserTest < ActiveSupport::TestCase
 
   test 'contributor_count' do
     contributor_count = User.contributor_count_for(Time.now-5.years, Time.now)
-    assert_equal contributor_count,4
+    is_positive = contributor_count>=0
+    assert is_positive
```
jywarren: You could try `assert_equal contributor_count, 0` and then read the test output to find what the correct value ought to be, then modify the test to match that.

jywarren: Although we may want to be careful -- any time someone adds more test data, this value could change, no? That could make adding new test data a bit of a pain.

ViditChitkara: Yep, I guess that's why I got a failed test when I kept it equal to 4. It worked perfectly on my pc which contains a bit of manipulated data (got 4 using the method you told) but fails on travis. I guess this should me kept more generic and not equal to a numerical value. What do you think??

jywarren: Sure, we could just leave it at not 0 for now, i guess. Hmm. Doing more risks being circular, you know?

jywarren: We could test that it increments by 1 though, when adding a new comment... what do you think?

jywarren: Like here: https://github.com/publiclab/plots2/blob/master/test/unit/comment_mailer_test.rb#L7

ViditChitkara: That's a good idea!! I'll have a look at it!

ViditChitkara:
@jywarren It's working pretty well now!! The comment creation was a good idea!! Please have a look.

jywarren:
OK, then! This looks great. Can you add the suggested explanatory comment I mentioned above, and remove the change to Gemfile.lock (as it seems unrelated) and we can move forward? Thanks!!

ViditChitkara:
Strange, why changes related to Gemfile.lock are tracked by git. Though this file is mentioned in gitignore!

jywarren:
That is odd, i agree. This looks perfect, thank you! Merging now.

PR #1935, opened 2018-01-08
Making comment answer

### LINKED ISSUE #1897: Move comments to answers in questions

Sometimes on the Q&A feature (www.publiclab.org/questions), people post answers to questions in the comments section. I think it would be good if there was an option to move a comment to an answer, (and maybe vice versa? but I don't think that's as important). I think it will help us curate the information and hopefully make the answers more visible.

Gauravano:
Fixes #1897 
I have implemented the whole functionality and have made some changes in views accordingly.
Demo:-
make_answer

### DIFF
--- a/app/views/questions/_comment.html.erb
+++ b/app/views/questions/_comment.html.erb
@@ -18,6 +18,11 @@
     <span class="content">
       <%= raw sanitize(auto_link(insert_extras(RDiscount.new(comment.body).to_html))) %>
     </span>
+    <div class="content-form" >
+      <form class="form-horizontal" action="/comment/make_answer/<%= comment.id %>" method="post" data-remote="true">
+        <button type="submit">move</button>
+      </form>
+    </div>
     <div class="content-form" style="display: none;">
       <form id="comment-<%= comment.id %>-edit-form" class="form-horizontal" action="/comment/update/<%= comment.id %>">
         <%= render partial: "questions/form", locals: { comment: comment } %>
--- a/app/views/questions/_comment.html.erb
+++ b/app/views/questions/_comment.html.erb
@@ -19,7 +19,8 @@
       <%= raw sanitize(auto_link(insert_extras(RDiscount.new(comment.body).to_html))) %>
     </span>
     <div class="content-form" >
-      <form class="form-horizontal" action="/comment/make_answer/<%= comment.id %>" method="post" data-remote="true">
+      <% byebug%>
+      <form class="form-horizontal" action="/comment/make_answer/<%= comment.id %>">
         <button type="submit">move</button>
       </form>
     </div>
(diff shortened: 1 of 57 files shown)


jywarren:
Wow! I was wondering if we could make a bootstrap tooltip that shows what the button does, and maybe we could use the plus sign icon? This is a big awesome new feature! Thanks!

Gauravano:
Actually jeff I was just about to ask you about the icon, I will soon send some good icon images to choose for this feature or we can just settle with + icon.

jywarren:
Cool, tell me what you come up with! Plus sign is a good one but i'm interested if you find a better one. Thanks!

Gauravano:
I have checked many but I have to agree that '+' icon with tooltip is the best option.so coming to tooltip I think according to button placement on extreme left tooltip should be generated towards left and message can be 'Make your comment answer' or 'Post your comment as answer'

jywarren:
Okay, that's great. How about "promote this comment to an answer"? Awesome!!

jywarren:
Hi, is this ready to merge? Awesome work!!

Gauravano:
just translation of tooltip title is left to be added to .yml file

Gauravano:
@jywarren plz add de.yml or tell the translation of tooltip title as I don't want to make any translation mistakes.With this work is complete on this.

jywarren:
Ah, it's OK, the translation will fall back to English if there's nothing else present. I think this is good to go!

jywarren:
Hmm, I was thinking, this is a delicate one where we delete content (after saving it of course!) -- do you think it's worth writing a functional test for it in `/tests/functional/comment_controller_test.rb`?

Gauravano:
sure ...writing test will be definitely good for it.Ok so I will add a functional test too

jywarren:
Awesome. Fantastic.

jywarren:
Perfect!! Bravo.

app/views/questions/_comment.html.erb
```diff
@@ -23,10 +23,13 @@
         <%= render partial: "questions/form", locals: { comment: comment } %>
       </form>
     </div>
-    
+
   </div>
   <div class="inline pull-xs-right" style="vertical-align: top">
-    <p><% if current_user && current_user.uid == comment.uid %>
+    <p><% if answer_id == 0 && current_user && current_user.uid == comment.uid %>
```
jywarren: Hi, @Gauravano -- could we expand this to include admins and moderators?

Gauravano: Hmm..actually I was thinking whether it would be appropriate to move any user's comment to answer section without any notice.We can add a suggest user feature and then it would be user's call.What do you think? But, if you want I can implement it like it's there for delete

jywarren:
Since it doesn't move to a different page, it's not a big change and we
often find it's helpful for new users who don't know how to use the system.
I think it's worth allowing without a check, but I appreciate your careful
thought on this! Thanks!

Gauravano:
Ok Jeff expanding this feature to admins and moderators too

PR #1943, opened 2018-01-10
showing basic user as a moderator corrected (fixes #1918 )

### LINKED ISSUE #1918: Showing a basic user as 'moderator' .

In [/people](https://publiclab.org/people) , basic user is shown as a 'moderator'  . 
Also do we have to show normal users the ban buttons ?

screen shot 2018-01-05 at 11 35 17 pm

Diksha2008:
fixes #1918 

### DIFF
--- a/app/views/users/list.html.erb
+++ b/app/views/users/list.html.erb
@@ -5,7 +5,7 @@
 			<p> This page shows a list of all Public Lab contributors and is sorted by those who have posted most recently. Click the profile link to view their posting activity. </p>
 		</div>
 	</div>
-  <% if current_user && current_user.role = "admin" && current_user.role = "moderator" %>
+  <% if current_user && current_user.role == "admin" && current_user.role == "moderator" %>
     <h2><%= t('users.list.user_moderation') %></h2>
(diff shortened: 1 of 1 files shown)


seafr:
Correct me if I'm wrong, but it seems that the code (line 8) checks if the current user is both admin and moderator.

Diksha2008:
@seafr yes! I didn't really notice it! Thanks for pointing it out!

app/views/users/list.html.erb
```diff
@@ -19,7 +19,7 @@
       <th><%= t('users.list.last_activity') %></th>
       <th><%= t('users.list.history') %></th>
       <th><%= t('users.list.joined') %></th>
-      <% if current_user && current_user.role = "admin" && current_user.role = "moderator" %>
+      <% if current_user && current_user.role == "admin" && current_user.role == "moderator" %>
```
jywarren: good catch here!!! But I think i agree w/ the above commenter that we need to use `||` instead of `&&` for this and line 8 -- so it's moderator OR admin. Thank you!!!

PR #1953, opened 2018-01-10
people maps

jywarren:
in progress...

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -396,7 +396,7 @@ DEPENDENCIES
   recaptcha
   rest-client
   rspec
-  rubocop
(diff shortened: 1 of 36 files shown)


Diksha2008:
Since the partial `_leaflet.html.erb` is template for  methods `notes_map` and `notes_map_by_tag` as well, I think we need to pass `people: false` or `people:true` in locals of these methods, i.e., here https://github.com/publiclab/plots2/blob/a63f10277070d568c948d676880a079d091251df/app/models/concerns/node_shared.rb#L169-L173 and similarly for `notes_map_by_tag`

jywarren:
Aha! Thanks, I added a "defined?" And it worked!

Issue #1951, opened 2018-01-10
 Add "Join" button to people lists


jywarren:
Hi, this is a [first-timers-only issue](https://github.com/publiclab/plots2/projects/2). This means we've worked to make it more legible to folks who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**. 

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general [help wanted](https://github.com/publiclab/plots2/labels/help-wanted) issues. Thanks!

## The problem

We can now list people who have a profile tag (for example "foo") by using the inline power tag `[people:foo]`:

Read about this and see it in action : https://publiclab.org/wiki/power-tags#Inline+People+Lists

But it's not easy to add yourself!

## Solution

Let's add a button below that which adds the current user to that group, by adding the corresponding profile tag. 

There's a controller method here for adding a profile tag, for reference: 

https://github.com/publiclab/plots2/blob/master/app/controllers/user_tags_controller.rb

And the template for the people lists is here; i've marked the line of code where we could insert the button, using:

```ruby
<p><a class="btn btn-primary" href="/profile/tags/create/<%= current_user.id %>?name=<%= tagname %>">Join this group</a></p>` tag. 
```

https://github.com/publiclab/plots2/blob/master/app/views/grids/_people.html.erb#L31

Finally, we'll need to make a fix to the user tag controller here: https://github.com/publiclab/plots2/blob/master/app/controllers/user_tags_controller.rb#L48

Let's change that to redirect to the profile page, so `redirect_to '/profile/' + user.username`

## Steps to Fix

- [ ] claim this issue with a comment here, below, and ask any clarifying questions you need
- [ ] set up a repository locally following the README instructions, and make sure that all tests pass
- [ ] try to fix the issue following the steps above, but even before you're done, you can:
- [ ] commit your changes and start a pull request (see [contributing to Public Lab software](https://publiclab.org/wiki/contributing-to-public-lab-software)) but mark it as "in progress" if you have questions or if you haven't finished
- [ ] alert someone via the developers list (see below) to have your pull request merged. This may take a few extra steps depending on when you do it, but we'll help you out!

[...]

PR #1954, opened 2018-01-10
[wip] #1951 Add "Join" button to people lists 

JahlHernandez:
Adds button to join people lists, as requested in #1951

Thanks!

### DIFF
--- a/app/controllers/user_tags_controller.rb
+++ b/app/controllers/user_tags_controller.rb
@@ -45,7 +45,7 @@ class UserTagsController < ApplicationController
       else
         flash[:notice] = I18n.t('user_tags_controller.tag_created', tag_name: @output[:saved][0][0]).html_safe
       end
-      redirect_to info_path, id: params[:id]
+      redirect_to '/profile/' + user.username
     end
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -34,8 +34,8 @@ module ApplicationHelper
     body = NodeShared.upgrades_grid(body)
     body = NodeShared.notes_map(body)
     body = NodeShared.notes_map_by_tag(body)
-    body = NodeShared.people_grid(body)
     body = NodeShared.people_map(body)
+    body = NodeShared.people_grid(body, current_user)
     body = NodeShared.graph_grid(body)
     body = NodeShared.wikis_grid(body)
     body
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -35,7 +35,7 @@ module ApplicationHelper
     body = NodeShared.notes_map(body)
     body = NodeShared.notes_map_by_tag(body)
     body = NodeShared.people_map(body)
-    body = NodeShared.people_grid(body, current_user || nil) # <= to allow testing of insert_extras
+    body = NodeShared.people_grid(body, @current_user || nil) # <= to allow testing of insert_extras
     body = NodeShared.graph_grid(body)
     body = NodeShared.wikis_grid(body)
     body
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -35,7 +35,7 @@ module ApplicationHelper
     body = NodeShared.notes_map(body)
     body = NodeShared.notes_map_by_tag(body)
     body = NodeShared.people_map(body)
-    body = NodeShared.people_grid(body, current_user)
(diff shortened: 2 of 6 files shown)


JahlHernandez:
I'm working on issue #1951, but I've run into a problem while trying to access current_user from the partial located at views/grid/_people.html.erb, whenever I try an reference that variable it returns 
`undefined local variable or method `current_user'` (you can see the line in 008652b)
How can I access that variable from the partial?

Thanks a lot for any help you can provide

JahlHernandez:
Hey there @jywarren, I tried that, but it throws the same error on the ```if```

jywarren:
Can you commit that addition and let's see how it does on this PR? Thanks!

JahlHernandez:
Done!
If it's worth something, if I use the defined? operator on the if, it always returns false, but doesn't throw any errors.

jywarren:
Wow, i had to solve a similar issue today - yes, let's use `defined?` -- great idea!

JahlHernandez:
Hey there, I still don't quite get when  would the `defined?` check return true, I'm logged in and it always returns false for current_user on that specific partial.

jywarren:
Hmm, i think i see what's happening. That partial is actually rendered unusually to a string insider of `/app/models/concerns/node_shared.rb`, where `current_user` is not defined. And, although it's not ideal here, the model shouldn't really "know" about stuff in the user session. 

Here's some possible solutions: https://stackoverflow.com/questions/2513383/access-current-user-in-model

What do you think?

jywarren:
I have a simpler idea. What if we just moved it to this line in the helper, where we /do/ have access to `current_user` -- 

https://github.com/publiclab/plots2/blob/ed1ef3b72c8a45bfd1d7b842d1d747db391b2a7f/app/helpers/application_helper.rb#L37

It's not quite as clean, but it'll just work and it'll be only one line. What do you think?

jywarren:

JahlHernandez:
Hey! sorry for the late reply.
I've done a couple more things:
89b0c2a I'm not sure if I understood what you meant by your previous comment, was I supposed to pass the `current_user` as a parameter like this?
4230f28 I added a button for when the user is not logged in, with a link to go to the login

jywarren:
Ah, a great solution! Perfect. There are some merge conflicts since others have submitted changes in the interim -- if you'd like to try rebasing your work over the latest master branch, give it a try! Otherwise I'm happy to help. 

In case it's helpful, here's some guidance on rebasing! https://www.atlassian.com/git/tutorials/merging-vs-rebasing

jywarren:
OK, that was a real tough rebase! I resolved the issues and made a couple small tweaks to the capitalization. It looks good now and I'll merge if the tests pass! 

Thanks for working on this -- it's great to see this feature finally get built!

jywarren:
Hmm, that caused some chaos in our tests, but I made some adjustments. Sorry this one ended up being more complex than we thought!

jywarren:
Hmm. I'm getting stuck here! @publiclab/reviewers -- is anyone able to help us out? Much appreciated!!

JahlHernandez:
I see, if there's anything else I can do to help, let me know!

ryzokuken:
@jywarren 

> I’ve seen this over and over again. Helpers relying on instance variables. Now, I’m not saying that it doesn’t work cause it does. If the view has the instance variable defined, the helper being mixed into the view context can use it. But it’s bad practice so don’t do it.

> Just treat helper methods like you do any other methods and pass them arguments. Don’t rely on instance variables because you won’t be able to reuse those helpers and it will make the helper very hard to read, test and maintain. The fact that Rails mixes everything in one giant view context is not something particularly helpful in this case so try to keep instance variables away from helper methods.

ryzokuken:
@jywarren better?

ryzokuken:
@jywarren could you please take a look at this?

jywarren:
Ooh, i agree! We've been gradually trying to move to more passed-in local arguments in partial templates as well. Any ideas on how to encourage this more?

Looking over this now.

jywarren:
Awesome, thank you!!!

ryzokuken:
@jywarren Perhaps we could add a rule in rubocop to enforce this behavior?

jywarren:
Cool idea -- maybe we say "encourage" not "enforce" despite rubocop being a cop! Maybe we should fork and rename "ruboencourager" :-P

This is live now! Thanks all!

https://publiclab.org/wiki/education#Education+group

PR #1964, opened 2018-01-12
added ability to exclude tags from an inline grid

### LINKED ISSUE #1340: ability to exclude tags from an inline grid with new syntax
## Problem

Using [inline power tags](https://publiclab.org/wiki/power-tags#Inline+power+tags), you can insert a table of content based on tags into a wiki page (see [these examples](https://publiclab.org/wiki/photo-monitoring)), with the synatx:

`[notes:balloon-mapping]`

But sometimes you want to show something with one tag, but excluding another tag, like "displaying questions about balloon mapping, but not ones that have been answered". 

## Proposed solution

We could make a new syntax, like using the `!` operator to indicate **exceptions**. That could look like this:

`[notes:question:balloon-mapping!answered]`

In this example, we'd see a list of questions on balloon mapping, but not ones that have been answered.

To do this, we'd need to look at the constructor methods for the inline grids themselves, here:

https://github.com/publiclab/plots2/blob/master/app/models/concerns/node_shared.rb#L13-L48

We could add a line after this section: https://github.com/publiclab/plots2/blob/master/app/models/concerns/node_shared.rb#L26-L29

We'd want to be sure an exclamation point would still be recognized by the regex expression here: https://github.com/publiclab/plots2/blob/master/app/models/concerns/node_shared.rb#L14

Then we could either rework the regex to separate out `$2` (the tagname) or split it with:

```ruby
if $2.includes?('!')
  tagname = $2.split('!').first
  exclude = $2.split('!').last
else
  tagname = $2
end
```

Before it gets used to form the query here:

https://github.com/publiclab/plots2/blob/master/app/models/concerns/node_shared.rb#L28

Then, if we found that there was an exclusion, we'd add an additional filter to that query, so we'd get only those which had the tag, but no answered ones, perhaps using code like this:

http://stackoverflow.com/a/42460499

For testing out Regex, please check out http://rubular.com -- very helpful!

This is quite a project, but would be really great - i'm happy to help some folks take this on!

ViditChitkara:
closes #1340

Thanks!

### DIFF
--- a/app/models/concerns/node_shared.rb
+++ b/app/models/concerns/node_shared.rb
@@ -31,11 +31,25 @@ module NodeShared
   def self.notes_grid(body, _page = 1)
     body.gsub(/[^\>`](\<p\>)?\[notes\:(\S+)\]/) do |_tagname|
       tagname = Regexp.last_match(2)
+      exclude = nil
+      if tagname.include?('!')
+        exclude = tagname.split('!') - [tagname.split('!').first]
+        tagname = tagname.split('!').first
+      end
+
       nodes = Node.where(status: 1, type: 'note')
                   .includes(:revision, :tag)
                   .references(:term_data, :node_revisions)
                   .where('term_data.name = ?', tagname)
                   .order('node_revisions.timestamp DESC')
+
+      if exclude.present?
+        exclude = Node.where(status: 1, type: 'note')
+                  .includes(:revision, :tag)
+                  .references(:node_revisions, :term_data)
+                  .where('term_data.name IN (?)', exclude)
+        nodes = nodes - exclude
+      end
       output = ''
       output += '<p>' if Regexp.last_match(1) == '<p>'
       a = ActionController::Base.new()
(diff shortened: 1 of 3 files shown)


ViditChitkara:
@jywarren This seems to be working for me! Still there are some cases which need to taken care of. For example:-
[notes:question:balloon-mapping!answered]
One way could be, add a tag named answered to questions which have been answered.
Your thoughts??

jywarren:
Wow, this is very impressive. Do you think it'd be possible to test it out with a unit test? There are some good examples here; you could search the output to ensure the excluded tag content does not appear: 

https://github.com/publiclab/plots2/blob/432a7cec5cb02c922e607a0649c26f9207eadf90/test/unit/node_shared_test.rb#L4-L12

What do you think?

jywarren:
I do think we'll have to follow up with another one for answered questions. let's open a separate issue!

ViditChitkara:
Sure, I'll write the tests and will open an issue regarding answered questions.

ViditChitkara:
@jywarren the changes are made!

jywarren:
This is so great. I do worry a little about the added complexity; would it be possible to duplicate your test to cover the remaining methods you've modified? Your PR is very thorough! 

Also, will this work for a text example like `note:balloon-mapping!question:balloon-mapping` to display `balloon-mapping` posts, but not questions with `question:balloon-mapping`? Maybe one extra test to that could be helpful. 

Thanks for your persistence and thoroughness!!!

ViditChitkara:
Sure I'll write the remaining tests!!
Also this will work if 'question:balloon-mapping' tag exists on question!! However one would just add balloon-mapping tag to a question and not question:balloon-mapping so this needs to be handled separately!
There are a couple cases which need to be followed up to this:-
1) answered questions (because answered is not a tag)
2) 'question:balloon-mapping' (because the tag maybe be balloon mapping and not question:balloon-mapping)
Should I make a follow-up issue for them, once this is merged?

jywarren:
Yes, that sounds great. Thanks!

jywarren:
This is awesome. Thank you so much!!! Big feature -- one we've wanted for a LONG time!

ViditChitkara:
Cool!! I'll make another follow up issue to cover all the remaining cases we discussed.

PR #1970, opened 2018-01-12
Few typing errors

SidharthBansal:
These are a few typing errors and few markdown features which were not implemented correctly


jywarren:
Thank you!!! :-) 👍

PR #1978, opened 2018-01-12
API for fetching nearby coordinates information

### LINKED ISSUE #1934: ability to refresh displayed inline maps' displayed data when you pan/zoom

You can embed inline maps using this inline powertag: https://publiclab.org/wiki/inline-maps

The code driving this is here:

https://github.com/jywarren/plots2/blob/184eced60d9d6842cad8df469bb1df2838528522/app/models/concerns/node_shared.rb#L102-L124

And the template is here: https://github.com/jywarren/plots2/blob/master/app/views/map/_leaflet.html.erb

However, when you drag the map, the markers aren't refreshed for the new viewing area. 

Let's break this in three:

1. [x] let's make an API call for data near a point, similar to https://publiclab.org/api/srch/notes?srchString=foo but perhaps https://publiclab.org/api/srch/notes?geo=41.0,-91.3. This could be built out from https://github.com/publiclab/plots2/blob/master/app/api/srch/search.rb#L52-L72, or a new method could be made
2. [x] let's rewire https://github.com/jywarren/plots2/blob/184eced60d9d6842cad8df469bb1df2838528522/app/models/concerns/node_shared.rb#L102-L124 to load a blank map which then loads the API call for the center point, and displays notes as points on hearing back from the API
3. [x] let's set the inline map to re-fetch notes near its center point each time the map is panned or zoomed

This is a big multi-part issue that we can solve in three or more steps. Anyone interested in breaking out the first one?

Gauravano:
Fixes #1934 
API working.Suggestions regarding naming and comments are welcome.

### DIFF
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -115,6 +115,27 @@ module Srch
         sresult
       end
 
+      desc 'Perform a search of particular latitude and longitude', hidden: false,
+                                                                                     is_array: false,
+                                                                                     nickname: 'srchGetLocations'
+      params do
+        requires :srchString, type: String, documentation: { example: 'Spec' }
+        optional :seq, type: Integer, documentation: { example: 995 }
+        optional :showCount, type: Integer, documentation: { example: 3 }
+        optional :pageNum, type: Integer, documentation: { example: 0 }
+      end
+      get :locations do
+        sresult = DocList.new
+        unless params[:srchString].nil? || params[:srchString] == 0 || !(params[:srchString].include? ",")
+          sservice = SearchService.new
+          sresult = sservice.search_nearbyLocations(params[:srchString])
+        end
+        sparms = SearchRequest.fromRequest(params)
+        sresult.srchParams = sparms
+        sresult
+      end
+
+
       # end endpoint definitions
     end
   end
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -131,7 +131,7 @@ module Srch
         sresult = DocList.new
         unless params[:srchString].nil? || params[:srchString] == 0 || !(params[:srchString].include? ",")
           sservice = SearchService.new
-          sresult = sservice.search_nearbyLocations(params[:srchString])
+          sresult = sservice.nearbyNodes(params[:srchString])
         end
         sparms = SearchRequest.fromRequest(params)
         sresult.srchParams = sparms
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -115,9 +115,12 @@ module Srch
         sresult
       end
 
-      desc 'Perform a search of particular latitude and longitude', hidden: false,
-                                                                                     is_array: false,
-                                                                                     nickname: 'srchGetLocations'
+      # Request URL should be /api/srch/locations?srchString=QRY[&seq=KEYCOUNT&showCount=NUM_ROWS&pageNum=PAGE_NUM]
+      # Note: Query(QRY as above) must have latitude and longitude as srchString=lat,lon
+      desc 'Perform a search of documents having nearby latitude and longitude tag values',
+                                                                    hidden: false,
+                                                                           is_array: false,
+                                                                           nickname: 'srchGetLocations'
       params do
         requires :srchString, type: String, documentation: { example: 'Spec' }
         optional :seq, type: Integer, documentation: { example: 995 }
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -183,7 +183,7 @@ class SearchService
     sresult
   end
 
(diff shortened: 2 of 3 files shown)


Gauravano:
When correct format is followed in URL
first

When correct format not followed
second

app/services/search_service.rb
```diff
@@ -182,4 +182,29 @@ def textSearch_questions(srchString)
     end
     sresult
   end
+
+  def search_nearbyLocations(srchString)
```
jywarren: This is looking great! Could we call this `nearbyNodes`?

app/api/srch/search.rb
```diff
@@ -115,6 +115,30 @@ class Search < Grape::API
         sresult
       end
 
+      # Request URL should be /api/srch/locations?srchString=QRY[&seq=KEYCOUNT&showCount=NUM_ROWS&pageNum=PAGE_NUM]
```
jywarren: Super! Perhaps we can add this also to `/doc/API.md`?

app/services/search_service.rb
Gauravano: doing changes now

app/api/srch/search.rb
Gauravano: Done

app/services/search_service.rb
```diff
@@ -182,4 +182,29 @@ def textSearch_questions(srchString)
     end
     sresult
   end
+
+  def nearbyNodes(srchString)
+    sresult = DocList.new
+    coordinates = srchString.split(",")
+    lat = coordinates[0]
+    lon = coordinates[1]
+
```
asahu8: latitude, longitude = coordinates 
directly we can assign, ruby will take care of it.

app/services/search_service.rb
```diff
@@ -182,4 +182,29 @@ def textSearch_questions(srchString)
...
+    nids = NodeTag.joins(:tag)
+               .where('name LIKE ?', 'lat:' + lat[0..lat.length - 2] + '%')
+               .collect(&:nid)
+
+    nids = nids || []
+
```
asahu8: nids ||= []

app/api/srch/search.rb
```diff
@@ -115,6 +115,30 @@ class Search < Grape::API
...
+      params do
+        requires :srchString, type: String, documentation: { example: 'Spec' }
+        optional :seq, type: Integer, documentation: { example: 995 }
+        optional :showCount, type: Integer, documentation: { example: 3 }
+        optional :pageNum, type: Integer, documentation: { example: 0 }
+      end
+      get :locations do
+        sresult = DocList.new
+        unless params[:srchString].nil? || params[:srchString] == 0 || !(params[:srchString].include? ",")
+          sservice = SearchService.new
```
asahu8: params[:srchString].nil? || params[:srchString] == 0
this particular code is kind of used in multiple places in this file. can't be move it to a private method and re-use it relevant place ?

Gauravano:
jeff could you plz help with the build logs 
build

app/services/search_service.rb
Gauravano: You are absolutely right but I think for sake of clarity of all writing one extra line is good

jywarren:
hmm, this is odd, no? let me take a look.

jywarren:
`Cannot find a merge base between danger_base and danger_head` -- well, searching for this while restarting the build.

jywarren:
https://github.com/danger/danger/issues/768

jywarren:
oh, do you have two branches of almost the same name?

app/api/srch/search.rb
jywarren: I think so -- but perhaps independently from this PR so we address code repetition as its own issues. Would you mind opening a new issue to address this? I agree that the services and api code in particular can be enormously compacted without loss of organization.  Thank you!

app/services/search_service.rb
jywarren: Good suggestion! I always forget to do this myself.

app/services/search_service.rb
jywarren: :-)

jywarren:
Tough one -- it says the name collision may originate in a different PR -- scary! I'll watch if other PR's travis tests are failing...

Gauravano:
same logs again

jywarren:
:-/ i'm not sure about this one! @publiclab/reviewers can anyone help debug this?

Gauravano:
There is warning about missing secret_key_base too.Do you think it has any role to play?

jywarren:
Hmm, that may be a question for @icarito ... It's a rails 4.1 thing!

Gauravano:
Jeff error was generated due to the wrong merging of code on my master branch.Its resolved now.Thanks

PR #1984, opened 2018-01-12
Inconsistencies in tag deletion

### LINKED ISSUE #1859: Inconsistencies in tag deletion (investigating)
On the https://publiclab.org/wiki/getting-started I observed some issues
1.Any user can add tags to this page. Like I added "balloon". 
2. It automatically tagged with "balloon" and "balloon mapping".
3.There are some tag names with 'X' and some without 'X' for deletion. The tag named "balloon" and "balloon mapping" is not having 'X' to click which I made.
4.On pressing the 'X' next to the tag name by Warren, a javascript dialog box appears, which should not. I tried to delete one but not able to. 
I think a person who has written the tag has the capability to it but here it is not happening.
image

bansal_sidharth2996
### Which operating system and browser are you using?
I am using Ubuntu 14.04 with Firefox 57.0.1 (64-bit)

SidharthBansal:
Fixes #1859

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -36,7 +36,7 @@ class TagController < ApplicationController
                .where('community_tags.date > ?', (DateTime.now - 1.month).to_i)
                .group(:name)
                .order('name')
-               .paginate(page: params[:page], per_page: 24)    
+               .paginate(page: params[:page], per_page: 24)
     end
   end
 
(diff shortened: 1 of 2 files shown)


SidharthBansal:
@jwarren Hope this will also correct the" balloon-mapping " tag which is only at one url.
Thanks

jywarren:
Perfect, thanks!

PR #1987, opened 2018-01-13
API for nearby node fetching

### LINKED ISSUE #1934: ability to refresh displayed inline maps' displayed data when you pan/zoom

Gauravano:
Fixes #1934

### DIFF
--- a/app/models/doc_result.rb
+++ b/app/models/doc_result.rb
@@ -1,6 +1,6 @@
 # A DocResult is an individual return item for a document (web page) search
 class DocResult
-  attr_accessor :docId, :docType, :docUrl, :docTitle, :docSummary, :docScore
+  attr_accessor :docId, :docType, :docUrl, :docTitle, :docSummary, :docScore, :latitude,:longitude
 
   def initialize; end
 
@@ -15,6 +15,19 @@ class DocResult
     obj
   end
 
+  def self.fromLocationSearch(idval, typeval, urlval, titleval, sumval, scoreval,latitude,longitude)
+    obj = new
+    obj.docId = idval
+    obj.docType = typeval
+    obj.docUrl = urlval
+    obj.docTitle = titleval
+    obj.docSummary = sumval
+    obj.docScore = scoreval
+    obj.latitude = latitude
+    obj.longitude = longitude
+    obj
+  end
+
   # This subclass is used to auto-generate the RESTful data structure.  It is generally not useful for internal Ruby usage
   #  but must be included for full RESTful functionality.
   class Entity < Grape::Entity
@@ -24,5 +37,7 @@ class DocResult
     expose :docTitle, documentation: { type: 'String', desc: 'Title or primary descriptor of the linked result.' }
     expose :docSummary, documentation: { type: 'String', desc: 'If available, first paragraph or descriptor of the linked document.' }
     expose :docScore, documentation: { type: 'Float', desc: "If calculated, the relevance of the document result to the search request; i.e. the 'matching score'" }
+    expose :latitude, documentation: {type: 'String',desc: "Returns the latitude associated with node"}
+    expose :longitude, documentation: {type: 'String',desc: "Returns the longitude associated with node"}
   end
 end
--- a/app/models/doc_result.rb
+++ b/app/models/doc_result.rb
@@ -1,6 +1,6 @@
 # A DocResult is an individual return item for a document (web page) search
 class DocResult
-  attr_accessor :docId, :docType, :docUrl, :docTitle, :docSummary, :docScore, :latitude,:longitude
+  attr_accessor :docId, :docType, :docUrl, :docTitle, :docSummary, :docScore, :latitude, :longitude, :blurred
 
   def initialize; end
 
@@ -15,7 +15,7 @@ class DocResult
     obj
   end
 
-  def self.fromLocationSearch(idval, typeval, urlval, titleval, sumval, scoreval,latitude,longitude)
+  def self.fromLocationSearch(idval, typeval, urlval, titleval, sumval, scoreval,latitude,longitude,blurred)
     obj = new
     obj.docId = idval
     obj.docType = typeval
@@ -25,6 +25,7 @@ class DocResult
     obj.docScore = scoreval
     obj.latitude = latitude
     obj.longitude = longitude
+    obj.blurred = blurred
     obj
   end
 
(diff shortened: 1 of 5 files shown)


Gauravano:
Tests are always good.will add tests...Thanks, jeff.

Gauravano:
Hey, @sagarpreet-chadha I have updated the items so latitude and longitude would be available now.

update

jywarren:
Hi, just a note - can we also specify in a field if `location:blurred` so we know whether people are sharing an exact location or a low-res one? Thank you!

Gauravano:
@jywarren could you please elaborate on what factors we can term location as blurred ? Is it related to latitude and longitude entered (srchString=lat,lon) that if lat=24.68 i.e.,2 digits after decimal then it is termed as blurred location ?

sagarpreet-chadha:
Hi @Gauravano  , i think if node contains tag `location:blurred` , the node is sharing low-res location . What do you think ? Thanks 😄 !

Gauravano:
Yes @sagarpreet-chadha it makes sense.Thanks

Gauravano:
@jywarren the first part of the issue is completed by me, plz review it.

app/models/doc_result.rb
```diff
@@ -15,6 +15,20 @@ def self.fromSearch(idval, typeval, urlval, titleval, sumval, scoreval)
     obj
   end
 
+  def self.fromLocationSearch(idval, typeval, urlval, titleval, sumval, scoreval,latitude,longitude,blurred)
```
jywarren: Hi, just wondering if we could tidy up the spacing towards the end of this line, and on lines 41-2 below. This really helps keep the code tidy over the long term, thanks!

app/models/doc_result.rb
```diff
@@ -24,5 +38,7 @@ class Entity < Grape::Entity
     expose :docTitle, documentation: { type: 'String', desc: 'Title or primary descriptor of the linked result.' }
     expose :docSummary, documentation: { type: 'String', desc: 'If available, first paragraph or descriptor of the linked document.' }
     expose :docScore, documentation: { type: 'Float', desc: "If calculated, the relevance of the document result to the search request; i.e. the 'matching score'" }
+    expose :latitude, documentation: {type: 'String',desc: "Returns the latitude associated with node"}
+    expose :longitude, documentation: {type: 'String',desc: "Returns the longitude associated with node"}
```
jywarren: Here, should it be "Float"? Or is it String because we want to preserve the # of significant digits? (the latter makes sense to me... just want to be sure)

app/services/search_service.rb
```diff
@@ -183,4 +183,37 @@ def textSearch_questions(srchString)
     end
     sresult
   end
+
+  #Search nearby nodes with respect to given latitude and longitude
```
jywarren: I /think/ this needs a space between the `#` and `S` to be recognized by the auto API generator which I've been testing in another PR. Sorry to be so detail-oriented!

app/services/search_service.rb
```diff
@@ -183,4 +183,37 @@ def textSearch_questions(srchString)
...
+               .collect(&:nid)
+
+    nids = nids || []
+
+    items = Node.includes(:tag)
+                .references(:node, :term_data)
+                .where('node.nid IN (?) AND term_data.name LIKE ?', nids, 'lon:' + lon[0..lon.length - 2] + '%')
+                .limit(200)
+                .order('node.nid DESC')
+items.each do |match|
```
jywarren: And here maybe slightly better indentation plus spaces around the `=` on line 206, 208 would be really great. 

Thanks again for all your hard work! This is going to be great.

doc/API.md
```diff
@@ -10,6 +10,7 @@ Per-model API endpoints are:
 * Questions: https://publiclab.org/api/srch/questions?srchString=foo
 * Tags: https://publiclab.org/api/srch/tags?srchString=foo
 * Notes: https://publiclab.org/api/srch/notes?srchString=foo
+* Locations: https://publiclab.org/api/srch/locations?srchString=lat,lon
```
jywarren: Awesome

app/models/doc_result.rb
Gauravano: correcting this..

app/services/search_service.rb
Gauravano: No problem at all

test/functional/search_api_test.rb
```diff
@@ -26,7 +25,31 @@ def app
     assert_equal nodes(:blog).path, json['items'][0]['docUrl']
     assert_equal "Blog post",       json['items'][0]['docTitle']
     assert_equal 13,                json['items'][0]['docId']
-    
+
+    assert matcher =~ json
+
+  end
+
+  test 'search nearby nodes functionality' do
```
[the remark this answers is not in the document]
Gauravano: Thanks Jeff

app/services/search_service.rb
Gauravano: Correcting..

Gauravano:
@jywarren plz review it.Thanks.

jywarren:
🎉 🎈 👍 💥

PR #1997, opened 2018-01-13
excluding fulltext indices from sqlite in schema.rb.example -- test

jywarren:
Related to sqlite issues with fulltext indices in schema.rb

### DIFF
--- a/db/schema.rb.example
+++ b/db/schema.rb.example
@@ -51,7 +51,7 @@ ActiveRecord::Schema.define(version: 20180103212804) do
     t.integer "aid",                          default: 0,  null: false
   end
 
-  add_index "comments", ["comment"], name: "index_comments_on_comment", type: :fulltext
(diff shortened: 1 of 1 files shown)


jywarren:
@StlMaris123 -- how about this?

jywarren:
Hi, @Gauravano -- what do you think, should I merge this in? We may have to manually maintain these lines in `schema.rb.example`, if we update the schema, though. Maybe let's do this and then think about maintainability? We could also add a line to the Dangerfile to help us remember:

https://github.com/publiclab/plots2/blob/033676b8abd28cd56a999e60934ff9f01bd0a9ac/Dangerfile#L13-L15

Thoughts?

Gauravano:
Sorry, @jywarren for responding late on this.I agree with your point to merge this for now and then think about maintainablility issues.Also, many new people might want to contribute to plots2 but can't continue due to this issue generating error in initial stages of repo setup.Thank you.

jywarren:
OK, i'll go ahead and do this and open a change to the Dangerfile!

PR #2000, opened 2018-01-13
added "Suggest new title" feature

### LINKED ISSUE #919: "Suggest a new title" feature to make friendly suggestions of more descriptive titles
This would need some UI design work, but the idea is that it'd be a comment which could actually have a button on it that says something like:

> [Jeff]() is suggesting an alternative title, "How to fix a Time Machine Kit". To accept this as a new title to your post, `click here`. 

Then that button would directly replace your title. Curious what people think of this idea -- it'd likely need both a new controller action in the [notes controller](https://github.com/publiclab/plots2/tree/master/app/controllers/notes_controller.rb), and some well-thought-through UI design leading into the comment display template: https://github.com/publiclab/plots2/blob/master/app/views/comments/_comments.html.erb

ViditChitkara:
closes #919

Thanks!

### DIFF
--- a/app/views/notes/_comment.html.erb
+++ b/app/views/notes/_comment.html.erb
@@ -64,7 +64,7 @@
   </script>
   
   <div id="c<%= comment.cid %>show">
-    <p><%= raw sanitize (RDiscount.new(title_suggestion(comment.body, comment.drupal_user.name, comment.nid, comment.id)).to_html), attributes: %w(class style href data-method) %></p>
+    <p><%= raw sanitize (RDiscount.new(title_suggestion(comment)).to_html), attributes: %w(class style href data-method) %></p>
     <% if comment.body.include?('?') %>
       <p class="alert alert-info">Is this a question? <a href="/questions/new?title=<%= comment.body %>">Click here</a> to post it to the <a href="/questions">Questions page</a>.
       </p>
--- a/app/views/notes/_comment.html.erb
+++ b/app/views/notes/_comment.html.erb
@@ -64,7 +64,7 @@
   </script>
   
   <div id="c<%= comment.cid %>show">
-    <p><%= raw sanitize (RDiscount.new(title_suggestion(comment)).to_html), attributes: %w(class style href data-method src) %></p>
+    <p><%= render_comment_body(comment) %></p>
     <% if comment.body.include?('?') %>
       <p class="alert alert-info">Is this a question? <a href="/questions/new?title=<%= comment.body %>">Click here</a> to post it to the <a href="/questions">Questions page</a>.
       </p>
--- a/app/views/notes/_comment.html.erb
+++ b/app/views/notes/_comment.html.erb
@@ -64,7 +64,7 @@
   </script>
   
   <div id="c<%= comment.cid %>show">
-    <p><%= raw sanitize(auto_link(RDiscount.new(comment.body).to_html)) %></p>
+    <p><%= raw sanitize (RDiscount.new(title_suggestion(comment.body, comment.drupal_user.name, comment.nid, comment.id)).to_html), attributes: %w(class style href data-method) %></p>
     <% if comment.body.include?('?') %>
       <p class="alert alert-info">Is this a question? <a href="/questions/new?title=<%= comment.body %>">Click here</a> to post it to the <a href="/questions">Questions page</a>.
       </p>
--- a/app/views/notes/_comment.html.erb
+++ b/app/views/notes/_comment.html.erb
@@ -64,7 +64,7 @@
   </script>
   
   <div id="c<%= comment.cid %>show">
-    <p><%= raw sanitize (RDiscount.new(title_suggestion(comment)).to_html), attributes: %w(class style href data-method) %></p>
+    <p><%= raw sanitize (RDiscount.new(title_suggestion(comment)).to_html), attributes: %w(class style href data-method src) %></p>
     <% if comment.body.include?('?') %>
       <p class="alert alert-info">Is this a question? <a href="/questions/new?title=<%= comment.body %>">Click here</a> to post it to the <a href="/questions">Questions page</a>.
       </p>
--- a/app/views/notes/_title_suggestion.html.erb
+++ b/app/views/notes/_title_suggestion.html.erb
@@ -1,0 +1,6 @@
+<blockquote>
+	<a href="/profile/<%= user %>"><%= user %></a> is suggesting an alternative title, "<%= title %>".
+	<% if is_creator %> 
+	To accept this as a new title to your post, <a href="/request_title_change/<%= nid %>/<%= title %>" data-method='post'>click here</a>
+	<% end %> 
+</blockquote>
--- a/app/views/notes/_title_suggestion.html.erb
+++ b/app/views/notes/_title_suggestion.html.erb
@@ -1,6 +1,6 @@
 <blockquote>
 	<a href="/profile/<%= user %>"><%= user %></a> is suggesting an alternative title, "<%= title %>".
 	<% if is_creator %> 
-	To accept this as a new title to your post, <a href="/reset-title?id=<%= nid %>&title=<%= title %>" data-method='post'>click here</a>
+	To accept this as a new title to your post, <a href="/node/update/title?id=<%= nid %>&title=<%= title %>" data-method='post'>click here</a>
 	<% end %> 
 </blockquote>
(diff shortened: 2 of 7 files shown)


ViditChitkara:
@jywarren I have a raw idea about this. Just like inline power tags there could be another syntax which could create the specified template for title suggestion (e.g {New title}). 
Does this make sense?

config/routes.rb
```diff
@@ -6,6 +6,7 @@
   post 'comment/create/token/:id.:format', to: 'comment#create_by_token'
 
   get 'searches/test' => 'searches#test'
+  post '/request_title_change/:id/:title' => 'notes#request_title_change'
```
jywarren: Maybe we could simplify this and also not put the title itself in the path, as it may contain punctuation? It could be using `?title=____` instead.

app/views/notes/_comment.html.erb
```diff
@@ -64,7 +64,7 @@
   </script>
   
   <div id="c<%= comment.cid %>show">
-    <p><%= raw sanitize(auto_link(RDiscount.new(comment.body).to_html)) %></p>
+    <p><%= raw sanitize (RDiscount.new(title_suggestion(comment.body, comment.drupal_user.name, comment.nid, comment.id)).to_html), attributes: %w(class style href data-method) %></p>
```
jywarren: This isi getting quite long; any way to simplify via formatting?

ViditChitkara: Maybe pass the comment object to title_suggestion method would be better rather than breaking it up into body, name, etc. 'Attributes' require href and data-method for the time being perhaps. I'll try something to shorten this!!,

config/routes.rb
ViditChitkara: Sure, we could do the same with id!! Also we'd need to have a check in the request_title_update method, since the route could easily be manipulated to change the id or title so an authorisation check is a must at the back-end.

jywarren:
PR #2000 -- cool! :-)

app/views/notes/_comment.html.erb
```diff
@@ -64,7 +64,7 @@
   </script>
   
   <div id="c<%= comment.cid %>show">
-    <p><%= raw sanitize(auto_link(RDiscount.new(comment.body).to_html)) %></p>
+    <p><%= raw sanitize (RDiscount.new(title_suggestion(comment)).to_html), attributes: %w(class style href data-method src) %></p>
```
jywarren: This starts to get quite long. See how we've batched these in this code?

https://github.com/ViditChitkara/plots2/blob/624aa873794f0d904fb44e0772a2fddf51309d5d/app/helpers/application_helper.rb#L30-L42

What do you think about making this more readable with an additional helper somehow, rather than a long in-line code block?

Looking good though!

app/views/notes/_title_suggestion.html.erb
```diff
@@ -0,0 +1,6 @@
+<blockquote>
+	<a href="/profile/<%= user %>"><%= user %></a> is suggesting an alternative title, "<%= title %>".
```
jywarren: Can we switch to 2-space indents here? sorry to be particular :-)

Also, can you share a screenshot of what this looks like?

config/routes.rb
```diff
@@ -6,6 +6,7 @@
   post 'comment/create/token/:id.:format', to: 'comment#create_by_token'
 
   get 'searches/test' => 'searches#test'
+  post '/reset-title' => 'notes#request_title_change'
```
jywarren: Does this actually change the title, or does it request a title change? I think it actually changes the title. Maybe we could do like `/node/update/title` or something more descriptive? Same for the method name `request_title_change` -- could that be `update_title` to be more accurate, or do i misunderstand?

app/views/notes/_comment.html.erb
ViditChitkara: Some helper method like:- comment_body
It could look like this:- 
```
def comment body(comment)
  RDiscount.new(title_suggestion(comment)).to_html), attributes: %w(class style href data-method src
end
```
Then we could use it in line 67 as:- 
```
<p><%= raw sanitize(auto_link(RDiscount.new(comment_body(comment)).to_html)) %></p>

```
Will this be good?

app/views/notes/_title_suggestion.html.erb
ViditChitkara: Oh I missed that!! I'll do!!

config/routes.rb
ViditChitkara: This actually changes the title. Yes, that's a cool idea!! I'll change the route name and method name to the above.

app/views/notes/_title_suggestion.html.erb
ViditChitkara: Yes, I'll share the screenshots!

ViditChitkara:
tsug-1
tsug-2
@jywarren  Here are the screenshots of the feature

jywarren:
aha - cool. Do you think perhaps it should be something like this, though, so we can expand out for other types of inline comment features in the future?

`[propose:title:_____________]`

ViditChitkara:
This syntax looks cool indeed!! I'll do the changes

jywarren:
Or:

```
[propose:title]This is a new title[/propose]
```

which do you think makes more sense?

ViditChitkara:
@jywarren I have made the changes!! May I write tests if the design seems to be finalised?

jywarren:
Let's see -- one thing is, perhaps we can use an `<div class="alert alert-warning"></div>` instead of a blockquote? That'll seem more like an official feature. 

Also, just to think through this, what if you press this twice? It'll just continually replace the title, right? No possibility of weird behavior I can think of... just want to be thorough. OK, let's go for tests then!

ViditChitkara:
Yes, it would just replace the title continuously, I have tried that too and no weird behaviours were seen. I'd like to try and replace the block quote with something mentioned abo.

ViditChitkara:
Duh! That was closed by mistake! I'll write the tests!

ViditChitkara:
@jywarren written the tests and also replaced the blockquotes with <div class="alert alert-warning"></div>.
This actually looks nice with the alert panel!! Thanks for the suggestion.
Could you please see if this is good to go now?

jywarren:
Wow, very cool! I think we are ready to go; do you think we should try this out once merged, then think about a user interface as a follow-up? We could for example add a button like the "promote to answer" button on comments -- which could use a JavaScript `prompt()` to ask for the proposed title. 

The button could appear as it does in this change by @Gauravano -- https://github.com/publiclab/plots2/pull/2001/files

What do you think, shall we merge this in, test it out, then think about UI as a follow-up?

ViditChitkara:
This could make a really good first timer issue. Let's test this out on
production and see if it works or not.
Then, should I create a first-timer-issue for the follow-up part? This
would be quite a task but would be very interesting for new comers!!

ViditChitkara:
@jywarren I was just thinking of another way of doing this! We would have a button in the comments section (which would say something like 'propose title'). When clicked, it would append the template for title suggestion like `[propose:title]Propose your title here[/propose]` in the comment editor.
This could be useful if a user wants to propose more than one title in a single comment. Then there would be no need of the prompt.
Your thoughts?

jywarren:
Oh, that sounds really nice -- and could still be a first-timers-issue --
super!

ViditChitkara:
Cool!! Do you think we should merge this? I will create a fto issue for the left over part as explained above!!
Thanks.

ViditChitkara:
@jywarren please see if there are any more required changes to it.

ViditChitkara:
That's interesting!! I'll do add the tests!!

ViditChitkara:
@jywarren the tests are written!! Please have a look!

jywarren:
Wow, this is so cool! 🎉 

Merging this! I hope we can publish tomorrow, we'll see!

jywarren:
Hi, @ViditChitkara -- just noting one follow-up issue -- i see an error:

```
/app/app/helpers/application_helper.rb:55: warning: regular expression has ']' without escape: /\[propose:title](.*?)\[\/propose]/
/app/app/helpers/application_helper.rb:55: warning: regular expression has ']' without escape: /\[propose:title](.*?)\[\/propose]/
```

Looks like maybe we need to escape that?

ViditChitkara:
Oh I had this warning too, slipped out of my mind. Will give a quick fix to update this!!

namangupta01:
@ViditChitkara @jywarren working on this in #2125

ViditChitkara:
Oops sorry didn't see that!! So are you covering the warning in fixing the robocup offences?

namangupta01:
@ViditChitkara yes i am covering all the rubocop offences there
see here: https://pastebin.com/raw/J2kxYX66

jywarren:
great, thanks folks!

PR #2002, opened 2018-01-14
Sort wikis by popularity corrected

### LINKED ISSUE #1928: Wiki Pages can't be sorted by the pageviews
Wiki pages do not get sorted in the descending order on https://publiclab.org/wiki/popular
image

SidharthBansal:
Fixes #1928

### DIFF
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -315,7 +315,7 @@ class WikiController < ApplicationController
                  .order('node_revisions.timestamp ASC')
                  .where("node_revisions.status = 1 AND node.status = 1 AND (type = 'page' OR type = 'tool' OR type = 'place')")
                  .page(params[:page])
-                 
+
     @paginated = true
     render template: 'wiki/index'
   end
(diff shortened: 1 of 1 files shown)


SidharthBansal:
@jywarren  Can you please review me and suggest if there are any suggestions

app/controllers/wiki_controller.rb
```diff
@@ -315,19 +315,19 @@ def stale
...
 
   def popular
     @title = I18n.t('wiki_controller.popular_wiki_pages')
     @wikis = Node.limit(40)
-                 .order('views DESC')
                  .joins(:revision)
                  .group('node_revisions.nid')
                  .order('node_revisions.timestamp DESC')
                  .where("node.status = 1 AND node_revisions.status = 1 AND node.nid != 259 AND (type = 'page' OR type = 'tool' OR type = 'place')")
+                 .sort_by(&:totalviews).reverse
```
jywarren: Would this conflict with (or could we combine it with) the above `order()` method? Thanks!!!

SidharthBansal: It is a totalview method, methods of models needs to be written with sort_by method. In order by we cant declare model methods. that is why I have written like this.
T

jywarren: I see. We could in theory join another table and do a MYSQL count there, but it's probably not worth it. OK, then is this ready to go?

SidharthBansal: On my local host it is working fine. I have pasted a photo under the issue section

SidharthBansal: Most probably it should work in production also :smile:

jywarren:
Great! Merged!

PR #2005, opened 2018-01-14
Update the text below MapKnitter Map

SidharthBansal:
I was reading your profile @jwarren. It is awesome. Just found a small error
image

The last line in the image it is broken. Someone missed raw keyword.
I hope it works fine now

### DIFF
--- a/app/views/users/profile.html.erb
+++ b/app/views/users/profile.html.erb
@@ -200,7 +200,7 @@
     </div>
 
     <div class="tab-pane" id="maps">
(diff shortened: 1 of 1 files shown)


SidharthBansal:
The travis tests are failing. As soon as I pushed the pull request it break. They are not even running. That was the reason, I close and reopen. But again they are not working.

jywarren:
Looks like one passed and one failed -- a sure sign that it's not your code, but some upstream server availability -- maybe a timeout. Restarting the one that failed. Thanks!

PR #2012, opened 2018-01-16
Redundant dashboard2 route 

### LINKED ISSUE #2006: Redundant dashboard2 route
I found there is a route /dashboard2 
  get 'dashboard2' => 'home#dashboard2'
I checked the controller's action it is redirecting to the dashboard.
Does it needs to be removed @jywarren ? Or is there any place we are using it at?

SidharthBansal:
Fixes #2006


jywarren:
Awesome, thanks. An artifact from when we did a public beta for a new dashboard.

Issue #1958, opened 2018-01-11
Adding images and decription to tweets


sagarpreet-chadha:
Hi, this is a [first-timers-only](https://github.com/publiclab/plots2/projects/2) issue. This means we've worked to make it more legible to folks who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**.

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general [help wanted](https://github.com/publiclab/plots2/labels/help-wanted) issues. Thanks!

# The Problem

Currently the **tweets** on [PublicLab account ](https://twitter.com/PublicLab)  have format  : 
<img width="588" alt="screen shot 2018-01-11 at 6 25 28 pm" src="https://user-images.githubusercontent.com/14952645/34826552-fde0b91c-f6fc-11e7-8569-f7da50890440.png">

The problem is that the **image is not shown** on tweets and the **description of the shared article** is not shown . 

Example :  For research note ,   [https://publiclab.org/notes/cutlan/01-09-2018/kilimanjaro-citizen-science-1](https://publiclab.org/notes/cutlan/01-09-2018/kilimanjaro-citizen-science-1) ,  the tweet is [this](https://twitter.com/PublicLab/status/950784615513493507) . 

# Solution 

We need to change the tweets to this **format** : 

<img width="594" alt="screen shot 2018-01-11 at 2 32 01 am" src="https://user-images.githubusercontent.com/14952645/34826725-bfec845a-f6fd-11e7-8510-e92f00d5d818.png">

The tweets are generated from the **recipe** (Learn more about recipes from [here](https://ifttt.com/discover))  [https://ifttt.com/applets/66362400d-rss-feed-to-post-a-tweet-with-image-and-url](https://ifttt.com/applets/66362400d-rss-feed-to-post-a-tweet-with-image-and-url) , which tweets whenever there is **change in the RSS feed of Publiclab.com site** [https://publiclab.org/feed.rss](https://publiclab.org/feed.rss) .

We need to change the **format of our RSS Builder** , [https://github.com/publiclab/plots2/blob/master/app/views/notes/rss.rss.builder](https://github.com/publiclab/plots2/blob/master/app/views/notes/rss.rss.builder) to produce correct RSS Feeds .

[...]

PR #2024, opened 2018-01-16
Made RSS feed a valid one + functional test for it 

sagarpreet-chadha:
Some changes to #1958 .

- Made changes to `rss.rss.builder` and made it a valid one according to [https://validator.w3.org/feed/check.cgi](https://validator.w3.org/feed/check.cgi) . 

- Functional test also added to check the same .

Thanks!

### DIFF
--- a/app/views/notes/rss.rss.builder
+++ b/app/views/notes/rss.rss.builder
@@ -1,14 +1,15 @@
-xml.instruct! :xml, :version => "1.0" 
-xml.rss :version => "2.0" do
+xml.instruct!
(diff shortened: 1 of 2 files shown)


jywarren:
Hmm, not sure why, but looks like an error. I just tweaked the spacing a bit, so it wasn't my change... great changes btw!

sagarpreet-chadha:
@jywarren ,thank you , also i was thinking about what would happen if the body/description  length is greater than the tweet character limit ?  What do you think ?

jywarren:
I think it'll auto-truncate, but tweets are also longer now :-)

Do you know why it's stalling?

jywarren:
Awesome!

PR #2031, opened 2018-01-17
Wildcard searches shows a message for lead wiki pages

### LINKED ISSUE #1982: Lead Wiki pages on Power tag search
Urls like www.publiclab.org/tag/balloon-mapping* end up in showing it leads to no wiki pages but balloon-mapping may be a wiki page.
image
image
Refer #1872

SidharthBansal:
Fixes #1982


jywarren:
:-) 🎉 👍

Issue #1977, opened 2018-01-12
Add comments to search


jywarren:
Our new search system is real nice! We have basics of comment search built:

https://github.com/publiclab/plots2/blob/01fac7cda6ab3df0ee9ba492f489edeef741ece9/app/services/typeahead_service.rb#L35-L46

But it needs to be integrated. 

We'll need a `search_comments` function similar to this one:

https://github.com/publiclab/plots2/blob/01fac7cda6ab3df0ee9ba492f489edeef741ece9/app/services/typeahead_service.rb#L130-L143

which uses the above `comments` method instead of `notes`...

Then we can add a line after this one to include comments into search results!

https://github.com/publiclab/plots2/blob/01fac7cda6ab3df0ee9ba492f489edeef741ece9/app/services/typeahead_service.rb#L90-L92

I think that should work... let's try it out!

ViditChitkara:
Shall I try?

jywarren:
Sure, that'd be great! Thanks!

PR #2045, opened 2018-01-18
added ability of power tags to exclude answered questions

ViditChitkara:
updates #1340

Make sure these boxes are checked before your pull request is ready to be reviewed and merged. Thanks!

* [X] all tests pass -- `rake test:all`
* [X] code is in uniquely-named feature branch, and has been rebased on top of latest master (especially if you've been asked to make additional changes)
* [X] pull request is descriptively named with #number reference back to original issue

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/wiki/contributing-to-public-lab-software

We have a loose schedule of reviewing and pulling in changes every Tuesday and Friday, and publishing changes on Fridays. Please alert developers on plots-dev@googlegroups.com when your request is ready or if you need assistance.

Thanks!

PR #2047, opened 2018-01-18
fixed search for comments

ViditChitkara:
updates #1977

Thanks!

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -10,6 +10,7 @@ class UsersController < ApplicationController
 
   def create
     @user = User.new(params[:user])
+    @user.status = 1
     using_recaptcha = !params[:spamaway] && Rails.env == "production"
     recaptcha = verify_recaptcha(model: @user) if using_recaptcha
     @spamaway = Spamaway.new(params[:spamaway]) unless using_recaptcha
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -10,7 +10,6 @@ class UsersController < ApplicationController
 
   def create
     @user = User.new(params[:user])
-    @user.status = 1
     using_recaptcha = !params[:spamaway] && Rails.env == "production"
     recaptcha = verify_recaptcha(model: @user) if using_recaptcha
     @spamaway = Spamaway.new(params[:spamaway]) unless using_recaptcha
--- a/app/models/drupal_user.rb
+++ b/app/models/drupal_user.rb
@@ -47,7 +47,6 @@ class DrupalUser < ActiveRecord::Base
   def moderate
     self.status = 5
     save
-    update_user_status(5)
     # user is logged out next time they access current_user in a controller; see application controller
     self
   end
@@ -55,7 +54,6 @@ class DrupalUser < ActiveRecord::Base
   def unmoderate
     self.status = 1
     save
-    update_user_status(1)
     self
   end
 
@@ -63,7 +61,6 @@ class DrupalUser < ActiveRecord::Base
     self.status = 0
     decrease_likes_banned
     save
-    update_user_status(0)
     # user is logged out next time they access current_user in a controller; see application controller
     self
   end
@@ -72,18 +69,11 @@ class DrupalUser < ActiveRecord::Base
     self.status = 1
     increase_likes_unbanned
     save
-    update_user_status(1)
     self
   end
 
   def email
     mail
-  end
-    
-  def update_user_status(status)
-    u = user
-    u.status = status
-    u.save!
   end
 
   def first_time_poster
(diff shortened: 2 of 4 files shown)


ViditChitkara:
@jywarren The problem was actually not due to comments but the notes method!! The problem regarding user search also seems to be fixed in this patch. Actually the default status for a user is '0' but we are searching for users with status 1. We may want to use DrupalUser instead of User!!
Please correct me if I am wrong! Do you think we could try this in production??

seafr:
The if block (line 16) seems to evaluate to the same thing whether the condition is true or false.

ViditChitkara:
Yes, I was actually trying to remove the failing tests that existed in the previous commit!! Will do the changes asap!! Thanks for the feedback.😊😊

ViditChitkara:
Also I am not really sure if this would work properly in staging. So waiting for Jeff's input on this. Let's see if it works!! Fingers' crossed!!

seafr:
You're welcome. I thought perhaps you have not noticed, but it seems that you have a reason for this. :+1:

jywarren:
Oh, this looks really great, and super teamwork. We are trying to deprecate `DrupalUsers` in favor of `User` however, but we can use `user.username` instead of `drupal_user.name` -- in the `where()` filters. Both should have "status" the same, and i think a swap should work here! Timestamp field is also named differently

Thanks a lot!!

ViditChitkara:
Oh alright, I'll do that!! I have another thing in mind, let's not filter the query using status: 1 (for the time being) since the default status of user is 0 and its not changed (however, status of drupal_user is used to determine a banned user). Let it be:-
`
 User.search(input).limit(limit).order('uid DESC')
`
What do you say?

ViditChitkara:
Or currently just keep the changes to notes and leave user method as it was?

ViditChitkara:
I just realised, the migration thing could also be applied to update the User status (which is 0 by default) in the searches perhaps would solve the problem related to username search! What do you think?

jywarren:
Oh, ack... that's bad -- you're right, on this line, we're still only updating `drupal_user`: https://github.com/publiclab/plots2/blob/master/app/controllers/admin_controller.rb#L108

we need to fix that actually. Can we think about how to address this in this or a separate PR?

jywarren:
Yes, so we could do the migration for user status too, and then try to fix the code that should be using User instead of DrupalUser

jywarren:
:-/ I think there's an advantage to doing only a single migration with both changes in it, since it simplifies deployment. But we'd have to rewire the `ban` method at the same time as we do the migration, otherwise it may go out of sync, right? 

Thanks for thinking through this!!!!

ViditChitkara:
By both changes do you mean migrations related to answer_tag (pr #2045) too? So let's just tackle changes related to notes here and stash user related changes!! I'll open a separate pr for the user_related searches!! or should we do it here only?

jywarren:
> let's just tackle changes related to notes here and stash user related changes!! I'll open a separate pr for the user_related searches!! 

this sounds perfect, thank you!

So then just tell me when this one is ready! :-)

ViditChitkara:
I have removed the changes related to user, please have a look!! It's ready according to me!!

ViditChitkara:
@jywarren made changes regarding user_status, please see!!

jywarren:
Hi! I see there are still a few changes to user status in the PR -- I'm just not sure about it so perhaps we can remove those too and stash them for a later PR -- I did gather up some links related to that topic though:

https://github.com/publiclab/plots2/blob/master/db/schema.rb.example#L344

https://github.com/publiclab/plots2/blob/ed1ef3b72c8a45bfd1d7b842d1d747db391b2a7f/app/models/drupal_user.rb#L4-L7

I think we need to restructure that system completely around User and not DrupalUser, including:

1. [ ] a migration to copy all `DrupalUser` status records to `User` status records
2. [ ] ensure (as you do in the extra code in this PR) that banning/unbanning affects the `User` records
3. [ ] ensure (as you do in the extra code in this PR) `User` status is properly set on creation
4. [ ] ensure all references to `status` are to the newer `user.status`
5. [ ] leave some decent documentation on what the status codes are and how they work
6. [ ] in yet another follow-up PR, stop using the `DrupalUser` status column and remove it

We can copy this checklist into a new issue!

Thanks for your work on this - the status system needs some care and I appreciate your attention to it!

ViditChitkara:
All right. let's do that in a new issue!!
So in this pr we remove all changes related to user and let's just keep the changes related to node search. Also let's keep the migrations in #2045 confined to add 'answered' tag only and remove user related changes. @jywarren please correct me if I am wrong!!

ViditChitkara:
made the changes, let's see if it works or not!

jywarren:
Great, i'll try to publish these changes later today. Thanks!

jywarren:
Actually i'm pushing them to staging.publiclab.org now -- you'll be able to see them in that preview mode shortly.

ViditChitkara:
Cool!!

PR #2055, opened 2018-01-20
contributor count on tags page fixed

### LINKED ISSUE #2042: List wiki revision authors as contributors on /contributors/tagname pages

I noticed that "Nova" just contributed a revision to the water sensors wiki page: https://publiclab.org/water-sensors

However, they don't appear on the contributors page for that tag: https://publiclab.org/contributors/water-sensors

How hard would it be to add all revision authors to that query? 

https://github.com/publiclab/plots2/blob/5bd6efacc416dc2a1b1235b386c44babe8602b72/app/controllers/tag_controller.rb#L315-L325

Could we make an implementation closer to how `contributor_count` works, in the Tag model?

https://github.com/publiclab/plots2/blob/5bd6efacc416dc2a1b1235b386c44babe8602b72/app/models/tag.rb#L55-L66

Could this implementation also live in the Tag model, and even use the same code as `contributor_count`?

Any help appreciated!

ViditChitkara:
fixes #2042

Thanks!

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -323,7 +323,7 @@ class TagController < ApplicationController
                  .references(:term_data, :node_revisions)
                  .where('term_data.name = ?', params[:id])
                  .order('node_revisions.timestamp DESC')
-    @users = User.where(id: Tag.contributors(@tagnames[0]))
+    @users = Tag.contributors(@tagnames[0])
   end
 
   # /contributors
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -113,7 +113,7 @@ class TagController < ApplicationController
                 .includes(:tag)
                 .references(:term_data)
                 .where('term_data.name = ?', params[:id])
-    @length = notes.collect(&:uid).uniq.length || 0
+    @length = Tag.contributor_count(params[:id]) || 0
 
     respond_with(nodes) do |format|
       format.html { render 'tag/show' }
@@ -323,7 +323,7 @@ class TagController < ApplicationController
                  .references(:term_data, :node_revisions)
                  .where('term_data.name = ?', params[:id])
                  .order('node_revisions.timestamp DESC')
-    @users = @notes.collect(&:author).uniq
+    @users = DrupalUser.where(uid: Tag.contributors(@tagnames[0]))
   end
 
   # /contributors
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -323,7 +323,7 @@ class TagController < ApplicationController
                  .references(:term_data, :node_revisions)
                  .where('term_data.name = ?', params[:id])
                  .order('node_revisions.timestamp DESC')
-    @users = DrupalUser.where(uid: Tag.contributors(@tagnames[0]))
+    @users = User.where(id: Tag.contributors(@tagnames[0]))
   end
 
   # /contributors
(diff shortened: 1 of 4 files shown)


ViditChitkara:
@jywarren please see if this solves the problem!

app/controllers/tag_controller.rb
```diff
@@ -321,7 +321,7 @@ def contributors
                  .references(:term_data, :node_revisions)
                  .where('term_data.name = ?', params[:id])
                  .order('node_revisions.timestamp DESC')
-    @users = @notes.collect(&:author).uniq
+    @users = DrupalUser.where(uid: Tag.contributors(@tagnames[0]))
```
jywarren: Is it possible to use `User.where(id: ...` here instead, since we're trying to stop using DrupalUsers? Thank you! This otherwise looks fantastic.

app/models/tag.rb
```diff
@@ -52,8 +52,9 @@ def belongs_to(current_user, nid)
     node_tag && node_tag.uid == current_user.uid || node_tag.node.uid == current_user.uid
   end
 
-  def self.contributor_count(tagname)
+  def self.contributors(tagname)
```
jywarren: Love this!!! 👍 👍

app/controllers/tag_controller.rb
ViditChitkara: Hi, I guess we'd have to refactor the Tag.contributors method then, which currently  gives  array of uid (drupal_user id)!! Let's give it a try.

ViditChitkara: @jywarren I guess we need to write a conversion function in user mode which fetches user_ids against drupal_user_ids. Then we could replace the DrupalUser with User in L324. Do you think this is the correct way, however it doesn't completely rules out DrupalUser.
The function would look something like (Would go in User Model):-
```
def self.user_ids(drupal_user_ids)
  user_ids=[]
  drupal_users = DrupalUser.include(:user).where(uid: drupal_user_ids)
  drupal_users.each do |du|
    user_ids<<du.user.id
  end
end
```
This would greatly increase the overload time since a separate query is required to fetch a user from drupal_user. Do you think we should do this???

jywarren: So, we've synced `User.id` and `DrupalUser.uid` and you can treat one like the other -- we did a big database migration a while back to ensure this. Does that help?

ViditChitkara:
Sure, I'll do!

ViditChitkara:
Done with the tests!

app/models/tag.rb
```diff
@@ -62,6 +63,11 @@ def self.contributor_count(tagname)
       uids+=n.revision.collect(&:uid)
     end
     uids = uids.uniq
+    uids
```
jywarren: Ah, so, do you think we should return an ActiveRecord collection here, instead of just uids? Like, replace this line with `User.find uids` instead -- and then the line above (https://github.com/publiclab/plots2/pull/2055/files#diff-a8eb2425ebeb0814d010864bb014e815R326) would be simpler -- make sense?

ViditChitkara: Yes, makes sense, will do!!

app/views/tag/contributors.html.erb
```diff
@@ -68,7 +68,7 @@
 
   <% end %>
 
-  <% if @notes.nil? || @notes.length == 0 %>
+  <% if @users.nil? || @users.length == 0 %>
```
ViditChitkara: Found some unexpected behaviour when @notes.length was 0 but @users still existed hence modified the condition a bit!!

app/models/tag.rb
ViditChitkara: @jywarren done with the suggested changes!

jywarren:
Super, is this ready to go, then? Thanks!

ViditChitkara:
Yes!!

jywarren:
Awesome!

PR #2058, opened 2018-01-21
List activities for subscriptions/digest

### LINKED ISSUE #2027: Display "activity from past day" in a list at /subscriptions/digest
We have a new feature in the User model called `user.content_followed_in_past_period(time_period)`

https://github.com/publiclab/plots2/blob/c6b2d3460d1fb9da28ce5e1cc0944daf690f9415/app/models/user.rb#L313-L315

Let's adjust this to be `user.content_followed_in_period(start, end)` so we can use it like:

`user.content_followed_in_period(Time.now - 1.day, Time.now)`

And make it not use `self.node.where()` but `Node.where()` so it returns /all/ content, not just the user's own content -- adjusting both the test and the model function, and the test description on this line:

https://github.com/publiclab/plots2/blob/3db7e63c37d10516576a4cf585e1d05584c06682/test/unit/user_test.rb#L92-L96

****

### Follow-up task

Then, let's make a new controller method in https://github.com/publiclab/plots2/blob/master/app/controllers/subscription_controller.rb to display this content. It can just be in a table, like this template:

https://github.com/publiclab/plots2/blob/master/app/views/wiki/_wikis.html.erb

I'm not sure, but it may be possible to simply render that template with a collection, so the overall `/app/views/subscriptions/digest.html.erb` file could be just:

```ruby
<h2>Content you've subscribed to in the past week</h2>

<%= render partial: "wiki/wikis" %>
```

And the controller method could be:

```ruby
# display the past day's followed nodes on one page
def digest
  # note these won't be wikis but we'll use the wiki partial template to display a list of nodes
  @wiki = current_user.content_followed_in_period(Time.now - 1.week, Time.now)
end
```

This may not work on the first try, but I'm happy to help troubleshoot!

****

### Follow-up task

Then, let's make a method to send the digest in an email!

Gauravano:
Fixes #2027


jywarren:
This looks fantastic, and is so exciting to see! This'll be a really cool page. 

Can we add a simple functional test to it, though? That way we know the page renders in tests. 

Thank you!!!

Gauravano:
Hey, @jywarren I have a doubt regarding this, kindly see https://github.com/publiclab/plots2/issues/2027#issuecomment-359258980 .Thanks.

jywarren:
I agree. We want all nodes which the user is subscribed to AND which have changed in that time period. 

This doesn't quite work for that:

https://github.com/publiclab/plots2/blob/93358c4b3aa38513e00be226800aa0194a6dce7d/app/models/user.rb#L314

I think we need to wrap this function:

https://github.com/publiclab/plots2/blob/93358c4b3aa38513e00be226800aa0194a6dce7d/app/models/user.rb#L169-L174

How's that sound? Good catch on that method, and if we change it it may also affect the unit test which we could update to match. Thanks so much!

Gauravano:
Hi @jywarren, the current method is working but I found something strange. node.changed time is different from node.latest.created_at i.e., from its latest revision.Could you suggest why is it so ? Also, this problem is highlighted from the view as view is using node.latest.created_at time and I have used node.changed time:-
activity

jywarren:
Ah, i see. Yes, the two can drift, which is a problem. I think we should be sure to update `.changed` on these lines:

https://github.com/publiclab/plots2/blob/master/app/models/node.rb#L533-L539

Be sure to change `changed` using this syntax (below) since `changed` is a reserved word in Rails, and this field dates from pre-Rails code: 

https://github.com/publiclab/plots2/blob/ceb7d05404b4619c7fa8cd8cae5e8215798cb02c/app/models/node.rb#L111

However, you can also change your query to use `revision.update_at` by adding these lines to the query:

```ruby
                 .includes(:revision, :tag)
                 .references(:node_revision)
```

As on these lines:

https://github.com/publiclab/plots2/blob/ceb7d05404b4619c7fa8cd8cae5e8215798cb02c/app/controllers/wiki_controller.rb#L370-L374

How does that sound? 

Thank you!!!

Gauravano:
Actually, @jywarren,node.changed is showing latest time but time shown by revisions is not updated, that's issue.I think its related to #2092.Anyways I'll  try running some more tests and thanks for helping Jeff.

jywarren:
Hmm, revisions should be auto set because rails auto sets some timestamp fields. But this is a legacy table so maybe not. The revisions should be getting timestamps in node.new_revision in node.rb. Hope that helps!

Gauravano:
Hi @jywarren, I have displayed node.changed time on digest page as we are comparing with it only in the controller , removing all conflicts.Please review it.Thanks.

jywarren:
OK, this is looking really fantastic. Can we add a functional test for the `digest` action, though, just to be sure that doesn't break? 

Are you limiting ability to view to only people logged in, with `require_user`? 

Awesome!

app/controllers/subscription_controller.rb
```diff
@@ -5,7 +5,7 @@
 class SubscriptionController < ApplicationController
 
   respond_to :html, :xml, :json
-  before_filter :require_user, :only => [:create, :delete, :index]
+  before_filter :require_user, :only => [:create, :delete, :index, :digest]
```
Gauravano: Yes @jywarren here is the line for checking current_user

Gauravano:
> Are you limiting ability to view to only people logged in, with require_user?
Yes @jywarren its already handled

Gauravano:
Hey, @jywarren I have added the functional tests.here's preview : 

weekly_digest

Don't you think we need to change the page heading as we are showing users content that was changed or created with the tag they are subscribed to?

jywarren:
This looks great. Yes, sorry i guess we should explain more what the content is -- agreed. Maybe a new PR for that? 

Awesome!

jywarren:
"Content from the past week on topics you've subscribed to"?

Gauravano:
Sounds good ...

Issue #2028, opened 2018-01-17
send email to newly added coauthors explaining what it means


jywarren:
This could actually be relatively simply achieved via a comment -- by automatically leaving a comment as we do for the barnstar feature:

https://github.com/publiclab/plots2/blob/dccf61b4cf0aa816a3a906495ea6a0b2d5d847b0/app/controllers/tag_controller.rb#L178-L180

The comment could say `@_author_ has marked @_recipient_ as a co-author.` and everybody on the thread would get email notified.

This code would go just before the last line of this block:

https://github.com/publiclab/plots2/blob/033676b8abd28cd56a999e60934ff9f01bd0a9ac/app/models/node.rb#L660-L667

and the message could look like:

```ruby
"_@#{user.username} has marked @#{node.author.username} as a co-author._"
```

This should work but could take some troubleshooting -- happy to help, and any assistance in taking this task on is appreciated!

Gauravano:
Can I take this issue ?

jywarren:
that'd be super. Thank you!

Gauravano:
Hey @jywarren, the method in the tag_controller for solving this issue would work similar to method barnstar.The barnstar method is called by the form which is there in 'views/like/_like' but for making co-author there is no form present, we make co-author as for now by adding tags only, so should I add if statement in create method of tag_controller and redirect to separate method for handling creation of with:user tag or should I make a new form to trigger the method for this issue which would then perform comment creation part ?

Gauravano:
Also, presently automatic comment creation for barnstar take place if the barnstar is generated via Award barnstar form and not by just adding power tag [barnstar: type], so we can also make changes so that comment creation could take place if someone has added barnstar tag.What do you think @jywarren?

jywarren:
ah, i see. Yes, i think perhaps we should refactor these to trigger on the tag creation, not on the form submission, for both barnstars and co-authorship. 

Do you want to take a crack at this? Thank you!

jywarren:
Just watch out because you can't access `current_user` in the models -- they don't "know" about the user session. But we should be able to use the tag creator's uid, right?

Gauravano:
I will take care of this.Thanks, Jeff.

jywarren:
awesome!!!

Gauravano:
> The comment could say @_author_ has marked @_recipient_ as a co-author. and everybody on the thread would get email notified.

[...]

PR #2062, opened 2018-01-22
Contributor tab design consistent with other tabs on Tag page

### LINKED ISSUE #1736: apply same button treatment to all tabs on tag pages
hello everyone!

I noticed that this page https://publiclab.org/tag/pesticides has a very nice display of buttons for "Follow [topic]" "Ask a question about [topic]" and "Subscribe to answer questions on [topic]" that is consistent across most tabs (`Research Notes`, `Questions`, `Wiki pages`, `Maps` ):

screen shot 2017-10-27 at 8 59 37 am

....but not all tabs -- `Contributors` is very different, and I believe this may be an oversight:

screen shot 2017-10-27 at 8 59 20 am

Do we think this is fixable? Thanks!

Gauravano:
Fixes #1736


Gauravano:
@ebarry @jywarren  this is the updated view 

contributor_update1

ebarry:
wow awesome! 
Do you like the line of stats shown on the Contributor tab? 

I admit I got a feeling like it was a little repetitive to see the number of research notes above the contributors,.... what do you all think?

Gauravano:
Here comes 2nd view,

contri_update2

Gauravano:
I also think we should go with the 2nd view as the 1st view is not matching with tabs it is taking more space.Also, previous contributor's page route is still active https://publiclab.org/contributors/spectrometer, I haven't changed it, I have just changed view how we see contributor's page when visited through navbar.

jywarren:
I like the 1st view as we're trying to develop a consistent 3-column "at a glance" display which is more readable than the dense text description we've had on pages like this in the past. But we can circle back to it in a later issue so let's not let that hold this back.

I also wanted to preserve the route: https://publiclab.org/contributors/pm

because it may be linked to. We could redirect in the routes file:

https://github.com/publiclab/plots2/blob/ed1ef3b72c8a45bfd1d7b842d1d747db391b2a7f/config/routes.rb#L42

This is looking great, thanks!

Gauravano:
@jywarren https://publiclab.org/contributors/pm route is preserved as I have created an additional partial file which I am rendering at http://localhost:3000/contributors/tag/barnstar:basic so I guess after some code cleanup this is good to go.So, which view should I finally implement?

ebarry:
to reduce dense text, i can agree on view 1.
but is there an extra line pushing the content farther down on the page
than need be?

--

+1 336-269-1539 / @lizbarry <http://twitter.com/lizbarry> / lizbarry.net

Gauravano:
How about this ?
update3

jywarren:
By the way, awesome work on this @Gauravano -- it'll be great to close this one out. 👍

Gauravano:
Actually @jywarren the gap is due to <hr /> which is added after the end of tab content as the 
grid has 
`class = col-md-9` which is causing the line to come after.I have done some changes.

updatee

Gauravano:
Hey @jywarren its ready to merge.Also, plz check https://github.com/publiclab/plots2/issues/2028#issuecomment-359058548.Thanks.

app/views/tag/_contribute.html.erb
```diff
@@ -0,0 +1,36 @@
...
+            <div class="row" id="highlight" style=" text-align: center;">
+              <h3 class="col-md-4 col-xs-12 col-sm-4"><%= @users.length || 0 %> contributors</h3>
+              <h3 class="col-md-4 col-xs-12 col-sm-4"><%= @notes2.length %> research notes</h3>
+              <h3 class="col-md-4 col-xs-12 col-sm-4"><%= @wikis.length %> wiki pages</h3>
+            </div>
+        <% end %>
+
+    <% end %>
+
+    <% if @notes2.nil? || @notes2.length == 0 %>
```
jywarren: This looks great, thank you! 

I'm wondering if we could pull out this section or even more, which is duplicated with this:

https://github.com/publiclab/plots2/blob/ceb7d05404b4619c7fa8cd8cae5e8215798cb02c/app/views/tag/contributors.html.erb#L71-L89

And use a partial to provide both? Like, `/app/views/tag/_contributors.html.erb` -- using `<%= render partial: "tag/contributors" %>`? What do you think?

Gauravano: Good idea !!

Gauravano:
Hey, @jywarren I have done changes in partial _contribute.html.erb only such that it shows different view for both routes contributors/basic and contributors/tag/basic.Here are the final views:

share

jywarren:
I think this is super. Would it be possible to rename `_contribute.html.erb` to `_contributors.html.erb` and then we should be ready to go here? 

Just to clarify, there will now be two ways to see this page -- at `/contributors/TAGNAME` and `/tag/TAGNAME` under the "contributors" tab --- but we are largely using the same code to drive both?

app/controllers/tag_controller.rb
```diff
@@ -114,6 +115,16 @@ def show
                 .where('term_data.name = ?', params[:id])
     @length = Tag.contributor_count(params[:id]) || 0
 
+    # fetching nodes with revisions for contributor tab data
+    @tagnames = [params[:id]]
+    @tag = Tag.find_by(name: params[:id])
+    @notes2 = Node.where(status: 1, type: 'note')
```
jywarren: Here, do we need to name this `@noteCount` and pass on the `.count` of this rather than the whole notes list?

Let's also leave a comment here to `# later, move this into a method in the Tag model like Tag.taggedNodeCount(tagname)` -- what do you think? could be a modification of this maybe?

https://github.com/publiclab/plots2/blob/master/app/models/tag.rb#L85

Gauravano: I have made function also as you stated, and other changes are also done.

ebarry:
exciting!

Gemfile.lock
```diff
@@ -366,7 +366,7 @@ DEPENDENCIES
   mustermann (~> 0.4)
   mysql2 (~> 0.3.20)
   nifty-generators
-  paperclip (< 4.2.0)
+  paperclip (~> 4.2.2)
```
jywarren: Changes look great! But we just bumped this version -- can you revert this change? Thank you!

Gauravano: done @jywarren

jywarren:
Fantastic! Thank you @Gauravano -- well done!

jywarren:
Hi, @Gauravano - it looks like the contributors page is not quite working correctly -- see:

* https://publiclab.org/contributors/soc-2018-proposals
* https://publiclab.org/tag/soc-2018-proposals

The tab seems to link to the page, which now shows no contributors listed. I think there may have been a reversion in https://github.com/publiclab/plots2/pull/2180 which affected your code here. Following up in #2444 -- any help appreciated!

Gauravano:
hi @jywarren , I will look into it.Thanks.

PR #2066, opened 2018-01-22
Added Turbolinks to rails app for faster loading of Web Pages

### LINKED ISSUE #2064: Add Turbolinks for faster loading

We can add turbolinks for faster loading of our web-pages.
Turbolinks can make a  Rails app feel faster by using JavaScript to replace the page content when clicking a link. 

@jywarren What do you think of this?

namangupta01:
Closes #2064 

Thanks!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,9 @@ gem 'passenger'
 gem 'rails-i18n', '~> 4.0.0'
 # gem 'responders', '~> 2.0' # for Rails 4.2
 
+gem 'turbolinks'
+
+
 # run with `bundle install --without production` or `bundle install --without mysql` to exclude this
 group :mysql, :production do
   gem 'mysql2', '~> 0.3.20'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -311,6 +311,9 @@ GEM
     timecop (0.9.1)
     tins (1.16.3)
     tool (0.2.3)
+    turbolinks (5.1.0)
+      turbolinks-source (~> 5.1)
+    turbolinks-source (5.1.0)
     tzinfo (1.2.4)
       thread_safe (~> 0.1)
     uglifier (4.1.3)
@@ -391,6 +394,7 @@ DEPENDENCIES
   test-unit
   therubyracer
   timecop
+  turbolinks
   uglifier (>= 1.0.3)
   will_paginate (>= 3.0.6)
   will_paginate-bootstrap (>= 1.0.1)
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -49,3 +49,5 @@
 //= require short-code-forms/dist/short-code-prompts.js
 //= require wikis.js
 //= require chart.js/dist/Chart.js
+//= require turbolinks
+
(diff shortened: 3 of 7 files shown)


jywarren:
You'll be able to follow the build here: http://jenkins.laboratoriopublico.org/

and it'll finish up here: unstable.publiclab.org

jywarren:
I tried this out on http://unstable.publiclab.org and it's impressive and things seem to work. But I want to be cautious. I found some articles on "gotchas" we may want to consider:

* It can **be disabled for some pages** if we run into trouble: https://stackoverflow.com/questions/22326473/rails-4-disable-turbolinks-in-a-specific-page
* **Event handling in JavaScript** with Turbolinks: https://www.iexposure.com/industry-insights/rails-4-turbolinks-and-delegated-event-handling-in-jquery (not sure if we need to implement anything?)
* **Google Analytics** -- ours is in `setup.js` -- not in the body tag. Are we still tracked? We'll have to test this after deploying. 
* does `jQuery.ready()` work normally on page load? https://stackoverflow.com/questions/18770517/rails-4-how-to-use-document-ready-with-turbo-links

The last one seems relevant. But I can't find any pages which don't work ... yet. Sorry to be slow on this, i just want to be thorough! @icarito do you know about this Turbolinks system, which replaces the `<body>` of the page but not the `<head>` to save load time?

icarito:
I did check it out as soon as you published the staging unstable branch.

I was impressed! It is fast and feels like a great improvement. I had
seen this in Single-Page-Application style apps, but didn't know it was
available for rails.

Beyond what you listed I would check if it degrades gracefully for
browsers which don't support this or otherwise have Javascript disabled
or limited (not sure how we are doing otherwise in this area).

It is also one more layer/moving part, but a really cool one, +1 from me.//

namangupta01:
@jywarren  `jQuery.ready()` will work for normally when the page is loading for the first time but after that `jQuery.ready()` will not be called but we can use `document.addEventListener("turbolinks:load", function() {
  // your code here
})`.

For **Google Analytics** we can also use the same function on `turbolinks:load`

jywarren:
Here's where we do google analytics -- maybe we could name this function `function googleAnalyticsInit()` and run it here but also on `turbolinks:load`? 

https://github.com/publiclab/plots2/blob/ceb7d05404b4619c7fa8cd8cae5e8215798cb02c/app/assets/javascripts/setup.js#L64

namangupta01:
@jywarren I have updated the google-analytics compatibility with turbolinks. Please have a look!

jywarren:
Oh, interesting -- have you updated our google analytics code itself to the most recent? Would you like to delete the now-commented code? 

I think we should give this a go! And just be extra aware for bugs coming up if we go forward. Maybe we can alert all reviewers when it goes live in case we have to do any followup fixes. You never know!!

namangupta01:
@jywarren yes we can!
And can we keep the both commits without squashing in master branch? because it will help us in future to rectify bugs in case we got errors.

namangupta01:
@jywarren i think its ready now!

jywarren:
see https://github.com/publiclab/plots2/issues/2064 and https://github.com/publiclab/plots2/issues/2132 for next steps

PR #2067, opened 2018-01-22
Fix for #2045 answered migration

jywarren:
fixes #2045se alert developers on plots-dev@googlegroups.com when your request is ready or if you need assistance.


ViditChitkara:
That was built successfully! :+1: But I do not see the 'answered' tag on the accepted questions. Does it take some time to update the db??

PR #2085, opened 2018-01-24
Added the richWiki view

### LINKED ISSUE #2083: Add rich editor to New Wiki form
This is a multi-step issue to add the rich text editor to the new wikis page.

The form for making new wiki pages (for example if you're logged in, at https://publiclab.org/wiki/new) is the old editor -- we should replace it (for new pages only!) with the rich editor, as you see at http://publiclab.org/post (again, if logged in)

The rich editor uses this template:

https://github.com/publiclab/plots2/blob/master/app/views/editor/rich.html.erb

We could copy that to one called `wikiRich.html.erb` in the same directory. 

Then we'd need to adapt the wiki controller to use the new template:

https://github.com/publiclab/plots2/blob/master/app/controllers/wiki_controller.rb#L151

Here's the corresponding section for the main rich editor, where the template is used: https://github.com/publiclab/plots2/blob/ceb7d05404b4619c7fa8cd8cae5e8215798cb02c/app/controllers/editor_controller.rb#L39-L49 (this could be copied or adapted for our use).

This is a few steps but I'm happy to help guide someone through! Thanks!

(solves #1475)

SidharthBansal:
Fixes #2083 
- [x] Make a copy of https://github.com/publiclab/plots2/blob/master/app/views/editor/rich.html.erb to one called wikiRich.html.erb in the same directory.
- [x]  Adapt the wiki controller to use the new template
- [x] Write tests

### DIFF
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -156,6 +156,7 @@ class WikiController < ApplicationController
           node = Node.find(params[:n])
           params[:body] = node.body if node
         end
+        flash[:error] = "Please enter both body and title"
         render template: 'editor/wikiRich'
       end
     else
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -148,7 +148,14 @@ class WikiController < ApplicationController
         end
         redirect_to @node.path
       else
-        render action: :edit
+        flash.now[:notice] = "This is the new rich editor. For the legacy editor, <a href='/post?#{request.env['QUERY_STRING']}&legacy=true' class='legacy-button'>click here</a>."
+        if params[:main_image] && Image.find_by(id: params[:main_image])
+          @main_image = Image.find_by(id: params[:main_image]).path
+        end
+        if params[:n] && !params[:body] # use another node body as a template
+          node = Node.find(params[:n])
+          params[:body] = node.body if node
+        end
       end
     else
       flash.keep[:error] = I18n.t('wiki_controller.you_have_been_banned').html_safe
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -148,7 +148,7 @@ class WikiController < ApplicationController
         end
         redirect_to @node.path
       else
-        flash.now[:notice] = "This is the new rich editor. For the legacy editor, <a href='/post?#{request.env['QUERY_STRING']}&legacy=true' class='legacy-button'>click here</a>."
+        flash.now[:notice] = "This is the new rich editor. For the legacy editor, <a href='/wiki/new?legacy=true' class='legacy-button'>click here</a>."
         if params[:main_image] && Image.find_by(id: params[:main_image])
           @main_image = Image.find_by(id: params[:main_image]).path
         end
@@ -156,6 +156,7 @@ class WikiController < ApplicationController
           node = Node.find(params[:n])
           params[:body] = node.body if node
         end
+        render template: 'editor/wikiRich'
       end
     else
       flash.keep[:error] = I18n.t('wiki_controller.you_have_been_banned').html_safe
--- a/app/views/editor/wikiRich.html.erb
+++ b/app/views/editor/wikiRich.html.erb
@@ -21,30 +21,7 @@
 
   <div class="ple-header">
 
-    <h1>Share</h1>
-
-    <!-- selector module -->
-    <div class="container ple-type">
-
-      <div class="col-md-5 col-md-push-4">
-        <div class="row">
-          <div class="col-xs-10">
-            <select class="form-control input-lg">
-              <option selected value="">a research note</option>
-              <option value="" <% if params[:tags] && params[:tags].include?('activity') %>selected='selected'<% end %>>a step-by-step activity</option>
-              <option value="" <% if params[:tags] && params[:tags].include?('question') %>selected='selected'<% end %>>a question or request</option>
-              <option value="">an event</option>
-            </select>
-          </div>
-          <div class="col-xs-2" style="padding-top:10px;text-align:left;">
-            <a rel="tooltip" data-placement="bottom" title="Learn more about different posting types." class="ple-help-light" style="font-size:18px;"><i class="fa fa-question-circle"></i></a>
-
-          </div>
-        </div>
-      </div>
-
-    </div>
-    <!-- end selector module -->
+    <h1>New wiki page</h1>
 
   </div>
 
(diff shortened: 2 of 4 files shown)


SidharthBansal:
@jywarren  how shall I proceed in this?

app/controllers/wiki_controller.rb
```diff
@@ -148,7 +148,14 @@ def create
         end
         redirect_to @node.path
       else
-        render action: :edit
+        flash.now[:notice] = "This is the new rich editor. For the legacy editor, <a href='/post?#{request.env['QUERY_STRING']}&legacy=true' class='legacy-button'>click here</a>."
```
jywarren: Great -- here, we'll want to link to `/wiki/new?legacy=true` and then have a conditional letting people use the original template, on the line above. Make sense?

app/views/editor/wikiRich.html.erb
```diff
@@ -0,0 +1,296 @@
...
+          }
+        }
+      }
+    });
+
+  })();
+
+  $('.legacy-button').click(function onLegacyClick(event) {
+      event.preventDefault();
+      window.location = "/post?legacy=true&body=" + editor.richTextModule.value() + "&tags=" + editor.tagsModule.value() + "&title=" + editor.titleModule.value();
```
jywarren: Here too, switch to legacy via `/wiki/new?legacy-true ...`

app/views/editor/wikiRich.html.erb
```diff
@@ -0,0 +1,296 @@
...
+
+<link href="/lib/publiclab-editor/dist/PublicLab.Editor.css" rel="stylesheet">
+
+<script src="/lib/publiclab-editor/dist/PublicLab.Editor.js"></script>
+
+<div class="pl-editor">
+
+  <div class="ple-header">
+
+    <h1>Share</h1>
```
jywarren: Here we can say "New wiki page"

app/views/editor/wikiRich.html.erb
```diff
@@ -0,0 +1,296 @@
...
+
+    <!-- selector module -->
```
jywarren: lets remove the selector module

SidharthBansal:
@jwarren I have updated the existing functional test to use the rich wiki template instead of the edit template which demonstrates that we are not rendering the controller's edit action. I don't think we need an additional test to demonstrate that it is not the edit action.

test/integration/wiki_creation_test.rb
```diff
@@ -36,7 +36,7 @@ class WikiCreationTest < ActionDispatch::IntegrationTest
 
     assert_nil node
     assert_equal '/wiki/create', path
-    assert_select 'h2', '1 error prohibited this revision from being saved'
```
SidharthBansal: When I am adding this the tests are failing. I can't find out why this is happening. So, I have deleted the test in red and all tests are passing. On the localhost, it is displaying a red rectangle without text if I try to submit wiki page with no body or no title.

jywarren: Hmm, you mean it's showing a red rectangle without text and your PR fixes that, or that's a remaining problem the PR doesn't get to? 

Perhaps we can open a new issue with a sub-issue as described here.

jywarren: Could you post a screenshot of what you see from the code in this PR? Thank you! I just want to be careful with this one.

SidharthBansal: Yeah, we can open an issue for the red rectangle.

SidharthBansal:
image

SidharthBansal:
One more thing, sorry to be too critical of the website. 
image
Can you see that we have Getting started on both footer and the left sidebar? Would not be it better to remove the redundancy? If later on we are not using it, then there is no need of removing it. 
Also, it is very near to the footer so the user will be able to see both on the same page simultaneously.

jywarren:
No need to apologize -- there are lots of places we need to improve! This is a good one to open an issue for. 

> A red rectangle ie a flash message without an error.

Do you think we can modify the logic of the alert display to show a correct error message when this happens?

> Now it is rendering the new rich wiki page if something wents wrong in the old one. 

Ah, i see. Perhaps when we submit from a legacy page we can send the `legacy` parameter so we error to the correct page? Or would it be simpler to try to do some client-side validations and handle the errors locally? That sounds harder to me (double validation check, locally and remotely). What do you think? 

> Also autosaving the things. Is it what we are trying to achieve?

Not sure -- what do you mean autosaving? Thanks!

> Shouldn't be it better to directly link the url to the new rich wiki view?

Do you mean possibly just abandoning the legacy form? I worry that if people have trouble with the new one, which does have some bugs, we need to allow them to use the original form. 

Thanks for thinking through this one carefully! Let's get this in as good a state as we can, and tackle subsequent issues as separate PRs. But I do want to ensure there are no missing or confusing error messages if we merge this. Thanks!

SidharthBansal:
The rich wiki is working fine for not supplying any body or any title. It is showing us the error messages correctly. Only the problem is the errors are not propagating to the richWiki Page from the legacy wiki.
> Do you mean possibly just abandoning the legacy form? I worry that if people have trouble with the new one, which does have some bugs, we need to allow them to use the original form.

Yeah, I mean that. But you are right. There are a few issues which I have encountered with that.

>Ah, i see. Perhaps when we submit from a legacy page we can send the legacy parameter so we error to the correct page? Or would it be simpler to try to do some client-side validations and handle the errors locally? That sounds harder to me (double validation check, locally and remotely). What do you think?

I will try to do it with a shared partial folder which needs to be made for many stuff which we are using at multiple places.

SidharthBansal:
https://github.com/SidharthBansal/plots2/blob/4c299e0f5dadd93503b71efe1e8cbf1dd267ed0b/app/views/editor/wikiRich.html.erb#L53-L60
these are the lines of code which were creating an empty flash messages' red rectangle.
I have removed it. Now everything is working fine. 

It is a big commit and I would like to suggest that if there are big prs....like this .... then it is better to get review from @jywarren  as @jywarren knows the most of the public lab software instead of anyone else. The reason is simple, they are not familiar with all the code in public lab but you are. So, how shall we able to discriminate these kind of prs from rest by labels????

One more thing, shall I also delete these lines https://github.com/SidharthBansal/plots2/blob/4c299e0f5dadd93503b71efe1e8cbf1dd267ed0b/app/views/editor/wikiRich.html.erb#L53-L60
in the view from which I have copied it from in a new pr??

SidharthBansal:
Travis test cases were failing so closed and reopened it. Everything is working fine on my local

SidharthBansal:
@publiclab/reviewers @jywarren can you help me with this please? I can't detect the error. All tests are passing on my local machine

Gauravano:
hi @SidharthBansal , sorry for late reply, this issue happens when your branch is detached from the parent branch like when one uses git checkout [new-branch-name] to make a branch and doesn't include "-b". The easiest solution is to copy the changes to the new branch and raise new PR. Or, you can cherry-pick the commits too.

SidharthBansal:
@Gauravano thanks for your help. I have a query to the help you posted. 
You are saying I created the branch without -b ,if it was so why did I not get the travis error initially? I just got the error after fixing patches as told by jeff sir

Gauravano:
making branch without "-b" is one of the reason. I have also faced this error like your's even when travis build passed for earlier version of PR.

SidharthBansal:
Ok @jywarren @Gauravano moving to #2516 hope that works

PR #2087, opened 2018-01-24
List people by profile tag at /people/TAGNAME

### LINKED ISSUE #2071: List people by profile tag at /people/TAGNAME
We have a great list of people over at http://publiclab.org/people

But now we have profile tags -- see the bottom of: https://publiclab.org/profile/warren

We should be able to view people by what profile tags they've added. So, for example:

`https://publiclab.org/people/air-quality` - for people who've added the tag `air-quality` to their profile.

This could be achieved through a variation of this method, in the Users controller:

https://github.com/publiclab/plots2/blob/41fe0807cf39f05f2a75183313c67c946859510d/app/controllers/users_controller.rb#L83-L100

To collect such users, we'd want to add to routes.rb: https://github.com/publiclab/plots2/blob/master/config/routes.rb

And in the Users controller, we'd collect users by their profile tag using:

```ruby
User.find(UserTag.where(value: params[:tagname]).collect(&:uid))
```

Does this make sense? 

Any help appreciated!

This will support https://github.com/publiclab/plots2/issues/317 completion too.

rishabhptr:
Fixes #2071 

Thanks!

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -101,21 +101,26 @@ class UsersController < ApplicationController
 
     # allow admins to view recent users
     if params[:id]
-      @users = DrupalUser.joins('INNER JOIN rusers ON rusers.username = users.name')
-                          .order(order_string)
-                          .where('rusers.role = ?', params[:id])
-                          .where('rusers.status = 1')
-                          .page(params[:page])
+      @users = User.order(order_string)
+                    .where('rusers.role = ?', params[:id])
+                    .where('rusers.status = 1')
+                    .page(params[:page])
+    
+    elsif params[:tagname]
+      @users = User.where(id: UserTag.where(value: params[:tagname]).collect(&:uid))
+                    .page(params[:page])
+
     else
       # recently active
-      @users = DrupalUser.select('*, MAX(node.changed) AS last_updated')
-                          .joins(:node)
-                          .group('users.uid')
-                          .where('node.status = 1')
-                          .order(order_string)
-                          .page(params[:page])
+      @users = User.select('*, rusers.status, MAX(node.changed) AS last_updated')
+                    .joins(:node)
+                    .group('rusers.id')
+                    .where('node.status = 1')
+                    .order(order_string)
+                    .page(params[:page])
     end
-    @users = @users.where('users.status = 1') unless current_user && (current_user.can_moderate?)
+
+    @users = @users.where('rusers.status = 1') unless current_user && (current_user.can_moderate?)
   end
 
   def profile
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -89,6 +89,14 @@ class User < ActiveRecord::Base
     DrupalUser.find_by(name: username)
   end
 
+  def last
+    self.drupal_user.last
+  end
+
+  def node_count
+    self.drupal_user.node_count 
+  end
+
   def notes
     Node.where(uid: uid)
       .where(type: 'note')
--- a/app/views/users/list.html.erb
+++ b/app/views/users/list.html.erb
@@ -29,11 +29,11 @@
     <% @users.each do |user| %>
     <tr>
       <td class="visible-sm visible-xs" style="padding-top: 16px;">
-        <img class="img-circle" id="profile-photo" style="width:40px;margin-right:8px;border: 1px solid lightgrey;" src="<%= user.user.profile_image %>" />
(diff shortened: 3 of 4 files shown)


SidharthBansal:
Could you please change DrupalUser to User.

rishabhptr:
@SidharthBansal Yes I know :smiley:. But there is a problem using User here as we are using user.last in the list view page and it is not defined when using a User variable(using id instead of uid).

Also other cases in list function are using DrupalUsers model and issue didnt include changing the model so I went ahead with this.Maybe if @jywarren suggests we could work around to use User model for all cases.

jywarren:
We are trying to stop using `DrupalUser` -- it's true. You should be able to do:

```ruby
User.where(id: ... ).collect(&:id)
```

Make sense?

jywarren:
Thank you!!!! This is exciting!

jywarren:
And indeed, you should be able to swap `User` in for other uses in the same method. And we won't need `.joins('INNER JOIN rusers ON rusers.username = users.name')` anymore, as that bridged the models!

jywarren:
I see the issue on the view, too! Good catch. We can add a method to the `user.rb` model to alias this:

```ruby
def last
  self.drupal_user.last
end
```

The other spot we'd need to change would be:

```
      <td><%= time_ago_in_words(Time.at(user.last['changed'])) if user.last %></td>
```

Here, we could just switch to: `time_ago_in_words(user.updated_at) if user.last`

Hope that helps!!!

rishabhptr:
@jywarren There some more things I have encountered.

I think users status values are opposite in both models which I have reversed also aliased node_count method in User model.
But,
In the the third case when we normally list all people (/people)
`@users = User.select('*, MAX(node.changed) AS last_updated')`
                         ` .joins(:node)`
                          `.group('rusers.id')`
                          `.where('rusers.status = 1 AND node.status = 1')`
                         ` .order("last_updated DESC")`
                         ` .page(params[:page])`
There is a problem here because I think the node model has uid as the foreign key and Users model dont have a uid attribute so joining wont work.

Also I Read the comment in User model:
  #this doesn't work... we should have a uid field on User
  #has_one :drupal_users, :conditions => proc { ["drupal_users.name =  ?", self.username] }
  has_many :images, foreign_key: :uid
  has_many :node, foreign_key: 'uid'
So a migration is needed or we should use any other way around.

Also one last thing, people#id was earlier leading to 'legacy#profile' i.e. user profile so that would be overriden with this.

jywarren:
You're right -- there's some weirdness to this part of the code we need to work around. 

@publiclab/reviewers -- is anyone able to help @rishabhptr work through some of these issues? I'm occupied this afternoon and I know some of you are pretty familiar with this stuff. Thank you!!

jywarren:
So, take a look at this example of how to reference `revisions` instead of node, which should contain the latest updated revision of a node:

https://github.com/publiclab/plots2/blob/f52a4f56e3f1f5af56521d284dc3a0d0a2624992/app/models/user.rb#L97-L103

What do you think? Could you model on this to solve the issue? 

Thank you!!!!

rishabhptr:
@jywarren Thanks for the info also I was actually waiting for the ongoing work to update user status from DrupalUser status because at people list page we check to show only users not banned so after that fix we could easily use User model without any work-around for the same. :smile:

jywarren:
OK, that sounds good -- I'm trying to link to that issue or PR here, but can't find it. Based on https://github.com/publiclab/plots2/issues/2156, can we now rely on them being synced? 

@ViditChitkara -- what do you think? 

Thanks all!

ViditChitkara:
Yes, they should now be synced. Although we have a bit of encoding issue for the same (#2209), however it won't really effect here.

rishabhptr:
Yeah the User status seems to be synced now , thanks @ViditChitkara , @jywarren I have made the required changes, sorry for the delay :smile:

rishabhptr:
There was a small issue that on joining user and node table , status attribute of node was selected hence I have added rusers.status in select query, I hope that's the right way to go :smile:

jywarren:
Oh no! Sorry i should've tried merging this in earlier but didn't see it marked `ready`. Can you resolve the `users_controller` conflict and we can merge it in? Thanks for figuring this out, very exciting!

jywarren:

rishabhptr:
@jywarren Hey, sorry for the delay. I have resolved the conflicts :smile:

jywarren:
Awesome work!!! This is going to be great!!!

👍 👍 🎉

PR #2095, opened 2018-01-24
Opened port 25 in container to docker host (for SMTP)

### LINKED ISSUE #2065: Open port 25 in container to docker host (for SMTP)

Staging container should try to use localhost email facility. Currently it errors (see https://github.com/publiclab/plots2/issues/1570 )

namangupta01:
Closes #2065

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -57,9 +57,12 @@ Plots2::Application.configure do
   # Disable delivery errors, bad email addresses will be ignored
   # config.action_mailer.raise_delivery_errors = false
   # config.action_mailer.delivery_method = :sendmail
-  config.action_mailer.sendmail_settings = {
-    location: '/usr/sbin/sendmail',
-    arguments: '-i'
+  
+  config.action_mailer.delivery_method = :smtp
+
+  config.action_mailer.smtp_settings = {
+  :address => "smtp",
+  :port => 25,
   }
 
   # Enable threaded mode


icarito:
I was researching this and I think this could be a part of the solution.
In order to work in production I'd need to alias smtp to localhost.
Then in order to work in staging, we'd need to figure out a way to bind `smtp` hostname to the docker host. Some documentation on it here https://stackoverflow.com/questions/17770902/forward-host-port-to-docker-container .

Summarizing @jywarren it's a good idea but requires:
 [ ] - smtp alias in production
 [ ] - smtp alias dockerhost in docker-compose.yml file
 [ ] - configure host exim to listen on docker virtual network (currently only listens to localhost).

namangupta01:
@icarito so the problem is that :
We have a docker container for staging and we have another docker container where we have smtp server setup so all we want to do here is to access one docker container port from another docker container?

icarito:
@namangupta01 sorry, no. We do not have another container doing smtp. 
In staging, the app is running in a container. The docker _host_ is listening for SMTP on localhost.
In production, the app is running directly on the server (not yet in a container). The same server is listening for SMTP on localhost.

So we want to access the appropriate SMTP host in each instance.

This might become simpler once we have containers in production.

From the plan above, the one thing that needs to be done here is to somehow see if we can setup `docker-compose.yml` for the app to reach the _docker host_ smtp service in staging, without breaking production.

namangupta01:
@icarito I was researching over this issue, i want to know one thing where are you defining your smtp settings? In rails config files or in linux files?

icarito:
Currently since plots2 runs locally in production, it relies on `sendmail` command, if I understood the configuration. But the host itself is listening to SMTP on localhost, so it should be safe to switch from `sendmail` to SMTP.

namangupta01:
I think directly writing the smtp setting in config files should work everywhere:
`config.action_mailer.smtp_settings = {
address:              'smtp.gmail.com',
 port:                 587,
  domain:               'gmail.com',
  user_name:            'xyz@gmail.com',
  password:             'yourpassword',
  authentication:       :plain,
  enable_starttls_auto: true
}` what do you say?

icarito:
Yeah - a sensible default would be `localhost:25` which is standard SMTP.

I think having the other options as comments would be great so they are easy to figure out.

namangupta01:
I think using above configuration has nothing to do with whether it is running in container or directly.
And i think `action_mailer.smtp_settings` manages itself for standard SMTP but i'm not sure.

namangupta01:
I have been using these settings in my projects and everything working fine for me https://github.com/namangupta01/MCQ/blob/master/config/environments/development.rb#L44

icarito:
I don't know what you mean with `manages itself` :-)
The issue at hand is providing a mechanism for staging (`docker-compose up -d`) to send mail, without breaking production.
Thanks for trying to resolve it!

jywarren:
What's the next step for this, do you think? Perhaps we should consider an approach that sets up a "fallback" mailing system that would be overridden by the production setup, but would silently accept mail (but not send it) from staging. What might that look like?

jywarren:
Hi @icarito if you have some time to think about how this could work, we're hoping to test out #2257 based on fixing mail in staging; currently it doesn't let us test that out. What do you think? Thanks!

icarito:
Hi Jeff,

Previously I had thought this might need some kind of tunneling or
proxying but was reluctant to add complexity.

Since then we tested container in production and it had same problem,
needed access to the a docker host service (production MySQL engine in
this case).

Just a note that we need to add something like this to
`docker-compose-production.yml`

```

    network_mode: "host"
```

...in order to access Docker host smtp.

In sum, we have a path forward.

Regards,
Sebastian

jywarren:
OK, so should we first create a separate `docker-compose-production.yml` in Git? 

Thanks!

jywarren:
@icarito - do we have to start version tracking `docker-compose-production.yml` then? Thanks!

icarito:
No, we use docker-compose.yml in staging. We need to add a SMTP service
to docker-compose.yml and set networking to `host`.

jywarren:
ah, perfect!

jywarren:
Trying this in #2495 -- let's see if that passes! @icarito if it does, please go ahead and merge it so we can test in staging. Thanks!

icarito:
Apologies I thought this would be simpler. The problem is that with containers SMTP won't be 127.0.0.1 unless we do some tricks that make everything more complex. See https://github.com/moby/moby/issues/1143 for a discussion on this specific issue.
The simplest solution is to containerize the SMTP service.

jywarren:
Great work! Thank you!!! 👍

Issue #2057, opened 2018-01-20
node.replace() errors in tests


siaw23:
So I'm getting to know the project. But I have a few questions. 

1. The test results below have some failures, how come this project is passing in Travis on production? 
2. How come there are so few tests and how do I see the percentage of test coverage because I think the test coverage is very very small.

```
siaw23 ~/desktop/plots2 😯 $ rake test
[Coveralls] Using SimpleCov's default settings.
Emptying /Users/siaw23/Desktop/plots2/test/reports
/Users/siaw23/.rvm/gems/ruby-2.3.0/gems/activerecord-4.1.16/lib/active_record/associations/builder/collection_association.rb:38: warning: already initialized constant TagNodeAssociationExtension
/Users/siaw23/.rvm/gems/ruby-2.3.0/gems/activerecord-4.1.16/lib/active_record/associations/builder/collection_association.rb:38: warning: previous definition of TagNodeAssociationExtension was here

# Running tests with run options --seed 64498:

.....................................F................................FF.....................................................................................................................................................................................................................................................................................................F..F.........................................................................................................................................................................................................................................

Finished tests in 96.671273s, 6.2273 tests/s, 18.4646 assertions/s.


Failure:
Minitest::Result#test_previous_and_next_revisions [/Users/siaw23/Desktop/plots2/test/unit/revision_test.rb:55]
Minitest::Assertion: --- expected
+++ actual
@@ -1 +1 @@
-#<Revision vid: 3, nid: 2, uid: 1, title: "About", body: "All about Public Lab, ONCE MORE", teaser: nil, log: nil, timestamp: 0, format: 0, status: 1>
+#<Revision vid: 1, nid: 2, uid: 2, title: "About", body: "All about Public Lab", teaser: nil, log: nil, timestamp: 0, format: 0, status: 1>


Failure:
Minitest::Result#test_replacing_content_in_a_node_with_replace_action [/Users/siaw23/Desktop/plots2/test/functional/wiki_controller_test.rb:453]
Minitest::Assertion: --- expected
+++ actual
@@ -1 +1 @@
-"All about Private Lab"
+"All about Private Lab, ONCE MORE"


Failure:
Minitest::Result#test_replacing_content_in_a_node_with_replace_action_via_JavaScript/AJAX [/Users/siaw23/Desktop/plots2/test/functional/wiki_controller_test.rb:473]
Minitest::Assertion: --- expected
+++ actual
@@ -1 +1 @@
-"All about Private Lab"
[...]

PR #2096, opened 2018-01-24
Fix for node.replace() bug

jywarren:
Attempted fix for https://github.com/publiclab/plots2/issues/2057

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -186,12 +186,13 @@ class Node < ActiveRecord::Base
   def latest
     revisions
       .where(status: 1)
+      .order(timestamp: :desc)
       .first
   end
 
   def revisions
     revision
-      .order('timestamp DESC')
+      .order(timestamp: :desc)
   end
 
   def revision_count
--- a/test/fixtures/revisions.yml
+++ b/test/fixtures/revisions.yml
@@ -25,12 +25,12 @@ about_rev_3:
   title: About
   body: All about Public Lab, ONCE AGAIN
 
-about_rev_3:
+about_rev_4:
   nid: 2
-  vid: 3
+  vid: 4
   uid: 1
   title: About
-  body: All about Public Lab, ONCE MORE
+  body: All about Public Lab
 
 three:
   nid: 3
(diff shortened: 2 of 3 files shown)


siaw23:
yeah i did already. remaining one failure to be fixed. i posted in https://github.com/publiclab/plots2/issues/2057#issuecomment-360239575

jywarren:
It's concerning that these errors did not occur in Travis, but if they work for you, I'm all right with it!

siaw23:
all is fine now. if you merge I'll pull and continue work. i think there's something up with the setup with Travis. @jywarren

siaw23:
Travis failed but I think this should be an easy fix.

siaw23:
how about we specify ruby version on Travis? @jywarren

jywarren:
Actually we do already! v2.3 -- hang on, trying another tweak here --

siaw23:
OK. haven't checked Travis and setup though cause it I thought it's been taken care of. Let's hope this tweak fixes it...

jywarren:
There's an issue with the ordering -- we hadn't set order for `revision.latest` but i just added that -- i think it was accounting for some problems where it wasn't sorting by timestamp... well, we'll see here; 

https://github.com/publiclab/plots2/blob/a701eb8b4c91c42a8311aa2922f66aaadf3e878b/test/unit/revision_test.rb#L42-L74

app/models/node.rb
```diff
@@ -186,12 +186,13 @@ def likers
...
     revisions
       .where(status: 1)
+      .order(timestamp: :desc)
       .first
   end
 
   def revisions
     revision
-      .order('timestamp DESC')
+      .order(timestamp: :desc)
```
siaw23: btw this is one of the changes the style guide takes care of. if we implement it will save us a lot of headache.

jywarren:
There may also be a timestamp based issue, if these tests rely on timestamp ordering and the tests run in less than one second! But we can use Timecop to resolve this: 

https://github.com/publiclab/plots2/blob/a9ae22477ace5eb34c5e7883dbc48819b5d01637/test/functional/admin_controller_test.rb#L20

jywarren:
Hope this works for you!

PR #2129, opened 2018-01-26
Add omniauth(Facebook, Google, Twitter)

biswesh456:

Thanks!


biswesh456:
There are some issues which I have faced:
1) Password validation created a lot of problem. Hence for now I have given password as the uid of the person which comes directly from the provider.(Suggest some better method if there are any).
2) Username provided by Omniauth provider sometimes contain space in between. Such names are shown invalid right now. Can we make any change to that?
3) If a user with same name as an existing user creates an account then it shows "name exists". Should username be unique? Since many users can have same name.
4) Right now only Google works since I have given 'first_name' as the username instead of 'name' and Facebook only provides name. I would Like to give 'name' parameter as the username but then it contains both first name and last name(space in between) which is invalid. In case of twitter, it provides email only when we provide privacy,terms of services page link. Since I didn't find any page, so twitter doesn't work for now.

biswesh456:
I have made dummy client id and secret keys. Please make changes to them.

ryzokuken:
A few directions to keep in mind while doing this part:

1. This needs to be broken in (1 + N) steps (one step for adding omniauth generally, and 1 step for each provider). Don't rush it and don't make one large PR for all of them.

2. Don't hardcode the auth parameters, save them as environment variables and access them inside the app accordingly. Makes changing them (which we'd need to constantly, everyone needs a different one for their dev environments separately) much easier.

3. We will have to fine-tune the integration so that it works perfectly with the current auth system with minimal changes.

biswesh456:
Ya I hardcoded the auth parameters for the time being. Would have made them as env variable in the future. 
So should I create another PR where I break it into various commits?
Also can you help me with the issues which I faced?

ryzokuken:
@biswesh456 Indeed they are! I would suggest coordinating with others and breaking the problem into smaller ones and tackling them one at a time in cohesion with others as this problem could be quite intimidating otherwise (if you go deeper into this).

jywarren:
Hello! I'm going to close this if that's all right, for now. It looks like we're waiting on https://github.com/publiclab/plots2/pull/2137 and then parts of this may re-open? Thank you!!!

PR #2131, opened 2018-01-26
Stats on profile page

### LINKED ISSUE #928: Display completed/replicated Activities on profile pages
We now have "activities" that people can do on the PublicLab.org site: https://publiclab.org/notes/liz/08-30-2016/check-out-these-activity-grids

As people view them, they're prompted to post their results, kind of like a replication: https://publiclab.org/notes/stoft/09-16-2016/stability-upgrade-mockup-for-plab-spectrometer-3-0

This is my replication of the above activity, for example: https://publiclab.org/notes/warren/09-20-2016/building-the-stoft-stability-upgrade-for-desktop-spectrometry-starter-kit-3-0

It'd be interesting to think through and design a way to display what activities you've 'completed' on your profile page -- so that last one would be listed on https://publiclab.org/profile/warren, for example.

You can tell that something's a `replication` or a `build` (that is, a replication of an upgrade, rather than an activity), because it'll have a `replication:12345` power tag (or `build:12345`) -- where `12345` is the `nid` of the activity it's based on. 

That is also shown at the top of the replication post, using this code: https://github.com/publiclab/plots2/blob/master/app/views/notes/show.html.erb#L5-L6

Collecting the replications a user's done wouldn't be that hard, but how do we display them? This needs some design work first, and the profile page is a bit messy and in need of a redesign already.

SidharthBansal:
Fixes #928


jywarren:
Amazing! Thank you -- we've wanted this one for a while!!! 🎉

PR #2137, opened 2018-01-26
Facebook Authentication

### LINKED ISSUE #683: Add "Log in with Facebook" for new accounts
(later following up with twitter/other social media)

OmniAuth is a great starting point: https://github.com/omniauth/omniauth

Let's begin with Facebook!

See these "strategies" to support FB, Twitter, Google, etc:

https://github.com/omniauth/omniauth/wiki/List-of-Strategies

****

@steviepubliclab - starting this issue here :)

SidharthBansal:
Fixes #683 
To Dos
* [ ] User_session_controller update
* [x] Views update
* [ ] Write tests for the simple login through fb(without integration)
* [ ] Integration with the current model
* [ ] Write tests
* [ ] Write Documentation for the fiagro gem
* [ ] Write Documentation for omniauth 
* [ ] Write Documentaion for the omniauth-facebook

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -177,6 +177,7 @@
                     <a class="btn btn-default" href="/signup" tabindex="4"><%= t('layout._header.login.sign_up') %></a>
                   </p>
 
+                  <p><a href="/auth/facebook">Login with Facebook</a></p>
                   <p><a href="/reset/"><%= t('layout._header.login.reset_password') %></a></p>
 
                 <% end %>
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ test/reports/
 # Ignore NPM
 node_modules
 .byebug_history
+
+# Ignore application configuration
+/config/application.yml
--- a/Gemfile
+++ b/Gemfile
@@ -106,3 +106,5 @@ gem 'jbuilder'
 
 # Pin mustermann to Ruby 2.1 compatible
 gem 'mustermann' , '~> 0.4'
+gem 'figaro'
+gem 'omniauth-facebook'
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -4,68 +4,74 @@ class UserSessionsController < ApplicationController
   end
 
   def create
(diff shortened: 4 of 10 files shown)


biswesh456:
Interesting. For me in the method create_with_omniauth(auth) caused problem related to password validation hence I had to include user.password_confirmation in it. Does it not give you any error @SidharthBansal ? ( Just curious !!!)

SidharthBansal:
I am currently working on it. A lot of things are pending in this pr

jywarren:
Wow, this is a big one. If there's any way to implement some of these systems in smaller pieces, I'm open to that -- but I understand if it seems impractical. I guess, i'm wondering also if we could implement it only offer login by facebook when a URL flag is set like `?fb=true` (checking `params[:fb]` in the view) so that we can test this out for a bit before presenting it to everyone on the live site by default. Make sense? 

Good luck with this!!!

app/views/layouts/_header.html.erb
```diff
@@ -177,6 +177,7 @@
                     <a class="btn btn-default" href="/signup" tabindex="4"><%= t('layout._header.login.sign_up') %></a>
                   </p>
 
+                  <p><a href="/auth/facebook">Login with Facebook</a></p>
```
jywarren: Here -- we could do `<% if params[:fb] %> ... <% end %>` to only show this when that flag is set, you know?

SidharthBansal: Yeah!!!   I also want to test this feature on production before launching to public. Thanks for the hint. I will try to do it.

SidharthBansal:
We can write the documentations in separate prs. But all other things needs to be done in a single PR.
Thanks

jywarren:
OK, great -- just tell me how I can help keep these organized and if you get stuck!

jywarren:
Hi @SidharthBansal how's this coming along? is it worth trying to see if we could merge in functional subsections of this project in smaller parts? Like adding gems, etc etc, adding back-end infrastructure... what do you think?

SidharthBansal:
We can do that. But that will not result in functionality until all commits/ prs are merged. 
Ok. I am closing this now and pushing subsections one by one.
Thanks

PR #2139, opened 2018-01-27
Order by subscription fix

Gauravano:
If there is no current_user then also the button was shown generating the error.Also, I can make table heading links so the button will no longer be needed.What you say @jywarren .


jywarren:
> Also, I can make table heading links so the button will no longer be needed

That sounds super. Please do! Is this ready for merging though? Thanks!!!

Gauravano:
@jywarren links added in table headings and Sort button removed.Here's the preview 

subscriptionfix
Do you think tooltip should be added? Also, plz suggest if any other changes are required.Thanks.

jywarren:
This looks great! What do you think about making additional routes that are more readable -- like, `/tags/uses` `/tags/subscribers` -- or using a GET param like `/tags?sort=subscribers` -- to make the routes more readable? 

We could merge this and do that in a separate PR, if you think that's best? I'm open to either way. Thank you, this is so great.

👍

Gauravano:
ok @jywarren, I will open a PR soon setting new routes

jywarren:
Awesome. Thanks!!!

PR #2142, opened 2018-01-27
Error messages for rich editor

### LINKED ISSUE #2141: Error in flash messages condition
The condition inside the if will give error if any of the @node or @revision is nil
We need to tackle this in separate if-else loop
https://github.com/publiclab/plots2/blob/097605afe74fa63f6fa5c3de3fbd081fa0d2e2c1/app/views/editor/rich.html.erb#L221-L223
https://publiclab.org/ ...

bansal_sidharth2996

## Thank you!

SidharthBansal:
resolved #2141
Without this it will give an error --> undefined method errors on nil object as it may be that one of the two is nil

### DIFF
--- a/app/views/editor/rich.html.erb
+++ b/app/views/editor/rich.html.erb
@@ -38,7 +38,7 @@
           </div>
           <div class="col-xs-2" style="padding-top:10px;text-align:left;">
             <a rel="tooltip" data-placement="bottom" title="Learn more about different posting types." class="ple-help-light" style="font-size:18px;"><i class="fa fa-question-circle"></i></a>
-            
+
           </div>
         </div>
       </div>
@@ -163,7 +163,7 @@
     </div>
     -->
     <!-- end multi module -->
-  
+
 
   </div>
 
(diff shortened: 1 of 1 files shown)


jywarren:
This looks cool but is a little tough to read! Do you think this is the most efficient way to implement it? Is there any way to merge them even if they're empty, and then display the merged errors no matter what, instead of having 3 possible options? 

Just an idea. If not, i'm happy to merge this. Thank you!!!

SidharthBansal:
OK. I will redactor it.

SidharthBansal:
The main problem is we cant apply the `.errors` on `nil`. If by any way I try to merge them together then I need to use the errors method which we should not.
One solution is to use ternary operator but for some, it is hard to read while others like it. Shall I use the ternary operator if you like them?

jywarren:
No, i guess i agree this is more readable than ternary. Shall we go ahead and merge then?

jywarren:
Thanks for giving it a little extra thought!

SidharthBansal:
Yeah sure.

PR #2163, opened 2018-01-29
Redundant "method" fixture

SidharthBansal:
I have encountered a fixture named "method" inside the node.yml file. It is not at needed as we already have another fixture named "method" in the same file with same nid. 
I have checked that all tests pass after removing the fixture.


jywarren:
Thanks!!

PR #2197, opened 2018-02-02
comments on profile page fixed

### LINKED ISSUE #2196: Comments not loading on certain profile pages

Unable to fetch comments for certain profile pages (e.g: https://publiclab.org/profile/liz, https://publiclab.org/profile/warren). 
comments-1-1

https://publiclab.org/profile/liz, https://publiclab.org/profile/warren (click on the comments tab)

vidit

ViditChitkara:
fixes #2196

Thanks!

### DIFF
--- a/app/controllers/answers_controller.rb
+++ b/app/controllers/answers_controller.rb
@@ -41,7 +41,7 @@ class AnswersController < ApplicationController
        current_user.role == 'admin' ||
        current_user.role == 'moderator'
       respond_to do |format|
-        if @answer.delete
+        if @answer.destroy
           format.html { redirect_to @answer.node.path(:question), notice: 'Answer deleted' }
           format.js
         else
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -233,7 +233,7 @@ class NotesController < ApplicationController
   def delete
     @node = Node.find(params[:id])
     if current_user.uid == @node.uid && @node.type == 'note' || current_user.role == 'admin' || current_user.role == 'moderator'
-      @node.delete
+      @node.destroy
       respond_with do |format|
         format.html do
           if request.xhr?
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -67,7 +67,7 @@ class Comment < ActiveRecord::Base
     if aid == 0
       node
     else
-      answer && answer.node
+      return answer.node unless answer.nil?
     end
   end
 
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -5,8 +5,8 @@ class Comment < ActiveRecord::Base
     :subject, :hostname, :comment,
     :status, :format, :thread, :timestamp
 
-  belongs_to :node, foreign_key: 'nid', touch: true,
-                    dependent: :destroy, counter_cache: true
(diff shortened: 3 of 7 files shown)


ViditChitkara:
@jywarren could you have a look at it?

app/models/comment.rb
```diff
@@ -5,8 +5,8 @@ class Comment < ActiveRecord::Base
     :subject, :hostname, :comment,
     :status, :format, :thread, :timestamp
 
-  belongs_to :node, foreign_key: 'nid', touch: true,
-                    dependent: :destroy, counter_cache: true
+  belongs_to :node, foreign_key: 'nid', touch: true, counter_cache: true
```
jywarren: Hmm, does this have to do with the comment changes? What was the issue? I wasn't sure `counter_cache` was working and would like to demonstrate this with a test if that's OK... a unit test?

app/models/comment.rb
```diff
@@ -67,7 +67,7 @@ def parent
     if aid == 0
       node
     else
-      answer.node
+      answer && answer.node
```
jywarren: Here, we have to be careful because `parent` should return the record, not just a boolean. Can you think of how to split this out more clearly? Or maybe we make an alternate way to refer to an answer instead of the parent node?

app/models/comment.rb
ViditChitkara: Hi @jywarren. Actually I haven't removed counter_cache but 'dependent: :destroy'.
Actually the error was occurring because some comments did not belong to an answer or a node. Perhaps there is no need of dependent: :destroy here but the node model (has_many comments, dependent: :destroy). Just realized that dependent destroy also needs to be added to answer model here:-
https://github.com/publiclab/plots2/blob/b978d2bd8912a3c4cbffd3a320f44f05deb8764d/app/models/answer.rb#L9
Please correct me if I am wrong!

app/models/comment.rb
ViditChitkara: Oops, my mistake, you're right. I just need to check if answer existed then return answer.node

ViditChitkara: I'll make the changes right away!

ViditChitkara:
@jywarren thanks for the review!! I,m writing the tests now along with minor tweaks.

app/models/comment.rb
ViditChitkara: Hi @jywarren , The tests are written. Please have a look

app/controllers/answers_controller.rb
```diff
@@ -41,7 +41,7 @@ def delete
        current_user.role == 'admin' ||
        current_user.role == 'moderator'
       respond_to do |format|
-        if @answer.delete
+        if @answer.destroy
```
ViditChitkara: destroy method triggers dependent: :destroy callback, hence, delete was changed to destroy here.

app/controllers/notes_controller.rb
```diff
@@ -233,7 +233,7 @@ def update
   def delete
     @node = Node.find(params[:id])
     if current_user.uid == @node.uid && @node.type == 'note' || current_user.role == 'admin' || current_user.role == 'moderator'
-      @node.delete
+      @node.destroy
```
ViditChitkara: Same as answer_controller

jywarren:
Awesome! Great work here.

PR #2205, opened 2018-02-02
Change heading of Weekly Digest page

Gauravano:
Heading changed to "Content from the past week on topics you've subscribed to" as suggested in #2058


jywarren:
:-) 👍  perfect!

PR #2108, opened 2018-01-25
Various pages for tagname and authorname

SidharthBansal:
Make sure these boxes are checked before your pull request is ready to be reviewed and merged. Thanks!

* [x] all tests pass -- `rake test:all`
* [x]  Wiki page with tagname and authorname is functional
* [x] Map page with tagname and authorname is functional
* [x] Questions page with tagname and authorname is functional

PR #2207, opened 2018-02-02
Added tests for show_for_author

SidharthBansal:
These are the remaining tests in addition to #2108 for maps, wikis, and questions
Is there any need for the integration tests?
I have made some rearrangements in the controller so that it will be in analogy with the show method.
I have changed the model a bit in accordance with the show controller.
This is ready to be merged.

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -138,15 +138,10 @@ class TagController < ApplicationController
   end
 
   def show_for_author
+    # try for a matching /wiki/_TAGNAME_ or /_TAGNAME_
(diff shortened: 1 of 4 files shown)


jywarren:
Hi, what do @publiclab/reviewers think of this one? It looks good to me; thanks @SidharthBansal !!!

SidharthBansal:
@jywarren @sagarpreet-chadha has reviewed it. Can you please review it again?

jywarren:
Super. Merging! and many thanks -- this is a big one. It makes me want to try to reduce the code complexity at some point... but glad to have all these tests to cover it!

jywarren:
🎉 👍

PR #2232, opened 2018-02-05
Even treatment of heading on /wiki

SidharthBansal:
On /wiki we have uneven headings. Notice the star symbol only for the likes. It looks ugly.
image
I have made it like this.
image


jywarren:
Awesome. 🎉 👍

PR #2297, opened 2018-02-11
remove extra dashboard route

ViditChitkara:

Thanks!

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -180,7 +180,6 @@ Plots2::Application.routes.draw do
   get 'feed/liked' => 'notes#liked_rss'
 
   get 'dashboard' => 'home#dashboard'
-  get 'dashboard2' => 'home#dashboard'
   get 'comments' => 'comment#index'
   get 'profile/comments/:id' => 'users#comments'
   get 'nearby' => 'home#nearby'


SidharthBansal:
I have changed it earlier to in one of my PRS. Actually, @jywarren told me we use it somewhere I don't remember where ,so was asking me to keep it. So, I have refactored it a lot in that PR.

SidharthBansal:
In #2006 @seafr and @jywarren suggested to keep both of the routes.

ViditChitkara:
Hi @SidharthBansal , dashboard2 might have been used to test the beta version of the dashboard. Now that both routes point to the same controller and action, it does the same thing as dashboard does, so we must remove the extra one in order to reduce the unnecessary code. What do you think?? @seafr, @jywarren your thoughts on this?

seafr:
@ViditChitkara According to @SidharthBansal's comment above, the route in question is still in use so it cannot be removed (although I'll admit that it bothers me to see it there - maybe you can add a comment to explain things while you are at it). And some unnecessary code has already been removed in the form of the action it used to point to.

ViditChitkara:
Hi @seafr , how have you been? 
Actually I searched dashboard2 in the entire project but could not find a usage. Maybe it was used before, however its usage is gradually removed now.

ViditChitkara:
What do you think??

seafr:
Hi @ViditChitkara I'm fine, thank you. And I hope you are well too :+1: 
Regarding dashboard2, an action by that name was removed from the `home` controller in a earlier refactoring by @SidharthBansal. This means that the `home/dashboard2` was in use at some point in the past and may have been bookmarked. Maybe we should check the logs to see whether or not people still access the site using this url.

ViditChitkara:
Yep, I guess it was used before to test for a new dashboard (maybe not used by normal users). 
I agree, we need to check the logs for the same and then proceed further.

jywarren:
Hi folks - we left that in for folks who had been beta testing, but it's been over a year now, so I think we're OK to remove it. In the future when we add temporary routes, let's add a comment indicating roughly on what date we want to deprecate them. Great!

Issue #2304, opened 2018-02-12
PLANNING ISSUE: Spam comments moderation 


Gauravano:
A user when banned by moderators, the alert text said that banning the author would remove all their content from the site -- but it did not. This is a bug.So, for handling this a comment moderation system needs to be built as discussed in #1785 .

* [x] make a "status" column for comments, default `1`, spam = `0`
* [x] make a "spam" method like this, that only moderators or admins can use (`user.role == admin`): https://github.com/publiclab/plots2/blob/master/app/models/node.rb#L158
* [x] a "publish" method which sets `status` back to 1
* [x] unit test the above 2 methods
* [x] display comments (on dashboard, /research, in answers, wiki pages and notes) only if they have `status == 1`
* [ ] display spammed comments to admins and moderators (`user.role == admin || user.role...`) but with text hidden, with a notice "A comment by @______ was put in moderation. _Read the full comment_" which displays the text, and a "republish" button

Gauravano:
Hi, @jywarren as per the first part status column needs to be added but it's already there with the default value of 0 (for published). So, I am starting with 2nd part and giving spam comment as status = 2.

jywarren:
This is looking awesome! Want to add "Planning issue" to the title? :-) 🎉

Gauravano:
Hi @jywarren , sorry it was still open. Let's complete this ASAP.

Gauravano:
Raised PR at #2957 . Also, I have gone through the work done by me in past. Just last 2 parts remaining.

The comments generated after PR https://github.com/publiclab/plots2/pull/2373 would be having `status = 1` but comments before that would be carrying `status = 0` . Could you please check stats @jywarren? 

We need to run a migration on production DB to change the status of all comments to 1 and the next step would be to merge #2957. Thanks!

Gauravano:
> display spammed comments to admins and moderators (user.role == admin || user.role...) but with text hidden, with a notice "A comment by @______ was put in moderation. Read the full comment" which displays the text, and a "republish" button

@jywarren , any more details you wanna add to this part of issue?

jywarren:
This is great. Maybe also that when a user is banned their comments should be moderated as well!

Gauravano:
Adding this to checklist. @jywarren have you run migration for changing comment status to 1?

jywarren:
Uhhhhh, hmm, sorry i think i need a refresher -- but any migrations that've been merged have definitely been run!

[...]

PR #2305, opened 2018-02-12
Comment spam moderation Part 2

Gauravano:
Fixes part 2 of #2304 

* [x] code is in uniquely-named feature branch and has been rebased on top of latest master (especially if you've been asked to make additional changes)

Thanks!

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -141,7 +141,7 @@ class AdminController < ApplicationController
   def publish_comment
     if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
       @comment = Comment.find params[:id]
-      if @comment.status == 0
+      if @comment.status != 2
         flash[:notice] = 'Comment already published.'
       else
         @comment.publish
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -123,6 +123,23 @@ class AdminController < ApplicationController
     end
   end
 
+  def mark_comment_spam
+    @comment = Comment.find params[:id]
+    if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
+      if @comment.status == 0
+        @comment.comment_spam
+        flash[:notice] = "Comment has been marked as spam."
+        redirect_to '/dashboard'
+      else
+        flash[:notice] = "Comment already marked as spam."
+        redirect_to '/dashboard'
+      end
+    else
+      flash[:error] = 'Only moderators can moderate comments.'
+      redirect_to '/dashboard'
+    end
+  end
+
   def publish
     if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
       @node = Node.find params[:id]
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -126,7 +126,7 @@ class AdminController < ApplicationController
   def mark_comment_spam
     @comment = Comment.find params[:id]
     if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
-      if @comment.status != 2
+      if @comment.status == 0 || @comment.status == 1
         @comment.spam
         flash[:notice] = "Comment has been marked as spam."
       else
@@ -141,7 +141,7 @@ class AdminController < ApplicationController
   def publish_comment
     if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
       @comment = Comment.find params[:id]
-      if @comment.status != 2
+      if @comment.status == 0 || @comment.status == 1
         flash[:notice] = 'Comment already published.'
       else
         @comment.publish
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1,5 +1,5 @@
 class AdminController < ApplicationController
-  before_filter :require_user, only: %i(spam spam_revisions)
+  before_filter :require_user, only: %i(spam spam_revisions mark_comment_spam publish_comment)
 
   def promote_admin
     @user = User.find params[:id]
@@ -126,7 +126,7 @@ class AdminController < ApplicationController
   def mark_comment_spam
     @comment = Comment.find params[:id]
     if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
-      if @comment.status == 0
+      if @comment.status != 2
         @comment.spam
         flash[:notice] = "Comment has been marked as spam."
       else
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -127,15 +127,30 @@ class AdminController < ApplicationController
     @comment = Comment.find params[:id]
     if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
       if @comment.status == 0
-        @comment.comment_spam
+        @comment.spam
         flash[:notice] = "Comment has been marked as spam."
-        redirect_to '/dashboard'
       else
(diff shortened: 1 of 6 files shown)


app/controllers/admin_controller.rb
```diff
@@ -123,6 +123,23 @@ def mark_spam
     end
   end
 
+  def mark_comment_spam
+    @comment = Comment.find params[:id]
+    if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
+      if @comment.status == 0
+        @comment.comment_spam
+        flash[:notice] = "Comment has been marked as spam."
+        redirect_to '/dashboard'
```
seafr: IMHO, since the user is redirected in any case, the redirect statement can be placed at the end of the action (just after line 140).

Gauravano: Hi @seafr, yes you are absolutely right, actually, it's just a basic function to show @jywarren to get approval that I am going correctly with the issue and to ask if the status code used is fine or not.But still thanks for highlighting, I will definitely correct it in next commit.

app/models/comment.rb
```diff
@@ -142,4 +142,11 @@ def answer_comment_notify(current_user)
     notify_users(uids, current_user)
     notify_tag_followers(already + uids)
   end
+
+  def comment_spam
```
jywarren: Oh cool! Let's name this just `def spam` to match node.rb, and add a corresponding `def publish` method? 

https://github.com/publiclab/plots2/blob/fa16a69f8f475c101bc23325f4df7411c2172872/app/models/node.rb#L151-L161

How does that sound? Then we can unit test this and functional test the above. This is looking great!!!

jywarren:
Very excited for this!!! 👍 👍

app/models/comment.rb
Gauravano: will do

app/controllers/admin_controller.rb
```diff
@@ -123,6 +123,38 @@ def mark_spam
...
+    else
+      flash[:error] = 'Only moderators can moderate comments.'
+    end
+    redirect_to '/dashboard'
+  end
+
+  def publish_comment
+    if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
+      @comment = Comment.find params[:id]
+      if @comment.status != 2
```
seafr: I thought you'd check to see if `status == 0`

Gauravano: I wanted to but nodes and comments have a different status system.Active nodes has status = 1, spam has status 0.But, for comments, an active comment has status = 0 and so I have used status = 2(for spam).

app/controllers/admin_controller.rb
```diff
@@ -123,6 +123,38 @@ def mark_spam
...
+        flash[:notice] = 'Comment already published.'
+      else
+        @comment.publish
```
seafr: Or maybe you can instead check here if `status != 0`, and if `true` set it to 0. This would be parallel to the strategy you used above to mark a comment as spam.

seafr: These are comments on clarity, not correctness. I think the code itself has no problem.

test/fixtures/comments.yml
```diff
@@ -112,3 +112,11 @@ question_tag:
   comment: 'Question #everything'
   timestamp: <%= Time.now.to_i + 13 %>
   thread: /04
+
+spam_comment:
+  uid: 1
+  nid: 1
+  status: 2
+  comment: This thing is spam.
```
seafr: :+1: Great comment!

app/controllers/admin_controller.rb
Gauravano: Oops, I understood your question now, actually, while writing tests I found that we have one more status for comment i.e., status = 1 - it represents what unknown to me, so I used condition 
status!= 2 so that other status values wouldn't change.

test/fixtures/comments.yml
Gauravano: Thanks :smile:

Gauravano:
hi, @jywarren could you please clarify what comment with status = 1 represent as when I checked for comment's status in my rails console they all have status = 0 but fixtures have some comments with status = 1.

app/controllers/admin_controller.rb
seafr: I see your point.
Can status be set to a negative number?

app/controllers/admin_controller.rb
Gauravano: thanks, @seafr I really appreciate your comments and will correct this.

app/controllers/admin_controller.rb
Gauravano: Got your point !!!

app/controllers/admin_controller.rb
```diff
@@ -123,6 +123,38 @@ def mark_spam
...
 
+  def mark_comment_spam
+    @comment = Comment.find params[:id]
+    if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
+      if @comment.status == 0 || @comment.status == 1
+        @comment.spam
+        flash[:notice] = "Comment has been marked as spam."
+      else
+        flash[:notice] = "Comment already marked as spam."
+      end
```
seafr: I'd like to make one last suggestion. The idea is to first check if the status is as desired, in which case you move on, or else you set it as desired. I think it'll look as follows:
```
if status is as desired
  flash the "already" msg
else
  set status as desired
  flash msg about what you did there
end
```
This is what you did in the publish_comment method, and it is clearer IMO.

jywarren:
^ as to your question -- i thought status = 1 means published. But if you check, maybe I had made a mistake on that. Let's have it match the status for nodes! And let's document it as we do here:

https://github.com/publiclab/plots2/blob/0a29154decc8e51c36c46447aed023998498f838/app/models/drupal_user.rb#L4-L7

Thanks!

Gauravano:
For clarity,
# Comment status:
0 : normal
2 : spam
1 : unknown

jywarren:
haha, yeah i guess that's the case currently but can we try go get it to
match the status of node.rb? Where on these lines we see that status = 4
means a first-time poster, and 0 is spam and 1 is published.

https://github.com/publiclab/plots2/blob/master/app/models/node.rb#L143-L161

How about that?

Gauravano:
@jywarren that would imply changing all fixtures and if fixtures are changed then tests would also need to be changed accordingly.
What you say should we change the system or adjust for now with the current situation.

Gauravano:
Actually, I just checked the comment_controller and there is no trace of status there, so I guess a try can be given.So, @jywarren I will raise a PR as soon as possible for this one.

jywarren:
Sorry, indeed! I don't think comment status has been used at all previously! So I don't think it's a big change. Right now, I don't think any comments have a meaningful status set, so we could just add a migration to /all/ comments within this PR that sets them to status = 1 -- would that be possible?

Sorry to scare you! I didn't mean to say we should do a major refactoring, I just didn't think it'd be that much since we haven't yet ever used the comment status field -- best to do it right before we start using it, right?

Thank you!

Gauravano:
Agreed @jywarren it would be better to have the same system for both node and comment.So writing a migration is good option

jywarren:
Just removing `ready` label until #2373 is merged! 👍

Gauravano:
@jywarren I have optimized all the tests and controller as per our new comment status

jywarren:
Awesome! Is next step to build a button that's displayed to only mods? 

Great work!!!

Gauravano:
Yes, Jeff next step is button and routes addition.

PR #2373, opened 2018-02-20
Migration for changing comment status

### LINKED ISSUE #2304: PLANNING ISSUE: Spam comments moderation

Gauravano:
Fixes #2304 Part 1 and based on discussion in #2305

Thanks!

### DIFF
--- a/db/migrate/20180220180926_add_default_value_to_comment_status.rb
+++ b/db/migrate/20180220180926_add_default_value_to_comment_status.rb
@@ -1,0 +1,6 @@
+class AddDefaultValueToCommentStatus < ActiveRecord::Migration
+  def change
+    change_column :comments, :status, :integer, default: 1
+    Comment.update_all ["status = ?", 1]
+  end
+end
--- a/db/schema.rb.example
+++ b/db/schema.rb.example
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180128162224) do
+ActiveRecord::Schema.define(version: 20180220180926) do
 
   create_table "answer_selections", force: true do |t|
     t.integer "user_id"
--- a/db/schema.rb.example
+++ b/db/schema.rb.example
@@ -42,7 +42,7 @@ ActiveRecord::Schema.define(version: 20180128162224) do
     t.text    "comment",   limit: 2147483647,              null: false
     t.string  "hostname",  limit: 128,        default: "", null: false
     t.integer "timestamp",                    default: 0,  null: false
-    t.integer "status",    limit: 1,          default: 0,  null: false
+    t.integer "status",    limit: 1,          default: 1,  null: false
     t.integer "format",    limit: 2,          default: 0,  null: false
     t.string  "thread"
     t.string  "name",      limit: 60


Gauravano:
Hi @jywarren please see if the migration statements are correct.Thanks.

db/migrate/20180220180926_add_default_value_to_comment_status.rb
```diff
@@ -0,0 +1,6 @@
+class AddDefaultValueToCommentStatus < ActiveRecord::Migration
+  def change
+    change_column :comments, :status, :integer, default: 1
```
jywarren: Are you just changing the default here then? Can you add a comment that only the default is changing?

Otherwise this looks good and it'll need an update to the schema.rb.example to pass tests. Good work!

Gauravano: This is changing default to 1

db/migrate/20180220180926_add_default_value_to_comment_status.rb
```diff
@@ -0,0 +1,6 @@
...
+    Comment.update_all ["status = ?", 1]
```
Gauravano: this is updating the status of all existing comments in the database to  1

db/migrate/20180220180926_add_default_value_to_comment_status.rb
Gauravano: >Are you just changing the default here then? Can you add a comment that only the default is changing?

Both default value and existing comment's status are updating with this as we have discussed

Gauravano:
@jywarren I just followed logs and found this 

errorsource

Its due to rake  db:setup

jywarren:
Ah, you'll have to run that locally and copy the checksum into the top of schema.rb.example, so it knows it's got the latest version!

jywarren:
Awesome!

db/migrate/20180220180926_add_default_value_to_comment_status.rb
Gauravano: by changing the version number in schema.rb.example will only affect new comments by setting their status =1 but this migration would have no effect on current data unless `rake db:migrate` is run on production. What do you think?

Gauravano:
hi @jywarren , by changing the version number in schema.rb.example will only affect new comments by setting their status =1 but this migration would have no effect on current data. Although Comment.update_all line I have included in my migration to change the status but it will be in effect if the migration would be run on production. What do you think, am I right?

jywarren:
Yes, @Gauravano this will have been run on all comments in the db as soon as it was deployed, so we can assume comment status default for historical and new comments is 1 now.

PR #2380, opened 2018-02-22
Replace spaces with hyphens (list of suggested tags and query submitted in the form)

### LINKED ISSUE #2335: Insert hyphens into tag search input query
(via @ebarry) On this page: https://publiclab.org/questions/ in the sidebar, we let people filter by tag, but when you type `water quality`, it doesn't offer the tag name `water-quality` (because all tagnames are hyphenated). 

The logic for the query is here:

https://github.com/publiclab/plots2/blob/d078466a6dc1c1c4ae3b39e45a8ccb861b7d5860/app/views/tag/_sorting.html.erb#L15

Let's insert a filter to replace spaces with hyphens as people type. So we should be able to replace `query` with:

```js
query.replace(' ', '-')
```

This'll improve matching overall and make it default that it selects a good related tag for many more possible inputs.

Any help appreciated on this!!! 👍

milaaraujo:
fixes #2335, replacing spaces with hyphens as people type.

Thanks!

### DIFF
--- a/app/views/tag/_sorting.html.erb
+++ b/app/views/tag/_sorting.html.erb
@@ -27,7 +27,7 @@
 
       $('#tagform').submit(function(e) {
         e.preventDefault();
-        window.location = "/questions/tag/" + ($('#taginput').val()).replace(' ', '-');
+        window.location = "/questions/tag/" + ($('#taginput').val()).replace(/\s/g, '-');
       });
 
     });
--- a/app/views/tag/_sorting.html.erb
+++ b/app/views/tag/_sorting.html.erb
@@ -14,7 +14,8 @@
         items: 8,
         minLength: 3,
         source: function (query, process) {
-          return $.post('/tag/suggested/' + query, {}, function (data) {
+          var replaced_query = query.replace(' ', /-/g);
+          return $.post('/tag/suggested/' + replaced_query, {}, function (data) {
             return process(data)
           })
         },
@@ -26,7 +27,7 @@
 
       $('#tagform').submit(function(e) {
         e.preventDefault();
-        window.location = "/questions/tag/" + $('#taginput').val();
+        window.location = "/questions/tag/" + ($('#taginput').val()).replace(' ', '-');
       });
 
     });


milaaraujo:
Hi, @publiclab/reviewers! Can anyone review this for me?

jywarren:
Looking good! Have you confirmed locally that this works when you type in?

Also, would you mind changing the title to briefly sum up the fix so it's easy to see in a list of PRs? Thanks a lot this is almost good to go!

milaaraujo:
Hi @jywarren! Well, what I did was replace the spaces with hyphens in the tags suggested for users, while the user types, and in the query submitted when the user clicks 'Search'. In this way, if the user types 'water quality', in the suggestions it will show 'water-quality'. And if the user types 'water quality' and clicks 'Search', the submitted query will be 'water-quality'. Is it right? Now I'm not sure if that's what you'd like me to do!

jywarren:
That's correct, thank you!!!

jywarren:
This is great. Thanks so much for your work! We really appreciate it! 👍 👍 🎉

milaaraujo:
You're welcome!

app/views/tag/_sorting.html.erb
```diff
@@ -14,7 +14,8 @@
         items: 8,
         minLength: 3,
         source: function (query, process) {
-          return $.post('/tag/suggested/' + query, {}, function (data) {
+          var replaced_query = query.replace(' ', /-/g);
```
seafr: Compare the arguments of the `replace` function here with those of the same function on line 30 below. Their types differ. Here the second argument is a regular expression. Shouldn't the first argument be a regular expression?

jywarren: This is true -- if you'd like to file a follow-up issue to make them identical, I'm fine with that! Either of you -- thanks all!

seafr: @milaaraujo will decide!

milaaraujo: Hi @jywarren and @seafr! I'm reviewing the code and the arguments I'm using in that `replace` function on line 17 aren't making sense. Sorry! Thank you for noticing that, @seafr.

By typing `'water '`, the substitution would cause an error and the suggested terms were the terms suggested only for `'water'`, without the space. Once the suggestions tags were there and `rake test` did not accuse any error, I thought it was working properly.

Changing the arguments of the replace function on line 17 (for the same ones on line 30), it seems that the replacement of spaces by hyphens happens correctly, but for some reason the list of suggestions does not appear.

captura de tela 2018-02-27 as 19 27 04

Do you have any idea why this is happening? @jywarren, can I continue working on this issue here? Or should I open a new issue?

seafr: @milaaraujo have you tried stripping any trailing spaces before the substitution process. 
There is also the fact that `' '` is no the same thing as `\s` as the latter also includes the *tab* and *newline* characters.

milaaraujo:
Yes, @seafr! Now I'm running the code with the `query.replace(/\s/g, '-')` on line 17. I think the replacement is occurring correctly, because when I search for 'test test test' on the terminal I see that `Parameters: {"id" => "test-test-test"}`.

captura de tela 2018-02-28 as 10 50 08

I see exactly the same result (SQL) when I type 'test-test-test' and the window with the suggested tags appears... The problem is that when I type the words with spaces the dropdown window with the suggestions do not appear, even with the replacement apparently working well.

app/views/tag/_sorting.html.erb
seafr: @milaaraujo have you tried using: `query.replace(/ /g, '-')`?

milaaraujo: Yep, @seafr. It's the same thing: when I search for 'test test test' on the terminal I see that `Parameters: {"id" => "test-test-test"}`. That's why I think the problem is not the replace function. Now I'm trying to understand how the suggestion process works...

seafr: :( That's strange. 
The problem is apparently not with the replace function. But the suggestion dropdown window stopped working after you changed it from its previous form: `query.replace(' ', /-/g)`
Let's see what @jywarren has to say on this.

app/views/tag/_sorting.html.erb
```diff
@@ -14,7 +14,8 @@
...
+          return $.post('/tag/suggested/' + replaced_query, {}, function (data) {
             return process(data)
```
seafr: Are we missing a semicolon here?

milaaraujo: On line 19? I did not change that part of the code, but I tested here with the semicolon and the result is the same. I'm almost sure that in javascript the semicolon isn't needed when you have only one statement in a block.

seafr: Yes, you're right, it's not necessary, but it's not good practice. However, since you have confirmed that adding it makes no difference and that's not code you've written, let's leave it at that.
Regarding the solution, I think we'll have to wait for @jywarren to debug this. (I'm not very comfortable with JS anyway.)

milaaraujo: Thanks for your help! :)

PR #2459, opened 2018-03-09
Caching impressions

### LINKED ISSUE #2389: Cache impressions for optimizing SQL query

We are noticing that there's a query taking a lot of time and it could be causing small outages by consuming available threads.
The query is `publiclab@172.18.0.1: SELECT COUNT(*) FROM impressions WHERE impressions.impressionable_id = :int AND impressions.impressionable_type = :string` 

The suggestion by @jywarren is to **Optimize impressionist counts with a cache**:

- wrap this line in cache block: 
https://github.com/publiclab/plots2/blob/58bb6010e28d91106556b2cf07c9a100e3324c46/app/models/node.rb#L133

- example cache block that'll need a new "name":
https://github.com/publiclab/plots2/blob/245d6bc11383d220101f098d25b3843ce59e4c7d/app/views/home/home.html.erb#L133-L135

****

## Thank you!

jywarren:
Fixes #2389 -- let's see if this works! It may break some tests if we test `node.totalcount` but we can adjust them or use TimeCop to trigger a cache refresh.

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -130,7 +130,9 @@ class Node < ActiveRecord::Base
   def totalviews
     # disabled as impressionist is not currently updating counter_cache; see above
     # self.views + self.legacy_views
-    impressionist_count(filter: :ip_address) + legacy_views
+    cache("totalviews-#{self.id}", expires_in: 60.minutes, skip_digest: true) do
+      impressionist_count(filter: :ip_address) + legacy_views
+    end
   end
 
   def self.weekly_tallies(type = 'note', span = 52, time = Time.now)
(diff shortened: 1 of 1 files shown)


app/models/node.rb
```diff
@@ -130,7 +130,9 @@ def setup
   def totalviews
     # disabled as impressionist is not currently updating counter_cache; see above
     # self.views + self.legacy_views
-    impressionist_count(filter: :ip_address) + legacy_views
+    cache("totalviews-#{self.id}", expires_in: 60.minutes, skip_digest: true) do
```
seafr: Maybe `Rails.cache` should be used instead.

jywarren: aha! Let me try.

jywarren:
I guess i don't think this optimization is necessary anymore... I'm going to close it!

PR #2448, opened 2018-03-08
Add methods to User model for verifying user roles

seafr:
Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [x] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [x] code is in uniquely-named feature branch and has no merge conflicts
* [x] PR is descriptively titled

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software

We have a loose schedule of reviewing and pulling in changes every Tuesday and Friday, and publishing changes on Fridays. 

Thanks!

PR #2480, opened 2018-03-12
Features controller refactor

seafr:
No issue opened.

Thanks!

### DIFF
--- a/app/controllers/features_controller.rb
+++ b/app/controllers/features_controller.rb
@@ -19,9 +19,9 @@ class FeaturesController < ApplicationController
   end
 
   def edit
-    if current_user.role != 'admin'
+    if !current_user.admin?
       flash[:warning] = 'Only admins may edit features.'
-      redirect_to '/features'
+      redirect_to features_url
     else
       @node = Node.find params[:id]
     end
--- a/app/controllers/features_controller.rb
+++ b/app/controllers/features_controller.rb
@@ -12,9 +12,9 @@ class FeaturesController < ApplicationController
   end
 
   def new
-    if current_user.role != 'admin'
+    unless current_user.admin?
       flash[:warning] = 'Only admins may edit features.'
-      redirect_to '/features'
+      redirect_to features_url
     end
   end
 
--- a/app/controllers/features_controller.rb
+++ b/app/controllers/features_controller.rb
@@ -14,14 +14,14 @@ class FeaturesController < ApplicationController
   def new
     unless current_user.admin?
       flash[:warning] = 'Only admins may edit features.'
-      redirect_to features_url
+      redirect_to '/features'
     end
   end
 
   def edit
     if !current_user.admin?
       flash[:warning] = 'Only admins may edit features.'
-      redirect_to features_url
+      redirect_to '/features'
     else
       @node = Node.find params[:id]
     end
--- a/app/controllers/features_controller.rb
+++ b/app/controllers/features_controller.rb
@@ -28,7 +28,7 @@ class FeaturesController < ApplicationController
   end
 
   def create
-    if current_user.role != 'admin'
+    if !current_user.admin?
       flash[:warning] = 'Only admins may edit features.'
       redirect_to '/features?_=' + Time.now.to_i.to_s
     else
@@ -63,7 +63,7 @@ class FeaturesController < ApplicationController
   end
 
   def update
-    if current_user.role != 'admin'
+    if !current_user.admin?
       flash[:warning] = 'Only admins may edit features.'
       redirect_to '/features?_=' + Time.now.to_i.to_s
     else
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -137,6 +137,18 @@ class User < ActiveRecord::Base
     role == r
   end
 
+  def admin?
+    role == 'admin'
+  end
+
+  def moderator?
+    role == 'moderator'
+  end
+
+  def can_moderate?
+    admin? || moderator?
+  end
+
   def tags(limit = 10)
     Tag.where('name in (?)', tagnames).limit(limit)
   end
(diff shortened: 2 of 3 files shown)


seafr:
@jywarren this is related to the refactor in #2448

app/controllers/features_controller.rb
```diff
@@ -12,23 +12,23 @@ def embed
   end
 
   def new
-    if current_user.role != 'admin'
+    unless current_user.admin?
       flash[:warning] = 'Only admins may edit features.'
-      redirect_to '/features'
+      redirect_to features_url
```
jywarren: Is this auto-generated from the routes, then? Thanks!

seafr: Yes, `features_url` and `features_path`. The first is recommended for redirection as in contains the full URL.

seafr:
@jywarren, please not that the first 2 commits have already been merged (#2448).

app/controllers/features_controller.rb
seafr: I want to add that I rarely come across named routes in the codebase. And I don't see Rails' `link_to` method. Is there a specific reason for this?

jywarren: In general, I've felt that it is more readable for people who just do HTML (which is a lot of people) to just use `<a>` tags, and the `_url` Rails helpers are a little mysterious to people. I can see the advantage to having managed routes, but I think there's something good about having a literal link where you can see the URL, in a standard HTML format. 

But I'm open to discussion on it! What do you think?

seafr: I see your point, esp. regarding templates. However, controllers are just Ruby so named routes would probably suit them better. The idea is to enable any future change of URLs without breaking things, like what would happen in #2484 (/blog -> /wiki/stories). But I don't want to be disruptive here. Maybe you think it's better to keep things as they are (for now at least). Or maybe we should get input from other contributors too. Let me know your thoughts and I'll be glad to revert relevant changes back if needed.

jywarren: How about in order to move forward, we stick with static string routes for now, and then as you said 

> get input from other contributors too

That sounds great. I do think we have trouble with route maintenance -- #1841 was trying to think through that one -- and very occasionally routes break since they are not thoroughly tested :-(

jywarren: Thanks for thinking this through! I agree re: controllers being in ruby, but the named routes are just so mysterious to newcomers! If we could smooth that learning curve a bit I'd be more comfortable using them.

seafr: Yes, the routes file needs a bit of tidying up.
OK I'll use static strings for now. I agree that named routes can be a mystery for newcomers. Very relevant to our newcomer-friendly policy!
Last, this is not a promise, but I'll try to look into the routes issue and see if I can help. Anything you think you should mention?

jywarren:
Looks great, is this ready to go? :-)

seafr:
Yes, I guess so. But there has been no formal review by anybody, hence the review tag.

app/controllers/features_controller.rb
```diff
@@ -12,14 +12,14 @@ def embed
   end
 
   def new
-    if current_user.role != 'admin'
+    unless current_user.admin?
```
Gauravano: hi @seafr , in all other places, you have used `if !current_user.admin?`. how about following the same syntax here too, only if you want to?

app/models/user.rb
```diff
@@ -137,6 +137,19 @@ def has_role(r)
...
+
+  def moderator?
+    role == 'moderator'
+  end
+
+  def can_moderate?
+    # use instead of "user.role == 'admin' || user.role == 'moderator'"
+    admin? || moderator?
+  end
+
```
Gauravano: Nice work !!!

test/unit/user_test.rb
```diff
@@ -164,4 +164,19 @@ class UserTest < ActiveSupport::TestCase
...
+
+    moderator = users(:moderator)
+    assert moderator.moderator?
+    assert moderator.can_moderate?
+
+    basic_user = users(:newcomer)
+    assert_not basic_user.admin?
+    assert_not basic_user.moderator?
+    assert_not basic_user.can_moderate?
+  end
```
Gauravano: test seems fine to me

app/controllers/features_controller.rb
seafr: I used `if !` elsewhere due to the presence of the `else` clause. I find it hard to understand `else` after `unless`, else I'd use `unless` everywhere.

P.S. I have a feeling there is an overuse of *else* in the above text though.

app/models/user.rb
seafr: Thanks!

test/unit/user_test.rb
seafr: Thank you for review. :+1:

app/controllers/features_controller.rb
jywarren: yeah `unless/else` trips me up too. Interestingly -- I think double negatives also vary in use across languages!

jywarren:
Thanks @Gauravano for such a thorough review! Thank both of you for your thorough work! I also notice both of you helping others out a lot and with great attitude 😄 

Merging away! ✈️

Gauravano:
Thanks, Jeff . Just performing our part as a member of the PublicLab community !!!

seafr:
@jywarren thank you for the kind words. It's a pleasure to be part of the PL community. We are just trying to give back the welcome we received as newcomers. Ideal place to get into OSS.

app/controllers/features_controller.rb
seafr: Could you tell me why the current time is (sometimes) appended to the redirect URL?

PR #2495, opened 2018-03-14
Update docker-compose.yml for staging mail tests (WIP)

jywarren:
Re #2095!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
   - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then gem install danger danger-junit; fi'
 
 script:
-  - docker-compose run -e CI=TRUE web bash -c "GENERATE_REPORT=true rake test:all"
+  - docker-compose exec web bash -c "CI=TRUE GENERATE_REPORT=true rake test:all"
   - echo -e '<?xml version="1.0" encoding="UTF-8"?>' > output.xml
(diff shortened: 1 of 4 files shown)


jywarren:
Hmm, `ERROR: for web  Cannot create container for service web: conflicting options: host type networking can't be used with links. This would result in undefined behavior`

jywarren:
Did I put it in the wrong place?

jywarren:
@icarito what do you think? Thanks!

icarito:
Yeah this should work. Checking the logs.

icarito:
Oh I see. 
`ERROR: for plots2_web_1  Cannot create container for service web: conflicting options: host type networking can't be used with links. This would result in undefined behavior`
We'll have to drop the links but in that case we need to reach the database at `localhost` instead of `db`.

jywarren:
Hmm, still not going... thanks for chipping away at this though!!!

icarito:
Sorry I took so long. We might have to start tracking docker-compose variants for production and staging also. For now it works in testing, hopefully it will work in staging too.

icarito:
I have also setup a SMTP server listening on localhost:25 on staging - but come to think of it, it still shouldn't be reached at 127.0.0.1 - the reason for this change in the first place. I had tried running everything under the host network and this exposed the SQL port so it was a security flaw. So I put the db port directly in the rails container.

PR #2512, opened 2018-03-19
Offset deeplinks on headers by floating navbar width

patcon:
When following a deeplink to a specific wiki header, the header is obscured under the floating header/navbar. The tactic used in this PR -- separating linked "chain" icon from a separate invisible anchor that is vertically offset -- should fix it. (Unsure how this looks on different screen sizes.)

screenshot of hidden header

Sorry, untested except in browser inspector :)

Reference: [StackOverflow: offsetting an html anchor to adjust for fixed header](https://stackoverflow.com/questions/10732690/offsetting-an-html-anchor-to-adjust-for-fixed-header)


jywarren:
Awesome, thank you!

Issue #2421, opened 2018-02-28
Brainstorming search results that are autosuggested and shown on results page


ebarry:
Update: this is a long conversation and there are some next steps being broken out. Please continue to use this issue for brainstorming! Thanks :)

Original issue continues below:

### Please describe the problem

The system by which autosuggested results seems to choose and rank content suggestions is mysterious, and seems like a black box.  

Autosuggested results have a display limit of 15 assorted content types, but do not provide an overview of Public Lab resources on a topic. 

> What did you expect to see that you didn't?

I expect to understand what the results mean. 

### Please show us where to look

The Search box in the menu bar

jywarren:
It is actually a black box! Full text search is a complex problem and we solve it with the "fulltext" module of MySQL, our database system; some pretty arcane (but thorough) documentation is here: https://dev.mysql.com/doc/refman/5.7/en/fulltext-search.html

It does seem we can tune/adjust it, though. There is, for example, a "natural language" option which attempts to algorithmically determine "relevance" -- https://dev.mysql.com/doc/refman/5.7/en/fulltext-natural-language.html

We use this fulltext feature on this line:

https://github.com/publiclab/plots2/blob/47b30044aa26027f7eb083bce56dd607a5f73d02/app/models/node.rb#L29

It does look like we could "turn on" natural language mode by making that say:

```ruby
    Revision.where('MATCH(node_revisions.body, node_revisions.title) AGAINST(? IN NATURAL LANGUAGE MODE)', query)
```

We may also need to then add ordering by `relevance` -- so, i /think/ that would be:

```ruby
Revision.select('node_revisions.body, node_revisions.title, MATCH(node_revisions.body, node_revisions.title) AGAINST(' + query.to_s + ' IN NATURAL LANGUAGE MODE) AS score')
  .where('MATCH(node_revisions.body, node_revisions.title) AGAINST(? IN NATURAL LANGUAGE MODE)', query)
```

It might take some testing out. 

Would you like to try this out? I have to point out that I do NOT know what will happen. The documentation for "natural language" says:

> Every correct word in the collection and in the query is weighted according to its significance in the collection or query. Thus, a word that is present in many documents has a lower weight, because it has lower semantic value in this particular collection. Conversely, if the word is rare, it receives a higher weight. The weights of the words are combined to compute the relevance of the row. This technique works best with large collections.

****

As to the second issue, -- 

> ...but do not provide an overview of Public Lab resources on a topic.

[...]

PR #2518, opened 2018-03-20
Search sorting

jywarren:
Working on #2421 search sorting by different parameters


jywarren:
Hmm, @publiclab/solvers  @publiclab/reviewers -- if anyone /really loves/ ActiveRecord and mysql, I'm having some trouble getting these tests to pass... I've done a lot of consolidation and cleanup here, but somehow the basic `Node.search` calls aren't picking up simple test queries anymore. Anything obvious I've missed?

jywarren:
Got it!!! I think, 🤞 -- i had used `.where()` instead of `.order()` -- fixed, let's see!

ebarry:
Awesome!!!!!!!

PR #2545, opened 2018-03-26
Modify edit in Users controller. Add functional tests for edit and …

seafr:
…update.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -63,7 +63,7 @@ class UsersController < ApplicationController
   def edit
     @action = "update" # sets the form url
     if params[:id] # admin only
-      @user = DrupalUser.find_by(name: params[:id]).user
+      @user = User.find_by(username: params[:id])
     else
       @user = current_user
     end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -42,7 +42,7 @@ class UsersController < ApplicationController
   end
 
   def update
-    if current_user
+    if current_user     # Is this really needed? See before filters above.
     @user = current_user
       @user.attributes = params[:user]
       @user.save({}) do |result|
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,6 +1,6 @@
 class UsersController < ApplicationController
   before_filter :require_no_user, :only => [:new]
-  before_action :require_user, :only => [:edit, :update]
+  before_filter :require_user, :only => [:update]
   before_action :set_user, only: [:info, :followed, :following, :followers]
 
   def new
@@ -74,11 +74,11 @@ class UsersController < ApplicationController
       @user = current_user
       @drupal_user = current_user.drupal_user
     end
-    if current_user.uid == @user.uid #|| current_user.role == "admin"
+    if current_user && current_user.uid == @user.uid #|| current_user.role == "admin"
       render :template => "users/edit"
     else
       flash[:error] = I18n.t('users_controller.only_user_edit_profile', :user => @user.name).html_safe
-      redirect_to "/profile/"+current_user.name
+      redirect_to "/profile/"+@user.name
     end
   end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -68,11 +68,9 @@ class UsersController < ApplicationController
   def edit
     @action = "update" # sets the form url
     if params[:id] # admin only
-      @drupal_user = DrupalUser.find_by(name: params[:id])
-      @user = @drupal_user.user
+      @user = DrupalUser.find_by(name: params[:id]).user
     else
       @user = current_user
-      @drupal_user = current_user.drupal_user
     end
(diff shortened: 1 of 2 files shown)


seafr:
@jywarren this a WIP. Please don't merge yet.

SidharthBansal:
@seafr I have changed the tag to in-progress from review me.

seafr:
@SidharthBansal I'd still like to receive reviews in the meantime though. Thanks.

seafr:
Thank you for the kind words @sagarpreet-chadha :+1:

seafr:
Thank you @Gauravano :+1:

app/controllers/users_controller.rb
```diff
@@ -68,11 +68,9 @@ def update
   def edit
     @action = "update" # sets the form url
     if params[:id] # admin only
```
seafr: Could anyone please tell me what the comment means here?

sagarpreet-chadha: Hi @seafr .
I guess if you will see comment in line number 77 , then it seems this route was made to give **admins** the power to edit profile of **any** user .
This route : 
screen shot 2018-03-28 at 11 49 56 pm

sagarpreet-chadha: I think currently admins do not require the ability to edit the profile of any user , hence it is commented out .

app/controllers/users_controller.rb
```diff
@@ -42,7 +42,7 @@ def create
   end
 
   def update
-    if current_user
+    if current_user     # Is this really needed? See before filters above.
```
seafr: I think this is taken care of by the before filter on line 3.

app/controllers/users_controller.rb
seafr: Yes that's what I was thinking. So I think this comment shoud be removed now as it is misleading.

sagarpreet-chadha: Ok...so if you are not signed in and you go to **/profile/edit** page , we get page not found error . Do you think we should be redirected to the **login-screen** or may be show a flash ?
Would you like to open a first-timer issue ?

seafr: According to the routes file, **/profile/edit** routes to `users#edit`, which is the same *edit* action above. Let me test this and get back to you.

seafr: Yes, I think the user should be redirected to **/login**. The flash should also be displayed. 
It's a good idea to open an FTO issue for that as it involves a tiny modification. I'll try it.

sagarpreet-chadha: Awesome !

seafr: I did it (not without glitches though :( ). Thank you for the suggestion.

app/controllers/users_controller.rb
seafr: @sagarpreet-chadha please let me know your thoughts on this one.

sagarpreet-chadha: I think it is important in fact i think it should be : 
`if current_user.email == params[:user][:email]` 
because if someone sends a **PATCH** request to **/users/update** and somehow sets the value of `current_user` then he/she can change the bio of any user . I am kinda new to Web Security and apologies if it may sound stupid :) 
What do you think @seafr ?

seafr: I don't think they can set the value of `current_user` in this case. They can't tamper with the `session`. What you said is true of `params` though.
Let's see what @jywarren has to say about this.

PS I'm not too familiar with web security either.

sagarpreet-chadha: Oh you are right , I didn't see the code below...one cannot tamper the session . Makes sense  :smile: !

seafr: So what do you think, should it be removed?

sagarpreet-chadha: This kind of makes the code more readable from the perspective of a first-timer .  What do you think ?

jywarren:
we could leave an inline comment like:

def update # login required

next to actions that are allowed, but reduce the number of lines of code...

On Fri, Mar 30, 2018 at 3:02 PM, Sagarpreet Chadha <notifications@github.com

seafr:
@jywarren I've implemented the change you suggested and I think this PR is ready to be merged now.
@sagarpreet-chadha you might want to do a final review. Thank you.

app/controllers/users_controller.rb
```diff
@@ -41,38 +41,31 @@ def create
...
-      redirect_to "/profile/"+@user.name
     end
   end
 
   def edit
     @action = "update" # sets the form url
     if params[:id] # admin only
-      @drupal_user = DrupalUser.find_by(name: params[:id])
-      @user = @drupal_user.user
+      @user = DrupalUser.find_by(name: params[:id]).user
```
jywarren: Ah, here we could just to User.find_by(username: _____) instead of using drupaluser at all.

seafr: Done! :+1:

jywarren:
Super work here! Thanks a lot!!

Issue #2388, opened 2018-02-23
PLANNING ISSUE:Facebook Authentication


SidharthBansal:
Planning issues for #683 
* [x] Add omniauth gem
* [x] Add omniauth-facebook gem
* [x] Add fiagro gem
* [x] Define routes
* [x] Add OmniAuth Configuration to initializer
* [x] Install Openssl
* [x] Add OmniAuth Capability to User Model
* [x] User_session_controller update
* [ ] Views update for login
* [ ] Views update for sign in
* [ ] Write tests for the login through fb
* [x] Write Documentation for the fiagro gem
* [x] Write Documentation for omniauth 
* [x] Write Documentaion for the omniauth-facebook
* [x] Write Documentaion for openssl

jywarren:
This is awesome!!! 👍 👍 👍 @ebarry 

What is the minimum chunk of this we could do in a single PR? Perhaps the gems, plus OmniAuth in initializer and user model, and routes, but not changing the existing login sequence or views YET. What do you think? All the back-end stuff.

jywarren:
Sorry, i don't mean put it all back into one big PR. Just that we can do some things that don't affect current workflows (like login, signup) and merge them in with tests, and know that they work well. Then in a followup PR we can start "Turning it on" for people logging into the site for real -- exposing the back-end work. Does this make sense?

jywarren:
Like, we could potentially enable this for login first, but not for signup. That would let us test this progressively in chunks.

One question is -- will existing users be able to log in with Facebook to their pre-existing accounts? Just wondering. 

Thanks so much!

SidharthBansal:
Sorry I could not do much last two weeks due to my mid sems.
Yeah, we will be doing this into chunks of separate prs so that it will be a quick, easy and a small merge.

SidharthBansal:
Regarding the tests. I think we will be able to write them at the end of the full functioning of login and signup

SidharthBansal:
First of all, I will make a simple facebook login compatible with the current publiclab.org
Then I shall add functionality so that the users who have the facebook account or the login on public labs directly can fetch their same profile in either way around.
I hope this makes sense.

SidharthBansal:
``` Then in a followup PR we can start "Turning it on" for people logging into the site for real -- exposing the back-end work ```
Yeah first of all we will test all the functionality on the local then on production, then we will make the frontend so that new users will be able to start logging in from there.

jywarren:
[...]

PR #2556, opened 2018-03-29
Identity model created

SidharthBansal:
#2388 

Thanks!

### DIFF
--- a/app/models/identity.rb
+++ b/app/models/identity.rb
@@ -1,0 +1,3 @@
+class Identity < ActiveRecord::Base
+  belongs_to :user
+end
--- a/app/models/identity.rb
+++ b/app/models/identity.rb
@@ -1,3 +1,4 @@
 class Identity < ActiveRecord::Base
   belongs_to :user
+  validates :user_id, presence: true
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -33,6 +33,7 @@ class User < ActiveRecord::Base
   has_many :passive_relationships, class_name: 'Relationship', foreign_key: 'followed_id', dependent: :destroy
   has_many :following_users, through: :active_relationships, source: :followed
   has_many :followers, through: :passive_relationships, source: :follower
+  has_many :identities, dependent: :destroy
 
   validates_with UniqueUsernameValidator, on: :create
   validates_format_of :username, with: /\A[A-Za-z\d_\-]+\z/
@@ -41,7 +42,7 @@ class User < ActiveRecord::Base
   before_save :set_token
   after_destroy :destroy_drupal_user
 
-  def self.search(query)
+  def self.search(query) 
     User.where('MATCH(username, bio) AGAINST(?)', query)
   end
 
@@ -216,15 +217,15 @@ class User < ActiveRecord::Base
       (1..span).each do |day|
           time = Time.now.utc.beginning_of_day.to_i
           days[(time-day.days.to_i)] = Node.select(:created)
-                                           .where(uid: self.uid, 
+                                           .where(uid: self.uid,
                                                   type: 'note',
-                                                  status: 1, 
-                                                  created: time - (day-1).days.to_i..time - (day - 2).days.to_i) 
+                                                  status: 1,
+                                                  created: time - (day-1).days.to_i..time - (day - 2).days.to_i)
                                            .count
       end
       days
   end
- 
+
 
   def weekly_comment_tally(span = 52)
     weeks = {}
--- a/db/migrate/20180329085153_create_identities.rb
+++ b/db/migrate/20180329085153_create_identities.rb
@@ -1,0 +1,11 @@
+class CreateIdentities < ActiveRecord::Migration
+  def change
+    create_table :identities do |t|
+      t.string :provider
+      t.string :uid
+      t.references :user, index: true
+
+      t.timestamps
+    end
+  end
+end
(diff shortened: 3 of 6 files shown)


SidharthBansal:
@publiclab/reviewers I have ran rake db:migrate on my local machine but travis shows that the migrations are pending. How shall I run migrations on travis?

app/models/identity.rb
```diff
@@ -0,0 +1,3 @@
+class Identity < ActiveRecord::Base
+  belongs_to :user
```
seafr: You should probably use a validation here to check for the presence of `user_id`

SidharthBansal: Oh thanks @seafr 
I did change it.

jywarren:
You can update the db/schema.rb.example file and it'll use the latest schema. Looking good!

SidharthBansal:
Hi @jywarren still the Travis is failing what I need to do??

Gauravano:
hi @SidharthBansal , you just need to copy the checksum into the top of schema.rb.example and test will pass.

SidharthBansal:
Thanks @Gauravano .
@jywarren, @publiclab/reviewers  can you kindly review it?

jywarren:
Looks awesome! Should I merge now, or are you going to add some unit tests? Or will that be in a follow-up PR?

SidharthBansal:
I will add them in following prs

jywarren:
Hi, I was thinking a bit here -- is there a reason to create a new database model for this, or do you think we could store this information in a `UserTag`? I think the key is - is any of the information now or in the future "secret" -- like, an auth token, or a secret key -- because then a unique table makes sense. 

Sorry this took me a few days of thinking to articulate this question. But I don't think going the other route would be too much trouble given how well you've modularized your planning. What do you think, and are there pros/cons to the `UserTag` approach? Would it make your life easier or harder?

SidharthBansal:
Hi @jywarren there is a need to create a new table to hold the provider (Facebook,twitter ,google ,github etc) ,uid for the id given by the provider of the account and the user-id to associate the identity with the user model. We would login through identity. We require multiple providers aka identities for a single user. When try to login or logout we will use these models. 
We can't do it in usertag model. If I am not wrong, usertag is the tags which are displayed on the profile page rt? We require those tags for all the providers. We require a separate model for login and logout.

Yes we require future information the uid given by the provider and the name of the provider. The app secret and the app id we need to write them in the config files.
The user tag model has uid , value ,updated at and created at fields in the schema. We do require other parameters uid given by the provider and name of provider as in multi party login two or more author service may give us the same uid. Does this makes sense?
Also, there would be ability to delete the identity at time of unlinking of the account and making new one at on linking new provider.

jywarren:
ohhhhh wow! There's so much info there!  Exciting, although also want to be
considerate of people /not/ wanting to share it all.

I think we could add fields like `time_zone:______` and `location:______`
and `facebook-picture:_______` if we want, but one thing I'm thinking of
that would /save/ us time is if the profile tag UI were what we used to
store all of these, making it unnecessary to develop a separate Identities
storage/management UI. I think most of these I can see are ones we could
adapt to a UserTag in one way or another.

The picture URL maybe wouldn't fit into a key:value format, now that I
think of it. But we can do that in a follow-up; we don't have to come up
with a way to use *every* piece of data here, at least not all at once.

What do you think? we could automatically import a few of these into
UserTags, then prompt people to remove the ones they don't want to share?

Pros/cons to that idea?

SidharthBansal:
Actually the user will grant the permissions for these fields. Some of the user grant some permission while the other grant other permissions. Also the various informations which we can get are different in case of different providers like we discussed the email case of twitter.
If we are using usertags then we need to migrate the database table to use these different information whenever needed in future.

SidharthBansal:
One problem is if any malicious user(like me :smile: ) creates a usertag named "facebook:jwarren" same as yours then the OAuth logic will get corrupted. 
image

We have given the capability to the user to add user tags we also need to change that.

jywarren:
Aha -- so if we were to use `UserTag` we'd have to perhaps forbid creation of ones prefixed with `facebook:___` `twitter:_____` etc UNLESS via the Oauth process. Could that work, hypothetically? 

DB optimization is a good point, but relative to the complexity of searching the NodeTag table, this is not so bad, and indexing helps a lot. 

What do you think of the above? We already prevent some types of powertags on nodes on these lines: 

https://github.com/publiclab/plots2/blob/979532802910dc035847b0efdc6e101e77fa7a6f/app/models/node.rb#L835-L845

SidharthBansal:
Yeah this way, we can do. I was also thinking of this way. Yeah, indexing helps you are right. So, let's close this PR and first quickly write validation for provider:uid first. 
Thanks

SidharthBansal:
Hi, I think the code which you have posted is not working
image
image
@jywarren

SidharthBansal:
I spotted the error. It took a little time. We are not using the can_tag method in the create action of the usertag controller but we are using it in the tagcontroller. So, we are able to create tags with the 'format:rss' etc which we should not be able to make.

jywarren:
Aha - great catch. I think we should implement can_tag in UserTag as well!

jywarren:
Great, deep work here, @SidharthBansal !!!

SidharthBansal:
Yeah working on this today. Was not able to do it because I was having exams.

Issue #2445, opened 2018-03-07
Need to add "Ascending" sort for `#of uses` on publiclab.org/tags 


ebarry:
Hi! This *might* be a first timers only issue, i'm going to tag it `FTO-candidate`

### Please describe the problem (or idea)

> What problem could this idea solve?

I love the new tag sorting, but i noticed that even though we have the "up-down" arrow icon next to `# of uses`, the sort function only works in one direction. Watch the gif below to see how the `# of uses` is only able to sort the tags in Descending order, demonstrated by attempting to change the sort order after sorting alphabetically by the `tag` column: 

![tag-sorting](https://user-images.githubusercontent.com/161439/37112777-73ddd980-2211-11e8-83c9-eaf2b468a9ee.gif)

> What did you expect to see that you didn't?

I expect to be able to sort in ascending and descending order by `# of uses` and in fact by all columns. 

### Please show us where to look

https://publiclab.org/tags
https://publiclab.org/tags?sort=uses

### What's your PublicLab.org username?

> This can help us diagnose the issue: liz

Thank you!!!!!!

jywarren:
Hi! This will involve changes to the controller on lines like this:

https://github.com/publiclab/plots2/blob/ca820a87e083ed7aff95e6c903732d2d4fca7949/app/controllers/tag_controller.rb#L22

Where we'll want to determine order via a URL parameter, like: 

`https://publiclab.org/tags?sort=uses&order=desc`

We can access the order with the variable `params[:order]` within the controller. So let's indicate order with `params[:order] == "desc"` and then update the template to enable this toggling in the table header code here:

https://github.com/publiclab/plots2/blob/ca820a87e083ed7aff95e6c903732d2d4fca7949/app/views/tag/index.html.erb#L26-L33

We'll want to be able to toggle sorting ascending/descending for each of the columns! Any help appreciated, thanks!!!

Gauravano:
hi @jywarren @ebarry , I guess this issue relates to #2352 .

ebarry:
wow yes, @Gauravano -- thank you for directing my attention at this GREAT PR in progress by @sukhbir-singh !!! It's very exciting. Feel free to close this issue in favor of  #2352 .

Gauravano:
[...]

PR #2557, opened 2018-03-29
Fixed issue#2445 added both direction sort order

sukhbir-singh:
Fixes issue #2445

Thanks!


app/controllers/tag_controller.rb
```diff
@@ -11,6 +11,12 @@ def index
 
     @title = I18n.t('tag_controller.tags')
     @paginated = true
+
+    order_string = "count DESC"
```
shauncarland: I would put this into a private function within the controller.  e.g. 
```
def order_string
  params[:order] === "asc" ? "count ASC" : "count DESC"
```

I would also perhaps abstract `count ASC` and `count DESC` into constants themselves.  e.g. `COUNT_ASC = "count ASC"` and `COUNT_DESC = "count DESC"`.  Or put them also into private functions, e.g. 

```
def count_asc
  "count ASC"
end 
```

This way, if we ever need to change `count DESC` to some other string, we'll have it all in one place, either in the constant or in the function, and we won't have to chase it around everywhere. :)

app/views/tag/index.html.erb
```diff
@@ -21,8 +21,12 @@
 
   <table class="table">
     <tr>
-      <th><a href = "<%= tags_path %>?sort=name"><%= t('tag.index.tag') %></a> <i class="fa fa-arrows-v"></i></th>
-      <th><a href = "<%= tags_path %>?sort=uses"><%= t('tag.index.number_of_uses') %></a> <i class="fa fa-arrows-v"></i></th>
+      <% order_type = "desc" %>
```
shauncarland: Can we set `order_type` in the controller like we're setting `@tags`?  I think the controller would be a better place to do that.  In general, from my understanding, logic should be kept out as much as possible from the ERB template.  It should primarily stay in the controller and model.

app/controllers/tag_controller.rb
```diff
@@ -19,31 +25,39 @@ def index
...
     @tags = Tag.joins(:node_tag, :node)
       .select('node.nid, node.status, term_data.*, community_tags.*')
       .where('node.status = ?', 1)
       .where('community_tags.date > ?', (DateTime.now - 1.month).to_i)
       .group(:name)
-      .order('count DESC')
+      .order(order_string)
       .paginate(page: params[:page], per_page: 24)
     elsif @toggle == "name"
+    order_string = "name DESC"
```
shauncarland: Yeah, if you do what I was saying, creating a function for defining `order_string`, this would be come more DRY. 🎉

app/controllers/tag_controller.rb
```diff
@@ -19,31 +25,39 @@ def index
...
     @tags = Tag.joins(:node_tag, :node)
       .select('node.nid, node.status, term_data.*, community_tags.*')
       .where('node.status = ?', 1)
       .where('community_tags.date > ?', (DateTime.now - 1.month).to_i)
       .group(:name)
-      .order('name')
+      .order(order_string)
       .paginate(page: params[:page], per_page: 24)
     else
+      order_string = "name DESC"
```
shauncarland: Same here

sukhbir-singh:
@shauncarland I have made all the required changes. so please review again. 
I will keep the DRY thing in mind while making further contributions to public labs. :smile:

jywarren:
This looks good, can I get one more +1 from @publiclab/reviewers? And thank you @shauncarland, are you interested in being a reviewer? If so would you mind reading over http://publiclab.org/software-outreach? Cool!!

app/controllers/tag_controller.rb
```diff
@@ -436,4 +440,15 @@ def location_modal
...
+  private
+
+  def get_order_string
+    if params[:search] || @toggle == "uses"
+      params[:order] == "asc" ? "count ASC" : "count DESC"
+    else
+      params[:order] == "asc" ? "name ASC" : "name DESC"
+    end
+  end
+
```
shauncarland: [nit] extra line

app/controllers/tag_controller.rb
```diff
@@ -11,6 +11,10 @@ def index
 
     @title = I18n.t('tag_controller.tags')
     @paginated = true
+    @order_type = params[:order] == "desc" ? "asc" : "desc"
+
+    order_string = get_order_string()
```
shauncarland: [nit] you don't need a `()` after a method invocation in ruby, unless you are passing in arguments. :)

shauncarland: Also, I would just rename `get_order_string` and replace that function with `order_string`.

e.g. 
```
def get_order_string 
 ...
end
```

Because ruby methods are invoked simply by calling them, you can keep the `order_string` the same as below.  all `order_string` will do there is call the `order_string` function, returning the order string.  thus, there is no need to assign the value of `get_order_string` to a variable.  just call that function wherever you were using `order_string`.

Does that make sense?  It's a little bit hard to explain, but im happy to go further in detail if you need it. :)

shauncarland:
Good job keeping the code DRYer @sukhbir-singh!  DRY is a good practice to keep in mind, whether you're working in this repo or in another one.  Here's a good resource on why DRY is important:  https://softwareengineering.stackexchange.com/questions/103233/why-is-dry-important

app/controllers/tag_controller.rb
shauncarland: (and make sure to keep `order_string` a private method! :D )

app/controllers/tag_controller.rb
```diff
@@ -11,6 +11,10 @@ def index
 
     @title = I18n.t('tag_controller.tags')
     @paginated = true
+    @order_type = params[:order] == "desc" ? "asc" : "desc"
```
shauncarland: I'm a little bit confused... if the `order` param is `desc`, we want to set the `@order_type` to `desc`...?

sukhbir-singh: @shauncarland The logic behind this is that, we have to invert the order when a user click on sorting header again. So lets say if current shown order is according to "name descending" then we want front end to request on url with order type be ascending (opposite of descending). I hope you got my point :smile:

app/controllers/tag_controller.rb
sukhbir-singh: The idea of removing get_order_string and directly calling order_string is really cool :sunglasses: Thanks !!

sukhbir-singh:
@shauncarland @jywarren I did all the remaining changes to this PR. Please review !!

app/controllers/tag_controller.rb
```diff
@@ -436,4 +438,14 @@ def location_modal
   def gridsEmbed
     render layout: false
   end
+
+  private
```
shauncarland: I would just do `private :order_string` at the end of the `order_string` method.  That will make `order_string` private.  The way you have it now, anything method below this line will be private.  Let's just make `order_string` private so no future developer needs to worry about it. :)

app/controllers/tag_controller.rb
shauncarland: 🎉 🎉 🎉 🎉 🎉 🎉

app/controllers/tag_controller.rb
sukhbir-singh: But all other controllers in this project have used same pattern to specify private methods in a class.
You can check out the home_controller.rb, application_controller.rb files, etc. So i think its better to do it the same way. : smile

ebarry:
I'm so excited to have this code running live on the website soon!

jywarren:
Looks great! Completed!!! Thanks everybody :-)

PR #2565, opened 2018-03-30
Replace footer with db-based feature

jywarren:
Using this:

https://gist.github.com/jywarren/78c77ed3e47b64a5ed3e5461cff03b92


Gauravano:
hi @jywarren, just wondering have you pushed this to stable as I just came across translation dropdown in the footer. Thanks  :smile:

translate1

jywarren:
I did replace it -- and the dropdown is now part of a db-stored feature.
Shall I remove german for now? I'm not sure how to manage this exactly --
maybe we need to break that into a separate code block below the footer, so
it's back in git-tracked code... thoughts?

PR #2569, opened 2018-03-30
Emojify!

jywarren:


jywarren:
We could do `ActionController::Base.helpers.emojify(str)` if need be!

PR #2581, opened 2018-04-02
Add methods to User model to change/test status [WIP]

seafr:

Thanks!


seafr:
There is one method missing to complete the trio of whether the user is banned, moderated or none of these. I am not sure 'active' would do and I need help determining an appropriate name for this. Please suggest some names. Thanks.

SidharthBansal:
@seafr how about 'normal' ?

seafr:
@SidharthBansal I've thought about this. And about 'active' too. The "problem" is when you realise what their opposites are.

app/models/user.rb
```diff
@@ -159,6 +159,22 @@ def can_moderate?
     admin? || moderator?
   end
 
+  # User status
+  # 0: banned
+  # 5: moderated
+  def mark_as!(new_status)
+    self.status = 0 if new_status == :banned
+    self.status = 5 if new_status == :moderated
```
siaw23: let's use ternary operator here for simplicity if this is only going to be about 2 states.

shauncarland: Did you consider using an enum instead of adding this definition?
https://bendyworks.com/blog/rails-enum-sharp-knife

shauncarland:
Did you consider using an enum instead of adding this definition?
https://bendyworks.com/blog/rails-enum-sharp-knife

Adding something like `enum status {banned: 0, moderated: 5}` to the class model.  If you do that, you'll get access to `user.banned?`, `user.moderated?` methods automatically thanks to Rails

app/models/user.rb
seafr: In fact, 1 more state is missing. Check my first comment above (and please help find a good name :) ). 
Thank you for review.

test/unit/user_test.rb
```diff
@@ -184,6 +185,16 @@ class UserTest < ActiveSupport::TestCase
     assert_not basic_user.moderator?
     assert_not basic_user.can_moderate?
   end
+
+  test 'user statuses' do
```
Gauravano: if possible, keep it status only, totally your call

app/models/user.rb
```diff
@@ -159,6 +159,22 @@ def can_moderate?
     admin? || moderator?
   end
 
+  # User status
+  # 0: banned
+  # 5: moderated
+  def mark_as!(new_status)
```
Gauravano: hi @seafr,  changes look good, I would agree with 'normal' as status as for comments we use 'normal' only

jywarren:
Thanks for the careful and thorough teamwork here! It's a tough naming choice and I was thinking of `basic` for a bit, but I think a moderated user is `basic` too -- the word we need is to contrast with a moderated status. So, I agree that `active` seems best here. 😄 👍

seafr:
Sorry for delay guys. 

So, I'm going with @jywarren's choice of `active`. Thank you all for valuable input.

seafr:
@jywarren I'm stuck here. Can you please take a look?

app/models/user.rb
```diff
@@ -41,6 +41,12 @@ class User < ActiveRecord::Base
   before_save :set_token
   after_destroy :destroy_drupal_user
 
+  # User status
+  # 0: banned
+  # 1: active
+  # 5: moderated
+  enum status: {banned: 0, active: 1, moderated: 5}
```
jywarren: Wow, i have never used `enum` like this. Can you link me to some documentation to follow how it works?

jywarren:
Ah, it looks like the tests are finding a `logged out` error instead of a `The user 'rob' has been placed in moderation...` message! Can you figure out why?

```
 FAIL["test_users_are_logged_out_and_alerted_when_moderated,_and_notes_are_not_accessible", Minitest::Result, 8.989677048000033]
 test_users_are_logged_out_and_alerted_when_moderated,_and_notes_are_not_accessible#Minitest::Result (8.99s)
        --- expected
        +++ actual
        @@ -1 +1 @@
        -"The user 'rob' has been placed in moderation; please see <a href='https://www.example.com/wiki/moderators'>our moderation policy</a> and contact <a href='mailto:moderators@www.example.com'>moderators@www.example.com</a> if you believe this is in error."
        +"You must be logged in to access this page"
        test/integration/moderate_and_ban_test.rb:68:in `block in <class:ModerateAndBanTest>'
```

jywarren:

gauravano:
Hi @jywarren do you think this PR is required as we have deprecated drupal user? Let me know,if there is still some potential in this PR for optimizing codebase, I may be able to help. Thanks!

jywarren:
This enum approach to user status does look pretty useful and efficient. Perhaps we can try to implement just that part?

gauravano:
Hi everyone, thanks for working and contributing to this PR. @siaw  did some cleanup in https://github.com/publiclab/plots2/pull/4894/files which cover the same goal as this PR, so I think, we can close this one. Feel free to open this PR again(actually, prefer to open new PR as we have many changes since then), if you think, there's still more for improvement.  

Thanks all! 

cc @jywarren

PR #2609, opened 2018-04-11
Update _tags.html.erb to refine tags popup

jywarren:


jywarren:
@Fastie this'll resolve the link issue as I described. Thanks for reporting!

PR #2603, opened 2018-04-09
Add token documentation

stefannibrasil:
This PR adds the token documentation discussed in #2018 about people or bots commenting via API using the token system

PR #2614, opened 2018-04-12
Refine token documentation

stefannibrasil:
As discussed in #2603, this PR refine the token documentation, adding the format for the POST request and adding the location of the token API in the codebase.

Closes #2603


stefannibrasil:
@publiclab/reviewers could you please review here? :)

PR #2621, opened 2018-04-16
Add test for questions count on methods page

jywarren:


jywarren:
@publiclab/reviewers - this is odd -- our Danger test failures for `assert_select` seem to be showing this error instead of indicating what was matched/not matched. Has anyone else seen this?

test/functional/wiki_controller_test.rb
```diff
@@ -558,6 +558,16 @@ def teardown
     assert_not_nil :topics
   end
 
+  test 'should get methods page and show questions count' do
+    nodes(:method).add_tag('questions:spectrometer', users(:bob))
```
sagarpreet-chadha: @jywarren ...see this line : 

https://github.com/publiclab/plots2/blob/34e78b8d42a6820cd17aa5633c4b3d18b8d72fbd/app/models/node.rb#L777

undefined local variable ,  right ?

sagarpreet-chadha:
Okay , so i think `nodes(:method)` does not contains tag `method` so i added this line : 
            **`nodes(:method).add_tag('method', users(:bob))`** .

Then i used **id** instead of **class** in the `<span>` tag in ` _notes.html.erb:44`

And then in assert_select i changed to : 
**`assert_select "#questions-count-#{nodes(:method).id}", "#{nodes(:method).questions.count} questions"`**   .

and the test passed ! What do you think ?

sagarpreet-chadha:
@jywarren ....when i used **id** instead of **class** , the travis showed the correct error of what was matched/unmatched. 
I guess when we used class then **assert_select** returns array of classes (in our case array has 1 item) and hence there was a different error .

jywarren:
Oh, this is awesome. Do you think you could add a commit to this PR? 

Thank you so much for helping out! 🙌 🤝

jywarren:
Oh wow you already did! My browser tab just hadn't refreshed. AMAZING TEAMWORK!!! 🎈🎈🎈🎈🎈🎈🎈🎈🎈

PR #2634, opened 2018-04-18
Add check for emoji inside grid tags

jywarren:
This should fail! Then we can fix it.

### DIFF
--- a/test/unit/node_test.rb
+++ b/test/unit/node_test.rb
@@ -21,6 +21,8 @@ class NodeTest < ActiveSupport::TestCase
     revision = node.latest
     revision.body = ':cat:'
     assert_equal "<p>🐱</p>\n", revision.render_body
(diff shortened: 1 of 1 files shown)


jywarren:
Hmm, does anyone LOVE regex (am I asking this again?) 

We have a situation where it tries to parse `[notes:question:test]` and then `:question:` gets emojified!

https://stackoverflow.com/questions/17405235/negative-lookbehind-and-square-brackets has some tips on "lookaheads"...

But I can't get this to work: http://rubular.com/r/AP5yqtYLvp

I can't figure out negative lookaheads, it seems! I can't get it to work with `X`s either: http://rubular.com/r/vvi56gC17U

@publiclab/reviewers any suggestions? Thank you!

jywarren:
Fixed in #3721!!

PR #2639, opened 2018-04-19
Added validation errors

SidharthBansal:
As asked in #2556 by @jywarren 

Thanks!


app/models/node.rb
```diff
@@ -851,6 +851,14 @@ def can_tag(tagname, user, errors = false)
...
     elsif tagname.split(':')[0] == 'redirect' && Node.where(slug: tagname.split(':')[1]).length <= 0
       errors ? I18n.t('node.page_does_not_exist') : false
+    elsif tagname.split(':')[0] == 'facebook'
+      errors ? "Only Oauth can create such tags" : false
+    elsif tagname.split(':')[0] == 'twitter'
+      errors ? "Only Oauth can create such tags" : false
+    elsif tagname.split(':')[0] == 'google'
+      errors ? "Only Oauth can create such tags" : false
+    elsif tagname.split(':')[0] == 'github'
+      errors ? "Only Oauth can create such tags" : false
```
seafr: By using `["facebook", ...].include? tagname.split(':')[0]` instead we can squeeze 8 lines into just 2 and avoid much repetition. What do you think?

SidharthBansal: @seafr I want to know whether we want the same error message for all 4 cases or different for these. If same message is OK. I will do the required patch.
Thanks

seafr: Ok, I see your point. I thought this was the final version as the `in progress` label was removed.

SidharthBansal: @jywarren what do you suggest?

jywarren: Hm, i like @seafr's suggestion! Let's do it!

seafr: @SidharthBansal I suggest moving all 4 options into 1 `elseif` as I said above. You can also customise the error message to something like: 
`"... create the #{tagname...[0]} tag"` 
if you want to make it different for each case.

SidharthBansal: Ok, updating in a minute. Thanks both of you

jywarren:
This looks good! Is it ready for merging? 👍 😄

SidharthBansal:
Yeah

jywarren:
I'm going to test CodeClimate integration by closing and reopening... let's see!

jywarren:
@publiclab/reviewers - I tried enabling CodeClimate -- it's configured in `.codeclimate.yml` in the root directory. It's a little strict... what do folks think about having it on? I can turn it off if it seems too unfriendly.

jywarren:
Maybe we should turn off some of the checks -- that way we can "ease in" more gradually. As files get cleaner, we can turn on more checks? https://github.com/publiclab/plots2/blob/master/.codeclimate.yml

jywarren:
Also notice I was able to log into CodeClimate by clicking "Details" and then approve it there despite the issues it complained about. So we can work around this... my main concern is that it looks too intimidating to newcomers -- and I wish it offered more advice to people on how to improve their code, rather than just complaining about it!

jywarren:
So now that I've approved it, to see the original error it complained about, click **Show all checks** and click on Details for CodeClimate.

jywarren:
OK, merging!! 😀

Issue #2647, opened 2018-04-21
Urgent: Home down because of invalid record (?)


icarito:
```
Started GET "/" for 47.31.242.72 at 2018-04-21 19:46:30 +0000
Processing by HomeController#home as HTML
Read fragment views/feature_home-carousel (0.3ms)                                                                                                              Read fragment views/feature_home-banners (0.1ms)                                                                                                               Read fragment views/home-activity (0.1ms)                                                                                                                      Read fragment views/trending-tags (0.1ms)                                                                                                                        Rendered dashboard/_node_comment.html.erb (9.2ms)                                                                                                              Rendered dashboard/_node_meta.html.erb (3.2ms)                                                                                                                 Rendered dashboard/_node_wiki.html.erb (7.5ms)                                                                                                                 Rendered dashboard/_node_wiki.html.erb (2.0ms)                                                                                                                 Rendered dashboard/_activity.html.erb (22.7ms)
  Rendered home/home.html.erb within layouts/application (25.9ms)
Completed 500 Internal Server Error in 37ms (ActiveRecord: 3.4ms)

ActionView::Template::Error (undefined method `author' for nil:NilClass):
    3:                                                                                                                                                             4:     <% if node.is_a?(Node) %>
    5:       <div class="header-icon"><i class="fa fa-book"></i></div>
    6:       <p class="meta"><span><%= t('dashboard._node_wiki.new_page_by') %></span> <a href="/profile/<%= node.latest.author.name %>"><%= node.latest.author.name %></a></p>                                                                                                                                                   7:     <% else %>
    8:       <div class="header-icon"><i class="fa fa-pencil"></i></div>
    9:       <p class="meta"><%= t('dashboard._node_wiki.new_edit_by') %> <a href="/profile/<%= node.author.try(:name) %>"><%= node.author.try(:name) %></a></p>
  app/views/dashboard/_node_wiki.html.erb:6:in `_app_views_dashboard__node_wiki_html_erb___2292190031077852178_69979555127840'
  app/views/dashboard/_activity.html.erb:67:in `block in _app_views_dashboard__activity_html_erb__2360015329309354006_69979556424080'
  app/views/dashboard/_activity.html.erb:65:in `each'
  app/views/dashboard/_activity.html.erb:65:in `each_with_index'
  app/views/dashboard/_activity.html.erb:65:in `_app_views_dashboard__activity_html_erb__2360015329309354006_69979556424080'
[...]

PR #2648, opened 2018-04-21
test revision spam flows

jywarren:
This should fail for the same reason as #2647 -- spamming the sole revision of a wiki page! 

Looking for, to resolve this:

redirect to the Revisions view with a notice that: 

`You can't delete the last remaining revision of a page; try deleting the wiki page itself (if you're an admin) or contacting moderators@publiclab.org for assistance.`

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -190,6 +190,14 @@ class AdminController < ApplicationController
 
   def mark_spam_revision
     @revision = Revision.find_by(vid: params[:vid])
+    @node = Node.find_by(nid: @revision.nid)
+
+    if(@node.drupal_node_revisions_count == 1)
+      flash[:warning] = "You can't delete the last remaining revision of a page; try deleting the wiki page itself (if you're an admin) or contacting moderators@publiclab.org for assistance."
+      redirect_to @node.path
+      return
+    end
+    
     if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
       if @revision.status == 1
         @revision.spam
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -192,12 +192,12 @@ class AdminController < ApplicationController
     @revision = Revision.find_by(vid: params[:vid])
     @node = Node.find_by(nid: @revision.nid)
 
-    if(@node.drupal_node_revisions_count == 1)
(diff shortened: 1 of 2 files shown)


jywarren:
Awesome, I added a test for the `flash[:warning]` -- let's see!

jywarren:
Wow, our revision spam tests are not great, i.e. had been non-existent! We should build them out.

Gauravano:
I guess it's complete now :smile: !!! What do you think @jywarren ?

sagarpreet-chadha:
I think we cannot used `assert_select` in integration tests or can we ?
What do you think @Gauravano ?

@jywarren ... All tests pass if we remove assert_select statement .

Gauravano:
hi @sagarpreet-chadha, we can use assert_select in integration tests, actually in our codebase we are using. hmm, ok I am gonna try one more time with assert_select.

Souravirus:
@Gauravano assert_select gives some problems in rails 4.2. So, what I have used during rails 4.2 update is that I css_selected the element and saw what was the number of elements by storing it in a variable. 
Something like this 
selector = css_select 'div.note'
 assert_equal selector.size, 15
So, here just print the selector.size once to see the number of elements and then use assert_equal. I guess this is helpful!!

Souravirus:
In your guess I guess this will not work just remove the class name alert warning and test with alert. That will work as ruby tests will find assert with the sentence you want to find. Something like this:

`assert_select ".alert" , "× You can't delete the last remaining revision of a page; try deleting the wiki page itself (if you're an admin) or contacting moderators@publiclab.org for assistance."`

This would work.

sagarpreet-chadha:
Also one thing ...as that node does not have any tag , so there would be two alerts on the page after we delete that node , hence `assert_select` will select two alerts (no tag alert and only 1 revision alert) .
I wonder if we have to loop over this selection and check for both the alerts , but we can open another PR for this .
What do you think @Gauravano , @Souravirus ?

Souravirus:
Did that work what I said above?

jywarren:
I think we have to keep the alert-warning class because it's a specific yellow Bootstrap style. I think for now we can skip the assert_select for now although I appreciate the idea! 

Wow, a real collaboration here :-)

jywarren:
I'm going to merge especially because we should do our best to prevent another error from the original issue. 

Thanks everyone!!!! Great work. :raised_hands:

PR #2675, opened 2018-04-28
enable method=post for data-url image upload at /post

jywarren:


jywarren:
Whoa weird errors for what I added... wonder why?!

Issue #2716, opened 2018-05-11
Search API error after installation


E-L-T:
### Please describe the problem (or idea)

After installation we ran the tests, and we got this error : 
"701 tests, 2061 assertions, 0 failures, 9 errors, 0 skips"

> What did you expect to see that you didn't?
no errors

### Browser, version, and operating system
Firefox 59.0.2 (64 bits), mac os 10.13.3

E-L-T:
After running the tests, we got 9 errors. Here is the first one. The 8 others concern the same file, app/models/node.rb
![capture d ecran 2018-05-11 a 21 40 46](https://user-images.githubusercontent.com/29131404/39943806-7dae8140-5564-11e8-82ac-1d28aa0ac5f9.png)

jywarren:
Hi, sorry it took me a little while but I left a comment on your PR with a proposed fix. I hope this gets us closer, THANKS SO MUCH!!! 👍 👍 👍

jywarren:
Glad to hear in #2717 that this isn't a problem anymore!

PR #2717, opened 2018-05-11
fix a functional test concerning node model (#2716)

E-L-T:
`fixes #0000`

We know that tests don't pass but we are trying to fix this in this pull request

Hi @publiclab/reviewers , can you check this ?

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -43,9 +43,7 @@ class Node < ActiveRecord::Base
           .order(orderParam)
       end
     else
-      nodes = Node.limit(limit)
-        .where('title LIKE ?', '%' + input + '%', status: 1)
-        .order(orderParam)
+      nodes = Node.where(status: 1).order(orderParam)
     end
   end
 


jywarren:
Hi, was this test giving you trouble locally?

E-L-T:
Hi @jywarren , 
yes i had a problem with this test locally. Actually, when i launched the tests locally, i had 9 errors (I opened and issue : #2716)
After rewriting some code, I got only 1 failure and 4 errors. 
It is surprising the Travis CI build passed anayway, isn't it ?

jywarren:
Sorry i was away traveling and got behind on this! 

It is surprising! But local tests run on sqlite (to make installation easier) and Travis tests on MySQL -- we should do better to standardize or be sure both pass. I'll merge this so we get a bit closer, thank you so much! We'd love some help getting all to pass.

app/models/node.rb
```diff
@@ -43,9 +43,7 @@ def self.search(query, order = :default)
           .order(orderParam)
       end
     else
-      nodes = Node.limit(limit)
-        .where('title LIKE ?', '%' + input + '%', status: 1)
```
jywarren: Ah, but we do need this extra title matching -- otherwise we won't match the title against search input. This is just backup code if MySQL/MariaDB are not installed, but I still want to be sure it works. 

I think the issue here is we never defined the `limit` parameter! Not sure how that happened, but if you re-add the original code but remove the `.limit(limit)` we should be OK here. Thanks so much for looking into this!!!

E-L-T:
Hey, it seems that all the tests pass locally now. Good job !
So you can close this pull request (and the corresponding issue: #2716 )?

jywarren:
Oh really! Gosh, i wonder what it was. Maybe now on Rails 5? I'm happy to close this but would love a bit more info if you have it! Thanks!

E-L-T:
Yes, I think the changes have been made during the rails 5.0 upgrade.
It concerns the search method of node model that was mising parameters. Method which is called in app/services/typeahead_service.rb.

PR #2737, opened 2018-05-20
User should not be able to create Oauth based user_tags

SidharthBansal:
Part of #2388

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -53,8 +53,8 @@ class AdminController < ApplicationController
         PasswordResetMailer.reset_notify(user, key).deliver_now unless user.nil? # respond the same to both successes and failures; security
       end
 
-      flash[:notice] = "#{user.name} should receive an email with instructions on how to reset their password. If they do not, please double check that they are using the email they registered with." 
-      redirect_to "/profile/" + user.name
+      flash[:notice] = "#{user.name} should receive an email with instructions on how to reset their password. If they do not, please double check that they are using the email they registered with."
+      redirect_to URI.parse("/profile/" + user.name).path
     end
   end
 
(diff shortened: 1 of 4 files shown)


SidharthBansal:
image
It is working properly.
@jywarren where do I need to write documentation for Oauth, Omniauth gems and figaro gem. That I forget to write in previous commits.I could not find the exact place where the information will suit best.
Thanks

jywarren:
This looks super! I do wonder if we should leave these on until we enable the OAuth system -- we could comment out the line where it checks and just put `true` for now -- because people are currently still adding their accounts to their profiles, without Oauth linking. OR we could have a more specialized tag like `oa-github:_____` so that we don't collide with this use case. Maybe the second solution makes more sense and we can merge this feature without commenting out?

Great work!!!

SidharthBansal:
I am in favour of second option 'oa-github:____'.

SidharthBansal:
image

SidharthBansal:
image

jywarren:
This looks great. Seems like we just need to resolve the `octokit` bug now, in https://github.com/publiclab/plots2/pull/2738#issuecomment-393226759 -- see if you can coordinate with Vidit!

SidharthBansal:
Yeah sure.

jywarren:
Ah, i think this was affected by the previous issue of Dockerfile config. If you rebase it over the latest master branch, it should pass!

SidharthBansal:
@jywarren Can we ignore CodeClimate these 2 errors. I have fixed other CodeClimate errors?
Thanks

jywarren:
Yes, that's fine! And if we want, we can try to look at how to tweak the CodeClimate suggestions to be less strict? Or give more people access to approve them?

I updated the settings so it'll leave a comment, too - like this: https://docs.codeclimate.com/docs/github-pull-requests#section-summary-comments

Is it helpful or not welcoming enough?

jywarren:
I think we could relax some of the CodeClimate suggestions using this guide and our config file: https://docs.codeclimate.com/docs/advanced-configuration

jywarren:
This looks perfect -- i hope it's all good? I merged it :-) 👍 👍 👍

SidharthBansal:
thanks

PR #2738, opened 2018-05-21
Add sidekiq gem

ViditChitkara:
part of #2104 

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [ ] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [ ] code is in uniquely-named feature branch and has no merge conflicts
* [ ] PR is descriptively titled
* [ ] PR body includes `fixes #0000`-style reference to original issue #
* [ ] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software

We have a loose schedule of reviewing and pulling in changes every Tuesday and Friday, and publishing changes on Fridays. 

Thanks!

PR #2745, opened 2018-05-25
Openssl added for https

SidharthBansal:
@jywarren new Facebook settings allow the login through the localhost only if we are using https. I found that we can provide the https in development by creating the certificates by openssl gem and running the 
``passenger start --ssl --ssl-certificate localhost.crt --ssl-certificate-key localhost.key --ssl-port 3000`` command instead of passenger start.

This blog might be helpful. https://www.devmynd.com/blog/rails-local-development-https-using-self-signed-ssl-certificate/

Also, I have changed the installation guidelines a little. Check it out.


jywarren:
I think this is OK and if it does not interrupt the existing SSL config on production (@icarito can tell us it's ok or not?) then we should be fine and I"m OK merging this in. I think we're still seeing some issues with the travis containers so that may hold us up a bit first. Thanks!

SidharthBansal:
OK great to hear that. Please check this twice on production before merging @icarito. Also, we need to notify the existing developers about the command to launch the passenger server is changed.

SidharthBansal:
@jywarren  I tried to understand the Travis issues but was not able to. Any help is greatly appreciated.

Gemfile
```diff
@@ -130,3 +130,4 @@ gem 'omniauth-github', '~> 1.1', '>= 1.1.2'
 gem 'activerecord-tableless'
 gem 'figaro'
 gem 'sanitize'
+gem 'openssl', '~> 2.0.0.beta.1'
```
jywarren: Could we move this into the `development` gems only, since we won't be using this in production? Thanks!!!

jywarren:
And if you could change the docs -- the README or corresponding file -- but just say that `passenger start` is fine for default, but /if you're working on Facebook integration or something that needs SSL/, then you should use the command `passenger start --ssl --ssl-certificate localhost.crt --ssl-certificate-key localhost.key --ssl-port 3000` -- then we should be fine and won't complicate things for most developers. Thanks!!!

Gemfile
SidharthBansal: yes for sure.

SidharthBansal:
@icarito I have made the changes suggested by @jywarren can you please check the behaviour at production and development once?
Thanks

SidharthBansal:
image
image

README.md
```diff
@@ -56,7 +56,8 @@ Please read and abide by our [Code of Conduct](https://publiclab.org/conduct); o
 5. Make a copy of `config/database.yml.sqlite.example` and place it at `config/database.yml`
 6. Run `rake db:setup` to set up the database
 7. Install static assets (like external javascript libraries, fonts) with `bower install`
-8. Start rails with `passenger start` from the Rails root and open http://localhost:3000 in a web browser.
+8. By default, start rails with `passenger start` from the Rails root and open http://localhost:3000 in a web browser.
+But if you are willing to establish a secure connection in development mode then use 'passenger start --ssl --ssl-certificate localhost.crt --ssl-certificate-key localhost.key --ssl-port 3001'. Then open up https://localhost:3001. Add security exceptions from the advance settings of the browser. Secure connection is needed for OAuth etc.
```
jywarren: Let's say just `(for local SSL work, see [SSL](#ssl+in+development) below)` and have an "SSL in development" section -- is that OK? just to keep the basic install as simple as possible to keep newcomers in mind!

localhost.crt
```diff
@@ -0,0 +1,21 @@
+-----BEGIN CERTIFICATE-----
```
jywarren: Could we store these in a subfolder somewhere -- the more files we add to the root directory, the more it pushes down the README, you know? Maybe in `config`?

Thank you!!!!

jywarren:
Great work here, almost done!!!

SidharthBansal:
@jywarren can you please review it again? I have fixed the patches. 
Also, I want to know 
1. where to add the details of the 'Figaro' gem, 'omniauth' gem which we have installed previously.
2. do I need to make a developers account as a community account on facebook, twitter, google and Github so that we both can share password of that account and you can change the settings, later on, whenever required. This way the future developers on the public lab can change the settings. Or will you make the settings in your account and notify me with the app_id and app_secret.
Another option is giving you as admin role from my account.
I am in favour of the first option.
Thanks

jywarren:
awesome!

jywarren:
Now just need to resolve `octokit` bug --  https://github.com/publiclab/plots2/pull/2738#issuecomment-393226759

jywarren:
Ah, i think this was affected by the previous issue of Dockerfile config. If you rebase it over the latest master branch, it should pass!

SidharthBansal:
@jywarren  can we merge this?

jywarren:
Perfect, thank you!!!! 👍 👍 👍

PR #2825, opened 2018-06-13
Change like count to avoid flickering to fix issue #2656

### LINKED ISSUE #2656: non urgent: inconsistencies on count of Likes VS list of Likers on notes/wikis

> What happened just before the problem occurred? 

The number of Likes seems to flickr upon page reloading. I observed it yesterday on publiclab.org/evaluation, but actually can't reproduce what i documented in this video today: 
https://publiclab.org/system/images/photos/000/024/696/original/like_count.mov

Perhaps you can see it on a wiki page like https://publiclab.org/balloon-mapping. Try reloading the page and watching the Like count, and then seeing how many names are listed when you click the expanded view.

etalamante:
This fixes #2656 by changing the like count method `show` on the [like](https://github.com/etalamante/plots2/blob/fixing_issue_2656/app/controllers/like_controller.rb) controller to exclude banned/moderated users.

Also, I added two tests on the [like_controller_test](https://github.com/etalamante/plots2/blob/fixing_issue_2656/test/functional/like_controller_test.rb) file to replicate the bug and to show the like count is working.

With those fixes, the like count would not flick after the page reloads and the number of likes will correspond to the number of users shown on the expanded view.

### DIFF
--- a/app/controllers/like_controller.rb
+++ b/app/controllers/like_controller.rb
@@ -11,7 +11,7 @@ class LikeController < ApplicationController
   # return a count of likes for a given node
   # This does not support non-nodes very well
   def show
-    render json: Node.find(params[:id]).cached_likes
+    render json: Node.find(params[:id]).likers.length
   end
 
   # for the current user, return whether is presently liked or not
--- a/app/controllers/like_controller.rb
+++ b/app/controllers/like_controller.rb
@@ -11,7 +11,7 @@ class LikeController < ApplicationController
   # return a count of likes for a given node
   # This does not support non-nodes very well
   def show
-    render json: Node.find(params[:id]).likers.length
+    render json: Node.find(params[:id]).likers.count
   end
 
   # for the current user, return whether is presently liked or not
--- a/test/functional/like_controller_test.rb
+++ b/test/functional/like_controller_test.rb
@@ -83,7 +83,7 @@ class LikeControllerTest < ActionController::TestCase
     UserSession.create(User.find(2))
     note = Node.where(type: 'note', status: 1).first
 
-    get :create, id: note.id #first liked
+    get :create, params: { id: note.id } #first liked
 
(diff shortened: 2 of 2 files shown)


etalamante:
@jywarren, is there something that needs to be changed on the PR or the way to solve it was wrong?

jywarren:
Hi, I think this is fine - but I think we originally thought to use the cached value as an optimization. Do you think this would be a tough query if there were lots of likes or something? It seems unlikely. Do you have the query it generates from your local log, so we can think about that?

Thanks and sorry it took us a couple days to get back to you!

etalamante:
Hello, I also think it shouldn't be a problem using the `likers` collection, but I honestly do not have much experience in optimization, so I can't assure you that : /, sorry.

Regarding the query, are you referring to the log that explains how the query was executed? If I understood correctly, I attached you the [test.log](https://github.com/publiclab/plots2/files/2106957/test.log) file with both `moderated_likes` and `moderated_not_included` tests, where the second is the one you would want to watch to see how the query worked.

Thanks for answering and don't worry, I mentioned you to be sure if I did something wrong.

jywarren:
Hi! Apologies again for slow reply... I love this PR for how simple and readable it is, and the great tests. 

Can we switch to .count instead of . length, which I think will use SQL counting, instead of ruby, so it should be slightly more efficient? Thanks!!

etalamante:
No worries and thanks for the feedback : D. I will update the PR with your request and make an adjustment on the test to use keyword arguments.

etalamante:
@jywarren if it is something that I need to change, I will get it done. Pending.

jywarren:
Looks super! Thank you both!!!

jywarren:
🎉 !!!

etalamante:
Thanks to you @jywarren and @Gauravano for all your feedback and support :+1:

PR #2771, opened 2018-06-05
Google Authentication 

SidharthBansal:
MAIN POINTS TO PONDER IN THIS PULL REQUEST FOR #2676
* [x] User able to link/unlink their providers account
* [x] Log In
* [x] Log out
* [x] Sign Up / Sign in
* [x] Consideration of corner cases listed in https://publiclab.org/notes/bansal_sidharth2996/02-20-2018/gsoc-proposal-oauth-authentication

Facebook api is working on a test app. But they are changing their settings very often nowadays. So, firstly I am sending pr for Google. Most of the steps are common between different providers. So, once we are done with this. We will quickly have other providers.
@jywarren I will not be able to break into parts this pr. It will be a little longer pr as all code is linked heavily to each other.

This pr is in progress.

PR #2855, opened 2018-06-19
Documentation added for OAuth

### LINKED ISSUE #2848: Documentation for OAuth
# Checklist
Add the following to /doc/OmniauthLogin.md
* [x] Define the different variables defined in #2771 
* [x] Write the explanation for the different method used in the #2771 
* [x] Omniauth logic explanation
* [x] Minimal setup instructions
* [x] Links to code segments
* [x] links to external documentation such as where to generate a key
* [x] Testing documentation
Part of  #2676

SidharthBansal:
Fixed #2848


SidharthBansal:
@jywarren kindly review this too

jywarren:
This is really nice. Shall we update it to reflect the environment variables instead of using `application.yml`?

SidharthBansal:
This file is important because we don't want the developer to push his app_id and app_secret onto the github. We have enabled .gitignore to not include application.yml file while we push onto plots2 and we have enabled figaro too. It will be easiest to write the keys and values in application.yml in such cases.

jywarren:
Hi, Sidharth - we've set this up to use environment variables from the host
system directly in your code in
https://github.com/publiclab/plots2/pull/2771, so there should be no
situation where developers need to input their secrets into a file. If
you'd like to rewire this so that it looks more like this file:

https://github.com/publiclab/plots2/blob/master/config/initializers/recaptcha.rb

That's fine too, but you can see there also we don't explicitly put real
codes in... they're passed in through environment variables. This really
reduces the chances that a secret will be pushed. Does this system make
sense? @icarito may be able to explain a bit more as well. Thank you for
bearing privacy in mind!

SidharthBansal:
OK. I am pushing some changes to the documentation.
Thanks

jywarren:
This looks great. Is it ready? I'd ask perhaps that you change `he/she has` to `they have` and generally use `they` -- thank you! 

VERY good documentation!!!

SidharthBansal:
Thanks for the suggestion. I will keep this in mind.
Will push the required changes tonight

SidharthBansal:
@jywarren I have changed the file according to the changes suggested. 
It is ready for merge

jywarren:
Awesome!!

PR #2859, opened 2018-06-19
Google button added to header for OAuth

### LINKED ISSUE #2849: Google OAuth header view
Create view for header for google provider to sign up/log in
Part of #2676 and #2771

### LINKED ISSUE #2866: OAuth user interface design on login screen
@SidharthBansal - i wanted to share how Oauth looks on some other sites, so once we confirm it's perfectly working, we have a good workflow for login. See this example:

screen shot 2018-06-20 at 9 42 43 am

Note a couple things. First, since OAuth is easier, it's presented first. Then, I wonder if we can have the remainder of the non-OAuth screen be hidden until you interact with the first input... so as to keep things simple. Once you click or type in it, then the rest is shown. That way also it's clear that you can ignore it if you're using OAuth. 

What do you think?

SidharthBansal:
Fixes #2866  
image

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -183,7 +183,10 @@
                   </p>
 
                   <p><a href="/reset/"><%= t('layout._header.login.reset_password') %></a></p>
-
+                  <h5>Log In with</h5>
+                  <div class="loginbutton" >
+                    <a href="/auth/google_oauth2" id="connect-google"><span class="btn btn-default"><i class="fa fa-google" style="font-size:20px;color:red"></i></span></div></a>
+                  </div>
                 <% end %>
               </li>
             </ul>


jywarren:
This looks good, if you're not logged in maybe we should say "log in with..."?

SidharthBansal:
I wrote ``connect with`` instead of ``log in`` because on clicking on the google icon 
* New client can sign up
* Existing client can log in.
We use the same route(/auth/google_oauth2) for both signing a new account via provider and for logging in. So, I could only find one suitable word ``Connect``
If you still wish to make it login tell me, I will send the patch.

jywarren:
I think it's ok to say log in, because to a new person it's similar to
"logging in with your Google account" and doesn't seem like making a new
one, does that sound ok?

SidharthBansal:
OK, I am changing it in 5-7 minutes.
@jywarren have you sent the mail to Venkat?
Thanks

jywarren:
I haven't yet but it's on my list this week and I have a note about the deadline too (July 6!)

SidharthBansal:
Thanks @jywarren. It is July 6 IST that means July 5 at your time.
Thanks a lot

jywarren:
Gotcha, thanks!

Issue #2676, opened 2018-04-29
PLANNING ISSUE: Multi party authentication


SidharthBansal:
This is the checklist for the OAuth Login 
# Implementation common to all providers
* [x] Add omniauth gem #2381
* [x] Add fiagro gem #2531
* [x] Install Openssl #2745 
* [x] Make localhost work under https connection #2745
* [x] Write Documentaion for openssl #2745 
* [x] Write Documentation for omniauth and related gems #2855 #2848
* [x] Define routes #2531
* [x] Modify UserTag to acts as identiy model #2737 #2639
* [x] Add OmniAuth Capability to User Model #2771
* [x] Link provider to a user #2771
* [x] Delete provider from a user #2771
* [x] User_controller update #2771
* [x] User_session_controller update  - #2771
* [x] Wiki Page https://publiclab.org/oauth

# Google Provider
* [x] Add omniauth-google-oauth2 gem #2771
* [x] Add OmniAuth Configuration to initializer #2771 
* [x] OAuth testing hash #2771 
* [x] Write user session controller tests for the login through Google #2771 
* [x] Write integration tests for the login through Google #2771 
* [x] Set up developers app (discussion in #2771)
* [x] Docker file changes #2771 
* [x] header  #2866 
* [x] /profile/edit page  #2850
* [x] login page  #2851 
* [x] sign up page  #2852 

# Github Provider
* [x] Add omniauth-github gem #2686
* [x] Add OmniAuth Configuration to initializer #2873
* [x] Set up developers app #2856 
* [x] Docker file changes #2875
* [x] OAuth testing hash #2874
* [x] Write user session controller tests for the login through Github  #2915 
* [x] Write integration tests for the login through Github #2914
* [x] header  #2975 
* [x] /profile/edit page #2975 
* [x] login page  #2975 
* [x] sign up page  #2975 

# Twitter Provider
* [x] Add omniauth-twitter gem #2686 
* [x] Set up developers app #2893
* [x] Docker file changes #2947 
* [x] Add OmniAuth Configuration to initializer #2948
* [x] OAuth testing hash #2961
* [x] Write user sessions controller tests for the login through Twitter #2962
* [x] Write integration tests for the login through Twitter #2963
* [x] header  #2986
* [x] /profile/edit page #2986
* [x] login page #2986
* [x] sign up page  #2986

# Facebook Provider
* [x] Add omniauth-facebook gem #2381 
* [x] Set up developers app #2867
* [x] Docker file changes #2960
[...]

PR #2873, opened 2018-06-20
Add OmniAuth Configuration to initializer for Github

SidharthBansal:
Fixes ``Github`` Provider's ``Add OmniAuth Configuration to the initializer`` listed in the checklist #2676


SidharthBansal:
These initializer changes are for accessing the auth hash from the next provider github which we are going to tackle. 
I have chosen Github because it's developer's account settings is quite easy. Thus, we can figure out how to make other providers use the existing authentication system made in #2771.

jywarren:
Cool, is this ready to go? It looks OK!

PR #2916, opened 2018-06-25
dep update for sprockets

jywarren:


jywarren:
Strange, this did not resolve the dependency. I'll add the `~>` too. #2918

PR #2918, opened 2018-06-25
sprockets ~> 3.7.2

jywarren:


jywarren:
@icarito i don't get it but we still see an error although:

https://github.com/publiclab/plots2/blob/ea0268cad586540e7a5e6ae50a5d0f110a741e00/Gemfile.lock#L396

And this PR seems to be failing from bad formatting? A bit confused here.

jywarren:
Oh good, the notification went away. I guess it took a little while to show up? Weird but all good!

Issue #2913, opened 2018-06-25
Add autocompleting usernames with @ symbol


cheneyshreve01:
### Please describe the problem (or idea)

As a part of making the markdown editor more friendly, this will add autocompleting usernames with @
Implementation will be similar to emoji Auto complete #2832

Issue #2665, opened 2018-04-27
Error on new wiki page creation with emoji


jywarren:
@NiklasJordan said:

> yes I tried that too ;-) Thanks to create this page for me.

> But for this issue I go to https://publiclab.org/wiki/first-contribution

![bildschirmfoto 2018-04-27 um 17 07 51](https://user-images.githubusercontent.com/715033/39369818-8f5b5daa-4a3d-11e8-9a5c-6cfb45934c07.png)

> Then I put my content on it...

![bildschirmfoto 2018-04-27 um 17 08 28](https://user-images.githubusercontent.com/715033/39369848-a2c1d072-4a3d-11e8-83f9-1575a98057b9.png)

> ...press "publish" and get this error page:

![bildschirmfoto 2018-04-27 um 17 09 10](https://user-images.githubusercontent.com/715033/39369888-ba460600-4a3d-11e8-966c-3a43d5ff2bcb.png)

> I am off for the next two hours. But I try to check my phone for your response! Hope this help to find the bug.

****

Let's try to reproduce this -- @Gauravano has already. I'll look through logs but we should try to make a functional test that catches this...

Thanks all!

jywarren:
Maybe if we can get the full text we can try filtering out the sections
that are causing trouble...

Gauravano:
ok, I will try this

jywarren:
@NiklasJordan - can you send us the full text in a Gist using http://gist.github.com so we can try to post it? It's strange but there seem to be characters in the text causing the app to choke... not your fault but we'd like to try to reproduce it to fix it! Thanks for your help with this!

NiklasJordan:
Hej @jywarren, sorry for the late answer. Sure, I've created the gist here: https://gist.github.com/NiklasJordan/da02086827b6f01c0429bacf8cc5ea96

jywarren:
awesome, thank you!!!

namangupta01:
I got it why this is happening. This is happening because of the using of emojis. Emojis take four byte to store data in mysql and is of **utf8mb4** encoding which is a utf8 that take 4 byte. Normally by default it uses 3 byte to store data and use normal utf8 encoding. We can use utf8mb4 for this particular table to support emojis.

jywarren:
Great investigative work! @icarito - is there any downside to changing the
encoding of the revisions table to accommodate emojis?

In the meantime, we could replace the emojis with `:smiley:` format
[...]

PR #2925, opened 2018-06-26
[WIP] (please do not merge yet) Add autocompleting usernames with @ symbol #2913

cheneyshreve:
Hi  @publiclab/reviewers, @ViditChitkara 

I was able to get this working using jquery's atwho library. I was not able to get it to work using the emoji autocomplete approach. If you can give me more details on that and prefer to have it implemented differently, I'm happy to revise this. I do think I can get the Emoji approach to work with the hashes (I was planning on opening a different issue for that) because you can use the hash emoji, so that one should be very similar to the smiley emojis. 

The reason I've marked this PR as WIP is because, when I did an upstream fetch and rebase of the master (which is not reflected in this current branch), I get the error 

```
ActionView::Template::Error (couldn't find file 'noty/lib/noty.js' with type 'application/javascript'
```

Checked in these paths:
Even though the appropriate files seem to be in-place:

```
app/assets/javascripts/application.js
//= require noty/lib/noty.js
//= require noty_notification.js
//= require jquery-confirm/js/jquery-confirm.js
```

And I have run bundle update, bundle install, and restarted the server. Any idea why this error still persists or have you gotten this error? I'm not sure if you can still merge this PR with the branch being behind master because I have not rebased this branch per the error above. 

here is the demo of how the current branch/changes work to see if you want to keep it or maybe have me do some revisions, thanks so much for your help. 

demo_atwho


cheneyshreve:
@ViditChitkara Thank you!! Bower install noty did the trick.

ViditChitkara:
> here is the demo of how the current branch/changes work to see if you want to keep it or maybe have me do some revisions, thanks so much for your help.

The autocomplete for usernames is working good I guess. However atwho's default styling is being overridden. Can you explore something about atwho's manual styling? @jywarren any ideas on this??

cheneyshreve:
Hi @ViditChitkara, @jywarren . @ViditChitkara thanks for the tip on noty, I was able to fetch upstream and rebase changes for this branch. I will look into styling and see if I can find anything, but yes, let me @jywarren if you have  any ideas about better styling.  Thanks both :)

cheneyshreve:
Hi @ViditChitkara, @jywarren I had not yet added atwho to application.css. now the styling works as intended :) 
demo_atwho

ViditChitkara:
Awesome!!

jywarren:
Hi! I'm wondering if it could be solved in a way very similar to @ViditChitkara's solution using the `horsey` library:

https://github.com/cheneyshreve/plots2/blob/74e894ff92813ce6d7b630c6423768ee72440882/app/views/comments/_form.html.erb#L78-L86

I think you can likely refactor your code just a little bit so that we have a similar solution. Do you think that would be OK? I love your implementation though!

One thing I worry a bit about is autocompleting any user name from the whole database -- we have a lot of users (like, 300k+)! Instead, we might use the API to narrow the results:

https://github.com/publiclab/plots2/blob/master/doc/API.md

For example you might start typing @jywarren with https://publiclab.org/api/srch/profiles?srchString=jy

The Horsey library is here: https://bevacqua.github.io/horsey/

This is looking really good! Thanks for tackling it -- i'm happy to offer more suggestions too!

cheneyshreve:
@ViditChitkara, @jywarren  ahh, I think see what you mean now. I originally thought I couldn't use the Emoji approach because I couldn't find the @ for emoji, but it looks like I should be able to find a solution using just the horsey library and plucking the usernames, if I'm interpreting this right, and also limit the dropdown of users by setting a limit. I will get to work on the revision, thanks to both of you for the feedback :)

jywarren:
awesome, and keep up the great work!

cheneyshreve:
hi @jywarren 
I'm trying to get the horsey textarea example from the demo site to work, they give the code below at https://bevacqua.github.io/horsey/ 

horsey(document.querySelector('textarea'), {
  source: [{ list: [
    { value: '@michael', text: 'Michael Jackson' },
    { value: '@jack', text: 'Jack Johnson' },
    { value: '@ozzy', text: 'Ozzy Osbourne' }
  ]}],
  getText: 'text',
  getValue: 'value',
  anchor: '@'
});

I first tried adding that code snippet to /comments/_form.html.erb, checking both 'textarea' and 'textarea#text-input' to see if I typed in the @, if it would populate the input list as it does in the demo, but it doesn't. Also, when that code snippet is added, the emoji autocomplete stops working.

I also tried to edit the one above to reflect the actual username data, using the code below:
/application_helper.rb

def user_names_list
    usernames = User.all.collect(&:username)
    list = []
    usernames.each do |name|
      val = "@#{name}"
      list << { value: val, text: name }
    end
    [{ list: list }]
end

Returns this running locally, so it's the same format as the textarea output on the horsey demo site.

[{:value=>"@admin", :text=>"admin"}, {:value=>"@moderator", :text=>"moderator"}, {:value=>"@user", :text=>"user"}, {:value=>"@cheneyshreve", :text=>"cheneyshreve"}]

And to render:
/comments/_form.html.erb

    <% usernames = user_names_list %>
    horsey(document.querySelector('textarea#text-input'), {
      source: <% usernames %>,
      getText: 'text',
      getValue: 'value',
      anchor: '@',
      limit: 10
    });
    
But no success on that either yet.  Any ideas? Thanks for any feedback you can give :)

jywarren:
Maybe @viditchitkara could offer some help?

cheneyshreve:
Hi @ViditChitkara, @jywarren  Unfortunately, I'm still unable to get things working with horsey. I doubled back on atwho just to see if I could do both the autocomplete username and autocomplete tags without interfering with emoji, and it's easy to add multiple callouts with a configuration setting. I'll double back on horsey again too ( should be same case there I'd imagine) and see if I can figure that out, just haven't had luck with that yet. I was hoping playing around with the other library might give me insight into horsey. 

demo_atwho_v2

with code to ensure limits on dropdown

```
    names = <%= raw User.pluck(:username).compact.to_json %>;
    tagnames  = <%= raw Tag.pluck(:name).compact.to_json %>;
    var mentions_config = {
      at: "@",
      data: names,
      limit: 6,
      tpl: "<li data-value='@${names}'>${names}</li>"
      },
      hashtags_config = {
      at: "#",
      data: tagnames,
      limit: 6,
      tpl: "<li data-value='#${tagnames}'>${tagnames}</li>"
     };

    $('textarea#text-input').atwho(mentions_config).atwho(hashtags_config);
```

jywarren:
Hmm, interesting! One reason I was hoping we could use horsey is that it
can be made to work with 'woofmark', our rich text editor.

But, we've had some serious issues getting that to work smoothly. I wonder
if atwho would work with woofmark, and we could switch everything to atwho?

Would you mind seeing if you could get atwho working on a page like /post
-- with the rich editor?

Thanks for such a deep dive on this!

cheneyshreve:
@jywarren sure thing, I'll see if I can get atwho working on a page like/post with the rich editor. It's interesting to learn about this, especially with these kinds of editors becoming pretty popular.

cheneyshreve:
Hi @ViditChitkara @jywarren Looks like it does work in the rich text editor if I am testing this properly (I added earlier code snippet to rich.html.erb)

testing_atwho_richtexteditor

I think that it might be possible to also use atwho to do the emojis, building on @ViditChitkara's code. I found this code snippet https://gist.github.com/conan007ai/9503021#file-gistfile1-txt-L44 but it is not as sophisticated as @ViditChitkara's because it doesn't do the substitution of the :: around the image so I'm not sure if it would work with markdown. I can try it out if you want. I'd probably need to build on @ViditChitkara's code he has there already. What do you think?

cheneyshreve:
Hi @jywarren I checked out the rich text editor and markdown editor and it is possible to use atwho to do autocomplete for the usernames, hashtags, and emojis together. Just let me know if you want me to make issues/PRs to move forward with this.

in the markdown editor

testing_atwho_emoji_hash_username

in rich text editor

testing_atwho_emoji_hash_username_richtext

jywarren:
OMGGGGG This is the BEST 🎉 🎉 🐴 

Wow. Ok, how about this. For this PR, let's use atwho. If you think it's ready to go, that's great. 

Next, let's refactor the emoji one into atwho, so it's consistent. 

Third, let's go into the rich editor and implement that feature over there too! That'll be a little more complex but I can help you out. How does that sound? 

Amazing work, @cheneyshreve !!! We're so lucky to have your contributions!!!!!

🙌🙌🙌 ⚡️

jywarren:
Let's test one more thing -- in the rich text editor, can you switch the mode using the `Rich` button to bottom right, and see what happens when you keep autocompleting? We may have to "reconstruct" atwho for both modes. But it should be OK.

cheneyshreve:
Hi @jywarren many thanks for your support, I'm very excited to get to contribute to this project! Regarding toggling the rich text button & your earlier comments.  Based on testing in rich.html.erb, you are right, it will need a bit of refactoring. It does okay with keeping the hashes separate from markdown (e.g. one hash is header 1) vs. Rich text (it keeps it as a tag, yay!) but it identifies a bogus username as a real username if it follows the @ symbol (see gif below) so will need to be refactored in the users controller, maybe adding a if !@user.nil? statement. Or maybe make it fancier and give the user a message that that person is not in the database, and then we'd still need a way to make sure it does not try to link to a bogus user page from the user's controller.

 Let me know how you'd like me to proceed. I've refactored the atwho calls into one statement, but obviously we'll want to fix this first.

Editor behavior tests
atwho_editor_tests

jywarren:
We can break out and address these issues, indeed. Let's make a new issue for the rich editor! In the meantime, is this PR ready to go? What follow-up issues do we want to make? Let's make a /checklist/ ✅✅✅✅ 😄

config/database.yml.sqlite
```diff
@@ -0,0 +1,11 @@
+development:
```
jywarren: Ah, i think this has to be removed -- see this for some help with that!

https://stackoverflow.com/questions/215718/reset-or-revert-a-specific-file-to-a-specific-revision-using-git

👍

cheneyshreve:
Hi @jywarren the current PR [WIP] has the 3 autocomplete commands (username, tag, emoji) chained in the comments _form.html.erb, which I tested again locally to confirm however I still have some questions. 
- This code works, but as the gif above shows, the markdown editor interprets anything following @ as a username and if  a user clicks on that later on, it will throw an error from the user's controller.
- I have deleted the earlier emoji code (only from the comments form, as I believe the helpers within application_helper for emoji are currently being used elsewhere). 
- Tests pass locally, but due to earlier commit you mention, perhaps I still need to do some git-related revisions? It might be easier to delete this branch and just make a new one since it's not a lot of code. I was following a guide on open source, but perhaps not the best one, or probably I misinterpreted or missed a step. 

In terms of a checklist
- where are we dealing with potential username error? (In this PR, or elsewhere?)
- open a new issue for adding to Rich text editor 
- (perhaps) open a new issue to determine if other parts of emoji code could be refactored

thanks again for your patience and support :)

jywarren:
I think we can solve the non-existent username issue in another, and it'll be fine, thanks! And the git mergability looks perfect. I'm going to request a review from someone else and then we should be good!

Let's also open a new issue to see if we can optimize the emoji code - especially putting all emoji into a separate file which the browser can cache since it's a lot of code for each page load.

app/views/comments/_form.html.erb
```diff
@@ -66,38 +66,54 @@
...
-    horsey(document.querySelector('textarea#text-input'), {
-        suggestions: (<%= raw emojis.to_json %>),
-        anchor: ':',
-        render: function (li, suggestion) {
-            var image = image_map[suggestion.text];
-            li.innerHTML = "<b>"+image+"</b><span>"+suggestion.text+"</span>";
-        },
-        limit: 6
+
+    names = <%= raw User.pluck(:username).compact.to_json %>;
```
jywarren: Ah, so here, does atwho allow for remote loading of suggestions and especially ones base on what letters you've already typed? That would do a lot to optimize this for when it runs on the full db which has 300k usernames.

jywarren: So for this, see my note above about the API and 

https://publiclab.org/api/srch/profiles?srchString=jy

I think atwho will probably have a way to fetch remote results!

Thanks and great work!

cheneyshreve: Hi @jywarren awesome to learn how to deal with this kind of transaction for a large user community (I haven't run into that before  now ;) I think it should be possible to fetch remote data (e.g. from atwho docs):

When you set data as URL string, At.js will launch Ajax request to the server to get the JSON data

   $('textarea').atwho({
        at: "@", 
        data: "http://www.atjs.com/users.json",
        limit: 7
    });

so I would need to combine that with the API you mention to get the proper url. If that is not possible, I should at least be able to put the code in application_helper, as with the emoji example, but I think the API approach should be feasible. As a side note, my  less than a year old Mac has decided it's display wants to fail so in the next few days, I have to send it to Apple for repairs. If I fall off the radar for a few days, it is because of this, but I'm hoping it'll hold out for a few more days so I can get some work done before sending it off.

cheneyshreve:
Hi @jywarren  I refactored the code a bit so that usernames, tags, emoji-names are have methods in application_helper (as the horsey examples has). Per earlier note, I do think it's possible to fetch data remotely with atwho, but it's unclear to me how to fetch usernames from PublicLab's API from the docs, or did you mean there is some other way to narrow the search via API? So I've left this revision also as a [WIP], sorry that is still unclear to me at this point, will keep thinking about it though!

jywarren:
Sorry I didn't get to this today! I'll take a look tomorrow morning!

jywarren:
So ok - instead of `def user_names_list` we can make a remote request to this URL:

https://publiclab.org/api/srch/profiles?srchString=wa

(documentation here: https://github.com/publiclab/plots2/blob/master/doc/API.md)

For example you might start typing `@warren` with just the letters `wa`, triggering a remote call to:

https://publiclab.org/api/srch/profiles?srchString=wa

Then you might keep typing to narrow things down, so `warr` would generate:

https://publiclab.org/api/srch/profiles?srchString=warr

See how the results narrow?

https://publiclab.org/api/srch/profiles?srchString=warren still doesn't return `warren` but it returns a lot of other usernames including @warren. 

(I guess we should ask @milaaraujo and @stefannibrasil if they could change the sorting of the results in the API to most recently updated people... so [changing this line](https://github.com/publiclab/plots2/blob/master/app/api/srch/search.rb#L45) from [this](https://github.com/publiclab/plots2/blob/73a16d10a7e1bc03c8921f593a98a639687d91a2/app/services/search_service.rb#L236) instead of [this](https://github.com/publiclab/plots2/blob/73a16d10a7e1bc03c8921f593a98a639687d91a2/app/services/search_service.rb#L122) --- a good break-out issue!)

Sorry... tangent. 

But to fetch remote data, it looks like there is an example here: https://github.com/ichord/At.js/wiki/How-to-use-remoteFilter#normal-usage

The returned data will look like this:

```
{"items":[{"docId":0,"docType":"user","docUrl":"/profile/dewabetsport","docTitle":"dewabetsport","docSummary":"","docScore":0},{"docId":0,"docType":"user","docUrl":"/profile/wdewaqqcom","docTitle":"wdewaqqcom","docSummary":"","docScore":0},{"docId":0,"docType":"user","docUrl":"/profile/paciaciakowa84129","docTitle":"paciaciakowa84129","docSummary":"","docScore":0}],"srchParams":{"srchString":"wa","seq":null,"showCount":null,"pageNum":null,"tagName":null}}
```

So you will need to read the result like:

```js
$.getJSON("https://publiclab.org/api/srch/profiles?srchString=" + query, {}, function(data) {
        callback(a.items.map(function(i) { return i.docTitle }));
      });
```

That should map the usernames to an array like `['name', 'name', 'name']` -- does that make sense?

This is a cool feature, and quite complex! We'll have to take it one step at a time.

cheneyshreve:
Hi @jywarren some successes, but not quite there yet. Can get this code to work for users that are returned from query but throws an error if users are not in query 

```
  var at_config = {
     at: "@",
     callbacks: {
       remoteFilter: function(query, callback) {
       $.getJSON("/api/srch/profiles?srchString=" + query, {}, function(data) {
            callback(data.items.map(function(i) { return i.docTitle }));
          });
        }
      }
   }
    $('textarea#text-input').atwho(at_config);
```
The error is that data.items.map is undefined, which makes good sense, when you type in a user that does not appear in the automated dropdown list, so I was trying various ways to revise the callback to handle null data (e.g. function (data, status) { if (status == "success") {proceed as planned... } type of edits  and also adding in a conditional with self.xhr.abort() if the data was null, but probably wasn't getting the syntax quite right.

The other thing I was concerned about is how many times the error is thrown, it looks like every time you type a character after the @ if it is not for a user that pops up in the list, you get the type error thrown. So, I'm trying to get rid of these error, but no success just yet. Gif below describes in more detail (apologies if display noise shows in gif, need to get my computer fixed in next couple of days here).  I also tried to implement the cache example in the link https://github.com/ichord/At.js/wiki/How-to-use-remoteFilter#normal-usage you sent earlier to see if I could get that to work, but no success yet.

need_to_handle_nonuser_error

jywarren:
Awesome. Push up your latest changes?

This is looking really good! I think you should be able to check if if there are items with:

```js
if (data.hasOwnProperty('items') && data.items.length > 0) {
  callback(data.items.map(function(i) { return i.docTitle }));
}
```

Make sense?

You could also decide to only send the `getJSON` request if someone's typed more than 2 characters... that might be useful anyways?

jywarren:
Also, the CodeClimate check is just complaining about formatting... it does help to keep things tidy though! Check out its recommendations here: https://codeclimate.com/github/publiclab/plots2/pull/2925

Thanks again!

cheneyshreve:
Hi @jywarren yes, that worked, thanks for the tips! I just pushed most recent changes. Works locally and looks like codeclimate issues are fixed, just waiting now to see travis cli results.

fixes_errors

config/database.yml.sqlite
jywarren: Looking great - can you try removing this file from the PR? Thanks!

www.example.com/db/openid-store/associations/http-localhost-Sve98kiqQxHfzN8CdZK0Uuu1aIs-79QTAXpF_XIxzDrDmpcZxJStiSQ
```diff
@@ -0,0 +1,6 @@
+version:2
```
jywarren: Ah, and these two files also don't need to be included. I'll modify our `.gitignore` to not detect these anymore... thanks, but in the meantime would you mind removing them too?

jywarren:
OK, this is looking great! I think we need one more thing and we can merge it -- let's also tie the hashtag completions to a remote call. 

Looking at this, i think this API call from the API docs returns /content/ for a tag, instead of a list of tags:

https://publiclab.org/api/srch/tags?srchString=balloon

However, I /believe/ this URL should work:

`https://publiclab.org/tag/suggested/______`

I think it only works over a JavaScript AJAX call, but that's what we're doing, so it should be OK! Want to give it a try?

cheneyshreve:
hi @jywarren, thanks for the message, I will get started on the edits you mention for hashtags & removing those unnecessary files. A little slow today getting setup on a different computer, but should be setup soon!

cheneyshreve:
[WIP] Hi @jywarren I tried several variations to get the tag list via api routes but was unable to find the data we want, e.g. I tried using remote filter with:
```
/api/srch/tags?srchString= 
/tag/suggested/____
/api/srch/tag/suggested/
/api/typeahead/notes?srchString=
and looked through other api routes in the docs to see if any of them have the tag names in the response body we want, but I did not see that
```
I wonder though why /tag/suggested isn't working (either syntax error) or possibly because that data isn't seeded so it just won't work in the development context? 

I added new tags to test the current code (that calls the Tag model in application helper) and it looks like it's doing what we want, but there is some occasional odd behavior. Sometimes, if the hashtag is the first item added to the comment editor, it does not retain the hashtag in publish/preview. Other times it does. Seems like it's fine if the hashtag is not the first item, but obviously kind of an annoyance (gif below demonstrates). I'm not sure if the error relates to the atwho library (I could follow up with the author) or if its something elsewhere in code. 

I removed the files you asked, but see a conflict now with .gitignore. Happy to revise .gitignore and any other files and push again, just let me know. Thanks again!
testing-code-bits-pluck-example

jywarren:
aha - the bug you've found is that "# sometext" is Markdown format for a
header, so it makes everything big (like a header) and removes the #
character. I'm not sure how to solve this but let's... leave it for now
maybe?

OK, let me review this a bit and i'll see if we can get us past the
suggested tag url issue... thanks and fantastic work!!!

config/database.yml.sqlite
jywarren: Ah, still this one file to be removed!

app/views/comments/_form.html.erb
```diff
@@ -66,38 +66,55 @@
...
+            if (data.hasOwnProperty('items') && data.items.length > 0) {
+              callback(data.items.map(function(i) { return i.docTitle }));
+            }
+           });
+          }
+        }
+      },
+      hashtags_config = {
+        at: "#",
+        data: tagnames,
```
jywarren: OK, so here, I think you can use this syntax - it looks like it requires a POST request:

```js
$.post('/tag/suggested/' + query, function(response) {
  // do something with response, which should be an array
});
```

cheneyshreve:
Hi @jywarren -- interesting, so the underscores after /tag/suggested/___ must influence the number of characters being searched or where it starts the search? The current code seems to give desired behavior, but marked as [WIP] in case we need any additional changes. 

tag-demo

jywarren:
This looks great. I want to do one more thing here and then we'll be perfect. Right now it looks like the emoji are hard-coded and there are only about 5 available. But if you look at this page, you'll see (near the bottom) that we're including hundreds in each page load currently:

[page source](view-source:https://publiclab.org/notes/MaggPi/07-09-2018/computer-vision-color-detection)

This is because of this line in `_form.html.erb`:

`-        suggestions: (<%= raw emojis.to_json %>),`

Can we, instead of doing this all inline, add this emoji.json file to the PR and load it externally? 

https://gist.github.com/jywarren/02793d1ec65d3935ee7da61151c9ac68

cheneyshreve:
Hi @jywarren happy to continue working on reconfiguring the emojis but a couple questions. The current version of _form.html.erb does not contain the line ```suggestions: (<%= raw emojis.to_json %>)``` Do you mean to look in a different file for that? I'm not sure where the source code is for the notes page you referenced, so perhaps that's the piece I'm missing.

As you mention, the emojis are currently hard coded in the view as ```var emojis = <%== atwho_emojis %>``` which are defined in the application_helper. If I'm interpreting your request correctly, I would need to add the emoji json file to the public folder, then read it into the view. I may need to reformat it a little (e.g. to have an id or count and just the emoji name to look up). How many of the emojis would you like to be visible? It would be cool if there was a way to display them more like they are on a cell phone where you click on a smiley and all the smiley variations pop up. I'm not sure how to do that but happy to look into it (perhaps you guys are already doing that on the page I'm missing ;)

Let me know if I'm understanding you right and I'll get to work on this, thank you :)

jywarren:
Ack, i'm so sorry i took a while to write back - we just had [this big event](https://publiclab.org/barnraising). Looking now!!!

app/views/comments/_form.html.erb
```diff
@@ -66,38 +66,60 @@
...
-    <% image_map = emoji_names[:image_map] %>
-    var image_map = <%= raw image_map.to_json %>
-    horsey(document.querySelector('textarea#text-input'), {
-        suggestions: (<%= raw emojis.to_json %>),
-        anchor: ':',
-        render: function (li, suggestion) {
-            var image = image_map[suggestion.text];
-            li.innerHTML = "<b>"+image+"</b><span>"+suggestion.text+"</span>";
+
+    var emojis = <%== atwho_emojis %>
```
jywarren: OK, so here is where we are currently loading the emojis, and I believe it's compiled from a static list in this controller line:

https://github.com/publiclab/plots2/pull/2925/files#diff-127a850b7d7b20173ad3fba88f8a25e3R56

What we could do instead is use a `remoteFilter` as we do in the other two atwho configs, but point it at a static file that we create from the Gist I linked to. It could be stored in `/public/emoji.json` and so that would be the route as well. Does that makes sense? Thanks!!!

jywarren: I'm not sure the list I linked to is in the right format either, but it should be relatively easy to parse it.

jywarren:
Finally, we need to resolve the conflicts by rebasing this PR, and we should be good to go!

app/views/comments/_form.html.erb
cheneyshreve: Hi @jywarren  the barnraising event looks very cool, wish I could have attended. I did a project in the UK on  community emergency preparedness and we definitely need people and software to work on this!  

 I've been looking at loading the emoji data from the json file. It's easy enough to just load the emoji characters like so, and they populate the dropdown menu:

```
Option A
      emojis_config = {
        at: ':',
        callbacks: {
          remoteFilter: function(query, callback) {
            $.getJSON("/emoji.json" + query, {}, function(data) {
                callback(Object.values(data).map(function(name) { return name }));
             });
            }
          },
         displayTpl: "<li>${name}</li>",
         insertTpl: "${name}",
         limit: 25
     }
```
But 2 concerns: a) is this OK for the other editor to render the emoji as character? b) can we do something fancier? The at.js demo shows a searchable dropdown with names and emoji images displayed so I know that is possible but perhaps we'd need a different setup, or is it possible to do this with static files? I'd tried reading in the emoji_values.json into the remote filter, then trying to have "data: '/emoji.json' and that displays the emoji names properly, but I'm not sure how you'd display the emoji images in the dropdown via that route:

```
Option B
       emojis_config = {
         at: ':',
         callbacks: {
           remoteFilter: function(query, callback) {
           // this render the names of emojis properly, but we have no "search" functionality
             $.getJSON("/emoji_values.json" + query, {}, function(data) {
                 callback(Object.values(data).map(function(i) { return i.text }));
              });
             }
           },
// unsure how/if we can edit the displayTpl here to display the emoji characters, 
//since those are not in the emoji_json file we read in during callback---maybe if I merged these 2 files?
          displayTpl: "<li>${name}</li>",
          insertTpl: ":${name}:",
         // I don't think this is ever reading from this file but relying on the preexisting gemoji code, when you delete this line, it still renders
          data: '/emoji.json',
          limit: 10
      }
   ``` 

So I could use the Option A above and physically rearrange the json so we get more than just smileys showing in the dropdown, but maybe like 10 most popular emojis or so, then rebase and submit, or try something else. Let me know what you think. Thanks again for your feedback!

Comment (jywarren) [reply app/views/comments/_form.html.erb]: > for the other editor to render the emoji as character

do you mean like showing the emoji plus the name in the autocomplete, like:

screen shot 2018-07-19 at 6 00 19 pm

? If so, i like the idea of rearranging the JSON... would it be a huge pain? I think i could do it in a script if you want me to try?

Comment (jywarren) [reply app/views/comments/_form.html.erb]: And i love the idea! But if it's a lot of work we can just do the simple version first and solve that in a follow-up issue... so we can see this go into production first!

Comment (jywarren) [reply app/views/comments/_form.html.erb]: I think we could use `emoji.json` and make it like:

```
':smile:': "😄 smile",
':smile:': "😄 smile",
':smile:': "😄 smile",
```

(but without repetition, just for example)

Is that what you're thinking?

**Comment by cheneyshreve** [comment]
HI @jywarren I just did the rebase with the emojis so they just show about 8 popular ones for now, but would love to open a new issue, as it might take some time to figure out how to refactor the json and populate the dropdown properly, per your recommendations above. I marked this as WIP just in case there are any issues with rebase, etc. since I was working on this from 2 different computers, but should be sorted now.  Gif below of visual changes. Thanks again for the feedback, looking forward to working on this some more!

atwho_tests

File: app/views/comments/_form.html.erb [hunk 6]
```diff
@@ -66,37 +65,64 @@
...
+          }
         },
         limit: 6
-    });
+      },
+      hashtags_config = {
+        at: "#",
+        callbacks: {
+          remoteFilter: function(query, callback) {
+            $.post('/tag/suggested/____' + query, function(response) {
```
jywarren: Ah, oops, just need to remove this `____` here, and let it say just "query" - thanks!

app/views/comments/_form.html.erb
```diff
@@ -66,37 +65,64 @@
...
+             });
+            }
+          },
+        limit: 6
+      },
+      emojis_config = {
+        at: ':',
+        callbacks: {
+          remoteFilter: function(query, callback) {
+            $.getJSON("/emoji.json" + query, {}, function(data) {
```
jywarren: Ah, interesting, so here, we could remove the ` + query` part, because we're going to manually dig through the returned results to find what we want. 

You're noting that it returns 8 of them? I'm a little surprised it doesn't return the whole list, but maybe `atwho` is filtering it or something? I think we'd need to somehow filter this list for ones that match the query... not totally sure how to do that... looking at https://github.com/ichord/jquery-atwho-rails

jywarren: I think i see! I think it's returning the first 8. Once we remove ` + query`, we can optimize this in a follow-up!

jywarren: And we'll be able to refine what's displayed in the menu vs. inserted using these template settings! https://github.com/ichord/At.js/wiki/Base-Document#custom-template

jywarren:
Great work! We are ALMOST THERE 👍 👍 👍

cheneyshreve:
Hi @jywarren 

So, the "limit" keyword determines how many emojis, hashes, etc. will be returned, I had just set it to 8 but we could change that to a higher/lower number. Oddly, if you don't set it, it doesn't return all the values so there must be some default number baked in.

Regarding removing the query and/or the underscores:

//This does not throw any errors 
  $.post('/tag/suggested/____' + query, function(response) {}
//Nor does this 
  $.post('/tag/suggested/____', function(response) {}
  
  
//When you remove the underscore you get a 404 error
//2jquery.self-bd7ddd393353a8d2480a622e80342adf488fb6006d667e8b42e4c0073393abee.js?body=1:10255 POST http://localhost:3000/tag/suggested/ 404 (Not Found)
   $.post('/tag/suggested/' + query, function(response) {}
   
//Also throws a 404 error without the query
     $.post('/tag/suggested/', function(response) {}
     

Oddly, when you take out the underscore, you also get different behaviors, it doesn't search the body of the string in the same way:
   
 e.g. without underscorore or query
no_query_no_underscore

 e.g with underscore and query
with_query_and_underscore


 
 
 Happy to revise, just let me know what revisions you'd like, so far I have: 
 
 removed the query from emojis and set limit to 600, but we could do more/less (think there must be like 700 or 800).     

I can take the underscore out of the hash autocomplete if the 404 error is not an issue, just let me know.

  ```
        emojis_config = {
        at: ':',
        callbacks: {
          remoteFilter: function(query, callback) {
            $.getJSON("/emoji.json", function(data) {
                callback(Object.values(data).map(function(name) { return name }));
             });
            }
          },
         displayTpl: "<li>${name}</li>",
         insertTpl: "${name}",
         limit: 600
     }```
     note: if you remove query as an argument from function entirely, it does not work

jywarren:
I think it's showing an error when `query` is an empty string... does it keep showing an error as you keep typing?

jywarren:
Oh, indeed it does. Hang on...

jywarren:
Hmm, i'm a little mystified at why `query` is an empty string. Still skimming the docs here: https://github.com/ichord/At.js/wiki/How-to-use-remoteFilter

jywarren:
OH wait, so in this case, we need `query` but not the underscore. Can you try typing a few characters with that set up? For the emoji, we don't need `query` because we'll be filtering it in JavaScript on the client side. Make sense?

cheneyshreve:
Hi @jywarren  so with the query and without the underscore, e.g. 

  $.post('/tag/suggested/' + query, function(response) {}

still throws 404 error. 

I think what is going on is related to the sequence that functions are called by the atwho object but not sure what _exactly_ is going on with the search string
[callbacks](https://github.com/ichord/At.js/wiki/Callbacks)

jywarren:
Does it still result in just /tags/suggested/ in the shown errors, or are
you starting to see, for example,

/tags/suggested/h
/tags/suggested/he
/tags/suggested/hel
/tags/suggested/hell
/tags/suggested/hello

which would indicate that it's getting passed through properly?

jywarren:
Maybe just above the `$.post('/tag/suggested/' + query` line, you could do
`console.log(query)` to just see what's in there?

cheneyshreve:
ok so with console.log(query) for statement without underscore, it does throw a 404 error first, but then it starts populating the letters..

screen shot 2018-07-20 at 12 19 38 pm

jywarren:
aha! OK, so we could add an `if (query != '') {` around the $.post() bloc,
and it won't start searching until you type... how does that sound?

cheneyshreve:
Hi @jywarren yes! The if condition gets rid of the 404 error so we can drop the underscore. Thanks for the tip!

jywarren:
Awesome. As soon as you push a new commit I can do a final review and
probably merge this in! ⚡⚡👍🏼🙌🏼

cheneyshreve:
Hi @jywarren sorry for any confusion, the [last commit] (https://github.com/publiclab/plots2/pull/2925/commits/d51276e402f2612109655a443f77a7811bb37b7b) does have the changes we talked about with the conditional and no underscore.

jywarren:
OK, i used this a bit and ran into an encoding issue, and a few other things, so I made some small adjustments -- started including the emoji.json file in the page, rather than via a remote JS call. Then I wrestled with the display vs insertion formatters -- quite tough! But finally seemed to figure it out. The final changes are pretty small but I tested it out and it works /almost/ as well as we'd hoped. 

I think the next step will be to deal with this encoding issue so we can display the emoji in the editor instead of just in the published comments: #2665

But for now this is ready to merge -- VERY EXCITING!!!

We have a few follow-ups, if you're interested (though, take a moment to celebrate, you've done a fantastic job here!)

* [ ] we could move this JS code into its own `/app/assets/javascripts/emoji.js` file
* [ ] we could generalize it so it could ALSO be used, or at least some parts re-used, for the rich editor, as you demoed above

In any case, THANK YOU!!!!! 🎉 

Let's let this final version pass tests and I'll merge and publish it.

jywarren:
Super, merging -- and done!

OK, if you'd like to do a few follow-up things, we should open a new issue and copy in the checklist from above (and maybe your animated gif!) to start a new thread separate from this very long one!

I hope you felt great about this and got a chance to build some useful skills! Thanks again!

jywarren:
This is now live at PublicLab.org. Try it out! Say, on one of my posts: 

https://publiclab.org/notes/warren/03-20-2018/collaboratively-document-an-event-with-a-mini-newspaper

cheneyshreve:
HI @jywarren awesome! I am definitely learning and grateful to have your support! I will copy/paste the checklist above and open a new issue. Thanks again!

app/views/comments/_form.html.erb
```diff
@@ -65,42 +64,70 @@
...
-            var image = image_map[suggestion.text];
-            li.innerHTML = "<b>"+image+"</b><span>"+suggestion.text+"</span>";
+
+    (function() {
+
+    var at_config = {
+      at: "@",
+      callbacks: {
+        remoteFilter: function(query, callback) {
+          $.getJSON("/api/srch/profiles?srchString=" + query, {}, function(data) {
```
rexagod: Shouldn't this be `/api/srch/profiles?query=`?

jywarren:
I think so - the api has changed in the past few months!

PR #2933, opened 2018-06-27
 incoming email parsing feature for gmail and yahoo

namangupta01:

Thanks!


db/migrate/20180627060150_add_column_reply_to_content_in_comment.rb
```diff
@@ -0,0 +1,5 @@
+class AddColumnReplyToContentInComment < ActiveRecord::Migration[5.1]
+  def change
+  	add_column :comments, :extra_content, :text
```
jywarren: Could we just keep it in the regular comment body, but use a filter to remove it upon sending a notification email, and use JavaScript to hide it in the HTML views? Just wondering if we can solve this without a db change and the maintenance that may come with?

jywarren:
How's this coming along, Naman? Maybe we could get an early version published today?

namangupta01:
Yeah sure!

db/migrate/20180627060150_add_column_reply_to_content_in_comment.rb
namangupta01: I was thinking, it will become complex if we try to fit both the content into one attribute..of course we can use filter but i was thinking we both the forms will have some maintenance but i am happy doing either ways....what do you think?

namangupta01: @jywarren any thoughts?

jywarren: I guess I think we might use the `render` methods to offer a "full" or "trimmed" display for use in different places, and filter on display rather than filter on save. The filter could be like:

```ruby
comment.filter_body(:gmail)
# vs
comment.body
```

maybe?

namangupta01:
Okay sure! I will implement this way!!!

namangupta01:
Hey @jywarren, is this good now?

jywarren:
Hi @namangupta01 -- i think there are a few recommendations from CodeClimate that you could integrate -- not every one, but see if there are some you could resolve in code and the rest we can "accept". 

Exciting, thanks!

namangupta01:
Yeah sure! I will fix them. Btw i was asking about the current implementation.

app/models/comment.rb
```diff
@@ -13,6 +13,9 @@ class Comment < ApplicationRecord
   self.table_name = 'comments'
   self.primary_key = 'cid'
 
+  COMMENT_FILTER = "\n@@$$%% Trimmed Content @@$$%%\n"
```
jywarren: Interesting and clever! What about putting it into an HTML comment as well - like `<!-- $$%% TRIMMED CONTENT %%$$ -->` so that it can be parsed out with an HTML filter if need be? I dunno, just seems potentially extra helpful in the future...

namangupta01: This way would be great. Putting this in html comment didn't come to my mind. 👍 🎉

jywarren:
Awesome, this looks good! Did you want to perhaps include sample reply emails from diff. services in tests, and confirm they get stripped properly? This will also have the added benefit that if email vendors change their formats, we can compare to the test examples to update the solution. 

Great work! Staff were VERY excited to hear about this feature -- open mouthed! 😮

namangupta01:
Yes, i was already planning to do these and some more testing if we can find a way to test them in another pr.

jywarren:
OK - do you want to merge this first then add tests in another PR? I think it can be very good to have tests be in the same PR so there's a relationship between the feature and the test that confirms it... what do you think?

jywarren:
Even just adding some basic tests of this -- they can be expanded later!

namangupta01:
Yeah no problem...i will add tests in this pr for email parsing.

namangupta01:
Just for demo:

ezgif com-video-to-gif 8

jywarren:
oooooooooohhhhhhhh 👍 👍 👍  fantastic!!

jywarren:
I think this may need to be rebased following @Souravirus upgrade to Rails 5.2 -- @Souravirus what do you think?

namangupta01:
Got it...never mind

namangupta01:
@jywarren I have added the test... can you please have a look so that i can finalize it?

namangupta01:
@jywarren Is it good now?

namangupta01:
Also added readme for google and yahoo parsing.

namangupta01:
Yeah sure and we will also expand this feature for all email services.

Issue #2928, opened 2018-06-26
Emoji in reply-by-email


jywarren:
First, want to say that reply-by-email is AWESOME!!!!!!!

One small thing -- I tried sending a 🎉 in reply by email and got this error:

```

--000000000000cc5f67056f9393a7--

E, [2018-06-26T23:12:07.218298 #1] ERROR -- : Mysql2::Error: Incorrect string value: '\xF0\x9F\x8E\x89\x0A\x0A...' for column 'comment' at row 1: INSERT INTO `comments` (
`nid`, `uid`, `comment`, `timestamp`, `format`, `thread`, `comment_via`, `me
ssage_id`) VALUES (16574, 1, 'Testing response!\n  \n\nwith **formatting** &nbsp;<U+1F389>\n\n  \n\nOn Tue, Jun 26, 2018 at 5:01 PM \\<[do-not-reply@publiclab.org](mailto:do-not-reply@publiclab.org)\\> wrote:  \n\n> Public Lab contributor [warren](https://publiclab.org/profile/warren) just posted a new research note entitled \' **Are there lists of plants and the chemicals they can remediate?**\':\n> \n> Read and respond to the post here: [https://publiclab.org/notes/warren/06-26-2018/are-there-lists-of-plants-and-the-chemicals-they-can-remediate](https://publiclab.org/notes/warren/06-26-2018/are-there-lists-of-plants-and-the-chemicals-they-can-remediate)\n> \n> * * *\n> \n> I\'ve heard of compiled lists of plant species and the different things they can absorb from the air or soil or water. Does anyone know of such a list?\n> \n> * * *\n> \n> You received this email because you are subscribed to some or all of the following tags: .\n> \n> To change your preferences, please visit [https://publiclab.org/subscriptions](https://publiclab.org/subscriptions).\n> \n> Report spam and abuse to: [moderators@publiclab.org](mailto:moderators@publiclab.org)\n> \n> \r \r Check out the blog at [https://publiclab.org/blog](https://publiclab.org/blog) | Love our work? Become a Public Lab Sustaining Member today at [https://publiclab.org/donate](https://publiclab.org/donate)\n\n', 1530054727, 1, '01/', 1, 'CAGqfm_DFaA5h66ObwMsxFkgi5WwZHyByCniKbqpq1XfUAZj28A@mail.gmail.com')
/usr/local/bundle/gems/mysql2-0.3.21/lib/mysql2/client.rb:80:in `_query'
[...]

Issue #2209, opened 2018-02-02
Encoding issues with migrating old DrupalUser accounts to User records


jywarren:
We're seeing an issue with a recent migration https://github.com/publiclab/plots2/pull/2157 which tries to mitigate User vs DrupalUser status records. I tried to fix it here: https://github.com/publiclab/plots2/pull/2208 but we have a new problem. 

I believe it's due to this error from the [Jenkins build log](https://jenkins.laboratoriopublico.org/job/Plots-Stable/lastFailedBuild/console):

```
ActiveRecord::StatementInvalid: Mysql2::Error: Illegal mix of collations (latin1_swedish_ci,IMPLICIT) and (utf8_general_ci,COERCIBLE) for operation '=': SELECT  `rusers`.* FROM `rusers`  WHERE `rusers`.`username` = 'トリーバーチ クラッチ'  ORDER BY `rusers`.`id` ASC LIMIT 1
```

The username in question is: `トリーバーチ クラッチ`

Can we convert the string or otherwise mitigate when a username is ... not utf-8?

jywarren:
I think i can get past this by not matching `User` and `DrupalUser` by `name/username`, but instead by `id/uid` (did this in https://github.com/publiclab/plots2/pull/2210). But we will have to solve this if we ever want to convert /all/ `DrupalUsers` into `Users`, with the `drupal_user.migrate` command:

https://github.com/publiclab/plots2/blob/b73d598262f2132a57d4cd3f49f5331edafc8265/app/models/drupal_user.rb#L180-L193

`migrate` will break due to the above encoding problem. Once we fix that issue, we could ensure there are no "orphaned" `DrupalUsers` left with:

```ruby
DrupalUser.all.each do |du|
  du.migrate
end
```

https://stackoverflow.com/questions/17022394/how-to-convert-a-string-to-utf8-in-ruby

jywarren:
That'll get us one step closer to deprecating `DrupalUser`! #956

ViditChitkara:
Hmm, didn't anticipate this problem. will submit a patch for this!

jywarren:
I went ahead and fixed the issue for the purposes of your migration -- no
problem! i should've seen this too. But I do think the future desire to
stop using DrupalUsers entirely, for which we'll need this "migrate"
method. Interested in working on that a bit? Thanks!!

ViditChitkara:
Yes, interested! Finally now the following needs to be done:-
[...]

PR #3007, opened 2018-07-07
[WIP] Migration utf8 -> utf8mb4 and test emojis for comment and answer

namangupta01:
#2928

Thanks!


namangupta01:
Hey @jywarren, as discussed in #2928 I have added the tests here

jywarren:
I restarted this, but if you see this error you can restart it too by closing and reopening the PR!

namangupta01:
@jywarren Tests passed in travis. But when i tried on local machine using mysql it failed.

namangupta01:
screen shot 2018-07-09 at 11 48 23 pm

jywarren:
Hmm, I think we may need to redo the mysql encoding as in https://github.com/publiclab/plots2/issues/2665

Do you want to try the suggestion @icarito made on changing encoding to `utf8mb4` here: https://github.com/publiclab/plots2/issues/2209#issuecomment-397354832 ?

namangupta01:
Sure! So should we change the encoding of the whole db or just the tables or columns we need?

jywarren:
Let's start by doing just the columns, and if they work, we can probably
move the whole db? But i think that's a Q for @icarito.

namangupta01:
Pushing to unstable

jywarren:
Ah, did this end up working?

SidharthBansal:
As the person is inactive for more than a month, I am closing the PR. In case you want to push changes please feel free to open a new PR.
Thanks for contributing at Public Lab

jywarren:
OK, now I'll try adding a node emoji test to see that the tests fail. If this is true, we can add the same encoding fix for other types (profile bios, nodes, answers)!

test/unit/answer_test.rb
```diff
@@ -57,4 +57,10 @@ class AnswerTest < ActiveSupport::TestCase
     deleted_answer = answer.destroy
     assert_equal answer.comments.count, 0
   end
+
+  test 'should save answer with emojis' do
+    answer = Answer.new(content: "Answer with Emojis 😎😎😎")
```
jywarren: hmm, i'd have thought this would fail...?

jywarren:
This should also address #4118 and maybe #3988

jywarren:
If these fail, we can add:

```ruby
    # changing table that will store unicode:
    execute "ALTER TABLE comments CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
    # changing string/text column with unicode content:
    execute "ALTER TABLE comments MODIFY comment VARCHAR(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"

    # changing table that will store unicode:
    execute "ALTER TABLE nodes CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
    # changing string/text column with unicode content:
    execute "ALTER TABLE nodes MODIFY title VARCHAR(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"

    # changing table that will store unicode:
    execute "ALTER TABLE revisions CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
    # changing string/text column with unicode content:
    execute "ALTER TABLE revisions MODIFY title VARCHAR(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
    execute "ALTER TABLE revisions MODIFY body VARCHAR(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
```

jywarren:
Hmm, still not failing... similarly to #4118

jywarren:
I just tried this locally with a 👍  and even in the rails console it crashed:

```
Loading development environment (Rails 5.2.2)
[1] pry(main)> a = Answer.new(content: "hi ����")
/home/warren/.rvm/gems/ruby-2.4.4/gems/pry-0.12.2/lib/pry/history.rb:131:in `write': "\xF0" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
```

icarito:
Hey so am I correct that this works? i.e the test works?
Here for reference is an [article](https://hanoian.com/content/index.php/24-automate-the-converting-a-mysql-database-character-set-to-utf8mb4) on the work needed to move the whole database. It could be delicate if we have to truncate fields 256->191?

jywarren:
No, unfortunately we are at a stage where it passes Travis but fails in development and, presumably, production. We'd like to be able to "catch" the error so that we can be sure that our fix works.

icarito:
"in development" are you using the db container or standalone mysql? there are version and configuration requirements... I'll review this as I think some changes to https://github.com/publiclab/plots2/blob/master/config/my.cnf may be necessary.

jywarren:
Ooh, ugh.. the mysql conf file has also gone out of sync. This is a bit more complex to rebase now.

icarito:
I love emojis too!
There is a good reason that this is delayed, I think I failed to express it clearly, is because I found that this is likely a long, delicate, manual process of migrating the database. It's not just a matter of merging this patch, unfortunately.
I'll start making a list of steps from below references:

Step 1: Create a backup [DONE]
Step 2: Upgrade the MySQL server [DONE]
Step 3: Modify databases, tables, and columns
Step 4: Check the maximum length of columns and index keys
Step 5: Modify connection, client, and server character sets
Step 6: Repair and optimize all tables

## References
* https://mathiasbynens.be/notes/mysql-utf8mb4

This tool is 404 but supposedly automates part of this:
* https://hanoian.com/content/index.php/24-automate-the-converting-a-mysql-database-character-set-to-utf8mb4

icarito:
Looking for freshed docs from MariaDB which we are using now.

icarito:
Here perhaps this will help with the tests?
From https://stackoverflow.com/questions/38363566/trouble-with-utf-8-characters-what-i-see-is-not-what-i-stored/38363567#38363567

## Test the data

Viewing the data with a tool or with SELECT cannot be trusted. Too many such clients, especially browsers, try to compensate for incorrect encodings, and show you correct text even if the database is mangled. So, pick a table and column that has some non-English text and do

SELECT col, HEX(col) FROM tbl WHERE ...

The HEX for correctly stored UTF-8 will be

    For a blank space (in any language): 20
    For English: 4x, 5x, 6x, or 7x
    For most of Western Europe, accented letters should be Cxyy
    Cyrillic, Hebrew, and Farsi/Arabic: Dxyy
    Most of Asia: Exyyzz
    Emoji and some of Chinese: F0yyzzww

icarito:
Found another reference [this one](http://dchua.com/2014/08/15/migrating-from-utf8-to-utf8mb4-in-mysql/)
- this one says we also need to alter `database.yml` to set encoding / collation (also note from elsewhere that the collation to use is `utf8mb4_0900_ai_ci`)

icarito:
I think this last reference that I shared looks the simplest / most modern.
So I say we merge this when we have this:

* [ ] Add to `my.cnf`
```
 # in /etc/mysql/my.cnf add the following in the correct sections
  [client]
  default-character-set = utf8mb4

  [mysqld]
  character-set-client-handshake = FALSE
  character-set-server = utf8mb4
  collation-server = utf8mb4_unicode_ci

  [mysql]
  default-character-set = utf8mb4
```
* [ ] write a migration like this:
```
# For each database:
ALTER DATABASE database_name CHARACTER SET = utf8mb4 COLLATE = utf8mb4_0900_ai_ci;
# For each table:
ALTER TABLE table_name CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci;
```
* [ ] modify `database.yml.example` (**note manual step in production**):
```
  # database.yml
  development:
    adapter: mysql2
    database: db
    username:
    password:
    encoding: utf8mb4
    collation: utf8mb4_0900_ai_ci
```

That should be about it!
To write the above migration here's some code in Ruby that could be used as a basis:
```
require 'rubygems'
require 'sequel'
require 'mysql2'
database = 'your_database_name'
username = 'your_user_name'
password = 'password'
host = 'localhost'

DB = Sequel.connect("mysql2://#{username}:#{password}@#{host}/#{database}")
sw_tables = DB["show tables"].all.map{|x| x.first[1]}
puts DB["ALTER DATABASE #{database} CHARACTER SET = utf8mb4 COLLATE = utf8mb4_unicode_ci"].all.inspect
sw_tables.each do |t|
  puts DB["ALTER TABLE #{t} CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;"].all.inspect
end
```

Also note that if there are indexed fields of type VARCHAR(255) then we may need the following initializer:
```
# config/initializer/mysqlpls.rb

require 'active_record/connection_adapters/abstract_mysql_adapter'

module ActiveRecord
  module ConnectionAdapters
    class AbstractMysqlAdapter
      NATIVE_DATABASE_TYPES[:string] = { :name => "varchar", :limit => 191 }
    end
  end
end
```

Credit for these instructions to [David Chua](http://dchua.com/2014/08/15/migrating-from-utf8-to-utf8mb4-in-mysql/).

db/migrate/20180721114724_change_comments_to_utf8mb4.rb
```diff
@@ -0,0 +1,8 @@
+class ChangeCommentsToUtf8mb4 < ActiveRecord::Migration[5.2]
+   def change
+    # changing table that will store unicode execute:
+    execute "ALTER TABLE comments CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
+    # changin string/text column with unicode content execute:
+    execute "ALTER TABLE comments MODIFY comment VARCHAR(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
```
icarito: Do we really feel comfortable just truncating the field?
Let's not truncate the field just yet (line 6), it shouldn't be necesssary.

Also let's use collation `utf8mb4_0900_ai_ci` it seems to be preferred (we run MariaDB 10.2 which supports this).

jywarren:
Thanks so much for pushing this along, Sebastian!!! I will plan to go
through it carefully.

SidharthBansal:

jywarren:
I wonder if this could be a hall-of-fame?

SidharthBansal:
Request accepted: Here is the task https://codein.withgoogle.com/dashboard/tasks/6653314876309504/

CodeSarthak:
> I love emojis too!
> There is a good reason that this is delayed, I think I failed to express it clearly, is because I found that this is likely a long, delicate, manual process of migrating the database. It's not just a matter of merging this patch, unfortunately.
> I'll start making a list of steps from below references:
> 
> Step 1: Create a backup [DONE]
> Step 2: Upgrade the MySQL server [DONE]
> Step 3: Modify databases, tables, and columns
> Step 4: Check the maximum length of columns and index keys [NOT NEEDED because we are using MariaDB > 5.7.7 (10.2)]
> Step 5: Modify connection, client, and server character sets
> Step 6: Repair and optimize all tables
> 
> ## References
> * https://mathiasbynens.be/notes/mysql-utf8mb4
> 
> This tool is 404 but supposedly automates part of this:
> 
> * https://hanoian.com/content/index.php/24-automate-the-converting-a-mysql-database-character-set-to-utf8mb4

Hey, I believe it would be better if the GCI task could also be broken down into multiple hard/hof tasks, according to the steps mentioned above, since it would be difficult to complete all the steps by a single person. 

Thanks.

icarito:
Jeff and I agreed to prioritize this so I'll be reviewing the reference https://mathiasbynens.be/notes/mysql-utf8mb4 again and try to break it down into smaller pieces so it is clear what needs to be done.

jywarren:
Models which could possibly contain emoji, as far as I can determine:

```
comment.rb
csvfile.rb
doc_list.rb
doc_result.rb
image.rb
node.rb
revision.rb
tag.rb
user.rb
user_session.rb
user_tag.rb
```

jywarren:
Hi @icarito - what if we, temporarily, screened for emoji using regex, using an ActiveRecord validation, before we save records. Here is some regex we might be able to use:

https://www.regextester.com/106421

https://github.com/mathiasbynens/emoji-regex

Actually i think we already have some ability to scan for emoji: 

https://github.com/publiclab/plots2/blob/591f59a720c4e173ea4ff07db53f8b64897ffe8e/app/helpers/application_helper.rb#L9-L46

Wait, we can use the REGEX provided by that lib, already installed, to replace them!

https://github.com/janlelis/unicode-emoji

Yes, something like this, in the `before_save` filter:

```ruby
require "unicode/emoji"
require "gemoji"

string = "String which contains all kinds of emoji:😴▶️🛌🏽🇵🇹🏴󠁧󠁢󠁳󠁣󠁴󠁿2️⃣🤾🏽‍♀️"

string.match(Unicode::Emoji::REGEX) do |match|
  string.gsub!(match.to_s,  ":" + Emoji.find_by_unicode(match.to_s) + ":")
end
```

Using `gemoji` to swap for the text name of the emoji: https://github.com/github/gemoji

jywarren:
Making a new issue for this: https://github.com/publiclab/plots2/issues/7469 🙌

icarito:
I guess this would be an interim solution but the thing is that emojis are not the only kind of invalid characters for this encoding. For instance acording to https://mathiasbynens.be/notes/mysql-utf8mb4 - he found the issue by trying to insert the U+1D306 TETRAGRAM FOR CENTRE (𝌆) symbol.

So I think it'd be a lot of effort for a few cases, and rather the encoding migration should be done.

icarito:
Hi @jywarren @namangupta01 I finally got around to working on this and made it into a new PR here https://github.com/publiclab/plots2/pull/7479 - also, after I did, I found a snippet that may be a nicer / more complete migration?
https://gist.github.com/amuntasim/f3b12f20a30e9a9f3fb0

Please check and review if we should test this in unstable. Migration is not reversible so I guess we should try to get our tests to work (get unstable to fail before pushing this?). Let me know what you think! I also found that not all emojis should trigger this problem, only those that require 4 bytes such as U+1F4A9 PILE OF POO (💩) or U+1D306 TETRAGRAM FOR CENTRE (𝌆)

Regards,
Sebastian

icarito:
Okay an update on this. I tried the latest patch on `unstable` and the mysql database crashed leaving the database corrupt.
I believe trying to rebuild the entire database is too heavy.
I see two ways forward. One is trying to do this differently / tweak the command and even upgrade mysql if it will help.
The other is looking for a way to change the encoding by dumping the whole database then changing the encoding and then uploading the dump back. This should be way less processing if it can handle, but likely it would mean we can't do it in a migration (can't we? maybe we can....). Looking for solutions here.

icarito:
Restoring a fresh dump from backup for unstable.

icarito:
I am feeling and channeling the focus on this issue.    :eyeglasses: :comet: :superhero: 

Here's my current assessment and plan.

Since Naman's well started PR, I expanded the code to convert every one of our database tables and fields to the new encoding, but the times we tried to run this in the staging database, it crashes and corrupts the database, probably because of the size of our data overflowing our RAM.

The plan I have now is as follows:
  - [ ] Drop the database of `unstable` staging instance
  - [ ] Recreate the database and make sure with the right encoding
  - [ ] Export (mysqldump) the data from production (see if can export subset, e.g. "SQL LIMIT")
    - [ ] Data may need to be converted to `utf8mb4` _before_ importing it to staging.
  - [ ] Perform tests in unstable.publiclab.org
    - [ ] Check unicodes and international characters already in the database
    - [ ] Write new nodes and comments with problematic characters and check results

icarito:
This article from Moodle more or less describes the procedure we want to follow:
https://docs.moodle.org/32/en/Converting_your_MySQL_database_to_UTF8

icarito:
Thanks @namangupta01 @jywarren and everyone who contributed to this. I'm closing it in favour of #8003 which is the solution that worked! Good try though, it would've worked with a smaller database.

PR #3068, opened 2018-07-15
Add CodeClimate documentation

stefannibrasil:
@jywarren what do you think about this? is it okay to add the images on that folder? Thanks!


jywarren:
It looks great but I think we can make a folder like /doc/images/ - how's that sound? Thanks!!!

jywarren:
Awesome, thank you!!!

Issue #3070, opened 2018-07-15
Refactor Search API


stefannibrasil:
- [x] Remove API Search endpoints duplication (#3045) 
- [x] Remove API Typeahead endpoints duplication (#3107)
- [x] Unify models scopes from Search and Typeahead services (#3108)

PR #3108, opened 2018-07-20
Move common scope from search api to a new file

stefannibrasil:
The third part of #3070.

(we commented the unit test for this for now because we are going to separate a time only for testing! but by running the api docs locally we could still get the same results as before, so apparently it didn't changed the fetched results)

### DIFF
--- a/app/services/typeahead_service.rb
+++ b/app/services/typeahead_service.rb
@@ -3,47 +3,15 @@
 # Though similar in operation to the SearchService, the implementation is separate, in that the goal of the response
 # is to provide _fast_ returns at a higher level than a general search.  In effect, TypeaheadService provides pointers to
 # better searches, while SearchService provides deep and detailed information.
-# TODO: Refactor TypeaheadService and SearchService so that common functions come from a higher level class?
+#
+# See SrchScope class for more details about the reusable scope
+# that Typeahead and Search services use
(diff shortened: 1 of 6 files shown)


app/services/typeahead_service.rb
```diff
@@ -125,7 +79,8 @@ def search_profiles(search_string, limit = 5)
   def search_notes(search_string, limit = 5)
```
jywarren: Maybe we should add to the comment to say "and package up as a TagResult" for clarity?

jywarren:
This is looking good - so when you add the test back in, and are ready for final review, just ping me! 

These files are getting much simpler, very nice. Keep an eye out for layers of abstraction that seem unnecessary! And keep in mind how we want to be able to tell new people about how this is organized - it should be easy for a newcomer to figure out how it's all wired together, and to make a change or addition. 

Thanks, great work!

app/services/typeahead_service.rb
stefannibrasil: yeah, good point!

jywarren:
Great! When you think it's ready, can you start putting the `ready` label on it? Thanks and great work!

stefannibrasil:
great, done, thanks for the help!

stefannibrasil:
closes #3070

PR #3045, opened 2018-07-11
Remove API Search endpoints duplication

stefannibrasil:
Hello, everyone!

We are opening this PR to get some feedback about our work with the API.

In order to refactor how the searches are being made, we decided to first try to remove some **duplications**. We have done some small changes but we wanted to know what do you think and see if there is another thing we can remove/change.

@publiclab/reviewers @publiclab/soc @jywarren

PR #3126, opened 2018-07-22
Search endpoint most recently people

### LINKED ISSUE #3069: Show most recently updated people on Search API
Part of #2755

Change the sorting of the results in the API to most recently updated people as discussed in #2925

milaaraujo:
Hey everyone!

We are opening this request to fix the issue #3069. 

1. We created a new endpoint `recentprofiles` that returns the profiles with most recent activity.
2. We fixed `recentPeople` method. In one of our latest PR (#3045) we did the following change that was not correct. Because in this part of the code the limit is for the node search. So we undid this change and added a count to control the number of profiles returned ([line 290](https://github.com/publiclab/plots2/pull/3126/files#diff-9546dd6074cdca7a128c2bf04fa5367aR290)).

>nodes = Node.all.order("changed DESC").limit(100).distinct -> nodes = Node.all.order("changed DESC").limit(srchString).distinct

3. We don't think the name of the `recentPeople` method is appropriate for the `peopleLocations` endpoint, since the method returns the location of people with most recent activity with a specific tag. Can we change it to `peopleLocations`? So maybe we can use `recentPeople` for our new method.

We would like some feedback about our strategy here, thank you! :)

fixes #3069 

### DIFF
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -34,6 +34,18 @@ module Srch
         Search.execute(:profiles, params)
       end
 
+      # Request URL should be /api/srch/recentprofiles?srchString=QRY[&seq=KEYCOUNT&showCount=NUM_ROWS&pageNum=PAGE_NUM]
+      desc 'Perform a search of profiles', hidden: false,
+                                           is_array: false,
+                                           nickname: 'srchGetRecentProfiles'
+
+      params do
+        use :common
+      end
+      get :recentprofiles do
+        Search.execute(:recentprofiles, params)
+      end
+
       # Request URL should be /api/srch/notes?srchString=QRY[&seq=KEYCOUNT&showCount=NUM_ROWS&pageNum=PAGE_NUM]
       # Basic implementation from classic plots2 SearchController
       desc 'Perform a search of research notes', hidden: false,
--- a/app/services/execute_search.rb
+++ b/app/services/execute_search.rb
@@ -13,6 +13,8 @@ class ExecuteSearch
        sresult = sservice.textSearch_all(search_criteria.query)
      when :profiles
        sresult = sservice.textSearch_profiles(search_criteria.query)
+     when :recentprofiles
+       sresult = sservice.textSearch_recentProfiles(search_criteria.query)
      when :notes
        sresult = sservice.textSearch_notes(search_criteria.query)
      when :questions
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -131,7 +131,6 @@ class SearchService
     sresult
   end
 
-  #####################################################################################
   # Search for more recently updated profiles for matching text
   def textSearch_recentProfiles(srchString)
     sresult = DocList.new
@@ -155,7 +154,6 @@ class SearchService
 
     sresult
   end
-  #####################################################################################
 
   # Search notes for matching strings
   def textSearch_notes(srchString)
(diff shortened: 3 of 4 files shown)


jywarren:
Wow this is great... Is this recentpeople call also able to accept a query
string, so recent people starting with ... characters?

If so, would you think it might be a good match for the new autocomplete
comment interface by @cheneyshreve in
https://github.com/publiclab/plots2/pull/2925 ?

Even if not, that autocomplete is returning a relatively unsorted set of
usernames so maybe there is an opportunity to refine those results!

Awesome!

stefannibrasil:
HI, @jywarren we believe so because right now it is receiving a searchString and we are not doing any restriction about the string format. 

We tested locally and it is returning the profiles with the most recent activity, so we think we can help with the #2925 new feature, yeah :)

stefannibrasil:
this week we will refine the method, try to return a better list of results and then we'll fix the codeclimate issues, we wanted first to know if that is what you needed for the mentioned PR :)

oh, what do you think of changing the method name for the `peoplelocations` endpoint? we think that `recentPeople` for this endpoint does not reflect what it is doing... let us know anything!

jywarren:
And did you want to change `recentPeople` or `peoplelocations`, and to what? Thanks!

jywarren:
Ah, sorry, i see you asked for clarification here: https://github.com/publiclab/plots2/issues/3069#issuecomment-405454231

Could we imagine these as a single endpoint, which has different sorting parameters, like a keyword to match usernames against (if there's no keyword, it just shows recent profiles?) and that the default sort is recency of activity, but it can be set to other sorts, like maybe created date? 

Just trying to think what would be simplest!

jywarren:
See how, for example, https://publiclab.org/tags gets additional parameters, like https://publiclab.org/tags?sort=uses&order=desc when you click the headers. Do you think this would be more compact to code? Then perhaps it could just be `/people` and the `recent` would be a sort, rather than the endpoint itself? 

Not set on this, just wondering what you think! Thanks!

milaaraujo:
Hi @jywarren,  

> What's the difference betwen `recentPeople` and `recentprofiles`?

`textSearch_recentPeople` is the method called for the `/api/srch/peoplelocations` endpoint. The method gets X number of latest contributors - with X equal to srchString - optionally we can also filter the search by tag. For example, if I call `/api/srch/peoplelocations?srchString=10&tagName=collaborator` the api will return the location of the 10 users with most recent activities and with the tag 'collaborator'. Or if I call just `/api/srch/peoplelocations?srchString=10` the api will return the location of the 10 users with the most recent activities. 

`/api/srch/recentprofiles` is the new endpoint we are creating to search for more recently updated profiles for matching text. It works like `profile` endpoint, however it sorts users and returns those with recent activity first.

milaaraujo:
@stefannibrasil and I are thinking that maybe, for now, it would it be simpler just to change the implementation of endpoint `/api/srch/profiles` by our implementation `/api/srch/recentprofiles`. So we would have only one endpoint that always returns the most recent users. Then we could open other pull requests to add other ways to sort the results as you suggested. But for now, it makes sense that by default we order by activity, right?

jywarren:
Agreed on ordering. And is there a possibility to alias one to the other? I
worry if anyone has begun using one of these already.

Have you tried requesting both and comparing their responses? I'd be
interested!

milaaraujo:
Hey @jywarren, I implemented a new version of `profiles` endpoint following your idea. 

1. If we call '/api/srch/profiles?srchString=Jeff' the api perform the default search.
2. If we call '/api/srch/profiles?srchString=Jeff&order=recentdesc' the api returns more recently updated profiles for matching text.

That way I do not take the risk of breaking anything that already uses the endpoint. I'm going to open a new PR with these changes.

PR #3134, opened 2018-07-24
Improve textSearch_profiles result

### LINKED ISSUE #3069: Show most recently updated people on Search API

milaaraujo:
Hey everyone!

We are opening this request to fix the issue #3069. 

1. We created a new endpoint `recentprofiles` that returns the profiles with most recent activity.
2. We fixed `recentPeople` method. In one of our latest PR (#3045) we did the following change that was not correct. Because in this part of the code the limit is for the node search. So we undid this change and added a count to control the number of profiles returned ([line 290](https://github.com/publiclab/plots2/pull/3126/files#diff-9546dd6074cdca7a128c2bf04fa5367aR290)).

>nodes = Node.all.order("changed DESC").limit(100).distinct -> nodes = Node.all.order("changed DESC").limit(srchString).distinct

3. We don't think the name of the `recentPeople` method is appropriate for the `peopleLocations` endpoint, since the method returns the location of people with most recent activity with a specific tag. Can we change it to `peopleLocations`? So maybe we can use `recentPeople` for our new method.

We would like some feedback about our strategy here, thank you! :)

fixes #3069 

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -33,6 +33,7 @@ class User < ActiveRecord::Base
   has_many :following_users, through: :active_relationships, source: :followed
   has_many :followers, through: :passive_relationships, source: :follower
   has_many :likes
+  has_many :revisions, through: :node
 
   validates_with UniqueUsernameValidator, on: :create
   validates_format_of :username, with: /\A[A-Za-z\d_\-]+\z/
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -43,7 +43,7 @@ class User < ActiveRecord::Base
   after_destroy :destroy_drupal_user
 
   def self.search(query)
-    User.where('MATCH(username, bio) AGAINST(?)', query)
+    User.where('MATCH(username, bio) AGAINST (?)', query)
   end
 
   def new_contributor
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -43,7 +43,7 @@ class User < ActiveRecord::Base
   after_destroy :destroy_drupal_user
 
   def self.search(query)
-    User.where('MATCH(username, bio) AGAINST (?)', query)
+    User.where('MATCH(username) AGAINST((?) IN BOOLEAN MODE)', query + '*')
   end
 
   def new_contributor
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -42,7 +42,7 @@ class User < ActiveRecord::Base
   before_save :set_token
   after_destroy :destroy_drupal_user
 
-  def self.search(query)
+  def self.search_by_username(query)
     User.where('MATCH(username) AGAINST((?) IN BOOLEAN MODE)', query + '*')
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -42,8 +42,8 @@ class User < ActiveRecord::Base
   before_save :set_token
   after_destroy :destroy_drupal_user
 
-  def self.search_by_username(query)
-    User.where('MATCH(username) AGAINST((?) IN BOOLEAN MODE)', query + '*')
+  def self.search(query)
+    User.where('MATCH(bio, username) AGAINST(? IN BOOLEAN MODE)', query + '*')
   end
 
   def new_contributor
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -44,16 +44,23 @@ class SearchService
   end
 
   # Search profiles for matching text and package up as a DocResult
-  # If order == "recentdesc", search for more recently updated profiles for matching text
+  # If order == "recentdesc", search for more recently updated profiles for matching text.
+  # Otherwise, show profiles without any ordering
   def textSearch_profiles(srchString, order = nil)
     sresult = DocList.new
 
-    sresult =
+    users =
       if order == "recentdesc"
-        recentProfiles(srchString)
+        getRecentProfiles(srchString)
       else
-        defaultProfiles(srchString)
+        SrchScope.find_users(srchString, limit = nil)
       end
+
+    # User profiles
+    users.each do |match|
+      doc = DocResult.fromSearch(0, 'user', '/profile/' + match.name, match.name, '', 0)
+      sresult.addDoc(doc)
+    end
 
     sresult
   end
@@ -170,7 +177,44 @@ class SearchService
   # X = srchString
   def recentPeople(srchString, tagName = nil)
     sresult = DocList.new
+    users = getRecentProfilesTag(tagName)
+    count = 0
 
+    users.each do |user|
+      next unless user.has_power_tag("lat") && user.has_power_tag("lon")
+      blurred = false
+      if user.has_power_tag("location")
+        blurred = user.get_value_of_power_tag("location")
+      end
+      doc = DocResult.fromLocationSearch(user.id, 'people_coordinates', user.path, user.username, 0, 0, user.lat, user.lon, blurred)
+      sresult.addDoc(doc)
+      count += 1
+      if count == srchString.to_i then break end
+    end
+
+    sresult
+  end
+
+  # Get users with matching name ordering by most recent activity
+  def getRecentProfiles(srchString)
+    sresult = DocList.new
+
+    nodes = Node.all.order("changed DESC").limit(100).distinct
+    users = []
+    nodes.each do |node|
+      next unless node.author.status != 0
+      if node.author.name.downcase.include? srchString.downcase
+        users << node.author.user
+      end
+    end
+
+    users = users.uniq
+  end
+
+  # If a tagName is provided,
+  # method returns users with the specific tag ordering by most recent activity.
+  # Otherwise, it returns users with most recent activity.
+  def getRecentProfilesTag(tagName = nil)
     nodes = Node.all.order("changed DESC").limit(100).distinct
     users = []
     nodes.each do |node|
@@ -184,58 +228,5 @@ class SearchService
     end
 
     users = users.uniq
-    count = 0
-
-    users.each do |user|
-      next unless user.has_power_tag("lat") && user.has_power_tag("lon")
-      blurred = false
-      if user.has_power_tag("location")
-        blurred = user.get_value_of_power_tag("location")
-      end
-      doc = DocResult.fromLocationSearch(user.id, 'people_coordinates', user.path, user.username, 0, 0, user.lat, user.lon, blurred)
-      sresult.addDoc(doc)
-      count += 1
-      if count == srchString.to_i
-        break
-      end
-    end
-
-    sresult
-  end
-
-  def recentProfiles(srchString)
-    sresult = DocList.new
-
-    nodes = Node.all.order("changed DESC").limit(100).distinct
-    users = []
-    nodes.each do |node|
-      next unless node.author.status != 0
-      if node.author.name.downcase.include? srchString.downcase
-        users << node.author.user
-      end
-    end
-
-    users = users.uniq
-
-    # User profiles
-    users.each do |match|
-      doc = DocResult.fromSearch(0, 'user', '/profile/' + match.name, match.name, '', 0)
-      sresult.addDoc(doc)
-    end
-
-    sresult
-  end
-
-  def defaultProfiles(srchString)
-    sresult = DocList.new
-
-    users = SrchScope.find_users(srchString, limit = nil)
-    # User profiles
-    users.each do |match|
-      doc = DocResult.fromSearch(0, 'user', '/profile/' + match.name, match.name, '', 0)
-      sresult.addDoc(doc)
-    end
-
-    sresult
   end
 end
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -54,9 +54,9 @@ class SearchService
                           .reorder('')
 
     user_scope =
-      if search_criteria.order_by == "recent"
-        user_scope.left_outer_joins(:revisions)
-                  .order("node_revisions.timestamp #{search_criteria.sort_direction}")
+      if search_criteria.sort_by == "recent"
+        user_scope.joins(:revisions)
+                  .order("node_revisions.timestamp #{search_criteria.order_direction}")
                   .distinct # do we need unique/distinct on user id ?
 
       else
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -44,7 +44,7 @@ class SearchService
   end
 
   # Search profiles for matching text and package up as a DocResult
-  # If order == "recentdesc", search for more recently updated profiles for matching text.
+  # If order == "recentdesc", search for more recently updated profiles for matching name.
   # Otherwise, show profiles without any ordering
   def textSearch_profiles(srchString, order = nil)
     sresult = DocList.new
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -202,10 +202,8 @@ class SearchService
     nodes = Node.all.order("changed DESC").limit(100).distinct
     users = []
     nodes.each do |node|
-      next unless node.author.status != 0
-      if node.author.name.downcase.include? srchString.downcase
-        users << node.author.user
-      end
+      next unless (node.author.status != 0) and (node.author.name.downcase.include? srchString.downcase)
+      users << node.author.user
     end
 
     users = users.uniq
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -202,7 +202,7 @@ class SearchService
     nodes = Node.all.order("changed DESC").limit(100).distinct
     users = []
     nodes.each do |node|
-      next unless (node.author.status != 0) and (node.author.name.downcase.include? srchString.downcase)
+      next unless (node.author.status != 0) && (node.author.name.downcase.include? srchString.downcase)
       users << node.author.user
     end
 
@@ -219,8 +219,8 @@ class SearchService
       if node.author.status != 0
         if tagName.blank?
           users << node.author.user
-        else
-          users << node.author.user if node.author.user.has_tag(tagName)
+        elsif node.author.user.has_tag(tagName)
+          users << node.author.user
         end
       end
     end
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -50,8 +50,7 @@ class SearchService
   # If no sort_by value present, then it returns a list of profiles ordered by id DESC
   # a recent activity may be a node creation or a node revision
   def textSearch_profiles(search_criteria)
-    user_scope = SrchScope.find_users(search_criteria.query, limit = nil)
-                          .reorder('')
+    user_scope = User.where('username LIKE ? AND rusers.status = 1', '%' + search_criteria.query + '%')
 
     user_scope =
       if search_criteria.sort_by == "recent"
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -215,14 +215,10 @@ class SearchService
   def getRecentProfilesTag(tagName = nil)
     nodes = Node.all.order("changed DESC").limit(100).distinct
     users = []
+
     nodes.each do |node|
-      if node.author.status != 0
-        if tagName.blank?
-          users << node.author.user
-        elsif node.author.user.has_tag(tagName)
-          users << node.author.user
-        end
-      end
+      next unless node.author.status != 0 && ((tagName && node.author.user.has_tag(tagName)) || tagName.nil?)
+      users << node.author.user
     end
 
     users = users.uniq
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -44,15 +44,16 @@ class SearchService
   end
 
   # Search profiles for matching text and package up as a DocResult
-  def textSearch_profiles(srchString)
+  # If order == "recentdesc", search for more recently updated profiles for matching text
+  def textSearch_profiles(srchString, order = nil)
     sresult = DocList.new
 
-    users = SrchScope.find_users(srchString, limit = 10) # don't return hundreds!!
-    # User profiles
-    users.each do |match|
-      doc = DocResult.fromSearch(0, 'user', '/profile/' + match.name, match.name, '', 0)
-      sresult.addDoc(doc)
-    end
+    sresult =
+      if order == "recentdesc"
+        recentProfiles(srchString)
+      else
+        defaultProfiles(srchString)
+      end
 
     sresult
   end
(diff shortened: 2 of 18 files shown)


milaaraujo:
Moved from #3126
----------------------
Hey @jywarren, I implemented a new version of profiles endpoint following your idea.

If we call '/api/srch/profiles?srchString=Jeff' the api perform the default search.
If we call '/api/srch/profiles?srchString=Jeff&order=recentdesc' the api returns more recently updated profiles for matching text.
That way I do not take the risk of breaking anything that already uses the endpoint. I'm going to open a new PR with these changes.

milaaraujo:
Hey @jywarren, regarding your question from the other PR (# 3126):
>Have you tried requesting both and comparing their responses?

I created a silly test to illustrate the difference between the two ways of using the `profiles` endpoint. 
1. First, I created users, in this order: user5, user4, user1, user2 and user3. 
2. Then I created an activity (question) for users, in this order: user5, user4 and user3.
3. If I call `http://localhost:3000/api/srch/profiles?srchString=test`, then:
>{
  "items": [
    {
      "docId": 0,
      "docType": "user",
      "docUrl": "/profile/test3",
      "docTitle": "test3",
      "docSummary": "",
      "docScore": 0
    },
    {
      "docId": 0,
      "docType": "user",
      "docUrl": "/profile/test2",
      "docTitle": "test2",
      "docSummary": "",
      "docScore": 0
    },
    {
      "docId": 0,
      "docType": "user",
      "docUrl": "/profile/test1",
      "docTitle": "test1",
      "docSummary": "",
      "docScore": 0
    },
    {
      "docId": 0,
      "docType": "user",
      "docUrl": "/profile/test4",
      "docTitle": "test4",
      "docSummary": "",
      "docScore": 0
    },
    {
      "docId": 0,
      "docType": "user",
      "docUrl": "/profile/test5",
      "docTitle": "test5",
      "docSummary": "",
      "docScore": 0
    }
  ],
  "srchParams": {
    "srchString": "test",
    "seq": null,
    "showCount": null,
    "pageNum": null,
    "tagName": null
  }
}
4. If I call `http://localhost:3000/api/srch/profiles?srchString=test&order=recentdesc`, then:
>{
  "items": [
    {
      "docId": 0,
      "docType": "user",
      "docUrl": "/profile/test3",
      "docTitle": "test3",
      "docSummary": "",
      "docScore": 0
    },
    {
      "docId": 0,
      "docType": "user",
      "docUrl": "/profile/test4",
      "docTitle": "test4",
      "docSummary": "",
      "docScore": 0
    },
    {
      "docId": 0,
      "docType": "user",
      "docUrl": "/profile/test5",
      "docTitle": "test5",
      "docSummary": "",
      "docScore": 0
    }
  ],
  "srchParams": {
    "srchString": "test",
    "seq": null,
    "showCount": null,
    "pageNum": null,
    "tagName": null
  }
}

jywarren:
Looks like a good moment to address this optimization issue too! https://github.com/publiclab/plots2/issues/3147

app/services/search_service.rb
```diff
@@ -190,8 +191,47 @@ def recentPeople(srchString, tagName = nil)
...
     end
 
     sresult
   end
+
+  # Get users with matching name ordering by most recent activity
+  def getRecentProfiles
+    sresult = DocList.new
+
+    nodes = Node.all.order("changed DESC").limit(100).distinct
```
jywarren: I'm a little worried about the speed of this query. Could we collect up the `uid`s, then run:

```ruby
User.find_by(uid: uids.uniq).where(status: 1)
```

Then, we may also want to get recent edits, so what if we instead of doing `Node`, we searched `Revision` - that way we get both notes and wiki revisions?

app/services/search_service.rb
```diff
@@ -44,10 +44,18 @@ def textSearch_all(srchString)
...
-  def textSearch_profiles(srchString)
+  # If order == "recentdesc", search for more recently updated profiles for matching name.
+  # Otherwise, show profiles without any ordering
+  def textSearch_profiles(srchString, order = nil)
     sresult = DocList.new
 
-    users = SrchScope.find_users(srchString, limit = nil)
+    users =
+      if order == "recentdesc"
+        filterMatchingUserName(getRecentProfiles, srchString)
```
jywarren: Here, I worry we might not get any results if we filter after selecting the most recent 100. Is there another way to do this, like finding matching usernames first, sorting by last activity using the user table -- i think there's a last activity column, but not sure when we update that...

jywarren: It might take some kind of clever `join` and `group` query, like, joining revisions to the node table, and sorting by the revision timestamp...

jywarren:
Hi, sorry, we've created a conflict while fixing https://github.com/publiclab/plots2/pull/3148 - but it should be pretty easy to resolve. Thanks!

stefannibrasil:
@jywarren agree with your comments! we will talk with our coaches today to help us improve the query, will keep you updated!

stefannibrasil:
@jywarren we started doing the sorting and ordering today, still need to work on some things, but we believe we are targeting your concerns on them.

We are passing the a `order_by` and `sort_direction` as a parameter to the `search/profiles?` so we can use it on them.

The approach that our coaches are suggesting us to so is searching the profiles, joining them with nodes and then ordering them by the latest activities (ASC or DESC).

stefannibrasil:
hi @jywarren could you review here, please? We worked today on improving the `textSearch_profiles`  query. We have some questions, could you please take a look at the comments along the code?

We will work on the `recentProfiles` later after we finish this one. Also tomorrow we will write a test.

jywarren:
Great! First, looks like a missing or extra "end" in this file?

> SyntaxError: /app/app/services/search_service.rb:251: syntax error, unexpected keyword_end

Reading now...

app/services/search_service.rb
```diff
@@ -44,11 +44,33 @@ def textSearch_all(srchString)
...
+    # recent eh o sort
+    # order profiles in desc order
+    # search profiles whre username Like = 'data' order by Node.changed DESC
+
+    user_scope = SrchScope.find_users(search_criteria.query, limit = nil)\
+                          .reorder('') # removing order by id from find_users()
+
+    if search_criteria.order_by == "recent"
+      user_scope = user_scope.joins(:revisions)\
+                             .order("timestamp #{search_criteria.sort_direction}")\
```
jywarren: I guess i have a couple thoughts here. In general, I think it's best not to be adding activeRecord filters or conditions to an object that's had them added at an earlier level of abstraction (like, within SrchScope.find_users), because you can't see them when you're editing, so there's a strong possibility of doing things redundantly or in conflict. Better to do all your ActiveRecord stuff in one place as much as possible. 

Second, I wonder if there's a way to reduce the # of levels of abstraction. Here it seems like we do:

1. A call arrives at `app/api/srch/search.rb` for `get :profiles do`
2. its sent to `app/services/execute_search.rb` for `execute()`
3. that's sent to `app/services/search_service.rb` for `textSearch_profiles()`
4. that's sent to `app/services/srch_scope.rb` for `SrchScope.find_users()`
5. finally an ActiveRecord call is made: `User.search()...`

Each file is now much more readable, but tracing the whole thing is a bit harder. The readability improvements are super, and the files are cleaner and less redundant. Great work! Now, I wonder if we could try flattening the system without losing those advantages? One way to try (and we must admit it's possible every layer is important! But let's try.) is to list why each layer is needed.

1. where incoming calls get routed
2. utility 'execute' function
3. back-end standard service available inside the app
4. scope?
5. do the ActiveRecord call

I'm not saying there is an obvious or easy answer here. And if we stick with 5 layers, that's fine! But this is how I'd recommend we look at the question of "what complexity is necessary" and how readable it is. Make sense? What do you think?

To your narrower question of the optimization, I think it makes sense that you're doing an initial call for users, then making a more complex version of it by joining revisions and then sorting using that new joined table. This looks great! Because the incoming query is already going to be pretty efficient, we think, and then we're just joining something and adding an ordering. How do you like the results it returns?

stefannibrasil:
hey, @jywarren we have been working on the textSearch_profiles. Now it searches for a matching text with optional `order_by=recent` param and `sorted_direction DESC` by default.

If no sort_by value present, then it returns a list of profiles ordered by id DESC.

stefannibrasil:
We still left two comments for things that we are not sure and we left them to show you. Could you please take a look at them? Thanks!

@jywarren The list results are being returned as expected, we also added more tests to reflect that. Maybe using it on a stable branch before pushing to production is a good thing to do?

stefannibrasil:
@jywarren About the `recentProfiles` endpoint, we are not working on that yet. I know that this name suggests that it returns the most recent profiles, but in fact, it returns the profiles locations. We are planning to work on all endpoints, especially this one, but only after we finish this one :)

And regarding the new classes added, we did that because are following the OOP patterns for breaking the classes by responsibility and turn them into smaller objects. But yeah, we can put them back into the way they were, later.

Also, we believe travis is tricking us, we ran the tests locally several times and we don't get any errors ><

milaaraujo:
Hey, @publiclab/reviewers and @jywarren!

We need some help here...Travis' tests are always finishing with errors, but as @stefannibrasil said, we ran the tests locally several times without any error. Do you have any idea what might be happening?

desenho sem titulo

jywarren:
Hi, we had a recurrence of a bug with Danger, which I think is now resolved. Restarting!

jywarren:
OK, at least the errors are now properly reported by `plotsbot` -- let's see...

test/functional/search_api_test.rb
```diff
@@ -91,14 +92,62 @@ def app
 
      json = JSON.parse(last_response.body)
 
-     assert_equal "/profile/jeff", json['items'][1]['docUrl']
-     assert_equal "jeff",          json['items'][1]['docTitle']
-     assert_equal "user",          json['items'][1]['docType']
+     assert_equal "/profile/steff3",   json['items'][0]['docUrl']
```
jywarren: OK, error 1 here: `NoMethodError: undefined method ‘[]’ for nil:NilClass`

test/functional/search_api_test.rb
```diff
@@ -91,14 +92,62 @@ def app
...
+         srchString: 'steff',
+         seq: nil
+       }.ignore_extra_keys!
+     }.ignore_extra_keys!
+
+     matcher = JsonExpressions::Matcher.new(pattern)
+
+     json = JSON.parse(last_response.body)
+
+     assert_equal "/profile/steff3",     json['items'][0]['docUrl']
```
jywarren: Error 2 here: `NoMethodError: undefined method ‘[]’ for nil:NilClass`

test/functional/search_api_test.rb
```diff
@@ -91,14 +92,62 @@ def app
...
+         srchString: 'steff',
+         seq: nil
+       }.ignore_extra_keys!
+     }.ignore_extra_keys!
+
+     matcher = JsonExpressions::Matcher.new(pattern)
+
+     json = JSON.parse(last_response.body)
+
+     assert_equal "/profile/steff1",     json['items'][0]['docUrl']
```
jywarren: Error 3 here: `NoMethodError: undefined method ‘[]’ for nil:NilClass`

test/functional/search_api_test.rb
jywarren: It looks like things are not nested the same way as they were in the original test. We should probably conform to the original output format... can you check it against what https://publiclab.org/api/srch/profiles?srchString=steff currently returns?

```
{"items":[{"docId":0,"docType":"user","docUrl":"/profile/iujiu86","docTitle":"iujiu86","docSummary":"","docScore":0}],"srchParams":{"srchString":"steff","seq":null,"showCount":null,"pageNum":null,"tagName":null}}
```

I think to debug this, the best thing to do is to do `puts json` above this line and read it when you hit https://localhost:3000/api/srch/profiles?srchString=admin (or something like that, on your local instance). Make sense? It ought to match the above, and if not, maybe the API output needs to be tweaked?

stefannibrasil:
hey, @jywarren it was because locally we are using SQlite3 and travis uses maria db... our coaches really helped us here, but we thought that maybe we could discuss more about this here on #3170

now we can focus on the logic for this endpoint! Did you find anything about this PR? We are working on the Author endpoint this week!

app/services/srch_scope.rb
```diff
@@ -3,12 +3,12 @@ class SrchScope
   def self.find_users(input, limit)
     if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
       User.search(input)
-          .order('id DESC')
-          .where(status: 1)
+          .order('id DESC') # do we need this line?
```
jywarren: This orders returned profiles roughly by creation date, since the `id` column is auto-incrementing. If we're ordering later, we can remove it from here.

app/services/search_service.rb
```diff
@@ -11,44 +11,60 @@ def initialize; end
...
+  # If no sort_by value present, then it returns a list of profiles ordered by id DESC
+  # a recent activity may be a node creation or a node revision
+  def textSearch_profiles(search_criteria)
+    user_scope = User.where('username LIKE ? AND rusers.status = 1', '%' + search_criteria.query + '%')
+
+    user_scope =
+      if search_criteria.sort_by == "recent"
+        user_scope.joins(:revisions)
+                  .order("node_revisions.timestamp #{search_criteria.order_direction}")
+                  .distinct # do we need unique/distinct on user id ?
```
jywarren: we should be sure that `.distinct` is acting on users, not revisions... we don't want lots of repeat users because there are lots of revisions!

app/services/search_service.rb
```diff
@@ -11,44 +11,60 @@ def initialize; end
...
+  # Search profiles for matching text with optional order_by=recent param and
+  # sorted direction DESC by default
+  # then the list is packaged up as a DocResult
 
-    users = SrchScope.find_users(srchString, limit = 10) # don't return hundreds!!
-    # User profiles
+  # If no sort_by value present, then it returns a list of profiles ordered by id DESC
+  # a recent activity may be a node creation or a node revision
+  def textSearch_profiles(search_criteria)
+    user_scope = User.where('username LIKE ? AND rusers.status = 1', '%' + search_criteria.query + '%')
```
jywarren: Here i'm a bit worried -- the full text search provided by the `User.search` as part of the previous `SrchScope.find_users()` was much more efficient, and this is an open ended text match which is very expensive when run against 300k+ users. 

See:

https://github.com/publiclab/plots2/blob/master/app/services/srch_scope.rb#L5-L8

Is it possible to keep using that `.search` method so that we don't get hit hard on a very big database?

app/services/search_service.rb
milaaraujo: Yes, it is working properly!

app/services/search_service.rb
milaaraujo: Hey @jywarren,

We will need some help here. In my last commit, I sent the code using `SrchScope.find_users()`. It is working fine locally, but it doesn't work here... Probably because here it gets in in this condition `"if ActiveRecord::Base.connection.adapter_name == 'Mysql2'` and then calls `User.search(input)`. And locally we just run `User.where('username LIKE ? AND rusers.status = 1', '%' + input + '%')`.

Analyzing Travis' results it seems that `User.search` is not returning anything and so the error: NoMethodError: undefined method `[]' for nil:NilClass. But I don't know what we can do now, how to debug this if we are using different databases and locally it is working fine?

jywarren:
are you able to install mysql locally to test this out? That'd be much
faster feedback cycle than relying on Travis. It's not too bad to install!

stefannibrasil:
hi @jywarren we installed MySQL and finally were able to find the error. That User.search method was not returning what we expected... We added a migration to add an full-text index only on username. Now we are having an error because of the migration, how do we move forward with this? thanks

stefannibrasil:
Hi, @jywarren we need help here with the migration. I am not so sure how to proceed here. Travis is complaining about not running `rails db:migrate` ...

stefannibrasil:
To summarize what we did: we were able to use the `User.search` method to retrieve the users, as you requested, but we had to add a fulltext index only on the username.

It wasn't working because the `User.search` method only worked with the user bio and we were passing the user username. Testing with MySQL, we got this error:

`Can't find FULLTEXT index matching the column list`

because the username didn't have the fulltext index. So the problem was on the `User.search` because it wasn't working when we passed with username, so the tests were returning nil because it didn't get any users.

Now it works, tested with MySQL and it's faster.

jywarren:
Ah, ok - so, apart from resolving conflicts, you need to run the migration locally, then copy your changes into `schema.rb.example`, and push those up too. Then this should run smoothly! Sorry it took me a couple days to get back to you!

app/models/user.rb
```diff
@@ -41,8 +42,8 @@ class User < ActiveRecord::Base
   before_save :set_token
   after_destroy :destroy_drupal_user
 
-  def self.search(query)
```
jywarren: Ah, but perhaps we want to keep the more generalized `.search()` too, for searching by both username and bio -- for situations where we want both...

db/migrate/20180804042601_add_full_text_index_on_username.rb
```diff
@@ -0,0 +1,13 @@
+class AddFullTextIndexOnUsername < ActiveRecord::Migration[5.2]
+  def up
+    if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
```
jywarren: This is awesome!!! 🐎

jywarren:
I also want to say that it's so exciting to be making refinements at this level of sophistication. It's everything we'd hoped for in the API and search systems. Thank you!!!!

stefannibrasil:
Hi, @jywarren that would be great if the `User.search` method was working previously. That is the problem here!

It does not match against username, only works with bio, it never worked the way it was before we fixed it on this PR. So, for me it makes more sense, to make it work by searching for username for now.

That sounds to me that it should have its own issue and honestly, I would love to help with that, but we have other things to work on the API and having tests passing locally and not on travis, even with MySQL installed is not helping at all...

So, what I suggest is, leave it working as it is, where it works searching by username (which makes more sense, since it's a profiles/user search) and open an issue for someone else that understands MySQL better to fix that, so it returns both username and bio.

And about the Typeahead, there is no difference at all between it and the Search, they both do the same thing. The only difference is that the Typeahead returns results with different info, just a heads up!

jywarren:
Hi, ok -- so I very well may be mixed up here, but I believe User.search is being used in the navbar typeahead, which is why I can get my own profile by searching for `jeremijenko` (a string which is present in my bio) --

image

But I can also get a good result when searching by username:

image

Perhaps this means we should just leave `User.search` as it is for now, so as not to break this useful pair of behaviors? 

Otherwise, I agree that the changes to `User.search` are outside the scope of this PR, and we can open another issue. 

Sorry, this seems like a pain. I really appreciate your thoroughness and hard work on it!

stefannibrasil:
It's just that it makes no sense to me. Why is that the tests don't pass on travis when we use that method using a username? If you pass a username it breaks, but with a text from a bio it doesn't. Is that possible that travis is running the tests in a different way? I really don't understand, appreciate any insight here, I don't see what else it could be :thought_balloon: 

Anyway, I'll get back to it later. Anything else about this PR to change?

Oh, tomorrow we are having a meeting with the team, it would be great if you could attend. We just need to know if you are able to show :)

`It's time to have our all-team monthly call. How about Friday *August 10th 5:30 pm EET* ?`

jywarren:
> It's just that it does not make any sense to me. 

Me either! Let me look at the Travis tests. But first, I really think I need a cup of ☕️ 

As to the meeting, is there any chance to move the meeting a little earlier, to 4:45 or even 5:00? I have a local workshop i'd planned to run at 5:30, so I'd need to go by then. I'd very much like to hear your feedback/input on the summer so far! I hope I've been able to be helpful often enough!

jywarren:
I'm going to open a new PR to get Travis to run against the with-bio version of `User.search`...

stefannibrasil:
on travis it works only with bio... so you need to pass a username as the param to see the errors. For example, if you pass "jeff" it returns nil, but if you pass "something interesting about jeff" it returns something. At least that's how it was before

stefannibrasil:
okay, about the meeting, tomorrow, Friday *August 10th 5:00 pm EET*, then? :)

jywarren:
Wait, i'm sorry -- does that mean Eastern European Time?

https://www.worldtimebuddy.com/?qm=1&lid=5,13,5128581&h=5&date=2018-8-10&sln=10-10.5

If so, that would be fine. I'm sorry, I had thought ET (eastern time).
Thanks!

jywarren:
I think i have an idea about searching for "jeff"... hang on...did it happen only for "jeff" or also for "steff"?

Do you think you could try to reproduce the Travis error in a separate PR? Or was it exactly the error we're seeing here: https://github.com/publiclab/plots2/pull/3224 ?

jywarren:
And was the error you saw in `test/unit/user_test.rb` or in `/test/functional/search_api_test.rb#L95` like in my PR?

stefannibrasil:
yes, it's Eastern European Time! So, tomorrow 5:30 EET. See you there!

jywarren:
great, thanks.

jywarren:
OK, I modified the test PR at #3224 to build on your latest commits from this PR, and see what the issue was with the nil results. But I see you've modified to go back to `user.search` searching bio and username... did that work for you? It's passing Travis, and we can do the "switch to only username searching via parameter" variant in a separate PR if we want -- let's just get this PR merged!

Can you give me an update and I am happy to check in on it once I land.. it'll be sometime tomorrow ✈️ 

If it's ready to go, I approve and can merge it!

stefannibrasil:
Hi, @jywarren thanks for the approval and for the help! Saturday we were able to make it work searching for username and bio!! You can try now with our last commit if you want to check it out too! We don't know exactly where things started going wrong. I guess the errors that Travis was showing confused us and we missed something small and everything became a snowball of errors... 

Yeah, I agree that it's better to create a new PR for the autocomplete feature, will do that now!

So... I can't believe I'm saying this, LOL, but yes this can be merged!! :tada: thank you!!

app/models/user.rb
```diff
@@ -42,7 +43,7 @@ class User < ActiveRecord::Base
   after_destroy :destroy_drupal_user
 
   def self.search(query)
-    User.where('MATCH(username, bio) AGAINST(?)', query)
+    User.where('MATCH(bio, username) AGAINST(? IN BOOLEAN MODE)', query + '*')
```
stefannibrasil: @jywarren gold lesson from this PR about the MySQL fulltext search: the 'IN BOOLEAN MODE' modifier with the * wildcard.

https://dev.mysql.com/doc/refman/8.0/en/fulltext-boolean.html

really glad that this is working now, thanks for the help!

jywarren:
Congratulations!!! 👍🏼⚡⚡

PR #3150, opened 2018-07-25
add user.search_usernames

jywarren:


jywarren:
Looks like we need to select the full-text index too! https://dev.mysql.com/doc/refman/5.7/en/fulltext-search.html

PR #3098, opened 2018-07-19
Migrated Rusers

SidharthBansal:
Part of #3032

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [ ] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [ ] code is in uniquely-named feature branch and has no merge conflicts
* [ ] PR is descriptively titled
* [ ] PR body includes `fixes #0000`-style reference to original issue #
* [ ] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software

We have a loose schedule of reviewing and pulling in changes every Tuesday and Friday, and publishing changes on Fridays. 

Thanks!

PR #3158, opened 2018-07-26
Notifications field added while creating users with omniauth

### LINKED ISSUE #3032: Password checker field in OAuth Login
A user creates an account by the legacy authentication system. The field password_checker is set as zero. Now he tries to login. The username will be taken. We will check if the user having this username has password_checker field as zero. Yes the user has it set as zero we will check if the password matches and logs him in.

Let us define the fields for provider
Facebook 1
Github 2
Google 3
Twitter 4
A new user creates his account using provider. Password checker will be initialised as mentioned above. Then an email is sent to his account to change the password. If the user resets his password then password checker field is set to zero.
Suppose he forgets to change his password. So it will be a non zero field until he resets his password.
Without resetting the password user tries to login with the username and password. He enters wrong password. Then the backend flow will be like follows. We will fetch the user from the username Then we will fetch the password checker field. We see that this is 1 that means the user has created his account via Facebook. So we will notify the user that he has to change his password by going to forget password.

We will set the field of all the existing users in our database as zero.
If the field is zero and a user links his account then the password checker field need not to be changed.

We don't have to do anything with the user tag model here. Just we need to migrate the user model.
Part of #2676

SidharthBansal:
Closes #3032

Thanks!


SidharthBansal:
@jywarren there is some error in production vs development.
I pushed https://github.com/publiclab/plots2/pull/3098/files
But my laptop plots folder schema.rb shows
image

 I have created integer in schema.rb.example. So, why is it showing me string in schema.rb?

SidharthBansal:
@jywarren we migrated users instead of rusers in #3098. Can you please rollback the db for changes of #3098 before merging this one? Sorry for the inconvenience.

SidharthBansal:
image

jywarren:
oh no, oops! So, we actually have to add a reversed migration, rather than
rolling back, now that it's in the commit history. Can you open a PR with
undoing the previous change and adding the new one?

SidharthBansal:
Hi @jywarren I have added the migration for removal of password_checker field in this pr. Can you please review it?

SidharthBansal:
I think we are getting octokit error again. Maybe it is due to change of key as it happened earlier

db/migrate/20180731124028_add_password_checker_to_rusers.rb
```diff
@@ -0,0 +1,5 @@
+class AddPasswordCheckerToRusers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :rusers, :password_checker, :integer, default: 0
```
jywarren: These could actually be in the same migration!

app/controllers/user_sessions_controller.rb
```diff
@@ -66,75 +66,82 @@ def create
...
-                elsif session[:return_to]
-                  return_to = session[:return_to]
-                  session[:return_to] = nil
-                  redirect_to return_to
-                elsif params[:return_to]
-                  redirect_to params[:return_to]
+      u = User.find_by(username: username)
+      if u && u.password_checker != 0
+        n = u.password_checker
+        hash = { 1 => "Facebook", 2 => "Github", 3 => "Google", 4 => "Twitter"  }
```
jywarren: Is it possible to use the style `1: "Facebook"` here?

jywarren:
Also where could we helpfully explain the usage of password checker?

db/migrate/20180731124028_add_password_checker_to_rusers.rb
SidharthBansal: Sorry, I could not understand what I need to change.

app/controllers/user_sessions_controller.rb
SidharthBansal: Pushing the changes.
Thanks

SidharthBansal:
I will send the documentation in a minute

db/migrate/20180731124028_add_password_checker_to_rusers.rb
SidharthBansal: There are two migrations. First is to delete password_checker from Users.
Second is to add password_checker to RUsers.

app/controllers/user_sessions_controller.rb
SidharthBansal: I can't make integers as symbols as at another place I am using the reverse hash. In reverse hash I am mapping providers with integer values.

db/migrate/20180731124028_add_password_checker_to_rusers.rb
jywarren: Oh just that you can perform both steps in the same migration file. They needn't be separated... Although it's not a big deal, over the lifetime of the project I think we should try not to have too many migration files. Mind changing this? Thanks!

SidharthBansal: I will change it

SidharthBansal: Thanks for explanation

SidharthBansal: Can you please add link to the code for example? I tried a lot searching on net and on plots2 migration files. We can have different dbms queries on same table in single migration. But I could not find a migration with two different tables involved. In this pr, we are involved with two tables Rusers(removal of password checker) & Users(addition of password checker).

jywarren: I believe it'll work to do:

```ruby
def change
  remove_column :users, :password_checker, :integer
  add_column :rusers, :password_checker, :integer, default: 0
end
```

I don't have an example but have you tried it? I think the table name is just specified in the one line, no? But honestly it's not a big deal, just a slight preference, so if you find that it doesn't work, 2 files is fine!

SidharthBansal:
@jywarren can you please review this?
Thanks

db/migrate/20180731124028_add_password_checker_to_rusers.rb
SidharthBansal: Thanks a lot. I was thinking other type of solution which was giving errors.
Thanks it worked.

app/controllers/user_sessions_controller.rb
jywarren: ok! thanks for trying!

jywarren:
This looks good - are you confident that we should merge? Thanks!

SidharthBansal:
Yes
Thanks

PR #3161, opened 2018-07-26
Emoji encoding 3123

cheneyshreve:
[WIP] Please do not merge yet.

HI @jywarren, to follow up on the atwho edits what I've done so far in this PR:
- move the atwho code out of comments form and into /app/assets/javascripts/emoji.js
- required emoji.js in /app/assets/javascripts/application.js
- added emoji.js to config/initializers/assets.rb
- imported emoji.js as a script into
-  /comments/_form.html.erb
-  /editor/rich.html.erb
This has the expected behavior, e.g. makes atwho autocompletes available. 

A couple questions I'm still working on:
- had trouble reading emoji file into /assets/javascripts. Right now, var emoji is defined within that file, which works fine, but I was trying to refactor it to get it to read in within the remoteFilter, but no success yet. 
- I haven't been able to get the atwho functionality to work yet when toggling to the Markdown editor, e.g. if I got to localhosts/post and am in Rich text editor, works fine, but not sure where I need to edit codebase to get this to work for Markdown? What specific file(s) do I need to look at?

Thanks for your feedback and the opportunity to keep working on this!


jywarren:
Restarting this build -- if it gets stuck with the same error, you can close and reopen to try to restart it - we're trying to figure out that bug but it's unrelated to your work, sorry!


The unrelated bug looks like: `There was an error with Danger bot’s Junit parsing: undefined method 'value' for nil:NilClass`

jywarren:
`to read in within the remoteFilter, but no success yet.` -- can you say more about this?

Maybe we should do the rich editor in a separate PR - and even on the PublicLab.Editor repository, so it's available "upstream"? Or we can implement it here first and then port it over once it's working. 

But I think that we may need to trigger something at the moment the editor mode changes from markdown to rich text... so that may mean we need to do it inside the editor codebase, where we have access to those events. Make sense? If this is the case, we can put off that part of the integration until the code separation one is complete. How does that sound? 

Great progress here!

cheneyshreve:
Hi @jywarren , regarding more details about remoteFilter, what I meant was trying to refactor atwho emoji config using remoteFilter versus using the `data` keyword within the atwho config (something like below). Ultimately, I don't think it will let us divide tasks up that way;  I think if you use the remote filter, you can't also use the displayTpl. Or perhaps I'm making an error in the callback. I can also ask atwho moderators to clarify regarding using remoteFilter with displayTpl.

      emojis_config = {
        at: ':',
        callbacks: {
        #but this doesn't work as hoped...had also tried /public/lib/emoji.js and different variations of the callback
          remoteFilter: function(query, callback) {
            $.getJSON('/lib/emoji.js', function(emoji) {
               callback(Object.keys(emoji).map(function(name){ return {'name': name, 'value': emoji[name]}}));
             });
          }
        },
        displayTpl: "<li>${value} ${name}</li>",
        insertTpl: ":${name}:",
        limit: 100
     }
      
Since these emojis are "static" in that they don't get updated very frequently, perhaps it's OK to just define them as var emojis (current code) and not worry about reading them in through remoteFilter from /public/lib directory? 

If so, I could:
- delete the emoji.json from /public/lib
- remove current script from rich.html.erb, since it only works on the rich text editor at this point and not the markdown,
- Resubmit this PR. 

We'd have this PR being just moving the atwho calls out of comments/_form.html.erb into a script and adding that script to comments/_form.html.erb. Then open a new issue for figuring out the issue regarding switching between rich/markdown in the editor. I think you are right that it might have to be done in the PublicLab.Editor, and I'd love to get some experience with that too. I'm happy to keep trying things out to see if it can be done without going into PublicLab.editor too, so just let me know what you prefer. Thanks again!

jywarren:
> Since these emojis are "static" in that they don't get updated very frequently, perhaps it's OK to just define them as var emojis (current code) and not worry about reading them in through remoteFilter from /public/lib directory?

Actually, what I ended up doing was similar to what you propose -- see, I included `emoji.js` as a local variable here:

https://github.com/publiclab/plots2/blob/master/app/views/comments/_form.html.erb#L67

And so they're in `/public/lib/emoji.js` -- the hope was to get those to cache in the browser, and skip remoteFilter. So I think we're OK on that count. 

I think we're OK to:

> remove current script from rich.html.erb

But yes, as you say, if you want to copy that code over to https://github.com/publiclab/PublicLab.Editor/, I can help you step-by-step integrate it over there, so we can work on that two-mode system. I think you could start by integrating it into the example file at https://github.com/publiclab/PublicLab.Editor/blob/master/examples/index.html, or copy that and make one called `autocompletion.html` (to keep the default example simple). What do you think?

> We'd have this PR being just moving the atwho calls out of comments/_form.html.erb into a script and adding that script to comments/_form.html.erb. 

Perfect! One function per PR is easier and better, anyways. 

Awesome!

cheneyshreve:
Hi @jywarren, awesome, I just removed the script from rich text editor and emoji file from public and I'll push up those changes. I assume the new issue should be opened in PublicLab.Editor? I'll go ahead and download that repo and start look at the file you mention above. Thanks! very excited to work on this!

cheneyshreve:
Thanks @jywarren, I will open issue on PublicLab.Editor regarding package-lock issue.

config/database.yml.sqlite
```diff
@@ -0,0 +1,11 @@
+development:
```
jywarren: Hi, this is looking great, but would you mind removing the this file from your PR, as it's unrelated to your changes?

https://stackoverflow.com/questions/215718/reset-or-revert-a-specific-file-to-a-specific-revision-using-git

Thanks!

app/assets/javascripts/application.js
```diff
@@ -58,3 +58,4 @@
 //= require noty_notification.js
 //= require jquery-confirm/js/jquery-confirm.js
 //= require jquery.atwho
+//= require emoji.js
```
jywarren: I think I'd prefer not to include the entire emoji.js file here, and to leave it in the `/lib/` folder. These files are compiled together, and I think we can treat the emoji file as a static asset, so the system doesn't have to process it (since it's so big!). But we should include `comment-autocomplete.js` here -- i actually don't see that one in the PR, did you forget to add it?

cheneyshreve: Hi @jywarren I think that I did a poor job naming the emoji script, in that it is named the same as the file that held the emoji json. In this version of the code, emoji.js is the script living in app/assets/javascripts/emoji.js that includes within it the defined emojis (not being read from public, just defined in a var), and deleted the file from public/lib/emoji.js since it would not be needed. I can remove the require statement from application.js.

 I'm not sure about the comment-autocomplete.js? I had not done anything with that, when I  add it in to application.js I get the error ```couldn't find file 'comment-autocomplete.js' with type 'application/javascript'``` so perhaps I need to fetch more recent changes and do a rebase? I believe I have the config/database.yml.sqlite removed properly now so once I clear up the details above, I'll resubmit.

Thanks again for all of your help! And the callout in the community checkin, it is really wonderful to have the opportunity to work with you on this :)

jywarren:
ah, ok -- so if comment-autocomplete.js doesn't exist, no worries, we can
just remove references to it.

As to the combined emoji.js - i think using `var` and a separate file may
be preferable here, because it's hard to review the code in GitHub when
it's a giant file... it doesn't like to even render the file contents in
the webpage view, you know? If you don't mind, would it be OK to just
include the individual emoji list file separately? That will also prevent
it from being "compiled" into the giant application.js file, which will be
easier on the server.

Thanks!

cheneyshreve:
Hi @jywarren, I'm pushing up revisions per comments above. I renamed emoji script to be atwho_autocomplete.js so that is more clear, added emoji (var) back to lib (and out of atwho_autocomplete), adjusted comments form to read in /lib/emoji and the atwho_autocomplete script, removed reference to comment-autocomplete in assets file. Hoping the git workflow is correct; combined 2 commits together because I'd forgotten to remove reference to comment-autocomplete initially. I did not merge the other commit that was for reverting an older commit to get rid of that config/database.yml.sqlite file because that references on that one commit and not later commits, if that makes sense. Thanks, and let me know if you need further edits.

app/views/comments/_form.html.erb
```diff
@@ -65,50 +65,13 @@
...
-        at: ':',
-        data: Object.keys(emoji).map(function(name){ return {'name': name, 'value': emoji[name]} }),
-        displayTpl: "<li>${value} ${name}</li>",
-        insertTpl: ":${name}:", 
-        limit: 100
-     }
-
-    $('textarea#text-input').atwho(at_config).atwho(hashtags_config).atwho(emojis_config);
-
     $('#comment-form').bind('ajax:success', function(e, data, status, xhr){
```
jywarren: Can you confirm that the error message here still works without the `(function() {` around it? Maybe best to add that back in, as it helps the whole thing run /after/ page load is complete?

jywarren:
Awesome, this is super. Just that one note about comment erroring and we should be good to go!

app/views/comments/_form.html.erb
jywarren: I think adding it back in may be easier than testing it... i know you can get it to error either by turning off your locally running app OR by submitting a real unicode emoji (not just a `:smiley:`) -- since we're trying to fix that! But... could be a bit tricky to test.

cheneyshreve: Hi @jywarren, I moved the code for error message back into the function athwho_autocomplete script! Quick question before I push this code --- so the public folder is in .gitignore-- how do we then make sure we're including /lib/emoji.js? I just realized it wasn't getting included when I made that code change above then freaked out momentarily because "emoji " wasn't defined then realized it was because the file is in .gitignore. Thanks again for all of your help!

jywarren:
Ah, oops, sorry, my bad. I actually think we should keep the `(function() { ... })()` in the `form.html.erb` template with the error code inside it -- it was just that I worried that taking out the `(function() { .. })()` from around it would cause it to fail.

Basically what the `function` code does is it creates a "singleton" or one-off anonymous function around code, then wraps it in a `()` and then by adding a `()` at the end, runs the newly created anonymous function. But, by having done all this, it essentially delays the running of the code until everything else on the page is done. Does that make sense? It's a little obscure, but I think putting the error code back on `_form.html.erb` and within an anonymous function like that should get us there.

cheneyshreve:
Thanks, @jywarren I was not aware of that but it is cool! Made changes and pushing them up now.

public/lib/emoji.js
```diff
@@ -1 +0,0 @@
..."puerto_rico":"🇵🇷","qatar":"🇶🇦","reunion":"🇷🇪","romania":"🇷🇴","ru":"🇷🇺","rwanda":"🇷🇼","st_barthelemy":"🇧🇱","st_helena":"🇸🇭","st_kitts_nevis":"🇰🇳","st_lucia":"🇱🇨","st_pierre_miquelon":"🇵🇲","st_vincent_grenadines":"🇻🇨","samoa":"🇼🇸","san_marino":"🇸🇲","sao_tome_principe":"🇸🇹","saudi_arabia":"🇸🇦","senegal":"🇸🇳","serbia":"🇷🇸","seychelles":"🇸🇨","sierra_leone":"🇸🇱","singapore":"🇸🇬","sint_maarten":"🇸🇽","slovakia":"🇸🇰","slovenia":"🇸🇮","solomon_islands":"🇸🇧","somalia":"🇸🇴","south_africa":"🇿🇦","south_georgia_south_sandwich_islands":"🇬🇸","kr":"🇰🇷","south_sudan":"🇸🇸","es":"🇪🇸","sri_lanka":"🇱🇰","sudan":"🇸🇩","suriname":"🇸🇷","swaziland":"🇸🇿","sweden":"🇸🇪","switzerland":"🇨🇭","syria":"🇸🇾","taiwan":"🇹🇼","tajikistan":"🇹🇯","tanzania":"🇹🇿","thailand":"🇹🇭","timor_leste":"🇹🇱","togo":"🇹🇬","tokelau":"🇹🇰","tonga":"🇹🇴","trinidad_tobago":"🇹🇹","tunisia":"🇹🇳","tr":"🇹🇷","turkmenistan":"🇹🇲","turks_caicos_islands":"🇹🇨","tuvalu":"🇹🇻","uganda":"🇺🇬","ukraine":"🇺🇦","united_arab_emirates":"🇦🇪","gb":"🇬🇧","us":"🇺🇸","us_virgin_islands":"🇻🇮","uruguay":"🇺🇾","uzbekistan":"🇺🇿","vanuatu":"🇻🇺","vatican_city":"🇻🇦","venezuela":"🇻🇪","vietnam":"🇻🇳","wallis_futuna":"🇼🇫","western_sahara":"🇪🇭","yemen":"🇾🇪","zambia":"🇿🇲","zimbabwe":"🇿🇼"}
```
jywarren: Here, I think we just need to leave this file as it was, instead of deleting. 

https://stackoverflow.com/questions/215718/reset-or-revert-a-specific-file-to-a-specific-revision-using-git

Thanks, and this time it really is it! Did this work for you when testing?

cheneyshreve: Hi @jywarren, yes on the testing. I don't know what is going on with the lib/emoji.js file--I had removed it in an earlier commit, but added it back in to more recent one. It is present in my local data. Is it because that folder is in .gitignore that it doesn't pick up the (re)add? I can go back through the commits and try to amend that specific file per the link above, guess that is how I have to (re)add it.

jywarren:
Hmm, yeah - you can check what happens by looking at the Files tab in the
PR, which is how I'm looking at it. Ideally if you can revert that one
file, you'll see that emoji.js is not listed in the PR (meaning it doesn't
have any proposed changes in this PR). If you get stuck, i can give it a
try?

cheneyshreve:
Hi @jywarren I just ran ```git checkout b0377da5f678be6205290dda854a25837bf8ee55~1 -- public/lib/emoji.js``` and that seems to have done the trick by letting me add back public/lib/emoji.js. I just tested it again to make sure it's working. I can push up those changes now if you'd like. I think that's what was needed.

jywarren:
Yes, try pushing that up!

cheneyshreve:
Ok, just pushed that up!

cheneyshreve:
Hi @jywarren  yes, it works for me locally! Just tested it again.

jywarren:
Super. Merging now, thanks for sticking with it! Next, i guess, over to the Editor project?

cheneyshreve:
Thanks, @jywarren feeling very fortunate to be working with you on this! I'm learning a ton, especially on git with this particular PR so I can't thank you enough! Looking forward to working on the PublicLab.Editor!

AngelicaOvesdotter (commented):
Nice

PR #3140, opened 2018-07-24
Email settings documentation

gauravano:
Fixes #3141 

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [x] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [x] code is in uniquely-named feature branch and has no merge conflicts
* [x] PR is descriptively titled
* [x] PR body includes `fixes #0000`-style reference to original issue #
* [x] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software

We have a loose schedule of reviewing and pulling in changes every Tuesday and Friday, and publishing changes on Fridays. 

Thanks!

PR #3114, opened 2018-07-21
Add some docs info

stefannibrasil:
Fixes #3115 
@publiclab/reviewers could you give me more tips to add to those docs?

When @milaaraujo and I were reading and studying the code a few weeks ago, we found some space for improvement. Would appreciate if you could give more tips here. For example, we don't know exactly how to explain more about the stable and unstable branches and how to use them.

Thanks!

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [ ✔️ ] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [ ✔️ ] code is in uniquely-named feature branch and has no merge conflicts
* [  ✔️] PR is descriptively titled
* [ ✔️ ] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software

We have a loose schedule of reviewing and pulling in changes every Tuesday and Friday, and publishing changes on Fridays. 

Thanks!

PR #3163, opened 2018-07-28
Link addition to email for marking comment as spam

### LINKED ISSUE #3137: Add link to approve/spam comments in comment email notifications to moderators
When I see a new comment in my email inbox, it'd be great to be able to spam + ban directly from there, if possible! 

But only moderators should receive this type of email. In any case, it could be at the bottom of the message, and if, in replying, the link gets posted to the site, it should be OK since others won't be able to use the function. @Gauravano -- any interest in this one? Should be pretty simple!

gauravano:
Fixes #3137

Thanks!

### DIFF
--- a/app/views/comment_mailer/notify.html.erb
+++ b/app/views/comment_mailer/notify.html.erb
@@ -14,6 +14,13 @@ Hi! There's been a response to a discussion you're involved in. Do NOT reply to 
 
 <hr />
 
+<p>Look like spam?
+  <% if @user.can_moderate? %>
+    Mark this as <a href="https://<%= ActionMailer::Base.default_url_options[:host] %>/admin/mark_comment_spam/<%= @comment.id %>">Spam</a>
+  <% else %>
+    <a href="mailto:moderators@publiclab.org?subject=Reporting+spam+on+Public+Lab&body=Hi,+I+found+this+comment+that+looks+like+spam+or+needs+to+be+moderated:+https://publiclab.org/<%= @comment.parent.path %>#c<%= @comment.cid %>+by+https://publiclab.org/profile/<% if @comment.author %><%= @comment.author.name %><% end %>+Thanks!" title="Flag as spam" class="btn btn-sm btn-default btn-flag-spam-<%= @comment.id %>" >Flag this</a> for moderators
+  <% end%>
+</p>
 <div style="color:#aaa;">
 
 <% if @comment.parent.has_power_tag('question') %>


Gauravano:
Hi @jywarren @publiclab/soc anyone facing this issue 

octokit

Thanks!

stefannibrasil:
we are having the same error on #3114 and #3134

Gauravano:
yes @stefannibrasil, my previous PR #3140 also failed. Let's wait for @jywarren and @icarito .

Gauravano:
@jywarren please review this PR. Thanks!

app/views/comment_mailer/notify.html.erb
```diff
@@ -14,6 +14,13 @@ Hi! There's been a response to a discussion you're involved in. Do NOT reply to
 
 <hr />
 
+<p>Look like spam?
+  <% if @user.can_moderate? %>
```
jywarren: So here's what I'm worried about, are we notifying in a bulk BCC, or do we actually have unique user record each time it's sent? 

I can't find where this is used to see what the `@user` variable is, but let's check!

https://github.com/publiclab/plots2/search?q=CommentMailer.notify&unscoped_q=CommentMailer.notify

gauravano: Each time a new comment is created this `CommentMailer.notify`  send emails to users subscribed to the tags or node. So, I have added these buttons only on normal notification emails.
Here's the effective mailer https://github.com/publiclab/plots2/blob/master/app/mailers/comment_mailer.rb#L7-L12

gauravano: @jywarren

jywarren: yeah, that's great! I just want to know where `user` is specified -- like, do you know where this `.notify()` is called from, so we can be sure it represents the recipient user instead of the comment author? I think we may want to rename the parameter a little bit to be sure of it, since `user` is ambiguous, you know?

gauravano: here it is @jywarren 
https://github.com/publiclab/plots2/blob/master/app/models/comment.rb#L124-L130

jywarren:
perfect! OK, do you want to change the parameter name to `recipient` or do
you think we're OK here? Thanks!

gauravano:
@jywarren I can change the name but the whole comment_mailer.rb file is using @user. So, should I change it to recipient in notify_method only or everywhere. Thanks!

jywarren:
ok, well we can also do that in a follow-up PR. OK, let's see how this does!

PR #3172, opened 2018-08-01
initial debounced typeahead with debug statements

### LINKED ISSUE #3472: Explore request timing optimizations for typeahead search and username/tag autocompletion
Our nice typeahead search is SUPER! (Thanks @milaaraujo @stefannibrasil !!)

I think we could speed it up a bit and also have it be a little less jittery by using a throttling strategy.

See how when you type the word "balloon" it makes 5 different queries? 

screen shot 2018-09-25 at 2 30 33 pm

We're using Bootstrap Typeahead: https://github.com/bassjobsen/Bootstrap-3-Typeahead to run this code. 

For starters, we can set the `minLength` and `delay` to 3 and 1, respectively, so it only starts searching after 3 characters are input, and only sends one request per second. This should slow things down a good bit (but test it out locally to see if it works for you!)

After merging those changes, we could explore more sophisticated "debounce" strategy -- https://davidwalsh.name/javascript-debounce-function

jywarren:
I haven't been able to get this to work; the debounced function is never sent! @ryzokuken perhaps you understand a bit better here? 

https://lodash.com/docs/4.17.10#debounce

Fixes #3472

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -15,6 +15,7 @@
 //
 //= require jquery
 //= require jquery_ujs
+//= require lodash/dist/lodash.min.js
 //= require bootstrap/dist/js/bootstrap.min.js
 //= require bootstrap-typeahead/bootstrap3-typeahead.min.js
 //= require bootstrap-datepicker/dist/js/bootstrap-datepicker.js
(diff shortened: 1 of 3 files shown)


gauravano:
hey @jywarren, closing this PR as we have added debounce in https://github.com/publiclab/plots2/pull/4904/. If you think, this PR can be taken forward, feel free to reopen. Thanks!

jywarren:
super, thank you!

PR #3199, opened 2018-08-04
update semver marks on leaflet-environmental-layers and leaflet-blurred-location

jywarren:
@mridulnagpal @sagarpreet-chadha https://semver.org and now that I'm looking at these docs:

https://github.com/npm/node-semver#x-ranges-12x-1x-12-

maybe we should use x-ranges to be more clear?

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -24,12 +24,12 @@
     "megamark": "~3.3.0",
     "i18n-js": "^2.1.2",
     "short-code-forms": "jywarren/short-code-forms#~0.0.1",
-    "leaflet-blurred-location": "publiclab/leaflet-blurred-location#master",
+    "leaflet-blurred-location": "publiclab/leaflet-blurred-location#^1.1.14",
     "chart.js": "v2.7.0",
     "typeahead.js-browserify": "Javier-Rotelli/typeahead.js-browserify#~1.0.7",
     "noty": "3.1.4",
     "jquery-confirm": "craftpip/jquery-confirm#3.3.2" , 
-    "leaflet-environmental-layers": "1.0.1" ,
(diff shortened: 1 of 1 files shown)


jywarren:
Technically since both these libraries are past `1.0.0` we should be fine to use carets `^`! that should allow `0.x.0` or `0.0.x` (minor or patch) updates.

jywarren:
Hi @icarito the gitlab tests seem to have failed with:

```

Service container logs:
2018-08-04T15:59:24.253834031Z ln: failed to create symbolic link '/sys/fs/cgroup/systemd/name=systemd': Operation not permitted

*********

Pulling docker image gitlab/dind ...
Using docker image sha256:cc674e878f23bdc3c36cc37596d31adaa23bca0fc3ed18bea9b59abc638602e1 for gitlab/dind ...
ERROR: Job failed (system failure): Error response from daemon: Cannot link to a non running container: /runner-e11ae361-project-7706596-concurrent-0-gitlab__dind-0 AS /runner-e11ae361-project-7706596-concurrent-0-predefined-0/gitlab__dind (executor_docker.go:1014:0s)
```

I haven't seen that before, any ideas?

jywarren:
Also, gosh, this is taking a while to run...

icarito:
No idea! I think we may still improve the Gitlab config by trying to use
the docker socket directly instead of Docker-in-Docker we are using
right now.

https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/

But no idea if this will fix this issue. Rebuilding worked :-/ so we
have another don't-know-how-to-reproduce bug here :-/

I did see this error once or twice before when setting up Gitlab CI build.

PR #3204, opened 2018-08-05
feature-power tag for notes grid with specific tagname

### LINKED ISSUE #1097: develop `notes:tagname` inline tag alternative to display a grid of notes, not a table
This is not very high priority... but:

Currently, on a wiki or note, we can insert a table of topical content using the syntax:

`[notes:tagname]` -- like, for example, for content tagged "balloon-mapping", we can do:

`[notes:balloon-mapping]` and that's swapped for a table of balloon mapping posts. 

But some content is very visual and it'd be great to see a grid of thumbnails, much like we show on tag pages like https://publiclab.org/tag/balloon-mapping

Let's make a new inline tag, similar to `[notes:tagname]` but call it `[notes:grid:tagname]` (which would have to be **above** the existing `[notes:tagname]` check, so that it doesn't try to search for content tagged `grid:tagname`, you know?). 

The current inline grids are listed here:

https://github.com/publiclab/plots2/blob/b98790631fe22edf34750de5bff5c641ff6ea7b7/app/helpers/application_helper.rb#L31

And the code for each lives in this document:

https://github.com/publiclab/plots2/blob/master/app/models/concerns/node_shared.rb#L31-L67

We could have a new section just above that, that uses this matcher:

```ruby
    body = body.gsub(/\<p\>\[notes\:(.+)\]/) do |tagname|
```

The template code could be laid out like what we use for the tag pages: 

https://github.com/publiclab/plots2/blob/master/app/views/notes/_notes.html.erb

We can actually just render a partial in a helper, as in this line: 

https://github.com/publiclab/plots2/blob/b98790631fe22edf34750de5bff5c641ff6ea7b7/app/models/concerns/node_shared.rb#L56

This is a little complex, but worth a try -- i'm happy to offer help. It could also be broken up a bit into smaller projects, for example just swapping in a holding string like `Inline thumbnail grid coming soon.` as the first step.

### Mockups

screenshot 2016-11-14 at 6 17 27 pm

ViditChitkara:
closes #1097

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -79,6 +79,7 @@ module ApplicationHelper
   end
 
   def insert_extras(body)
+    body = NodeShared.notes_inline_grid(body)
     body = NodeShared.notes_grid(body)
     body = NodeShared.questions_grid(body)
     body = NodeShared.activities_grid(body)
--- a/app/models/concerns/node_shared.rb
+++ b/app/models/concerns/node_shared.rb
@@ -186,6 +186,42 @@ module NodeShared
     end
   end
 
+  def self.notes_inline_grid(body)
+    body.gsub(/[^\>`](\<p\>)?\[notes\:grid\:(\S+)\]/) do |_tagname|
+      tagname = Regexp.last_match(2)
+      exclude = nil
+      if tagname.include?('!')
+        exclude = tagname.split('!') - [tagname.split('!').first]
+        tagname = tagname.split('!').first
+      end
+
+      nodes = Node.where(status: 1, type: 'note')
+                  .includes(:revision, :tag)
+                  .references(:term_data, :node_revisions)
+                  .where('term_data.name = ?', tagname)
+                  .order('node_revisions.timestamp DESC')
+
+      if exclude.present?
+        exclude = Node.where(status: 1, type: 'note')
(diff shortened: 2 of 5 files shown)


app/helpers/application_helper.rb
```diff
@@ -79,6 +79,7 @@ def locale_name_pairs
   end
 
   def insert_extras(body)
+    body = NodeShared.notes_inline_grid(body)
```
jywarren: Wow, this looks great... but I think there should be a lot of overlap with the `notes_grid` method... do you think we could consolidate those two? In theory it should be relatively close ActiveRecord queries, but displayed differently, don't you think?

ViditChitkara: Yes, makes sense!! I see the methods `notes_grid` and `notes_inline_grid` differ only in rendering the templates. Both render different templates having same active-record queries. So, let's keep a single function with a parameter (say thumbnails=false by default) and change the template accordingly?

jywarren:
that sounds perfect!

gauravano:
@ViditChitkara want to give this a push? Thanks!

gauravano:
Actually, the issue is solved in https://github.com/publiclab/plots2/pull/3970/. So, closing this PR. Thanks!

PR #3217, opened 2018-08-08
Improve peopleLocations endpoint: people_locations method

### LINKED ISSUE #3196: Improve peopleLocations endpoint
Part of #2755

milaaraujo:
Hey everybody!

We are opening this request to fix the issue #3196. Here we implemented a new version of `recentPeople` method. 

Now we first select the users that have lat/lon and then we join the users table with the revisions table to order them by most recent activities. We also change the name of the method from `recentPeople` to `people_locations`.

We would like some feedback, thanks! :)

fixes #3196

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -170,29 +170,25 @@ class SearchService
   def recentPeople(srchString, tagName = nil)
     sresult = DocList.new
 
-    user_scope = User.where('status <> 0')
-                          .reorder('')
+    user_scope = User.where('rusers.status <> 0')\
+                     .joins(:user_tags)\
+                     .where('value LIKE "lat:%"')\
+                     .joins(:revisions)\
+                     .order("node_revisions.timestamp DESC")
 
-    user_scope = user_scope.joins(:revisions)
-                           .order("node_revisions.timestamp DESC")
-                           .distinct
+    if tagName.present?
+      user_scope = User.joins(:user_tags)\
+                       .where('user_tags.value LIKE ?', tagName)\
+                       .where(id: user_scope.select("rusers.id"))
+    end
 
-    users = []
+    user_scope = user_scope.limit(srchString.to_i)
+
     user_scope.each do |user|
-      next unless user.has_power_tag("lat") && user.has_power_tag("lon")
+      blurred = user.has_power_tag("location") ? user.get_value_of_power_tag("location") : false
 
-      blurred = false
-      if user.has_power_tag("location")
-        blurred = user.get_value_of_power_tag("location")
-      end
-
-      if tagName.blank?
-        doc = DocResult.fromLocationSearch(user.id, 'people_coordinates', user.path, user.username, 0, 0, user.lat, user.lon, blurred)
-        sresult.addDoc(doc)
-      elsif !(tagName.blank?) && user.has_tag(tagName)
-          doc = DocResult.fromLocationSearch(user.id, 'people_coordinates', user.path, user.username, 0, 0, user.lat, user.lon, blurred)
-          sresult.addDoc(doc)
-      end
+      doc = DocResult.fromLocationSearch(user.id, 'people_coordinates', user.path, user.username, 0, 0, user.lat, user.lon, blurred)
+      sresult.addDoc(doc)
     end
 
     sresult
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -173,31 +173,7 @@ class SearchService
   def people_locations(srchString, tagName = nil)
     sresult = DocList.new
 
-    user_scope =
-    if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
-         User.select('rusers.*, node_revisions.timestamp')
-             .where('rusers.status <> 0')\
-             .joins(:user_tags)\
-             .where('value LIKE "lat:%"')\
-             .joins(:revisions)\
-             .order("node_revisions.timestamp DESC")\
-             .distinct
-      else
-        User.where('rusers.status <> 0')\
-            .joins(:user_tags)\
-            .where('value LIKE "lat:%"')\
-            .joins(:revisions)\
-            .order("node_revisions.timestamp DESC")\
-            .distinct
-      end
-
-    if tagName.present?
-      user_scope = User.joins(:user_tags)\
-                       .where('user_tags.value LIKE ?', tagName)\
-                       .where(id: user_scope.select("rusers.id"))
-    end
-
-    user_scope = user_scope.limit(srchString.to_i)
+    user_scope = SrchScope.find_locations(srchString, tagName)
 
     user_scope.each do |user|
       blurred = user.has_power_tag("location") ? user.get_value_of_power_tag("location") : false
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -167,7 +167,7 @@ class SearchService
 
   # GET X number of latest people/contributors and package up as a DocResult
   # X = srchString
-  def recentPeople(srchString, tagName = nil)
+  def people_locations(srchString, tagName = nil)
     sresult = DocList.new
 
     user_scope = User.where('rusers.status <> 0')\
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -170,26 +170,29 @@ class SearchService
   def recentPeople(srchString, tagName = nil)
     sresult = DocList.new
 
-    nodes = Node.all.order("changed DESC").limit(srchString).distinct
+    user_scope = User.where('status <> 0')
+                          .reorder('')
+
+    user_scope = user_scope.joins(:revisions)
+                           .order("node_revisions.timestamp DESC")
+                           .distinct
+
     users = []
-    nodes.each do |node|
-      if node.author.status != 0
-        if tagName.blank?
-          users << node.author.user
-        else
-          users << node.author.user if node.author.user.has_tag(tagName)
-        end
-      end
-    end
-    users = users.uniq
-    users.each do |user|
+    user_scope.each do |user|
       next unless user.has_power_tag("lat") && user.has_power_tag("lon")
+
       blurred = false
       if user.has_power_tag("location")
         blurred = user.get_value_of_power_tag("location")
       end
-      doc = DocResult.fromLocationSearch(user.id, 'people_coordinates', user.path, user.username, 0, 0, user.lat, user.lon, blurred)
-      sresult.addDoc(doc)
+
+      if tagName.blank?
+        doc = DocResult.fromLocationSearch(user.id, 'people_coordinates', user.path, user.username, 0, 0, user.lat, user.lon, blurred)
+        sresult.addDoc(doc)
+      elsif !(tagName.blank?) && user.has_tag(tagName)
+          doc = DocResult.fromLocationSearch(user.id, 'people_coordinates', user.path, user.username, 0, 0, user.lat, user.lon, blurred)
+          sresult.addDoc(doc)
+      end
     end
 
     sresult
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -173,12 +173,23 @@ class SearchService
   def people_locations(srchString, tagName = nil)
     sresult = DocList.new
 
-    user_scope = User.where('rusers.status <> 0')\
-                     .joins(:user_tags)\
-                     .where('value LIKE "lat:%"')\
-                     .joins(:revisions)\
-                     .order("node_revisions.timestamp DESC")\
-                     .distinct
+    user_scope =
+    if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
+         User.select('rusers.*, node_revisions.timestamp')
+             .where('rusers.status <> 0')\
+             .joins(:user_tags)\
+             .where('value LIKE "lat:%"')\
+             .joins(:revisions)\
+             .order("node_revisions.timestamp DESC")\
+             .distinct
+      else
+        User.where('rusers.status <> 0')\
+            .joins(:user_tags)\
+            .where('value LIKE "lat:%"')\
+            .joins(:revisions)\
+            .order("node_revisions.timestamp DESC")\
+            .distinct
+      end
 
     if tagName.present?
       user_scope = User.joins(:user_tags)\
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -165,8 +165,11 @@ class SearchService
     sresult
   end
 
-  # GET X number of latest people/contributors and package up as a DocResult
-  # X = srchString
+  # Returns the location of people with most recent contributions.
+  # The method receives as parameter the number of results to be
+  # returned and as optional parameter a user tag. If the user tag
+  # is present, the method returns only the location of people
+  # with that specific user tag.
   def people_locations(srchString, tagName = nil)
     sresult = DocList.new
 
@@ -174,7 +177,8 @@ class SearchService
                      .joins(:user_tags)\
                      .where('value LIKE "lat:%"')\
                      .joins(:revisions)\
-                     .order("node_revisions.timestamp DESC")
+                     .order("node_revisions.timestamp DESC")\
+                     .distinct
 
     if tagName.present?
       user_scope = User.joins(:user_tags)\
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -33,6 +33,7 @@ class User < ActiveRecord::Base
   has_many :following_users, through: :active_relationships, source: :followed
   has_many :followers, through: :passive_relationships, source: :follower
   has_many :likes
+  has_many :revisions, through: :node
 
   validates_with UniqueUsernameValidator, on: :create
   validates_format_of :username, with: /\A[A-Za-z\d_\-]+\z/
--- a/app/services/execute_search.rb
+++ b/app/services/execute_search.rb
@@ -20,7 +20,7 @@ class ExecuteSearch
      when :tags
        sresult = sservice.textSearch_tags(search_criteria.query)
      when :peoplelocations
-       sresult = sservice.recentPeople(search_criteria.query, search_criteria.tag)
+       sresult = sservice.people_locations(search_criteria.query, search_criteria.tag)
      when :taglocations
        if search_criteria.query.include? ","
          sresult = sservice.tagNearbyNodes(search_criteria.query, search_criteria.tag)
--- a/app/services/srch_scope.rb
+++ b/app/services/srch_scope.rb
@@ -69,7 +69,7 @@ class SrchScope
     user_locations = User.where('rusers.status <> 0')\
                          .joins(:user_tags)\
                          .where('value LIKE "lat:%"')\
-                         .joins(:revisions)\
+                         .includes(:revisions)\
                          .order("node_revisions.timestamp DESC")\
                          .distinct
     if user_tag.present?
--- a/app/services/srch_scope.rb
+++ b/app/services/srch_scope.rb
@@ -66,23 +66,12 @@ class SrchScope
   end
 
   def self.find_locations(limit, user_tag = nil)
-    user_locations =
-      if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
-        User.select('rusers.*,node_revisions.timestamp')
-            .where('rusers.status <> 0')\
-            .joins(:user_tags)\
-            .where('value LIKE "lat:%"')\
-            .joins(:revisions)\
-            .order("node_revisions.timestamp DESC")\
-            .distinct
-      else
-        User.where('rusers.status <> 0')\
-            .joins(:user_tags)\
-            .where('value LIKE "lat:%"')\
-            .joins(:revisions)\
-            .order("node_revisions.timestamp DESC")\
-            .distinct
-      end
+    user_locations = User.where('rusers.status <> 0')\
+                         .joins(:user_tags)\
+                         .where('value LIKE "lat:%"')\
+                         .joins(:revisions)\
+                         .order("node_revisions.timestamp DESC")\
+                         .distinct
     if user_tag.present?
       user_locations = User.joins(:user_tags)\
                        .where('user_tags.value LIKE ?', user_tag)\
--- a/app/services/srch_scope.rb
+++ b/app/services/srch_scope.rb
@@ -64,4 +64,34 @@ class SrchScope
       .joins(:tag)
       .where('term_data.name LIKE ?', 'question:%')
   end
+
+  def self.find_locations(limit, user_tag = nil)
+    user_locations =
+      if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
+        User.select('rusers.*, node_revisions.timestamp')
+            .where('rusers.status <> 0')\
+            .joins(:user_tags)\
+            .where('value LIKE "lat:%"')\
+            .joins(:revisions)\
+            .order("node_revisions.timestamp DESC")\
+            .distinct
+      else
+        User.where('rusers.status <> 0')\
+            .joins(:user_tags)\
+            .where('value LIKE "lat:%"')\
+            .joins(:revisions)\
+            .order("node_revisions.timestamp DESC")\
+            .distinct
+      end
+
+    if user_tag.present?
+      user_locations = User.joins(:user_tags)\
+                       .where('user_tags.value LIKE ?', user_tag)\
+                       .where(id: user_locations.select("rusers.id"))
+    end
+
+    user_locations = user_locations.limit(limit.to_i)
+
+    user_locations
+  end
 end
--- a/app/services/srch_scope.rb
+++ b/app/services/srch_scope.rb
@@ -66,24 +66,12 @@ class SrchScope
   end
 
   def self.find_locations(limit, user_tag = nil)
-    user_locations =
-      if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
-        User.select('rusers.*, node_revisions.timestamp')
-            .where('rusers.status <> 0')\
-            .joins(:user_tags)\
-            .where('value LIKE "lat:%"')\
-            .joins(:revisions)\
-            .order("node_revisions.timestamp DESC")\
-            .distinct
-      else
-        User.where('rusers.status <> 0')\
-            .joins(:user_tags)\
-            .where('value LIKE "lat:%"')\
-            .joins(:revisions)\
-            .order("node_revisions.timestamp DESC")\
-            .distinct
-      end
-
+    user_locations = User.where('rusers.status <> 0')\
+                         .joins(:user_tags)\
+                         .where('value LIKE "lat:%"')\
+                         .joins(:revisions)\
+                         .order("node_revisions.timestamp DESC")\
+                         .distinct
     if user_tag.present?
       user_locations = User.joins(:user_tags)\
                        .where('user_tags.value LIKE ?', user_tag)\
--- a/app/services/srch_scope.rb
+++ b/app/services/srch_scope.rb
@@ -66,12 +66,23 @@ class SrchScope
   end
 
   def self.find_locations(limit, user_tag = nil)
-    user_locations = User.where('rusers.status <> 0')\
-                         .joins(:user_tags)\
-                         .where('value LIKE "lat:%"')\
-                         .joins(:revisions)\
-                         .order("node_revisions.timestamp DESC")\
-                         .distinct
+    user_locations =
+      if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
+        User.select('rusers.*,node_revisions.timestamp')
+            .where('rusers.status <> 0')\
+            .joins(:user_tags)\
+            .where('value LIKE "lat:%"')\
+            .joins(:revisions)\
+            .order("node_revisions.timestamp DESC")\
+            .distinct
+      else
+        User.where('rusers.status <> 0')\
+            .joins(:user_tags)\
+            .where('value LIKE "lat:%"')\
+            .joins(:revisions)\
+            .order("node_revisions.timestamp DESC")\
+            .distinct
+      end
     if user_tag.present?
       user_locations = User.joins(:user_tags)\
                        .where('user_tags.value LIKE ?', user_tag)\
--- a/test/unit/helpers/search_helper_test.rb
+++ b/test/unit/helpers/search_helper_test.rb
@@ -4,7 +4,7 @@ class SearchHelperTest < ActionView::TestCase
 
 	test "search Recent People helper" do
 		sservice = SearchService.new
-		result = sservice.recentPeople(100)
+		result = sservice.people_locations(100)
 		assert_equal result.items.first.docType , "people_coordinates"
 		assert_equal result.items.first.docTitle , users(:bob).username
 		assert_equal result.items.first.docSummary , 0


jywarren:
This looks great. I'd love a review from one other person maybe? @publiclab/reviewers

app/services/search_service.rb
```diff
@@ -167,27 +167,26 @@ def tagNearbyNodes(srchString, tagName)
...
-        if tagName.blank?
-          users << node.author.user
-        else
-          users << node.author.user if node.author.user.has_tag(tagName)
-        end
-      end
+    user_scope = User.where('rusers.status <> 0')\
+                     .joins(:user_tags)\
+                     .where('value LIKE "lat:%"')\
+                     .joins(:revisions)\
```
jywarren: This might be a bit heavy if all you're wanting is the timestamp. I wonder if you could select just the timestamp, but you might have to do `.select('node_revisions.timestamp', ` plus every other column (Not sure if you can wildcard `*` multiple columns...) so then the way to get past that could be to get the absolute minimum number of columns with a `select()`, collect the `nid`s with `nids = query.collect(&:nid)` and then run a second query, like:

```ruby
User.where(nid: nids)
```

This way the original query is very narrowly restricted, then a second query collects the full User records. This is a bit messy looking but you can see how it might result in a more efficient initial query followed by a narrower second query which gets all the `User` table bits and pieces you're looking for, without the messy joins. Does this make sense?

milaaraujo: To be honest I don't know if it makes sense! :stuck_out_tongue_closed_eyes: 

Actually, under rails, the query being executed is: `SELECT  DISTINCT "rusers".* FROM "rusers" INNER JOIN "user_tags" ON "user_tags"."uid" = "rusers"."id" INNER JOIN "node" ON "node"."uid" = "rusers"."id" INNER JOIN "node_revisions" ON "node_revisions"."nid" = "node"."nid" WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT ?`. We are only selecting the data from the user table, which is what we need at the end of the day, and we use the timestamp just to sort the results. Of course, I do not know exactly how mysql runs this query, but it is unlikely that it actually returns the entire secondary table to only then select what the query explicitly asks for...

jywarren:
Hmm, well I guess I meant not the whole table meaning all the rows, but I
meant every column, and the revisions table has some bigger columns like
the node body content. So I think it would be a relatively big amount of
data to return. Maybe this optimization is more than we need, and I'm just
being paranoid...

@namangupta01 what do you think, you were recently thinking about a query
optimization?

milaaraujo:
But what I'm saying is that we are not returning every column of the revisions table. We are returning ->  `SELECT DISTINCT "rusers".*`. And I really don't know how mysql runs the joins, but I believe it has its own optimization. 

However, @namangupta01 what do you think, you were recently thinking about a query optimization? [2] :smile:

jywarren:
I think I'm at the limit of my knowledge of mysql too! I *thought* that joining essentially adds new columns for each of the joined tables columns. But I'm trying to test this to see what the raw msyql result is, and whether we're getting revision body content in the results... hang on a sec and apologies!

jywarren:
Hmm:

```ruby
> user_scope = User.where('rusers.status <> 0').joins(:user_tags).where('value LIKE "lat:%"').joins(:revisions).order("node_revisions.timestamp DESC")
# ActiveRecord::ConfigurationError: Can't join 'User' to association named 'revisions'; perhaps you misspelled it?
```

jywarren:
Oh wait, duh - `has_many :revisions, through: :node` -- adding this...

jywarren:
I see the query as:

```sql
  User Load (17.8ms)  SELECT  `rusers`.* FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT 11
```

let me think...

jywarren:
OK, full SQL output (just running the generated query in the command-line `mysql`) is:

```sql
+----+----------+-------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+-------------+--------------------+---------------------+------------------+---------------+------------------+---------------+---------------------+---------------------+-------------------+-------+-----------+-----------------+--------------------+-----------------+------+--------------+--------+
| id | username | email             | crypted_password                                                                                                                 | password_salt                                                                                                                    | persistence_token                                                                                                                | login_count | failed_login_count | last_request_at     | current_login_at | last_login_at | current_login_ip | last_login_ip | created_at          | updated_at          | openid_identifier | role  | reset_key | photo_file_name | photo_content_type | photo_file_size | bio  | token        | status |
+----+----------+-------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+-------------+--------------------+---------------------+------------------+---------------+------------------+---------------+---------------------+---------------------+-------------------+-------+-----------+-----------------+--------------------+-----------------+------+--------------+--------+
|  1 | Bob      | bob@publiclab.org | 4d7b3086073fdc9d7647f598d9e036e0e6e210b680bf6ea402631ae28afc1a92d8c9640923ed9d8d738c341b264ae147aa06bbf4024fc319ab76554fbd3d0b52 | a517e2d46393a2a1ae1fd29cc5153abee965973ed572f72bbeca977d5b81be51fdbb8a115b6d4ace6c6cefa05fecd2f5dbf57669092613cfa734b4b41c66ee82 | 0f4d6908dc76f59302b57fd35b77661a2938d7481727f611086e1ab347669337bb122e32b4b3f905f1547fb29725c10cf7133455cf8f2492888e67e64187b2f0 |           0 |                  0 | 2018-01-26 01:24:55 | NULL             | NULL          | NULL             | NULL          | 2018-01-26 01:24:55 | 2018-01-26 01:24:55 | NULL              | basic | NULL      | NULL            | NULL               | NULL            |      | abcdefg12345 |      1 |
+----+----------+-------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+-------------+--------------------+---------------------+------------------+---------------+------------------+---------------+---------------------+---------------------+-------------------+-------+-----------+-----------------+--------------------+-----------------+------+--------------+--------+
1 row in set (0.00 sec)
```

jywarren:
Oddly, it seems different when I do this:

```ruby
User.where('rusers.status <> 0').joins(:user_tags).where('value LIKE "lat:%"').joins(:revisions).order("node_revisions.timestamp DESC").select('rusers.*, node_revisions.timestamp').limit(2)
```
```sql
  User Load (12.4ms)  SELECT  rusers.*, node_revisions.timestamp FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT 2
```

jywarren:
hmm, things get stranger but maybe clearer?...

```

2.4.1 :004 > User.group('rusers.id').where('rusers.status <> 0').joins(:user_tags).where('value LIKE "
lat:%"').joins(:revisions).order("node_revisions.timestamp DESC").select('rusers.*, node_revisions.timestamp').limit(2)
  User Load (6.6ms)  SELECT  rusers.*, node_revisions.timestamp FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") GROUP BY rusers.id ORDER BY node_revisions.timestamp DESC LIMIT 2                                  
ActiveRecord::StatementInvalid: Mysql2::Error: Expression #25 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'plots.node_revisions.timestamp' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by: SELECT  rusers.*, node_revisions.timestamp FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") GROUP BY rusers.id ORDER BY node_revisions.timestamp DESC LIMIT 2
```

jywarren:
Hmm, so I think i'm now stuck. In response to the error in the last one, i think it's saying we need to add `node_revisions` to our group_by, but that's silly because then it wouldn't group by `User` and we get duplicate records (which I can confirm from running this locally).

jywarren:
I guess I suspect the original version without the extra `select()` may not have really been sorting by `timestamp`. Looking at this attempt to get a distinct response:

```ruby
2.4.1 :022 > User.where('rusers.status <> 0').joins(:user_tags).where('value LIKE "lat:%"').joins(:revisions).order("node_revisions.timestamp DESC").limit(2).distinct
```
```sql
  User Load (2.3ms)  SELECT  DISTINCT `rusers`.* FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT 2                                                                    
ActiveRecord::StatementInvalid: Mysql2::Error: Expression #1 of ORDER BY clause is not in SELECT list, references column 'plots.node_revisions.timestamp' which is not in SELECT list; this is incompatible with DISTINCT: SELECT  DISTINCT `rusers`.* FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT 2
```

jywarren:
This is hard!!! 😭

jywarren:
OK, we're also affected by this: https://dev.mysql.com/doc/refman/5.7/en/group-by-handling.html

I'm going to take a short break here, probs a good idea!

jywarren:
OK, thinking thinking. Although we could potentially solve this, our intent is to return people who are recently active. 

Thinking outside the box, what if we simply sorted by `user.last_request_at` or `user.last_login_at` ?

milaaraujo:
> Hmm, so I think i'm now stuck. In response to the error in the last one, i think it's saying we need to add node_revisions to our group_by, but that's silly because then it wouldn't group by User and we get duplicate records (which I can confirm from running this locally).

I think if you will use `group_by` you need to do something like `MAX(node_revisions.timestamp)` to keep only the row with most recent timestamp.

milaaraujo:
Did you test the exactly query our code uses `SELECT DISTINCT "rusers".* FROM "rusers" INNER JOIN "user_tags" ON "user_tags"."uid" = "rusers"."id" INNER JOIN "node" ON "node"."uid" = "rusers"."id" INNER JOIN "node_revisions" ON "node_revisions"."nid" = "node"."nid" WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT ? ` ?

jywarren:
I had actually missed your commit where you added `.distinct` but I did try that query exactly --

```ruby
user_scope = User.where('rusers.status <> 0').joins(:user_tags).where('value LIKE "lat:%"').joins(:revisions).order("node_revisions.timestamp DESC").distinct
```

```sql
SELECT  DISTINCT `rusers`.* FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT 11                                       


ActiveRecord::StatementInvalid: Mysql2::Error: Expression #1 of ORDER BY clause is not in SELECT list, references column 'plots.node_revisions.timestamp' which is not in SELECT list; this is incompatible with DISTINCT: SELECT  DISTINCT `rusers`.* FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT 11
```

jywarren:
I'm running `mysql  Ver 14.14 Distrib 5.7.21, for Linux (armv7l)` and running that query raw in the mysql console shows:

```sql
ERROR 3065 (HY000): Expression #1 of ORDER BY clause is not in SELECT list, references column 'plots.node_revisions.timestamp' which is not in SELECT list; this is incompatible with DISTINCT
```

Which I believe is the same `only_full_group_by` error in these docs: https://dev.mysql.com/doc/refman/5.7/en/group-by-handling.html

(lots of info on this here: https://duckduckgo.com/?q=ERROR+3065+(HY000+order+by+mysql&atb=v62-4&ia=qa)

jywarren:
As to why I think it's not ordering, I just don't see `timestamp` in the returned query (here reproduced without the `DISTINCT` to not trigger that error):

```sql
SELECT `rusers`.* FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT 1;
+----+----------+-------------------+----------------------------------------------------------
------------------------------------------------------------------------+----------------------
-----------------------------------------------------------------------------------------------
-------------+---------------------------------------------------------------------------------
-------------------------------------------------+-------------+--------------------+----------
-----------+------------------+---------------+------------------+---------------+-------------
--------+---------------------+-------------------+-------+-----------+-----------------+------
--------------+-----------------+------+--------------+--------+------------------+
| id | username | email             | crypted_password                                         
                                                                        | password_salt        
                                                                                               
             | persistence_token                                                               
                                                 | login_count | failed_login_count | last_requ
est_at     | current_login_at | last_login_at | current_login_ip | last_login_ip | created_at  
        | updated_at          | openid_identifier | role  | reset_key | photo_file_name | photo
_content_type | photo_file_size | bio  | token        | status | password_checker |
+----+----------+-------------------+----------------------------------------------------------
------------------------------------------------------------------------+----------------------
-----------------------------------------------------------------------------------------------
-------------+---------------------------------------------------------------------------------
-------------------------------------------------+-------------+--------------------+----------
-----------+------------------+---------------+------------------+---------------+-------------
--------+---------------------+-------------------+-------+-----------+-----------------+------
--------------+-----------------+------+--------------+--------+------------------+
|  1 | Bob      | bob@publiclab.org | 951f9a3ae04c0491b850bd0329f8de3e1454bf00ca82c30fdf9cd4c2b
e5da9ed4464274fda2266aea8234865a6150453d13b1bb286803312e3724dcd9244779d | b59cf1d8ffec4747ac017
c3a20d6109267e9e773cfdf321caf422f53d242ca0e9b6e8e6b5598f904d1d765efb8b1870610debc3d9695944a7d0f
fbd3d384861d | 1eac6265c043bd737cd2aabdb7a603163518a59440503c2dfda7781ec59a8d4bc96f4b24a13bf6fa
a7560f7950e702ab6baacc78ef770084e352e86d54730a2c |           0 |                  0 | 2018-08-0
9 20:29:41 | NULL             | NULL          | NULL             | NULL          | 2018-08-09 2
0:29:41 | 2018-08-09 20:29:41 | NULL              | basic | NULL      | NULL            | NULL 
              | NULL            |      | abcdefg12345 |      1 |                0 |
```


...while when I do select that in, I can get `timestamp` to appear:

```sql
SELECT `rusers`.*, node_revisions.timestamp FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT 1;
+----+----------+-------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+-------------+--------------------+---------------------+------------------+---------------+------------------+---------------+---------------------+---------------------+-------------------+-------+-----------+-----------------+--------------------+-----------------+------+--------------+--------+------------------+------------+
| id | username | email             | crypted_password                                                                                                                 | password_salt                                                                                                                    | persistence_token                                                                                                                | login_count | failed_login_count | last_request_at     | current_login_at | last_login_at | current_login_ip | last_login_ip | created_at          | updated_at          | openid_identifier | role  | reset_key | photo_file_name | photo_content_type | photo_file_size | bio  | token        | status | password_checker | timestamp  |
+----+----------+-------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+-------------+--------------------+---------------------+------------------+---------------+------------------+---------------+---------------------+---------------------+-------------------+-------+-----------+-----------------+--------------------+-----------------+------+--------------+--------+------------------+------------+
|  1 | Bob      | bob@publiclab.org | 951f9a3ae04c0491b850bd0329f8de3e1454bf00ca82c30fdf9cd4c2be5da9ed4464274fda2266aea8234865a6150453d13b1bb286803312e3724dcd9244779d | b59cf1d8ffec4747ac017c3a20d6109267e9e773cfdf321caf422f53d242ca0e9b6e8e6b5598f904d1d765efb8b1870610debc3d9695944a7d0ffbd3d384861d | 1eac6265c043bd737cd2aabdb7a603163518a59440503c2dfda7781ec59a8d4bc96f4b24a13bf6faa7560f7950e702ab6baacc78ef770084e352e86d54730a2c |           0 |                  0 | 2018-08-09 20:29:41 | NULL             | NULL          | NULL             | NULL          | 2018-08-09 20:29:41 | 2018-08-09 20:29:41 | NULL              | basic | NULL      | NULL            | NULL               | NULL            |      | abcdefg12345 |      1 |                0 | 1533846581 |
+----+----------+-------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------+-------------+--------------------+---------------------+------------------+---------------+------------------+---------------+---------------------+---------------------+-------------------+-------+-----------+-----------------+--------------------+-----------------+------+--------------+--------+------------------+------------+
```

jywarren:
Gosh this is not exactly fun... thanks for hanging in there!

jywarren:
Trying the `MAX()` idea, but maybe I have it wrong...

```ruby
User.group('MAX(node_revisions.timestamp)').where('rusers.status <> 0').joins(:user_tags).where('value LIKE "lat:%"').joins(:revisions).order("node_revisions.timestamp DESC").select('rusers.*,node_revisions.timestamp').limit(2)
```

```sql
  User Load (4.4ms)  SELECT  rusers.*,node_revisions.timestamp FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") GROUP BY MAX(node_revisions.timestamp) ORDER BY node_revisions.timestamp DESC LIMIT 2                                                                                  
ActiveRecord::StatementInvalid: Mysql2::Error: Invalid use of group function: SELECT  rusers.*,node_revisions.timestamp FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") GROUP BY MAX(node_revisions.timestamp) ORDER BY node_revisions.timestamp DESC LIMIT 2
```

jywarren:
maybe something like this... https://stackoverflow.com/questions/19548048/how-to-select-maxdate-and-group-by-client-id

milaaraujo:
> As to why I think it's not ordering, I just don't see timestamp in the returned query [...]

I don't think this means it is not working. It doesn't return the timestamp because we are just selecting `rusers.*`. We can order by timestamp without selecting/returning the timestamp...

jywarren:
maybe it can if there is an index of the column we want to sort by?

jywarren:
And my apologies, the reason I'm trying to work this problem is that when running this locally, i get this:

```ruby
2.4.1 :001 > SearchService.new.people_locations(2)
   (18.8ms)  SET NAMES utf8,  @@SESSION.sql_mode = CONCAT(CONCAT(@@sql_mode, ',STRICT_ALL_TABLES'), ',NO_AUTO_VALUE_ON_ZERO'),  @@SESSION.sql_auto_is_null = 0, @@SESSION.wait_timeout = 2147483                                                                                             
  User Load (6.2ms)  SELECT  DISTINCT `rusers`.* FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT 2                                        
ActiveRecord::StatementInvalid: Mysql2::Error: Expression #1 of ORDER BY clause is not in SELECT list, references column 'plots.node_revisions.timestamp' which is not in SELECT list; this is incompatible with DISTINCT: SELECT  DISTINCT `rusers`.* FROM `rusers` INNER JOIN `user_tags` ON `user_tags`.`uid` = `rusers`.`id` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT 2
        from app/services/search_service.rb:191:in `people_locations'
        from (irb):1
```

jywarren:
Is it possible this is something Mariadb can do, but Mysql cannot?

jywarren:
OK, I have a possible solution. How about (simplified for clarity, but basically replacing `joins` with `includes`):

`User.includes(:revisions).order("node_revisions.timestamp DESC").distinct.collect(&:username)`

http://blog.scoutapp.com/articles/2017/01/24/activerecord-includes-vs-joins-vs-preload-vs-eager_load-when-and-where

What do you think?

I think i need to go home now... 🏠       🏇 

I hope this works 🤞

stefannibrasil:
I checked here, we are using this version of mariadb:

```
+-----------------+
| VERSION()       |
+-----------------+
| 10.2.16-MariaDB |
+-----------------+
```

milaaraujo:
I think the last commit will solve the problem for mysql! :crossed_fingers: 

But Travis is failing with this new weird error: 
> ERROR: Service 'web' failed to build: The command '/bin/sh -c bundle install --jobs 4' returned a non-zero code: 5

:weary:

jywarren:
oh exciting... i'll try it out myself too -- let me see about the travis
error maybe we can restart it...

jywarren:
Looks like a timeout issue, maybe travis servers lost internet connection for a bit. Restarting!

`fatal: unable to access 'https://github.com/publiclab/inline-markdown-editor.git/': Failed to connect to github.com port 443: Connection timed out`

milaaraujo:
My last commit didn't work on Travis, so I'm removing the changes.

milaaraujo:
Hey Jeff, now I'm running mysql locally and I could reproduce the error you were getting. That fix I tried early really solved the problem for mysql... But it failed on Travis. 

1 - This query works for mariaDB and Sqlite:

```
User.where('rusers.status <> 0')\
    .joins(:user_tags)\
    .where('value LIKE "lat:%"')\
    .joins(:revisions)\
    .order("node_revisions.timestamp DESC")\
    .distinct
```

2 - This one works for mysql:

```
User.select('rusers.*,node_revisions.timestamp')
    .where('rusers.status <> 0')\
    .joins(:user_tags)\
    .where('value LIKE "lat:%"')\
    .joins(:revisions)\
    .order("node_revisions.timestamp DESC")\
    .distinct
```

3 - So I'm trying now this logic (last commit). It works for mysql and sqlite, but failed on Travis:

```
if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
	User.select('rusers.*,node_revisions.timestamp')
        .where('rusers.status <> 0')\
        .joins(:user_tags)\
        .where('value LIKE "lat:%"')\
        .joins(:revisions)\
        .order("node_revisions.timestamp DESC")\
        .distinct
else
    User.where('rusers.status <> 0')\
        .joins(:user_tags)\
        .where('value LIKE "lat:%"')\
        .joins(:revisions)\
        .order("node_revisions.timestamp DESC")\
        .distinct
end
```
The conclusion today is that it is not possible to make everyone happy! My question now is which database does Travis use and which database is used in production? If a code passes without errors by Travis, does that mean it will run smoothly in production? If this is true, can we leave the version 2 that works on Travis?

milaaraujo:
More insights here about this... I'm using mysql to run the site locally and a lot of things (not related to our work) are broken, for example when I try to create a new question: 

screenshot from 2018-08-11 15-15-59

So I don't think we are using mysql in production, right? I think it makes more sense to keep the version that works on Travis, what do you think?

icarito:
Just a note that we are using [Mariadb 10.2 in production](https://github.com/publiclab/plots2/blob/master/containers/docker-compose-production.yml#L4). Not sure why you are seeing this issue! Thanks for your hard work!

milaaraujo:
Hey @icarito! I was testing locally with mysql, because Jeff pointed out he was seeing some errors in my query using mysql... But then I discovered that other things also break when I test with mysql.

Now I'm focusing on leaving the code error-free testing with MariaDB (which is what's in production). My last commit seems to be working okay! 

Now that Jeff is on vacation, could you please review it for me? 😄

jywarren:
Thanks this is really thorough and good, I'm planning on doing a more
thorough review later today before finally taking off. I haven't forgotten!
😄

jywarren:
OK, going through this now. First, on the `includes()` solution, you can leave out the final `.collect(&:username)` and it'll return the full results. Ah, i see you figured that out. 

While `includes()` does select a lot, it is just more detailed in aliasing all the selected columns, and to my understanding of SQL, the results will include all the same columns. It's a long dense query when Rails writes it all out and makes all the aliases, but the effective response from the database is the same. I may be wrong about this! But that's my best understanding and the resources I've read show that `includes()` (using a different join type) is not necessarily less efficient than `joins()`.

Does `includes()` work everywhere? If so let's use it. Otherwise, let's open a new issue to very clearly tell everyone to use `mariadb` and not `mysql` and then go ahead with your solution that works in `maridab` -- i totally agree with your assessment. 

In general it's nice to use generic-enough SQL features that it'll work in all similar database systems. But we're already pushing this a bit since we use full-text search, for example. So if `includes()` doesn't work, I'm OK with formally adopting `mariadb` and dropping support for `mysql`. 

Thanks for looking through this so carefully! A very evidence-driven decision making process. If you try `includes()` and it doesn't work, just tell me when you're ready with the `mariadb` version (is it https://github.com/publiclab/plots2/pull/3217/commits/3d7c5e4f2cc826de32b84a84e9d2b4152036efd6 ?)

Thanks!!!

jywarren:
(also just want to note this error here for future reference; https://github.com/publiclab/plots2/pull/3217/commits/e4a72e577b874fe1963a9fecaa3e2fbf15726dca produces the following):

```sql
- [ ] There was a test error at:  Failure: test_search_recent_people_functionality_having_specified_tagName(Minitest::Result) [/usr/local/bundle/gems/mysql2-0.5.2/lib/mysql2/client.rb:131]: ActiveRecord::StatementInvalid: Mysql2::Error: Operand should contain 1 column(s): SELECT 'rusers'.* FROM 'rusers' INNER JOIN 'user_tags' ON 'user_tags'.'uid' = 'rusers'.'id' WHERE (user_tags.value LIKE 'tool:barometer') AND ('rusers'.'id') IN (SELECT DISTINCT rusers.*,node_revisions.timestamp, rusers.id FROM 'rusers' INNER JOIN 'user_tags' ON 'user_tags'.'uid' = 'rusers'.'id' INNER JOIN 'node' ON 'node'.'uid' = 'rusers'.'id' INNER JOIN 'node_revisions' ON 'node_revisions'.'nid' = 'node'.'nid' WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC) LIMIT 100 app/services/search_service.rb:178:in 'people_locations' app/services/execute_search.rb:23:in 'execute' app/services/execute_search.rb:3:in 'by' app/api/srch/search.rb:112:in 'execute' app/api/srch/search.rb:100:in 'block (2 levels) in <class:Search>' <a href='https://github.com/milaaraujo/plots2/tree/improve-peopleLocations-endpoint/test/functional/search_api_test.rb#L214'>test/functional/search_api_test.rb:214</a>:in 'block in <class:SearchApiTest>' 
```

milaaraujo:
Hey @jywarren! I know that the `includes()` solution definitely works, I tried it before! The problem with it is not all the aliases it creates, the problem is that it selects/returns the all the columns of the revision table:
```
SELECT DISTINCT "rusers"."id" AS t0_r0, "rusers"."username" AS t0_r1, "rusers"."email" AS t0_r2, "rusers"."crypted_password" AS t0_r3, "rusers"."password_salt" AS t0_r4, "rusers"."persistence_token" AS t0_r5, "rusers"."login_count" AS t0_r6, "rusers"."failed_login_count" AS t0_r7, "rusers"."last_request_at" AS t0_r8, "rusers"."current_login_at" AS t0_r9, "rusers"."last_login_at" AS t0_r10, "rusers"."current_login_ip" AS t0_r11, "rusers"."last_login_ip" AS t0_r12, "rusers"."created_at" AS t0_r13, "rusers"."updated_at" AS t0_r14, "rusers"."openid_identifier" AS t0_r15, "rusers"."role" AS t0_r16, "rusers"."reset_key" AS t0_r17, "rusers"."photo_file_name" AS t0_r18, "rusers"."photo_content_type" AS t0_r19, "rusers"."photo_file_size" AS t0_r20, "rusers"."bio" AS t0_r21, "rusers"."token" AS t0_r22, "rusers"."status" AS t0_r23, "rusers"."password_checker" AS t0_r24, **"node_revisions"."vid" AS t1_r0, "node_revisions"."nid" AS t1_r1, "node_revisions"."uid" AS t1_r2, "node_revisions"."title" AS t1_r3, "node_revisions"."body" AS t1_r4, "node_revisions"."teaser" AS t1_r5, "node_revisions"."log" AS t1_r6, "node_revisions"."timestamp" AS t1_r7, "node_revisions"."format" AS t1_r8, "node_revisions"."status" AS t1_r9** FROM "rusers" INNER JOIN "user_tags" ON "user_tags"."uid" = "rusers"."id" LEFT OUTER JOIN "node" ON "node"."uid" = "rusers"."id"
 LEFT OUTER JOIN "node_revisions" ON "node_revisions"."nid" = "node"."nid" WHERE (rusers.status <> 0) AND (value LIKE "lat:%") AND ("rusers"."id") IN (?, ?) ORDER BY node_revisions.timestamp DESC [["id", 6], ["id", 5]]
```
The `joins()` solution only selects/returns the user table.
```
SELECT DISTINCT "rusers".* FROM "rusers" INNER JOIN "user_tags" ON "user_tags"."uid" = "rusers"."id" INNER JOIN "node" ON "node"."uid" = "rusers"."id" INNER JOIN "node_revisions" ON "node_revisions"."nid" = "node"."nid" WHERE (rusers.status <> 0) AND (value LIKE "lat:%") ORDER BY node_revisions.timestamp DESC LIMIT ?
```
 If you don't think this is a problem, I can certainly send a new commit (using `includes()`)! :smiley:

milaaraujo:
And I see you are worried about the support for mysql. But I don't think at this point, considering the code we have in production now, we really support mysql. I run `rake test` yesterday in the master branch, using mysql, and it pointed out 43 errors (or something close to this, I can't remember)! Did you try to do this before? Or maybe I'm using a wrong mysql version...

milaaraujo:
Done.

jywarren:
Fantastic!

PR #3235, opened 2018-08-16
API Search profiles: adding parameter to search (only) by username 

### LINKED ISSUE #3228: API Search profiles by username endpoint
Last part of #3134.

We need to add the option for the endpoint to search only by username on the profiles endpoint to use it on the autocomplete feature.

milaaraujo:
Hey everybody!

We are opening this request to fix the issue #3228. 

We added a new parameter to the profile endpoint so now we can search only by username. We will use this new functionality on the autocomplete feature.

We would like some feedback, thanks! :)

fixes #3228

### DIFF
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -103,13 +103,8 @@ module Srch
 
     def self.execute(endpoint, params)
       sresult = DocList.new
-      search_query = params[:srchString]
-      tag_query = params[:tagName]
-      order_query = params[:order_direction]
-      sort_query = params[:sort_by]
-      field_query = params[:field]
       search_type = endpoint
-      search_criteria = SearchCriteria.new(search_query, tag: tag_query, sort_by: sort_query, order_direction: order_query, field: field_query)
+      search_criteria = SearchCriteria.new(params)
 
       if search_criteria.valid?
         sresult = ExecuteSearch.new.by(search_type, search_criteria)
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -21,14 +21,14 @@ module Srch
         Search.execute(:all, params)
       end
 
-      # Request URL should be /api/srch/profiles?srchString=QRY[&sort_by=recent&order_direction=desc&seq=KEYCOUNT&showCount=NUM_ROWS&pageNum=PAGE_NUM]
+      # Request URL should be /api/srch/profiles?srchString=QRY[&sort_by=recent&order_direction=desc&field=username&seq=KEYCOUNT&showCount=NUM_ROWS&pageNum=PAGE_NUM]
       # Basic implementation from classic plots2 SearchController
       desc 'Perform a search of profiles', hidden: false,
                                            is_array: false,
                                            nickname: 'srchGetProfiles'
 
       params do
-        use :common, :sorting, :ordering
+        use :common, :sorting, :ordering, :field
       end
       get :profiles do
         Search.execute(:profiles, params)
@@ -107,8 +107,9 @@ module Srch
       tag_query = params[:tagName]
       order_query = params[:order_direction]
       sort_query = params[:sort_by]
+      field_query = params[:field]
       search_type = endpoint
-      search_criteria = SearchCriteria.new(search_query, tag: tag_query, sort_by: sort_query, order_direction: order_query)
+      search_criteria = SearchCriteria.new(search_query, tag: tag_query, sort_by: sort_query, order_direction: order_query, field: field_query)
 
       if search_criteria.valid?
         sresult = ExecuteSearch.new.by(search_type, search_criteria)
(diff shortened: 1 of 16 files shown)


app/api/srch/search.rb
```diff
@@ -107,8 +107,9 @@ def self.execute(endpoint, params)
       tag_query = params[:tagName]
       order_query = params[:order_direction]
       sort_query = params[:sort_by]
+      field_query = params[:field]
       search_type = endpoint
-      search_criteria = SearchCriteria.new(search_query, tag: tag_query, sort_by: sort_query, order_direction: order_query)
+      search_criteria = SearchCriteria.new(search_query, tag: tag_query, sort_by: sort_query, order_direction: order_query, field: field_query)
```
jywarren: Do you think we could break this up over multiple lines? It could be indented so they line up. And is there an advantage to making variables to store these values in the above lines versus directly using params[] in the .new call?

app/services/search_service.rb
```diff
@@ -50,7 +50,12 @@ def textSearch_all(search_criteria)
   # If no sort_by value present, then it returns a list of profiles ordered by id DESC
   # a recent activity may be a node creation or a node revision
   def profiles(search_criteria)
-    user_scope = SrchScope.find_users(search_criteria.query, limit = 10)
+    user_scope =
+      if search_criteria.field == "username"
+        SrchScope.find_by_username(search_criteria.query, limit = 10)
```
jywarren: Here we might think of setting an optional limit parameter in profiles(search_criteria, limit = 10) and then we could override it if needed?

app/services/srch_scope.rb
```diff
@@ -6,7 +6,18 @@ def self.find_users(query, limit)
           .where('rusers.status = ?', 1)
           .limit(limit)
     else
-      User.where('username LIKE ? AND rusers.status = 1', '%' + input + '%')
+      User.where('username LIKE ? AND rusers.status = 1', '%' + query + '%')
+          .limit(limit)
+    end
+  end
+
+  def self.find_by_username(query, limit)
```
jywarren: Great!

test/functional/search_api_test.rb
```diff
@@ -148,6 +148,34 @@ def app
      assert matcher =~ json
   end
 
+  # search by username and bio, returns users by id when order_by is not provided and sorted direction default DESC
+  test 'search profiles by username and bio without order_by and default sort_direction' do
+    get '/api/srch/profiles?srchString=steff'
```
jywarren: Sorry, looking on my phone... is there also a text for only bio content? Something that has no username matches?

stefannibrasil: we did like this: the previous tests (that we first had in mind to search only by username) have the `field=username` and therefore, returns only the users searched by username. So we added another one that would search only by username, but yeah, good point, we added a new one to search only by bio!

stefannibrasil: @jywarren we pushed some changes, thanks for the feedback!

milaaraujo:
Hey @jywarren,

We've just realized that code in the master branch contains an error. In line 9 of plots2/app/services/srch_scope.rb we forgot to replace input for query.
desenho sem titulo 1
This error only occurs in development/test environment, when using sqlite. That's why Travis did not complain about it. We are fixing this problem here in this PR!

stefannibrasil:
@publiclab/reviewers anyone to review this PR? :)

test/fixtures/users.yml
```diff
@@ -129,7 +129,7 @@ test_user:
   password_salt: <%= salt = Authlogic::Random.hex_token %>
   crypted_password: <%= Authlogic::CryptoProviders::Sha512.encrypt("secretive" + salt) %>
   persistence_token: <%= Authlogic::Random.hex_token %>
-  bio: ''
+  bio: 'I love ruby'
```
jywarren: 😄 💎

test/fixtures/users.yml
```diff
@@ -210,3 +210,16 @@ steff3:
...
+data:
+  username: data
+  status: 1
+  email: data@pxlshp.com
+  id: 18
+  password_salt: <%= salt = Authlogic::Random.hex_token %>
+  crypted_password: <%= Authlogic::CryptoProviders::Sha512.encrypt("secretive" + salt) %>
+  persistence_token: <%= Authlogic::Random.hex_token %>
+  last_request_at: <%= Time.now %>
+  bio: 'data is a nice cat and he loves steff!'
```
jywarren: 😹

jywarren:
Hi, this looks great, and I'm going to merge it. Just for future reference, if you can also say something like "ready to merge if you approve" then I can do so straight away... here it looks fine but sometimes folks want feedback but not actually to have it merged. Maybe that'll help me speed things up on my end a bit -- many thanks!

Issue #3170, opened 2018-08-01
Use MySQL on development and production?


stefannibrasil:
@milaaraujo and I were having some errors on travis even having our checks all green locally. We asked our coaches to help us and we were able to fix the error (#3134). We were doing full text search, to give more context.

The problem was that locally, we were using SQLite3 and travis uses maria db, so the query was not returning the same results.

Is it possible to have a unique and consistent database type for development and production? What do you think? We believe this may improve the future contributions for the project :)

we found some links that might be useful for this [here](https://stackoverflow.com/questions/29743444/rails-databases-why-use-same-dbms-for-development-and-production) and [here](https://medium.com/@amliving/rails-db-lesson-learned-e416e2f2549c).

jywarren:
aha - sorry, i should've caught that! Hmm, so we've traditionally used
sqlite because it's such an easy install (by default already available on
Macs, I think?) but this has come up before, and I agree it's a frustrating
issue.

I wonder if we could think about how to remind ourselves about the database
differences - so when this does happen, there's some clue to what's
happening... i don't know, do you think there's an approach like that which
could work?

But i'm not set on either way, i just want to think through this carefully.

stefannibrasil:
I added this to the Testing docs file #3483

PR #3242, opened 2018-08-21
Fix schema-rb-example file for SQLite

stefannibrasil:
@milaaraujo  and I have found an error today on the `schema-example.rb` file.

When we were trying to understand what was happening with the Travis error on #3134, we tried several things to connect the app with MySQL and that might have caused it... and since I haven't updated my project to use SQLite, I didn't see the error before.

We realized today that it created a `schema.rb` that only works with MySQL and that was pushed to the master... So this PR fix this. It also fixes an error caused that only happens on SQLite (which we didn't find before because we are using MySQL from now on and Travis as well).

We hope this didn't cause much trouble, but it's really tricky to catch all these things.

Thanks!

### DIFF
--- a/db/schema.rb.example
+++ b/db/schema.rb.example
@@ -1,3 +1,4 @@
+# encoding: UTF-8
 # This file is auto-generated from the current state of the database. Instead
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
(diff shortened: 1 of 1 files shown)


stefannibrasil:
@publiclab/reviewers

jywarren:
Ah, ok - great. Can you resolve the conflicts here and rebase if needed, and then try `git push -f https://github.com/publiclab/plots2.git HEAD:unstable` or `git push -f git@github.com:publiclab/plots2.git HEAD:unstable` to push this onto the experimental `unstable` branch and test it out at https://unstable.publiclab.org to be sure it works in the exact production environment. `@icarito` can also help if `unstable` gets stuck so do reach out to him in that case.

Thanks!

stefannibrasil:
hi @jywarren, @icarito helped me and the https://unstable.publiclab.org  is running with this commit. I believe this is ready to merge when you approve? Thanks!

jywarren:
Great, thanks! Perhaps related to fully switching to mariadb, we had thought of the possibility of adding a second run of tests that run everything in sqlite... good to consider in #3170

PR #3257, opened 2018-08-25
Remove Typeahead Service

stefannibrasil:
This is the first major attempt to decrease the amount of duplicated and ambiguous work that the API currently has. We removed a bunch of unused lines as well.

If we query the Search services from API/docs we see that Typeahead results aren't super useful or meaningful for the project. So our goal is to focus entirely on the Search Service, that currently returns better results than Typeahead. Since the beginning, we haven't touched on this Typeahead service, which is a signal that it isn't being used...

The tests are all green. For now, the queries are not so great, they will be improved along the next weeks (hopefully). With that in mind, we are planning the advanced search now. Our main goal is to have one good Search Service that can be useful to the community and returns insightful results.


app/api/srch/api.rb
```diff
@@ -19,9 +18,7 @@ class API < Grape::API
                               models: [
                                 SearchRequest::Entity,
                                 DocResult::Entity,
-                                TagResult::Entity,
-                                DocList::Entity,
-                                TagList::Entity
```
jywarren: Just to confirm, does the Search Service return tags? I think I remember that it may, just as part of a DocList or something.... but note how tags are returned in the site header's autocompletion search: 

image

jywarren: Otherwise this cleanup looks great!

app/api/srch/search.rb
```diff
@@ -34,7 +34,7 @@ class Search < Grape::API
         Search.execute(:profiles, params)
       end
 
-      # Request URL should be /api/srch/notes?srchString=QRY[&seq=KEYCOUNT&showCount=NUM_ROWS&pageNum=PAGE_NUM]
```
jywarren: Just curious what the reason for removing these parameters... sorry, haven't looked at this code for a while, can you remind me if they're not used or something?

app/api/srch/api.rb
stefannibrasil: hey, @jywarren I pushed the changes to the unstable, but I don't know why the server isn't running... anyway, about that, I spent the night fixing this. DocResult also has those attributes, so I changed the typeahead script to use the items from DocResult instead of TagResult. I can't reproduce locally nor on the unstable to send to you a print, but the last commit shows how the icons are generated. 

Interesting because now I finally get what are those DocResult params for! xD thanks for the help, I believe now this is good to go!

jywarren:
OK, looks good then! Merging this now. Thanks! And we'll double check that the tags still appear in stable's autocomplete search.

PR #3255, opened 2018-08-23
Add links for different sorting options in searches (#2544)

gSanchezCiarrocchi:
Fixes #2544
 @jywarren hope somebody can review this, thanks

PR #3275, opened 2018-08-30
Don't display link for the current sort option in the results page's sort option links

gSanchezCiarrocchi:
When a particular sort option is selected, display the text of the selected option in the links bar with a different style and without an anchor element.

Screenshots:

default-sort

sort-by-likes

@jywarren Take a look at this PR, is something like this what you had in mind [here](https://github.com/publiclab/plots2/pull/3255#issuecomment-416227264)?

### DIFF
--- a/app/assets/stylesheets/dashboard.css
+++ b/app/assets/stylesheets/dashboard.css
@@ -374,15 +374,3 @@
   color: #75754e;
   background: #f3ffc1;
 }
-
-span#searchSortingOptionsLinks a {
-  margin-left: 5px;
-  margin-right: 5px;
-  font-size: small;
-}
-
-span#searchSortingOptionsLinks:not(a) {
-  margin-left: 5px;
-  margin-right: 5px;
-  font-size: small;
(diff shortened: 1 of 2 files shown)


jywarren:
Wow! This is really nice. I didn't know about `link_to_unless_current`! 

I have one small request -- this style is only used in this one context and really nowhere else. So I'd like to ask if you could actually move the CSS into an inline `<style>` here. 

If you can make that change, i'd love to merge this in. Thanks for a great contribution!

gSanchezCiarrocchi:
@jywarren Done, I've updated the PR. 

P.S: I didn't knew about `link_to_unless_current` either until now

jywarren:
Fantastic. Thank you so much! We'd love more help from you if you're interested! 

There's one that might be easy or might be tough here: https://github.com/publiclab/plots2/issues/3279

There are a lot of others here that might be a bit simpler: https://github.com/publiclab/plots2/labels/fto-candidate

Thanks!

PR #2547, opened 2018-03-26
Update node.search to add Boolean mode

jywarren:

PR #3282, opened 2018-09-01
Adding Boolean mode option to node.search()

milaaraujo:
1 - Update node.search to add Boolean mode: Done
2 - Update /search to use node.search(): Done

Moved from #2547

### DIFF
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -101,18 +101,6 @@ module Srch
       end
     end
 
-    # Testing fulltext search
-    desc 'Perform a search of nodes', hidden: false,
-                                      is_array: false,
-                                      nickname: 'srchGetNodes'
-
-    params do
-      use :common
-    end
-    get :nodes do
-      Search.execute(:nodes, params)
-    end
-
     def self.execute(endpoint, params)
       sresult = DocList.new
       search_type = endpoint
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -101,6 +101,18 @@ module Srch
       end
     end
 
+    # Testing fulltext search
+    desc 'Perform a search of nodes', hidden: false,
+                                      is_array: false,
+                                      nickname: 'srchGetNodes'
+
+    params do
+      use :common
+    end
+    get :nodes do
+      Search.execute(:nodes, params)
+    end
+
     def self.execute(endpoint, params)
       sresult = DocList.new
       search_type = endpoint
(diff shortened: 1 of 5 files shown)


jywarren:
Hi! Is it possible to offer different types of search as parameter settable options, like here: 

https://publiclab.org/search/community?order=likes

Then, you can demonstrate how one type is better than others by linking to and comparing actual results on production. But also acknowledging that for SOME uses, people may choose a non-default sort. Does this make sense? Sorry, just catching up here! It was a holiday in the US yesterday 🏖

jywarren:
image

milaaraujo:
Hey @jywarren! No problem, yesterday was a holiday here in Canada too. Don't you think that we should open a new issue/PR to compare the results? So maybe we can review #3282 first and then compare results to see which mode we will keep as default... Or do you think we should do the comparison here?

milaaraujo:
And we are also working on the view, so the search will be able to offer different types of search as parameter settable options. But again, we were planning on doing this in a different PR.

jywarren:
Oh yes, a separate issue/pr is fine, i just responded here because you
mentioned the optimal search type.

milaaraujo:
Sorry I was not so clear! :grimacing:
And about this PR? What do you think?

jywarren:
super, great!

milaaraujo:
Hey Jeff, I tried my best to make code climate happy, but it seems an impossible mission! :laughing: 

I've followed your sugestion here (and it worked):
```
.where('MATCH(node_revisions.body, node_revisions.title) AGAINST(? IN NATURAL LANGUAGE MODE)', query.to_s)
```
But using the same strategy with the `select` does not work. For some reason the '?' is not replaced by the query:

>ActiveRecord::StatementInvalid (Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '? IN NATURAL LANGUAGE MODE) AS score, test FROM `node_revisions` WHERE (MATCH(no' at line 1: SELECT node_revisions.nid, node_revisions.body, node_revisions.title, MATCH(node_revisions.body, node_revisions.title) AGAINST(? IN NATURAL LANGUAGE MODE) AS score, test FROM `node_revisions` WHERE (MATCH(node_revisions.body, node_revisions.title) AGAINST('test' IN NATURAL LANGUAGE MODE)))

Anyway the code that is in the master now also concatenates the string in the query, like I'm doing now, so I do not think it should be big problem!

https://github.com/publiclab/plots2/blob/master/app/models/node.rb#L38

milaaraujo:
@stefannibrasil also found this: https://gitlab.com/gitlab-org/gitlab-ce/issues/33887

milaaraujo:
Hey Jeff, finally the possible SQL injection is fixed! Code climate is still complaining because of the number of lines, but can we just leave it the way it is? I don't think refactoring this method should be a priority right now.

jywarren:
Super! Congratulations and agreed re: codeClimate. Merging now!

PR #3286, opened 2018-09-02
Adding sort_by=recent as default in search/all-> profiles

milaaraujo:
Adding sort_by=recent as default in search/all-> profiles

### DIFF
--- a/app/services/search_criteria.rb
+++ b/app/services/search_criteria.rb
@@ -18,6 +18,10 @@ class SearchCriteria
     sanitize_direction(@order_direction)
   end
 
+  def add_sort_by(value)
+    @sort_by = value
+  end
+
   private
 
   def sanitize_direction(direction)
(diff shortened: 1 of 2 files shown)


jywarren:
This looks good - ready to merge?

jywarren:
Any worries that this won't return very active users before new users? Just thinking about the ideal ordering... if you search for `warren` you rarely get me (`@warren`) even though i post a lot:

image

just wondering! we can def. address this in a follow-up issue.

jywarren:
I mean i guess this new default sorting may be better, i don't know!

milaaraujo:
I think it is ready to merge.

With `sort_by=recent` it will return users with most recent activity (not the most recent/new users): https://github.com/publiclab/plots2/blob/f2082151f65254825594157028fcf56516725d93/app/services/search_service.rb#L51

jywarren:
ah, ok perfect!

jywarren:
🎉 !

PR #3292, opened 2018-09-04
Removing hash from inline maps

### LINKED ISSUE #3272: Inline maps need url hash feature disabled from LEL
@sagarpreet-chadha would you mind submitting a fix for this? I think the URL hash feature is great but i think we need to make it be default-off, and re-enable it for pages where we want it (publiclab.org/maps, publiclab.org/people, mapknitter.org, etc) and not for inline maps. It looks like it causes some trouble esp. when there are multiple maps!

sagarpreet-chadha:
Fixes #3272 !

 This PR removes Hash from : 
* inline maps
* sidebar maps

### DIFF
--- a/app/views/map/_leaflet.html.erb
+++ b/app/views/map/_leaflet.html.erb
@@ -28,7 +28,7 @@
 
     setupFullScreen(map<%= unique_id %>  , map_lat , map_lon) ; 
     setupLEL(map<%= unique_id %> ) ;
-    var hash = new L.Hash(map<%= unique_id %> );  
+    // var hash = new L.Hash(map< unique_id > );  
     
     function onMapLoad(e){
       var lat = <%= lat %> ;


sagarpreet-chadha:
I think --- for sharing of maps , Hash library fulfills half of our need because we cannot share the overlay layers (only location is shared as of now in URL) . 
Open to brainstorm solution for this problem .

app/views/map/_leaflet.html.erb
```diff
@@ -28,7 +28,7 @@
 
     setupFullScreen(map<%= unique_id %>  , map_lat , map_lon) ; 
     setupLEL(map<%= unique_id %> ) ;
-    var hash = new L.Hash(map<%= unique_id %> );  
+    // var hash = new L.Hash(map< unique_id > );
```
siaw23: you might as well remove it completely, not just commenting it out.

jywarren:
I think we'll be trying to add hashing back in for things like layers... in any case i'd like to get this merged to resolve some URL issues I've been seeing and we can fix in follow up if we like. Thanks!

PR #3295, opened 2018-09-05
Apply UX concepts to the search in navbar 

stefannibrasil:
hi @publiclab/community-reps @publiclab/reviewers 

@milaaraujo and I are now working on the search results. This PR (and many others will come in the future :smile: ) attempts to start addressing the concerns discussed in [Brainstorm search results Issue](https://github.com/publiclab/plots2/issues/2421).

Here is a suggestion that we have studied and we would like to ask for your feedback.

In order to make the search on the navbar more clean and intuitive (we tried! :D) we thought that we could remove some stuff. We believe that the less the better, especially in a navbar. With so many buttons, it may be harder to find what you are looking for or make the user give up for trying to understand the website (I think we could add a CTA to the Donate and Log in buttons, but for now looks good).

The results are limited to 3 for the category because the idea is to **show the more accurate results** there for a keyword. This is something that we are still working on, but having the front end will be helpful for the next steps!

We pushed the changes to the [unstable branch](https://unstable.publiclab.org/),** you can all test it to have a better experience of those changes! **currently not working, don't know why :(**

Here are some screenshots of the changes but it would be awesome if you could test it.

Desktop 1
^ Desktop 

Desktop 2
^ Desktop with the results from the search bar more intuitive

Mobile 1

^ Same view but on mobile

Mobile 2

^ Mobile search results

--------------------------
We have some other things in mind for the overall search feature, stay tuned!


jywarren:
Can you try to ask @icarito in the chatroom about unstable? It's possible someone else pushed over it?

I love this. I'd love to hear input from others but the things I like most are:

1. removing the button -- BUT I'd like to see if the default placeholder text in that text area could include a magnifying glass? Like 🔍  perhaps... just to make it extra clear that this is a search box? 

image

2. I love the organizing by type. Maybe "coordinates" could become "places"?
3. `3 for the category` -- i wonder if this is too aggressively narrowing. What if we did 5? Would that be too much? Sometimes when I search, the result i'm looking for is more than 5 items down. Could we restrict to 3 for most, but 5 for nodes?

image

stefannibrasil:
@jywarren  great, thanks so much for the feedback!

About your comments, I don't think it's possible to select a different number of items for each category - at least for this library- let's stick to 5, then!

The only thing that I can commit to work later is on the search button. Because right now, if I use the icon, it breaks the mobile breakpoint. Maybe we can leave as it is now and I can create an issue for this?

stefannibrasil:
@icarito can't access the unstable yet. Maybe my rebase didn't work? I don't know anymore :sos:

icarito:
Hi - Last night I saw you did push to unstable successfully and it
booted! I assumed it worked? Perhaps someone pushed to unstable after you?

Please try again, I saw
https://jenkins.laboratoriopublico.org/job/Plots-Unstable/291/ booting
fine and saw it working at https://unstable.publiclab.org/ .

stefannibrasil:
yeah, yesterday it wasn't working after I pushed. I don't know exactly where things started going to an unexpected path, but now I have commits here that aren't mine... I'll try to fix here, thanks @icarito!

stefannibrasil:
moved to #3304

PR #3309, opened 2018-09-11
Comparing search modes (boolean and natural)

milaaraujo:
Hey @jywarren, I'm opening this PR to add the boolean mode option to the search view. So the community can compare the different search options and we can decide which one to keep as default.

desenho sem titulo 3

'Natural mode' was previously 'best match'. We changed the name for now to make clear that it's the natural mode search. After this period of tests we can choose better names.

### DIFF
--- a/app/views/searches/results.html.erb
+++ b/app/views/searches/results.html.erb
@@ -3,7 +3,9 @@
   <h3>
     Search results for <i style="color:#aaa;"><%= params[:id] %> </i>
     <span style="margin-left: 5px; margin-right: 5px; font-size: small;">
-      <%= link_to_unless_current 'best match', controller: :searches, action: :results, id: params[:id], order: 'natural' %>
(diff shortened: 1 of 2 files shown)


jywarren:
Super! I'll try to publish now.

jywarren:
This is live at https://stable.publiclab.org!

I think you can start asking feedback now. I have another comment unrelated before I start to push this to production.

jywarren:
This is live on publiclab.org now too! 

If you could get a kind of "challenge" together asking people to try different searches and compare, that would be great. IT might also be good to ask people to compile a list of keywords they want "good" results for, so we can do additional tuning!

If you posted a question on https://publiclab.org/questions this might be a good place to get feedback from folks!

Great work!

jywarren:
Perhaps you should ask people to document:

```
a) what you're looking for, 
b) what search terms you use, 
c) what url you see, 
d) what you see there and how it differs from what you expect, and 
e) what pieces of content (urls) you were expecting to come up in your search
```

That way you'll get structured feedback that you can work on tuning the search towards!

PR #3310, opened 2018-09-11
show more results on typeahead located at navbar

stefannibrasil:


jywarren:
Going to production now!

jywarren:
Hmm, intersting, getting some duplicates?

image

stefannibrasil:
huum, maybe some of the methods from the SearchService are not selecting distinct results :thinking: 

I'll add that tonight! Good to see that the grouping by category was solved!!

Just a curiosity, do you think the results are different now because of the Node.search? they do seem different :)

jywarren:
I think they are different but it's tough to quantify... i hope the
feedback from people will help!

PR #3313, opened 2018-09-12
Using node.search() in the search service methods

milaaraujo:
Using node.search() in the search service methods.
Removing duplicates from some endpoints (#3310).
Adding 'limit' parameter to methods.

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -24,7 +24,7 @@ class Node < ActiveRecord::Base
   self.table_name = 'node'
   self.primary_key = 'nid'
 
-  def self.search(query:, order: :default, type: :natural, limit:)
+  def self.search(query:, order: :default, type: :natural, limit: "25")
     order_param = if order == :default
                     { changed: :desc }
                   elsif order == :likes
@@ -39,11 +39,13 @@ class Node < ActiveRecord::Base
           query = connection.quote(query.to_s + "*")
           nids = Revision.select("node_revisions.nid, node_revisions.body, node_revisions.title, MATCH(node_revisions.body, node_revisions.title) AGAINST(#{query} IN BOOLEAN MODE) AS score")
             .where("MATCH(node_revisions.body, node_revisions.title) AGAINST(#{query} IN BOOLEAN MODE)")
+            .limit(limit)
             .collect(&:nid)
         else
           query = connection.quote(query.to_s)
           nids = Revision.select("node_revisions.nid, node_revisions.body, node_revisions.title, MATCH(node_revisions.body, node_revisions.title) AGAINST(#{query} IN NATURAL LANGUAGE MODE) AS score")
             .where("MATCH(node_revisions.body, node_revisions.title) AGAINST(#{query} IN NATURAL LANGUAGE MODE)")
+            .limit(limit)
             .collect(&:nid)
         end
         where(nid: nids, status: 1)
@@ -52,6 +54,7 @@ class Node < ActiveRecord::Base
         tnids = Tag.find_nodes_by_type(query, type = %w(note page)).collect(&:nid) # include results by tag
         where(nid: nids + tnids, status: 1)
           .order(order_param)
+          .limit(limit)
       end
     else
       nodes = Node.limit(limit)
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -40,12 +40,14 @@ class Node < ActiveRecord::Base
           nids = Revision.select("node_revisions.nid, node_revisions.body, node_revisions.title, MATCH(node_revisions.body, node_revisions.title) AGAINST(#{query} IN BOOLEAN MODE) AS score")
             .where("MATCH(node_revisions.body, node_revisions.title) AGAINST(#{query} IN BOOLEAN MODE)")
             .limit(limit)
+            .distinct
             .collect(&:nid)
         else
           query = connection.quote(query.to_s)
           nids = Revision.select("node_revisions.nid, node_revisions.body, node_revisions.title, MATCH(node_revisions.body, node_revisions.title) AGAINST(#{query} IN NATURAL LANGUAGE MODE) AS score")
             .where("MATCH(node_revisions.body, node_revisions.title) AGAINST(#{query} IN NATURAL LANGUAGE MODE)")
             .limit(limit)
+            .distinct
             .collect(&:nid)
         end
         where(nid: nids, status: 1)
@@ -55,6 +57,7 @@ class Node < ActiveRecord::Base
         where(nid: nids + tnids, status: 1)
           .order(order_param)
           .limit(limit)
+          .distinct
       end
     else
       nodes = Node.limit(limit)
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -66,12 +66,10 @@ class SearchService
   end
 
   # Search notes for matching strings and package up as a DocResult
-  def textSearch_notes(srchString, limit = 25, order = :natural, type = :boolean)
+  def textSearch_notes(srchString, limit = 25)
     sresult = DocList.new
 
-    nodes = Node.search(query: srchString, order: order, type: type, limit: limit)
-                .where("`node`.`type` = 'note'")
-
+    nodes = find_notes(srchString, limit)
     nodes.each do |match|
       doc = DocResult.fromSearch(match.nid, 'file', match.path, match.title, 'NOTES', 0)
       sresult.addDoc(doc)
@@ -81,11 +79,10 @@ class SearchService
   end
 
   # Search nodes package up as a DocResult
-  def textSearch_nodes(srchString, limit = 25, order = :natural, type = :boolean)
+  def textSearch_nodes(srchString, limit = 25)
     sresult = DocList.new
 
-    nodes = Node.search(query: srchString, order: order, type: type, limit: limit)
-                .where("`node`.`type` = 'page' OR `node`.`type` = 'place' OR `node`.`type` = 'tool'")
+    nodes = find_nodes(srchString, limit)
 
     nodes.each do |match|
       doc = DocResult.fromSearch(match.nid, 'file', match.path, match.title, 'PAGES', 0)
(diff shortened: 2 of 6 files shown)


milaaraujo:
Hey @jywarren, here
```
  def textSearch_nodes(srchString)
    sresult = DocList.new

    nodes = find_nodes(srchString, 25)
    puts nodes.inspect
    nodes.each do |match|
      *doc = DocResult.fromSearch(match.nid, 'file', match.path, match.title, 'NOTES', 0)*
      sresult.addDoc(doc)
    end

    sresult
  end
```
Should we use `'NOTES'` or something else?

jywarren:
I'm not that familiar with this section -- `DocResult.fromSearch` -- where is it looking for `NOTES`?

I guess the "nodes" should return node type "note" and "page", and maybe even "map". But I'm not sure where the logic is for that. Can you link to `fromSearch`?

stefannibrasil:
It isn't, 'NOTES' is new, to use on the Category header for the typeahead. This DocResult.fromSearch method is used to get the result type, path, etc:

https://github.com/publiclab/plots2/blob/75f094a5e4e224c26a40699d13e3fa262714e8bf/app/models/doc_result.rb#L7

What I am thinking is that maybe we could separate those node types. Previosly they were all cat

But Notes we already have:

https://github.com/publiclab/plots2/blob/e07f342c1a73dda42edcaa0d7af32ade46ece3ea/app/services/search_service.rb#L221

and maps as well (nearby nodes and people locations).

We can add 'pages', but what is a Node of type 'pages'. Is it a wiki? What is the difference between a page and a Note? That is not super clear to me.

Maybe we should have only a general find_nodes method that returns all of this together and use the 'NOTES' category for them? 

Recap: This category is only used for the typeahead. The /search will have its own way to filter by type, on its view sepparetly.

stefannibrasil:
I think what would be super helpful would to have a description of what those SearchService methods should return, in terms of type of nodes, especially.

@milaaraujo and I haven't worked on the majority of them, just a heads up.

milaaraujo:
+
Removing duplicates from some endpoints (#3310).
Adding 'limit' parameter to methods.

milaaraujo:
Code Climate is complaining about similar blocks of code. But I really think that in this case, it makes the code clearer.

jywarren:
Currently, as you guessed, 'page' type means wiki pages. Unfortunately, we have some legacy names guiding all this -- see more at https://github.com/publiclab/plots2/blob/master/doc/DATA_MODEL.md

However, eventually, the goal might be to break 'notes' into 'notes' vs. 'questions', but for now PAGES for wiki pages sounds great.

milaaraujo:
So, I think we are ready to merge?

milaaraujo:
I just made some last changes! And now, after your review, we're ready.

app/models/node.rb
```diff
@@ -24,7 +24,7 @@ class Node < ActiveRecord::Base
   self.table_name = 'node'
   self.primary_key = 'nid'
 
-  def self.search(query:, order: :default, type: :natural, limit:)
+  def self.search(query:, order: :default, type: :natural, limit: "25")
```
jywarren: lets have this be an integer here, and convert to a string when splicing it in -- is that ok?

milaaraujo: Sure. Actually, I don't think we need to convert it to a string.

jywarren:
Ok, last questions and we're probably ready. "nodes" was used because although in theory its great to be able to search wiki pages and notes separately, in practice we are mostly searching them together, although the separation is useful because we're likely to be able to do more sophisticated searching soon (!! 😁). What I want to confirm before merging is that in the current cases, we aren't omitting notes or pages in common searching scenarios?

app/controllers/searches_controller.rb
```diff
@@ -14,7 +14,7 @@ def results
     @title = 'Search'
     @tagnames = params[:id].split(',')
     @users = SearchService.new.find_users(params[:id], limit = 10)
-    @nodes = SearchService.new.find_nodes(params[:id], 100, params[:order].to_s.to_sym, params[:type].to_s.to_sym)
+    @nodes = SearchService.new.find_pages(params[:id], 100, params[:order].to_s.to_sym, params[:type].to_s.to_sym)
```
jywarren: Like here, shouldn't we search both notes and pages? We could combine two "find" calls, or does this mean we may still need a third "find_nodes"?

app/services/execute_search.rb
```diff
@@ -14,11 +14,11 @@ def execute(type, search_criteria)
      when :profiles
        sresult = sservice.profiles(search_criteria)
      when :notes
```
jywarren: Here, are we skipping pages?

app/services/search_service.rb
```diff
@@ -7,30 +7,28 @@ def textSearch_all(search_criteria)
     sresult = DocList.new
 
     # notes
-    noteList = textSearch_notes(search_criteria.query)
+    noteList = textSearch_notes(search_criteria.query, search_criteria.limit)
```
jywarren: Here I don't see pages represented. Maybe they should be?

app/services/search_service.rb
```diff
@@ -68,26 +66,39 @@ def profiles(search_criteria)
...
+    nodes.each do |match|
       doc = DocResult.fromSearch(match.nid, 'file', match.path, match.title, 'NOTES', 0)
       sresult.addDoc(doc)
     end
 
     sresult
   end
 
+  # Search nodes package up as a DocResult
+  def textSearch_pages(srchString, limit = 25)
```
jywarren: Do we need an equivalent combined nodes function? Could we concatenate the two existing queries somewhere to reduce redundancy?

app/controllers/searches_controller.rb
stefannibrasil: hi, Jeff, all of this concerns are fixed on the current branch that we are working and this was something that I mentioned at the time Camila opened this PR... because it makes no sense to separate things like it was implemented.

app/services/search_service.rb
stefannibrasil: So, we can use this as our unique method/endpoint to search for nodes, right?

```
def find_nodes(input, _limit = 5, order = :default)
    Node.search(query: input, order: order, limit: 5)
        .group(:nid)
        .where('node.status': 1)
        .distinct
  end

File: app/controllers/searches_controller.rb (hunk 2 above)
Comment (jywarren) [reply app/controllers/searches_controller.rb]: Hi, I'm sorry, it's a combination of issues on my end -- it's a big PR, doing multiple things, so it's a little hard to keep it all in my head at once when providing feedback. Maybe in the future, breaking ones like this into multiple parts (say, limits, distincts, and separation of node/notes/pages?) could help.  But also, I didn't understand originally that the central premise of the PR was to present separate results in the Typeahead for NOTES vs. PAGES, as in this display there are only NOTES:

image

Now I understand, and I think separate sections for NOTES and PAGES sounds wonderful. However, I still want to be sure that, for example, a call like:

`SearchService.textSearch_all()` 

will return both notes and pages. They could be together, merged, or concatenated separately, but we should be sure they aren't forgotten -- i'm mainly thinking if there are any cases where someone would search and expect to see a page, but somehow we've forgotten to display that type.

File: app/services/search_service.rb (hunk 5 above)
Comment (jywarren) [reply app/services/search_service.rb]: This looks good. Do you want to add it here?

Comment (jywarren) [reply app/services/search_service.rb]: er, to this PR (i didn't mean to this place in the code, to be clear)

**Comment by jywarren** [comment]
OK, I left a couple clarifying comments. If you're happy with how this works at both the typeahead and `/search/____` interfaces, I'm happy to merge it, so just tell me what you think we should do?

File: app/services/search_service.rb (hunk 4 above)
Comment (jywarren) [reply app/services/search_service.rb]: Ah, on line 13!

**Comment by jywarren** [comment]
OK, just read through everything one more time. I think this is good to merge, so if you 👍 i can do so. I looked to see what are "added on" questions or concerns and what are core to the purpose(s) of this PR. And I think it looks good.

File: app/services/execute_search.rb (hunk 3 above)
Comment (jywarren) [reply app/services/execute_search.rb]: I'm still curious about this but it doesn't seem that it needs to be resolved in this PR. But curious about your thoughts!

File: app/services/search_service.rb (hunk 5 above)
Comment (jywarren) [reply app/services/search_service.rb]: So on close reading, I don't think we NEED this in this PR, so i vote to merge, and if you think that there is a good use for a generalized `find_nodes` we can do it in a follow-up PR.

Comment (milaaraujo) [reply app/services/search_service.rb]: We don’t have a ‘/search/nodes’ endpoint. We had a find_nodes() method that I remove to use find_pages(). It is true it just search for pages. If we want to search for all type of nodes we just need:

def find_nodes(input, _limit = 5, order = :default)
    Node.search(query: input, order: order, limit: 5)
  end

So, it would be simpler just to call Node.search()?

**Comment by jywarren** [comment]
👍 thanks very much for your hard work on this one and apologies again for the ambiguity! I think it's great.

**Comment by milaaraujo** [comment]
I think it would be better to merge it! If we need to do any modification I can open a new PR. Stéfanni needs this to finish the frontend part.

File: app/services/search_service.rb (hunk 5 above)
Comment (jywarren) [reply app/services/search_service.rb]: Agreed, simpler.

**Comment by jywarren** [comment]
Great, done!!!

**Comment by jywarren** [comment]
Thanks, everybody!

File: app/services/execute_search.rb (hunk 3 above)
Comment (milaaraujo) [reply app/services/execute_search.rb]: Yes, here we are just using notes. But this is how the endpoint was working before:
```
def find_nodes(input, limit = 5, order = :default, type = :natural)
   Node.search(query: input, order: order, type: type, limit: limit)
            .where('type = "note" AND node.status = 1 AND title LIKE ?', '%' + input + '%')
[...]
```
And now we are using `Node.search()`:
```
def find_notes(input, limit = 25, order = :natural, type = :boolean)
   Node.search(query: input, order: order, type: type, limit: limit)
            .where("`node`.`type` = 'note'")
```

Issue #3040, opened 2018-07-10
SMTP testing system - planning


jywarren:
Logs show this error when I try to approve a new post:

```
[372b6dd2-719b-4dd1-8d3a-7146e2196e67] Sent mail to xxxxx@xxxxx.com (29.0ms)
[372b6dd2-719b-4dd1-8d3a-7146e2196e67] Completed 500 Internal Server Error in 743ms (ActiveRecord: 12.0ms)
[372b6dd2-719b-4dd1-8d3a-7146e2196e67]   
[372b6dd2-719b-4dd1-8d3a-7146e2196e67] Net::SMTPFatalError (550 relay not permitted
```

icarito:
Hi,
The issue is we had hardcoded / whitelisted the app container IP address in our mail gateway configuration. But Docker may change this IP address. The solution was to relay for the IP address range of docker containers. This should have fixed this issue!

jywarren:
Is this something we can develop monitoring for? i worry it failed silently...

icarito:
It would be tricky to monitor this. Not sure how to go about it. If we could setup a route in the application that tests this functionality we can setup an HTTP monitor for it by using our current monitoring solution.
This test would have to be coded as a rails app route.

jywarren:
Would the application trigger an email sent to a known address? Could we
monitor an inbox, maybe?

icarito:
We don't need to send an email. I've just manually [tested the SMTP protocol](https://community.spiceworks.com/how_to/11-test-email-flow-using-smtp-commands) and found that in case the relay isn't responding, it will fail immediately upon giving the sender and recipient:
![image](https://user-images.githubusercontent.com/199755/42858455-54124f3a-8a14-11e8-846c-54f12d659cc4.png) so it could be tested continually every few minutes or so (via sidekiq I guess) *without* actually emailing.

The tricky thing is, what should the system do in case it finds it is unable to send mail? Tweet? Notify `#publiclab`? We could start with a `/api/status` route that tests the email relay and gives a 500 error if it fails. Then we could monitor this route with the Rackspace monitoring tools, which emit alerts if it fails.

jywarren:
I like this last way - do you think you can open an issue for that?
Maybe @ViditChitkara could help code that up?

jywarren:
i think in ruby you can execute commandline things by wrapping a command in ` characters, OR

by running `system('ls')`

icarito:
Hi,
[...]

PR #3330, opened 2018-09-14
SMTP test start

jywarren:
@icarito -- here's a start! #3040

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -352,4 +352,31 @@ class AdminController < ApplicationController
       redirect_to '/dashboard'
     end
   end
+
+  def smtp_test
+    require 'socket'
+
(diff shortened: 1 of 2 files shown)


icarito:
Thanks for the start!

icarito:
We need to get SMTP settings from rails configuration: `config.action_mailer.smtp_settings` (from `config/environments/production`).

icarito:
I'll test it in the interactive console. Stack overflow [says](https://stackoverflow.com/questions/1916193/how-to-read-ruby-on-rails-config-values-from-within-the-application-controller) it's `ActionController::Base.session_options[:domain]`

jywarren:
This only had Code Climate recommendations blocking it. It looks good now! Do you think it's ready to merge?

jywarren:
Merged!!!

PR #3445, opened 2018-09-25
Profiles method only join nodes where(node_revisions.status = 1)

### LINKED ISSUE #3443: Optimize users query in Srch API
Looks like we're seeing some site slowdowns due to a search query:

```
Srch::Search [GET] srch/all
Response Time | 35,831ms
```

```SELECT  DISTINCT `rusers`.* FROM `rusers` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (MATCH(bio, username) AGAINST(? IN BOOLEAN MODE)) AND (rusers.status = ?) ORDER BY node_revisions.timestamp DESC LIMIT ?```

You can see it appearing in Skylight: https://oss.skylight.io/app/applications/GZDPChmcfm1Q/1536553680/12h/endpoints

Can we optimize this more? Could an index like this: https://github.com/publiclab/plots2/pull/3380/files

or is there something else we can do here? I'm seeing worst-case load times of 70-90 seconds... pretty bad! But I'm sure we can figure something out 😄 

@milaaraujo @stefannibrasil what do you think?

milaaraujo:
fixes #3443

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -36,6 +36,7 @@ class SearchService
     user_scope =
       if search_criteria.sort_by == "recent"
         user_scope.joins(:revisions)
+        .where("node_revisions.status = 1")
         .order("node_revisions.timestamp #{search_criteria.order_direction}")
         .distinct
       else


jywarren:
🎉

PR #3304, opened 2018-09-07
Search ui improvement

stefannibrasil:
Moved from #3295

PR #3474, opened 2018-09-25
get navbar back to how it was

stefannibrasil:
moving back nav bar as it was before this PR  #3304

### DIFF
--- a/app/assets/stylesheets/style.css
+++ b/app/assets/stylesheets/style.css
@@ -378,6 +378,40 @@ div.note.moderated h4 {
   padding-top: 7.5px;
 }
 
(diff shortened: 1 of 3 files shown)


jywarren:
Hi, do you think this is the easiest way forward for some of the resizing issues we're seeing? I do like the new search box. Could we do something hybrid, like maybe preserve the search box and keep it to the left? 

Also i just feel like the new design was so nice even though it didn't resize the same way... happy to discuss, and thanks!

stefannibrasil:
I think so, I should have asked first. But I can leave the new searchbar to the left and leave the rest as it was.

PR #3483, opened 2018-09-25
Add doc for running plots2 with MySQL

stefannibrasil:
This is a doc improvement to help anyone that might have problems like this #3170 so they can run plots2 with MySQL on both development and test environments.

### DIFF
--- a/doc/TESTING.md
+++ b/doc/TESTING.md
@@ -72,12 +72,12 @@ already using it.
 ## How to run plots2 with MySQL on development and test environments
 
 In development and test environments, the project uses SQLite3, but in production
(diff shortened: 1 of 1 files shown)


jywarren:
This is great! Anything you wanted to say about `mariadb` too?

stefannibrasil:
not really, do you have something? I can add here :)

I installed `mariadb` on my pc because I use fedora and it was easier, but I don't really know what to put there about this :D any help would be appreciated :D

jywarren:
ah, i think we can probably just say you can use either MySQL or MariaDB,
near the top?

stefannibrasil:
done :+1: :shipit:

jywarren:
Awesome 👍

PR #3491, opened 2018-09-26
Navbar runs out of space

### LINKED ISSUE #3464: Header navbar runs out of space and covers content
Our new header is very nice, but it needs to collapse a bit earlier; we use Bootstrap: https://getbootstrap.com/docs/3.3/

screen shot 2018-09-25 at 12 07 09 pm

We should have it collapse into the "minimal" look earlier:

screen shot 2018-09-25 at 12 11 13 pm

The code is here: https://github.com/publiclab/plots2/blob/master/app/views/layouts/_header.html.erb

This will take some testing with expanding and contracting the browser window. Please upload some screenshots of various states so we know it collapses smoothly! An animated GIF would be extra nice!

Thanks!

DamonClark:
This pull request is a proposed solution for fixing issue: Fixes #3464

It includes a css media rule that will collapse the nav-right into hamburger at 990 pixels.

navbar_fix

### DIFF
--- a/app/assets/stylesheets/style.css
+++ b/app/assets/stylesheets/style.css
@@ -113,6 +113,43 @@ a .fa-white,
   }
 }
 
+@media (max-width: 990px) {
+    .navbar-header {
+        float: none;
+    }
+    .navbar-left,.navbar-right {
+        float: none !important;
+    }
+    .navbar-toggle {
+        display: block;
+    }
+    .navbar-collapse {
+        border-top: 1px solid transparent;
+        box-shadow: inset 0 1px 0 rgba(255,255,255,0.1);
+    }
+    .navbar-fixed-top {
+        top: 0;
+        border-width: 0 0 1px;
+    }
+    .navbar-collapse.collapse {
+        display: none!important;
+    }
+    .navbar-nav {
+        float: none!important;
+        margin-top: 7.5px;
+    }
+    .navbar-nav>li {
+        float: none;
+    }
+    .navbar-nav>li>a {
+        padding-top: 10px;
+        padding-bottom: 10px;
+    }
+    .collapse.in{
+        display:block !important;
+    }
+}
+
 body { padding-top: 70px; }
 
 #header .collapse.in .hidden-xs,
--- a/app/assets/stylesheets/style.css
+++ b/app/assets/stylesheets/style.css
@@ -113,7 +113,7 @@ a .fa-white,
   }
 }
 
-@media (max-width: 1200px) {
+@media (max-width: 768px) {
     .navbar-header {
         float: none;
     }
@@ -147,6 +147,9 @@ a .fa-white,
     }
     .collapse.in{
         display:block !important;
+    }
+    #md-ham {
+        display: none;
     }
 }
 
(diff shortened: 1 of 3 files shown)


jywarren:
oh this is great... would you mind sharing a gif of it running? This is a
nice gif app: https://www.cockos.com/licecap/

Thanks!

avsingh999:
@DamonClark please provide screenshot/gif

DamonClark:
I am glad you asked for a gif because I realized that when a user is logged in. The nav-bar needs to collapse into a hamburger at 1200 pixels instead of what I originally had at 900. So I committed another small change to fix this and here is the gif you request below:

navbar_fix_1200

jywarren:
Ah, this is great! But I wonder if there are other possible "break points" such that we see a few more menu items even on medium screens. What do you think? 

Thanks!!!

DamonClark:
With this proposed change the navbar menu items will all be combined into a hamburger when the screen is less than 1200 pixels wide. I checked all different size screens for tablets and mobile devices, they all look good. I think this will definitely solve the problem of the navbar running out of space and covering content.

DamonClark:
The only negative I can see with adding this change is that your going to see a hamburger allot more often. Even when the screen size width is relatively large.

jywarren:
Ah, yes. You know, I guess I'm interested in whether we could look at this previous version and have not /everything/ collapse down, but just some things:

https://github.com/publiclab/plots2/blob/c9717a7675033af8eed9292e5dcf64dac5e5cb98/app/views/layouts/_header.html.erb

In previous versions, we had the "collapse to hamburger" moment at like 768px or so, and at larger sizes, we selectively hid menu items one at a time at different break points. I know this is a bit more complex to implement, but I think it'll result in a good solution to the problem you pointed out -- that we don't suddenly lose all the menu items at lower than 1200px wide. 

Would you be interested in giving this a try? Thank you!!!!

DamonClark:
Sure I can give that a try!

jywarren:
thank you!

DamonClark:
Ok so this time I was able to do as you suggested and hide different menu items at different breakpoints. Unfortunately I had to shorten the search bar from 280px to 220px because three of the menu items, "get involved", "about us" and "profile" contain their own drop-downs. I wasn't able to add any of these items into the custom drop-down until after 768px. It definitely looks much better now, let me know what you think.

navbar_fix_breakpoints

jywarren:
I LOVE THIS

thank you so much. Is this ready to be merged in, then? THANK YOU!!!!!!!!

DamonClark:
Yup! It's ready to be merged. Thanks for providing the opportunity to work on this issue.

jywarren:
Fantastic. This was sorely needed and you've done a tremendous and thorough job. We'd love your help with another issue if you're interested! Some of the maps (may be documented in the Geographic Features milestones) are seeing some user interface quirks that could likely use some help! Or we're happy to help you find another you find interesting. 👍

DamonClark:
I would definitely be interested in working on another issue. I looked at the Geographic Features milestone but didn't see any user interface related issues. I am happy to work on one though or any other issue. Are you able to point me to some open issues that need worked on. I could also look at the maps section of the app to find and create an issue myself if needed.

PR #3508, opened 2018-09-27
Update api docs

stefannibrasil:
Update the API docs after all the changes for the past months! :)

PR #3519, opened 2018-09-27
Change API params to ruby standard

stefannibrasil:
After writing the #3508 API docs yesterday, we noticed that we could rename some params from the API so they use the Ruby naming convention. And also, `srchString` was not a good name because some params aren't a string and that could lead to some confusion in the future.

We searched through the whole app to change the API calls. But we were wondering if other repositories also use the API? @jywarren


jywarren:
🎉🤯😁

Issue #3549, opened 2018-10-01
Stored XSS in comments section https://publiclab.org/questions


vsk4:
### Please describe the problem (or idea)

 The comment section is vulnerable to  stored - XSS 

> What happened just before the problem occurred? Or what problem could this idea solve?

I tried to comment in [here](https://publiclab.org/questions/) and pasted simple XSS payload and uploaded the comment, after refreshing the page it got executed.
While taking the input for the comment, filtering of input does not happen so that XSS script is executed.

> What did you expect to see that you didn't?

Filtering of script and no pop up when a payload(testing script) is applied.
See the deadly effects of Stored XSS [here](https://resources.infosecinstitute.com/deadly-consequences-xss/#gref)

### Please show us where to look

click on this [link](https://publiclab.org/questions/apparao/09-29-2018/with-which-android-tablets-is-the-mobius-action-cam-compatible) to see the Stored -XSS
Screenshots :- 
![screen shot 2018-10-01 at 8 40 36 am](https://user-images.githubusercontent.com/23310407/46268022-ca328280-c555-11e8-8713-314c1ed59989.png)
![screen shot 2018-10-01 at 8 40 15 am](https://user-images.githubusercontent.com/23310407/46268023-ca328280-c555-11e8-83b1-354f2ab1b038.png)

https://publiclab.org/questions

### What's your PublicLab.org username?
herovsk4
> This can help us diagnose the issue: 

please have a look at this [link](https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)) for more information on XSS.
Filter the input containing Script tags like <,>-,% and all other payload inputs. 

### Browser, version, and operating system
Safari, chrome and firefox

> Many bugs are related to these -- please help us track it down and reproduce what you're seeing!

****

## Thank you!

Your help makes Public Lab better! We *deeply* appreciate your helping refine and improve this site. 

To learn how to write really great issues, which increases the chances they'll be resolved, see:

https://publiclab.org/wiki/developers#Contributing+for+non-coders

jywarren:
Hi! Thank you! I'm going to move your comment to our staging server so it doesn't impact real people posting on PublicLab.org:

https://publiclab.org/questions/apparao/09-29-2018/with-which-android-tablets-is-the-mobius-action-cam-compatible#c20756

It's now at: https://stable.publiclab.org/notes/bkleist/04-04-2017/uwec-air-quality-monitoring#c17379

The offending string was:

[...]

PR #3553, opened 2018-10-01
Testing comment sanitizing -- help wanted!

jywarren:
This should fail, then we should fix it!

This demonstrates a vulnerability documented in #3549 -- that issue now should build on this one to eliminate the vulnerability. Thanks!


jywarren:
```
/app/test/unit/comment_test.rb:2:in `<top (required)>': uninitialized constant ActionView::Helpers::ApplicationHelper (NameError)
```

But also now #3555 has moved it so we can use `comment.render_body` instead in the test!

gauravano:
@jywarren please update the status of this PR. Thanks!

jywarren:
This needs rebasing, then we need to add `ApplicationController.helpers.sanitize()` to this line:

https://github.com/publiclab/plots2/blob/a95ba1f4cba03afc3f94b6c874793411d7dcb08a/app/models/comment.rb#L478

However we'll want to allow (whitelist) a fairly wide range of attributes following documentation here:

https://guides.rubyonrails.org/security.html#html-javascript-injection-countermeasures

Here are some sources for whitelists of HTML attributes:

https://help.manuscript.com/12064/embedding-html-in-wiki-articles

https://github.com/rickpmartin/sanitize-html#what-are-the-default-options

I would guess at least:

`href, class, id, width, height, border, src, placeholder, title` and quite a few more...

SidharthBansal:
@jywarren we can include this in GCI if you want
cc\ @VladimirMikulic @Uzay-G 
GCI students are advised to focus on the tasks on which they are working. Don't get overwhelmed by the emails.

jywarren:
That would be great, thank you!

SidharthBansal:
Who is willing to grab this issue?

VladimirMikulic:
I and @Uzay-G will come up with something.

SidharthBansal:
Oh you both are distributing issues at Gitter. I forget. Sorry

VladimirMikulic:
:+1:

VladimirMikulic:
This can be closed. #7282 resolved this.

PR #3555, opened 2018-10-01
Moved render_comment_body helper into Comment model.

### LINKED ISSUE #3552: Move comment_render_body helper into Comment model

****

https://github.com/publiclab/plots2/blob/6595ef39f25ca4f2d377cacfd92e86a7460214d6/app/helpers/application_helper.rb#L100-L107

Let's make this a method of Comment - in https://github.com/publiclab/plots2/blob/master/app/models/comment.rb - like:

```
def render_body

end
```

And remove it from `application_helper.rb`!

We'll also need to change this line:

https://github.com/publiclab/plots2/blob/6595ef39f25ca4f2d377cacfd92e86a7460214d6/app/views/notes/_comment.html.erb#L120

To `comment.render_body`

Finally, the `sanitize comment body` test in https://github.com/publiclab/plots2/blob/master/test/unit/comment_test.rb could be updated to test `comment.render_body` instead of reproducing the code from the helper.

Glad to help with this one! Could be a first-timers issue!

mauricetmeyer:
Fixes #3552

Thanks!

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -98,15 +98,6 @@ module ApplicationHelper
   end
 
   # we should move this to the Comment model:
-  # returns the comment body which is to be shown in the comments section
-  def render_comment_body(comment)
-    raw RDiscount.new(
-      title_suggestion(comment),
-      :autolink
-    ).to_html
-  end
-
-  # we should move this to the Comment model:
   # replaces inline title suggestion(e.g: {New Title}) with the required link to change the title
   def title_suggestion(comment)
     comment.body.gsub(/\[propose:title\](.*?)\[\/propose\]/) do
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -333,7 +333,7 @@ class Comment < ApplicationRecord
   end
 
   def render_body
-    raw RDiscount.new(
+    RDiscount.new(
       title_suggestion(self),
       :autolink
     ).to_html
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -332,7 +332,6 @@ class Comment < ApplicationRecord
     comment.include?(COMMENT_FILTER)
   end
 
-
   def render_body
     raw RDiscount.new(
       title_suggestion(self),
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -331,4 +331,12 @@ class Comment < ApplicationRecord
   def trimmed_content?
     comment.include?(COMMENT_FILTER)
   end
+
+
+  def render_body
+    raw RDiscount.new(
+      title_suggestion(self),
+      :autolink
+    ).to_html
+  end
 end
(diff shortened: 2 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -97,15 +97,6 @@ def render_map(lat, lon)
...
-  def render_comment_body(comment)
-    raw RDiscount.new(
-      title_suggestion(comment),
-      :autolink
-    ).to_html
-  end
-
   # we should move this to the Comment model:
   # replaces inline title suggestion(e.g: {New Title}) with the required link to change the title
   def title_suggestion(comment)
```
jywarren: I'm only a little worried that we will still need this filter, and not sure if it's available inside the comment.rb file. Would you mind starting up the app and seeing on any note page (say, click any post shown on http://localhost:3000/research) to ensure that it's getting properly displayed? 

I /think/ tests ought to trip if this did go wrong, but a screenshot of a note page would be comforting!

Thank you!!

jywarren: Another possibility is that if comments DO have trouble, we could ALSO move the `title_suggestion` method into comment.rb:

https://github.com/publiclab/plots2/blob/ee5a7b120cc07bf678df480c6f404cc96ee09d04/app/helpers/application_helper.rb#L100-L117

What do you think?

mauricetmeyer: I was actually thinking about moving title_suggestions too, but I thought it would be better to stay in the context of the issue, as I don't know much about the codebase yet. I'll startup the app in a sec and report back.

mauricetmeyer: I had to remove the ```raw``` statement from ```def render_body```other than that it seems to work, also attached a screenshot. A commit will follow shortly.

screenshot 2018-10-03 at 01 34 36

If you would like me to still move ```def title_suggestion```I can do that.

As a side note ```find . -name '*.rb' | xargs grep 'render_comment_body'```returned 0 results, so it isn't referenced anywhere else.

jywarren: Now that i think of it, I believe we need the `raw` segment for some JavaScript work we do to scrape the original markdown. Do you think you could add it back in? 

Good to hear about the grep. Thanks!

mauricetmeyer: @jywarren ```raw``` isn't available from a model, only from a controller.

jywarren:
oh hmmm

sorry, i had thought this was in the view, i'm sorry i'm a bit out of sync
here!

I guess maybe this `raw` doesn't matter?

I wonder why it was added in the first place. Can you check if inline html
like `<a class="btn btn-primary">Hello</a>` or a YouTube embed is scrubbed
out? Maybe we need a test for that to be extra sure. What do you think?

app/helpers/application_helper.rb
mauricetmeyer: No problem, embedded HTML works (```<a class="btn btn-primary">Hello</a>```). Not sure about a youtube video, as I don't know how one does embed it, but I think that the ```raw```statement probably was redundant.

jywarren:
Hi -- this looks perfect; merging now. 👍 🎉  If you're looking for a new challenge, please take a look at https://code.publiclab.org#r=all; we prefer to leave `first-timers-only` issues for complete newcomers, but there are plenty of others we need help with! Thanks!

PR #3580, opened 2018-10-03
Map full screen support removed and Hash removes as well .

### LINKED ISSUE #3574: Inline leaflet maps show URL hash collisions when more than one are on a page
https://publiclab.org/wiki/inline-maps shows this @sagarpreet-chadha - is there a version update we need to resolve this?

They are generated in this file:

https://github.com/publiclab/plots2/blob/master/app/views/map/_leaflet.html.erb

Do we have to disable the hashes, or just update our version number in package.json?

https://github.com/publiclab/plots2/tree/master/package.json

Thanks!

sagarpreet-chadha:
Fixes #3574 !

So this PR :  

1.) Removes Full-screen support (for now) from **peopleLeaflet** and **leaflet** partials .
2.) Removes Hash as well .

### DIFF
--- a/app/views/map/_leaflet.html.erb
+++ b/app/views/map/_leaflet.html.erb
@@ -26,7 +26,7 @@
     var map_lat = <%= lat %> ;
     var map_lon = <%= lon %> ;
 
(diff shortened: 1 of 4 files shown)


jywarren:
🎉

Issue #3390, opened 2018-09-22
[Facebook Like] allowed on question while logged in from Google Auth


YashishDua:
The user is allowed to Facebook Like the question when logged in from Google Auth and even count of likes updates.

![screen shot 2018-09-22 at 9 49 23 pm](https://user-images.githubusercontent.com/20666950/45919381-10327b00-beb2-11e8-93d4-0379132f3f67.png)

![screen shot 2018-09-22 at 9 49 37 pm](https://user-images.githubusercontent.com/20666950/45919390-1b85a680-beb2-11e8-87c5-6657c569ab73.png)

Two cases:
1. If this like button counts itself irrespective of auth:
    No point of having it since we already have a start button on the top right for liking
2. If this like button uses facebook Auth:
    It actually is not using the auth service.

Example Scenario: https://publiclab.org/questions/TK_Siddharth/09-20-2018/what-would-be-the-day-to-day-use-of-spectroscopy

SidharthBansal:
Are you looking forward to be a Google Code in mentor @avsingh999 ?
If so please fill the form https://docs.google.com/forms/d/e/1FAIpQLScSBS-ddZN2H-lviUJbKlbV2VwP21fdIutXOCBzigRhmXJybw/viewform

YashishDua:
Alright, and I am taking up this issue @SidharthBansal

YashishDua:
Things we can do:
1. Allow facebook like only if logged in with Facebook Auth
2. Replace Facebook like with Facebook Share
3. Remove this button, since we already have a feature to like the question.
@gauravano @SidharthBansal 
How should we take this further?

SidharthBansal:
Hi, you can work on it but please wait for jeff input.
We already have the facebook and twitter like buttons long before OAuth was introduced. I am seeing these buttons since last 3 years. Oauth is introduced since last 2 months.
When you click on like button then the PL post is posted on fb. This is the current functionality, I guess.

Now @YashishDua wants to connect it to OAuth. Tell me if I am wrong @YashishDua. It will be great if you will connect both of them and allow users to post to their facebook accounts

SidharthBansal:
@jywarren I want your input here before he starts working.
Whenever you are free please tell what you think we should do here.

YashishDua:
Clicking the facebook like just increments the counter, there is no posting.
Preferably, Either we should make this a share button or remove.

YashishDua:
[...]

Issue #3384, opened 2018-09-21
Ensure OAuth login preserves `return_to` parameter and goes there after logging in


jywarren:
In our regular signup process at https://publiclab.org/signup, we can set the destination AFTER signup using a `return_to` parameter, like this:

https://publiclab.org/signup?return_to=/post?title=My Title

This nice URL can even set a title on the post you are prompted to create! It works with this hidden input:

`<input type="hidden" name="return_to" value="/post?title=My Title" />` (code here: https://github.com/publiclab/plots2/blob/cdef0bd7978836212c112f2295b4561d3f6a57f0/app/views/user_sessions/new.html.erb#L35)

We even have a test for this:

https://github.com/publiclab/plots2/blob/a8e9ffaa34319a3c7924ff735dc5231cf69b05d9/test/integration/login_flow_test.rb#L33-L41

Here's the controller logic that drives that redirect after account creation:

https://github.com/publiclab/plots2/blob/010bb27741625269e8c73d68bc65dcb7325b88a2/app/controllers/user_sessions_controller.rb#L108-L121

But for OAuth login (using Twitter, FB, Google logins) this doesn't seem to work. 

I got the message on this line - but it didn't attempt to redirect me: 

https://github.com/publiclab/plots2/blob/010bb27741625269e8c73d68bc65dcb7325b88a2/app/controllers/user_sessions_controller.rb#L41

I think we should look in the `session[:return_to]` parameter to see if it's there -- and we may need to store it there so it persists past the redirects through the external OmniAuth service. 

I'd love some help with this one! I don't know if it'd be a quick one @SidharthBansal  but interested in your thoughts!

jywarren:
Hi @SidharthBansal -- i know you're busy with Code-In and Outreachy, but I'm increasingly concerned that people on our site are losing their place when the log in with OAuth, and I want to ensure they get redirected back to their original page after login. I think we can use the session to do this, and I opened a starter PR to get this process going. 

You know the OAuth code the best, though... what do you think about this approach?

https://github.com/publiclab/plots2/pull/3762

SidharthBansal:
Hi Jeff, I will be happy to help here too asap.
Right now a lot of things are going on and I want to give some time in analysing the wide range of Login related issues created by you. I don't want to hurry up things. Authentication is a really crucial part for running production website. So, I will try to take out 3-4 hours on Saturday for these issues.
Please ping me on Sunday if I forget to do so. 
[...]

PR #3613, opened 2018-10-04
Facebook share button added

YashishDua:
Solves #3390 

### DIFF
--- a/app/views/home/_social.html.erb
+++ b/app/views/home/_social.html.erb
@@ -4,7 +4,7 @@
   <a class="twitter-share-button" href="https://twitter.com/share" style="width: 60px;height: 20px;bottom: 7px;">Tweet</a>
   <script>
     !function(d,s,id){
-        var js,fjs=d.getElementsByTagName(s)[0];
+      var js,fjs=d.getElementsByTagName(s)[0];
       if(d.getElementById(id)) return;
       js=d.createElement(s); js.id=id;
       js.src="//platform.twitter.com/widgets.js";
--- a/app/views/home/_social.html.erb
+++ b/app/views/home/_social.html.erb
@@ -1,7 +1,22 @@
 <div id="social">
   <% if Rails.env == "production" %>
-  <div class="fb-like" data-send="false" data-layout="button_count" data-width="50" data-show-faces="false" style="vertical-align: middle; bottom: 6px;margin-right:10px" ></div>
-  <a style="width: 60px;height: 20px;bottom: 7px;" href="https://twitter.com/share" class="twitter-share-button">Tweet</a>
-  <script>!function(d,s,id){var js,fjs=d.getElementsByTagName(s)[0];if(!d.getElementById(id)){js=d.createElement(s);js.id=id;js.src="//platform.twitter.com/widgets.js";fjs.parentNode.insertBefore(js,fjs);}}(document,"script","twitter-wjs");</script>
+  <div class="fb-share-button" data-href="https://publiclab.org<%= path %>" data-layout="button_count" style="vertical-align: middle; bottom: 6px;margin-right:10px">Fb Share</div>
+  <a class="twitter-share-button" href="https://twitter.com/share" style="width: 60px;height: 20px;bottom: 7px;">Tweet</a>
+  <script>
+    !function(d,s,id){
+        var js,fjs=d.getElementsByTagName(s)[0];
+      if(d.getElementById(id)) return;
+      js=d.createElement(s); js.id=id;
+      js.src="//platform.twitter.com/widgets.js";
+      fjs.parentNode.insertBefore(js,fjs);
+    }(document,"script","twitter-wjs");
+    !function(d, s, id) {
+      var js, fjs = d.getElementsByTagName(s)[0];
+      if (d.getElementById(id)) return;
+      js = d.createElement(s); js.id = id;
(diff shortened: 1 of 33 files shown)


YashishDua:
@SidharthBansal Require your help here!
I am not able to add a description to facebook share when I am using their JS Share API.
However, if I explicitly handle the click, I might be able to do this.
What should I do?

SidharthBansal:
This issue is not a simple one.
First of all you need to understand the OAuth project which I did with Jeff
as Gsoc project. Then you need to consider the different corner cases. Then
you need to understand the omniauth gem.

Then you need to think about all the providers. Then you need to understand
plots2 authentication system
 We don't use devise or any other gem. We have our own system. So, this
project will obviously take time and a lot of work.

SidharthBansal:
I think the purpose now becomes to sign in the person with the provider x on which we need to post. Then to post the content.
There is requirements to link oauth and posting service
See koala gem.

jywarren:
I guess it would be great if the share button worked whether you are Oauth
connected or not. Sidharth, what do you think? Many users do have twitter,
but haven't connected via Oauth. Thanks!

jywarren:
oh i see -- sidharth, you're saying you'd like to intercept the request and
require them to link via OAuth. I can see that being benficial, but some
people may refuse -- suspicious, or whatever. Maybe we could make a "Link
your FB account" link appear on hover, so that the two are prompted in the
same place, but it's not /required/?

SidharthBansal:
Whenever we post something to twitter or Facebook we can let the user sign
up first. People will just need to login to the twitter in the foreground.
In the background we will create a pl account for them.
If someone wants to share a post, that means they know what content is
written in the post and want to be a long lasting member.
It seems kind of weird behaviour to me to get the articles posted by
clicking on buttons without sign up process.
Reason: I clicked on twitter button to share. But I don't have any twitter
account. May be a person has two twitter accounts. If I have a twitter
signed up pl account then I can assume where the post will go. Otherwise
its ambiguous behaviour according to me where the post goes.
This is the reason I want that oauth and posting services to be combined.
People do need to sign up before posting the stuff. So we just need to link
two separate entities.
This is my opinion. May be I am forgetting something?

jywarren:
> If someone wants to share a post, that means they know what content is written in the post and want to be a long lasting member.

I think we all want this to be true. But we do have to admit to ourselves that there may be people who just browse our site and want to share it on twitter... and as they do, they may not be outgoing and want to join at first, but they may more slowly decide to be part of the community. I think we should be aware that this can happen and not make it a hard requirement to sign up, but allow multiple paths and styles. 

However, i think we should do everything we can to /persuade/ people to join! And to make that process as easy as possible. One thing I'd like to work on a bit is ensuring that when people sign in with Oauth -- say, at the moment they want to subscribe, or leave a comment -- that they go through the OAuth process and end up back where they first were, in https://github.com/publiclab/plots2/issues/3384

The other thing we can do is, after the Share on Facebook action happens, we could display a "Link your FB" text back on the original PL page (hopefully the share on facebook happened in a new tab, right?) and then we have the opportunity to engage them at that moment. 

Thanks so much everyone for thinking through this and I know @SidharthBansal has the community's growth and health at heart ❤️  I hope this makes sense and we can agree to move forward on this! Great teamwork and thank you again! 🙌

SidharthBansal:
Yeah we can agree on this. For sure.
Thanks a lot for quick reply Jeff.

SidharthBansal:
Another suggestions hope you all like it.
A person has pl account linked with the facebook. So we don't need to show
them the button to link it to Facebook.

jywarren:
That's great. So, shall we break this into two issues? The first is this one - to modify and update the FB button itself. Once merged, we can post a follow-up issue to suggest linking with facebook. But that follow-up can also have a check to see if linking with FB is already done, which would hide the notice. 

@YashishDua can you post a screenshot of the changes in this PR? And once it's complete, would you like to open a follow-up issue about linking your account to FB?

Thanks, everyone!

SidharthBansal:
Great Jeff. I agree with you.

app/views/home/_social.html.erb
```diff
@@ -1,7 +1,22 @@
 <div id="social">
   <% if Rails.env == "production" %>
-  <div class="fb-like" data-send="false" data-layout="button_count" data-width="50" data-show-faces="false" style="vertical-align: middle; bottom: 6px;margin-right:10px" ></div>
-  <a style="width: 60px;height: 20px;bottom: 7px;" href="https://twitter.com/share" class="twitter-share-button">Tweet</a>
-  <script>!function(d,s,id){var js,fjs=d.getElementsByTagName(s)[0];if(!d.getElementById(id)){js=d.createElement(s);js.id=id;js.src="//platform.twitter.com/widgets.js";fjs.parentNode.insertBefore(js,fjs);}}(document,"script","twitter-wjs");</script>
+  <div class="fb-share-button" data-href="https://publiclab.org<%= path %>" data-layout="button_count" style="vertical-align: middle; bottom: 6px;margin-right:10px">Fb Share</div>
```
jywarren: Could we make some code to ensure there is a fall-back? So, perhaps we look to see if there's a @node defined, and use that, but if nothing is provided in `@node` or `path` we display nothing? Thanks!

SidharthBansal:
@YashishDua are you stuck somewhere?
We could help you if you want

SidharthBansal:
cc @jywarren @gauravano

PR #3700, opened 2018-10-15
Update docs and add tests for the taglocations API method - fix #3597

### LINKED ISSUE #3597: Inline map API calls a bit broken?
@milaaraujo @stefannibrasil I think you can probably offer some quick input here -- it looks like some of the calls going from different types of inline maps are malformed... could you help us untangle what they should be and we can slot in some fixes?

Thanks! Probably similar to, or related to, #3573 

image

stefannibrasil:
Fixes #3597

The API docs were showing a wrong param format for the `taglocations` endpoint. After investigating, I believe these changes attempt to fix that.

### DIFF
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -243,7 +243,8 @@ module Srch
               doc_title: model.title,
               score: model.answers.length,
               latitude: model.lat,
-              longitude: model.lon
+              longitude: model.lon,
+              blurred: model.blurred?
             )
           end
           DocList.new(docs, search_request)
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -244,7 +244,7 @@ module Srch
               score: model.answers.length,
               latitude: model.lat,
               longitude: model.lon,
-              blurred: model.blurred?
+              blurred: false
             )
           end
           DocList.new(docs, search_request)
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -243,8 +243,7 @@ module Srch
               doc_title: model.title,
               score: model.answers.length,
               latitude: model.lat,
-              longitude: model.lon,
-              blurred: false
+              longitude: model.lon
             )
           end
           DocList.new(docs, search_request)
--- a/doc/API.md
+++ b/doc/API.md
@@ -83,12 +83,14 @@ All the endpoints have the optional parameter `limit` (10 by default) where you 
 
 ### TagLocations:
 
-* **URL**:  `https://publiclab.org/api/srch/taglocations?query=18,-66`
+* **URL**:  `https://publiclab.org/api/srch/taglocations?query=18.00,-66.00`
 * **URL Params** :
 
   **Required:**
 
-  `query=[coordinates]`: search notes from users located near the query separated by `,`.
+  `query=[coordinates]`: search notes from users located near the query separated by `,`
+  and with at least one digit after the '.'. Taking a lat value as example, `lat:18.0`
+  or `lat:18.00` are valid but `lat:18` is not.
 
   **Optional:**
 
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -534,6 +534,10 @@ class Node < ActiveRecord::Base
     drupal_content_type_map.last
   end
 
+  def blurred?
+    has_power_tag('location') && power_tag('location') == "blurred"
+  end
+
   def lat
     if has_power_tag('lat')
       power_tag('lat').to_f
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -197,7 +197,7 @@ class User < ActiveRecord::Base
 
   def get_value_of_power_tag(key)
     tname = user_tags.where('value LIKE ?', key + ':%')
-    tvalue = tname.first.name.partition(':').last
+    tvalue = tname.first.name.partition(':').last if tname.present?
     tvalue
   end
 
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -82,11 +82,13 @@ class SearchService
   end
 
   # Search nearby nodes with respect to given latitude, longitute and tags
-  # and package up as a DocResult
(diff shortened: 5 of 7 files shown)


stefannibrasil:
@publiclab/reviewers @jywarren is that ok pass this codeclimate issue for later? I just added more validations to the method... or if you have any tips for this, would appreciate them. Thanks!

app/api/srch/search.rb
```diff
@@ -244,7 +244,7 @@ class Search < Grape::API
               score: model.answers.length,
               latitude: model.lat,
               longitude: model.lon,
-              blurred: model.blurred?
+              blurred: false
```
jywarren: ah, here, we do want to pass this `blurred?` value along because eventually we'll be displaying them differently using https://github.com/publiclab/leaflet-environmental-layers/ -- make sense? Thank you!

doc/API.md
```diff
@@ -83,12 +83,14 @@ All the endpoints have the optional parameter `limit` (10 by default) where you
...
 ### TagLocations:
 
-* **URL**:  `https://publiclab.org/api/srch/taglocations?query=18,-66`
+* **URL**:  `https://publiclab.org/api/srch/taglocations?query=18.00,-66.00`
 * **URL Params** :
 
   **Required:**
 
-  `query=[coordinates]`: search notes from users located near the query separated by `,`.
+  `query=[coordinates]`: search notes from users located near the query separated by `,`
```
jywarren: 👍

jywarren:
Also, sorry, I think you may have written this somewhere else but I'm a bit behind... is there a matching JS call which we have to tweak to conform to the `.00` format?

app/api/srch/search.rb
stefannibrasil: yes, make sense! To be fair, I don't even think this is the place for this. The `user` model already has an instance class to know that here:

https://github.com/publiclab/plots2/blob/1d5eeeb148cb87be251800b4c047f7c519fd15b3/app/models/user.rb#L204

So whenever it's necessary to know that, it's possible to just call that `user.blurred?` Can I remove it from this file? What do you think?

stefannibrasil:
hi, @jywarren thanks for the review!

I don't know about the JS call, to be honest. Was that already been done somewhere? I'm not aware...

jywarren:
Ah, here: 

https://github.com/publiclab/plots2/blob/79d65bf700214f828df7f76543d553f391fb0f37/app/views/map/_leaflet.html.erb#L37

Do we need to pad it or otherwise update it so it conforms to the API spec you've updated? Thanks!

stefannibrasil:
oh, I see! thanks for the link, I'll check it out this week!

jywarren:
Can't seem to get codeclimate to run, so gonna close and reopen! 

Also, i changed the JS query I'd mentioned above -- here! https://github.com/publiclab/plots2/pull/3940

app/api/srch/search.rb
jywarren: Ah, but so we actually will need to be able to know it's blurred status from a JavaScript call, not a Ruby call, which is why this is here. Make sense? Probably specifically the call I changed in https://github.com/publiclab/plots2/pull/3940

So once we preserve this we can merge it!

stefannibrasil:
hi, @jywarren thanks! I was going to take a look at it this holiday. Do you think this is ready, then? Thanks!

jywarren:
Ah, i think we just have to preserve the `blurred` attribute as mentioned above, and then it's good to go. Thanks! https://github.com/publiclab/plots2/pull/3700#discussion_r232426716

app/api/srch/search.rb
stefannibrasil: oh, yeah, I remember now why I removed this line.

https://github.com/publiclab/plots2/blob/master/app/services/search_service.rb#L109

You see, the way the method is defined, it only returns the nodes from users that don't have the blurred tag set. That was something that was already there, I just organized the code. That's why it was set to false before...

stefannibrasil: I don't know what's happening and I'm a little busy right now. Maybe I should close this since you already added the changes to the JS call?

jywarren: No problem at all, I think i can just make a small adjustment and get this passing -- actually i realized this sub thread of comments is referring to an earlier state, so that line deletion is not even present! Picking up below and I'll go ahead and get this resolved. Thanks for checking in, and apologies for bothering while you're busy, i just didn't want to neglect this one. 👍

jywarren:
I see the error:

```

ERROR["test_search_Tag_Nearby_Nodes_functionality_with_a_valid_query", #<Minitest::Reporters::Suite:0x000055902d927d00 @name="SearchApiTest">, 81.07464563299993]
 test_search_Tag_Nearby_Nodes_functionality_with_a_valid_query#SearchApiTest (81.07s)
NoMethodError:         NoMethodError: undefined method `blurred?' for #<Node:0x000055902d9d88f8>
            app/api/srch/search.rb:247:in `block (3 levels) in <class:Search>'
```

So, I think we need to add a `blurred?` method to nodes, like this one for users:

https://github.com/publiclab/plots2/blob/788ff6fb60d579089eb3f5935494c2e79ff0354a/app/models/user.rb#L206-L208

I'll do that now.

jywarren:
Cool, added that -- let's see: https://github.com/stefannibrasil/plots2/commit/75724f34bb01d2d02cd9f593f9fe5b66be1fc812

app/api/srch/search.rb
stefannibrasil: No need for you to apologize, sorry I couldn't finish this, but I have a lot going on right now and I can't separate a time to put everything together... hope it won't be too much problem. Cheers!

jywarren: Good luck with everything!!! fingers 🤞 and i think this commit will pass tests 👍

jywarren:
Yay!!!

app/api/srch/search.rb
stefannibrasil: awesome, thank you!

PR #3719, opened 2018-10-19
Search API: new endpoint /srch/nearbyPeople

milaaraujo:
As discussed here ([Question: What more data-layers can we show on map?](https://publiclab.org/questions/sagarpreet/07-01-2018/what-more-data-layers-can-we-show-on-map)) I implemented a new method to search nearby people with respect to given latitude, longitute and tag (optional). 

I'm not filtering by most recent users... That would make things a little bit more complicated! Do you think the way it is now this endpoint would be useful? Do you have suggestions?

@jywarren @stefannibrasil @sagarpreet-chadha @steviepubliclab (I hope I've mentioned the right people here! Usernames are different...)

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -133,7 +133,16 @@ class SearchService
     ids = user_locations.collect(&:id).uniq || []
 
     items = User.where('rusers.status <> 0').joins(:user_tags)
-                .where('rusers.id IN (?) AND value LIKE ?', ids, 'lon:' + lon[0..lon.length - 2] + '%').limit(limit)
+                .where('rusers.id IN (?) AND value LIKE ?', ids, 'lon:' + lon[0..lon.length - 2] + '%')
+    
+    # selects the items whose node_tags don't have the location:blurred tag
+    items.select do |item|
+      item.user_tags.none? do |user_tag|
+        user_tag.name == "location:blurred"
+      end
+    end
+      
+    items = items.limit(limit)
   end
 
   # Returns the location of people with most recent contributions.
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -133,16 +133,7 @@ class SearchService
     ids = user_locations.collect(&:id).uniq || []
 
     items = User.where('rusers.status <> 0').joins(:user_tags)
-                .where('rusers.id IN (?) AND value LIKE ?', ids, 'lon:' + lon[0..lon.length - 2] + '%')
-
-    # selects the items whose node_tags don't have the location:blurred tag
-    items.select do |item|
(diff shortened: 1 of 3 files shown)


app/services/search_service.rb
```diff
@@ -113,6 +113,38 @@ def tagNearbyNodes(query, tag, limit = 10)
...
+    if tag.present?
+      user_locations = User.joins(:user_tags)\
+                       .where('user_tags.value LIKE ?', tag)\
+                       .where(id: user_locations.select("rusers.id"))
+    end
+
+    ids = user_locations.collect(&:id).uniq || []
+
+    items = User.where('rusers.status <> 0').joins(:user_tags)
+                .where('rusers.id IN (?) AND value LIKE ?', ids, 'lon:' + lon[0..lon.length - 2] + '%')
```
sagarpreet-chadha: Hey 😄 !
How about using limit method over here ?
```suggestion
                .where('rusers.id IN (?) AND value LIKE ?', ids, 'lon:' + lon[0..lon.length - 2] + '%').limit(limit)
```

app/services/search_service.rb
```diff
@@ -113,6 +113,38 @@ def tagNearbyNodes(query, tag, limit = 10)
...
+
+    # selects the items whose node_tags don't have the location:blurred tag
+    items.select do |item|
+      item.user_tags.none? do |user_tag|
+        user_tag.name == "location:blurred"
+      end
+
+      user_locations.limit(limit)
```
sagarpreet-chadha: What is the purpose of this line ? 
This line seems to run again and again inside a loop , right ?

sagarpreet-chadha: I guess we can remove this line . What do you think 🤔 ? Thank You 😄 !

app/services/search_service.rb
milaaraujo: I want to use limit after removing users that have location:blurred tag. That makes sense? But I see I've added limit method in the wrong place! Can you take a look now?

To be honest I wasn't sure if I really need to filter users with location:blurred tag, I'm just doing because other methods are doing so. Can an user have lat/lon tags and location:blurred at the same time?

sagarpreet-chadha: Yes @milaaraujo , we can have lat/lon tags and location:blurred tags at the same time .
This signifies that a user wants to give a rough idea of his location rather than giving exact location , which is why we need to filter users with location:blurred tag .
Thank you !

jywarren:
Hi, I think this is ready and will merge, but I do think sorting by most recent sign up at least would be a good next step. Would one of you mind opening an issue for this? Thanks!! And great work here!

milaaraujo:
Yes, I can do it!

PR #3619, opened 2018-10-05
[WIP - help wanted] attempt at getting data-uri based image upload working

jywarren:
Guidance (not working): https://stackoverflow.com/questions/23675747/use-paperclip-for-saving-base64-images-obtained-from-an-api?noredirect=1&lq=1

https://github.com/thoughtbot/paperclip/blob/f384174392ce192c9d76bd447902fe8a3ecf70ad/README.md#io-adapters

From a note page, i'm trying (in javascript console):

```js
var token = $('meta[name="csrf-token"]').attr('content');
// sample image:
var data = "data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wBDAAYEBQYFBAYGBQYHBwYIChAKCgkJChQODwwQFxQYGBcUFhYaHSUfGhsjHBYWICwgIyYnKSopGR8tMC0oMCUoKSj/2wBDAQcHBwoIChMKChMoGhYaKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCj/wAARCAAQABADASIAAhEBAxEB/8QAFQABAQAAAAAAAAAAAAAAAAAAAAf/xAAUEAEAAAAAAAAAAAAAAAAAAAAA/8QAFQEBAQAAAAAAAAAAAAAAAAAABgj/xAAUEQEAAAAAAAAAAAAAAAAAAAAA/9oADAMBAAIRAxEAPwCdABykX//Z";

$.post('/images/create', { authenticity_token: token,  data: data, uid: 1, image: { title: '', notes: '' }  }, console.log);
```
But getting: 

```
Started POST "/images/create" for 127.0.0.1 at 2018-10-05 11:22:52 -0400
Processing by ImagesController#create as */*
[...]

PR #3722, opened 2018-10-19
added image-sequencer

jywarren:
@tech4GT adding this to test out ability to run it from the same hostname. Just merged #3619 which should allow data-uri based uploads.


tech4GT:
aha! Nice!!

PR #3723, opened 2018-10-20
Upgrade packages sources repositories

icarito:
I think it's a good idea to perform updates on container creation.


icarito:
I'm hitting this issue apparently when trying to deploy to staging:

https://unix.stackexchange.com/questions/371890/debian-the-repository-does-not-have-a-release-file#371907

jywarren:
Ah hmm, is this related to the `yarn` issues we'd seen?

icarito:
Not really but found it while looking to reproduce the yarn issue in staging.

PR #3762, opened 2018-10-24
[WIP - Help wanted!] Ensure preservation of return_to through OAuth login flow

### LINKED ISSUE #3384: Ensure OAuth login preserves `return_to` parameter and goes there after logging in

jywarren:
Fixes #3384

I think this is only the beginning -- i'm not setting the `session[:return_to]` in the right places, nor am I setting up the redirects thoroughly... and it could probably use some refinement to not have redundant code.

### DIFF
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -33,12 +33,20 @@ class UserSessionsController < ApplicationController
           redirect_to root_url, notice: "Already linked to another account!"
         end
       else # not signed in
+        session[:return_to] = params[:return_to] # ensure we preserve return_to even when passing through OAuth provider
         if @identity&.user.present?
           # The identity we found had a user associated with it so let's
           # just log them in here
           @user = @identity.user
           @user_session = UserSession.create(@identity.user)
-          redirect_to root_url, notice: "Signed in!"
+          if session[:return_to]
+            return_to = session[:return_to]
+            session[:return_to] = nil
+            flash[:notice] = "Signed in!"
+            redirect_to return_to
+          else
+            redirect_to root_url, notice: "Signed in!"
+          end
         else # identity does not exist so we need to either create a user with identity OR link identity to existing user
           if User.where(email: auth["info"]["email"]).empty?
             # Create a new user as email provided is not present in PL database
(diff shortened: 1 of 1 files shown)


Radhikadua123:
@jywarren Wow, this seems to be quite interesting issue. I was thinking about approach to it.

What about this: 

1. All the third party login icons will point to `/redirect?from=#{url}&to=/auth/#{provider}`.
2. In that GET view:
    2.1 if user is already logged in, redirect user to `from` url. (handling edge case)
    2.2 if user is not logged in, just get the existing session or create new one. In that session, save the `return_to` url and redirect user to `to` url from get parameters.
3. When callback from third party login is called, on success, we check if `return_to` is set in session or not. If it is set, then just redirect user to that url.

I'm not sure how well would it address the issue. I might have missed something.

jywarren:
That sounds like a great approach, although it did seem to me that we might
be able to set the session in the oath login route itself? But maybe I'm
mixed up. If you'd like to try this out, we'd love some help!!!

Radhikadua123:
Sure, I'll try it out tomorrow. Thanks!

jywarren:
Deferring to #3879

PR #3767, opened 2018-10-25
/api/srch/profiles: search by user tag 

### LINKED ISSUE #3506: Profiles search page: search by user tag
Right now, the Profiles option on the search page (https://publiclab.org/search/profiles/balloons/) is searching for users by username and bio.

In the endpoint, we would like to add the option to also search for user tags. [Here](https://github.com/publiclab/plots2/blob/master/app/api/srch/search.rb#L85) we already have the parameter/value `field=username`, it is used when we want to search only by username. One option would be to add `field=tag`.

milaaraujo:
Fixes #3506 

/api/srch/profiles: Adding the option to search for user tags.

`/api/srch/profiles?query=[user_name]&tag=[tag_name]`

### DIFF
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -88,7 +88,7 @@ module Srch
                                            nickname: 'search_profiles'
 
       params do
-        use :common, :sorting, :ordering, :field
+        use :common, :sorting, :ordering, :field, :additional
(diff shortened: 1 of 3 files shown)


jywarren:
⚡⚡⚡

PR #3769, opened 2018-10-25
del

### LINKED ISSUE #3340: Sign-up Form Validation message correction
**We are preparing to participate in Google Code-in, and have reserved this issue for participants in GCI - but we'd love to have your help with another one! Please check out https://code.publiclab.org to see more.**

# Problem

When a user tries to submit an empty sign up form, and also does not check off the recaptcha box, This output is shown: 

image

# Solution

Instead, we want all the relevant errors ("field cannot be empty" or so) to be listed along with the failed recaptcha verification. 

Each box in this problem is worth some points. We will accept answers that complete one part and not the other, but encourage you to go for all 3. Completion of all three will result in bonus points :)

- [ ] fix the `create` action so that the correct errors show
- [ ] fix the error messages themselves as well. If you see below, there are three different errors for email, but only one of them is correct. 
- [ ] write a test reflecting this fix. Guidelines: the test must fail without your changes, and your changes are what makes it pass.

image

thesparks:
del

### DIFF
--- a/app/views/users/_form.html.erb
+++ b/app/views/users/_form.html.erb
@@ -21,7 +21,7 @@
       <div class="form-group">
 
         <label for="username"><%= t('users._form.username') %></label>
-        <%= f.text_field :username, :required => true, { tabindex: 1, class: 'form-control', id: 'username'} %>
+        <%= f.text_field :username, { tabindex: 1, class: 'form-control', id: 'username'} %>
 
       </div>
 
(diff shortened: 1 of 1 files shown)


thesparks:
@sk9331657 @gauravano, please check

gauravano:
Hi @thesparks, as per GCI dashboard, you abandoned this task so we have to close this PR as some one else has claimed it. Thanks!

PR #3805, opened 2018-10-26
Allow admins to update profiles

jywarren:


Radhikadua123:
@jywarren I was thinking what if admin can not only change the profile, but mock a user ? Like we replace the user in the session(only allowed to admin). So, whole website will be accessed like that mocked user and to remove it from session, there would be a special end point.

jywarren:
Hi, that's pretty interesting, but I worry sometimes about creating very complex systems because they can become brittle -- hard to maintain! And while interesting, this one maybe isn't useful enough to merit the extra feature. 

Actually i really appreciate the challenging issues you've been taking on! I was wondering if you'd be interested in some of our priority issues, which are often a bit complex, and you seem so skilled at complex Rails issues!

https://github.com/publiclab/plots2/labels/priority

I know you're working on something now but we'd love your help with some of these if you're free later!

Radhikadua123:
> Hi, that's pretty interesting, but I worry sometimes about creating very complex systems because they can become brittle -- hard to maintain! And while interesting, this one maybe isn't useful enough to merit the extra feature.

Oh, I understand.

> I was wondering if you'd be interested in some of our priority issues, which are often a bit complex

Sure, I would love to check them out.

Thanks!

PR #3809, opened 2018-10-26
Update en.yml

### LINKED ISSUE #3515: [multiple-use issue!] Move a translation text into a centralized file

****

We're starting a big project to (#3262) revamp and refresh our internationalization systems, and one basic task we'd LOVE help with is to start moving translations around the site into a more centralized location t help make them easier to manage. 

It's pretty easy! Try this:

## Find a single translation

First, look through files in this set of folders: https://github.com/publiclab/plots2/blob/master/config/locales/

Dig into the subfolders and find an `en.yml` - full of English text

We want to move all the lines in these files into the main `en.yml` file at the root of these folders; this file:

https://github.com/publiclab/plots2/blob/master/config/locales/en.yml

So take one translation, like this line:

https://github.com/publiclab/plots2/blob/35449fd5da4975aa235a49beae9f7b248d674c8c/config/locales/controllers/application_controller/en.yml#L3

How do we move it?

## Move the single line of text

Cut it from the original file and paste it into the top of the base file just below `en:` -- so between lines 3 and 4 in this example:

https://github.com/publiclab/plots2/blob/35449fd5da4975aa235a49beae9f7b248d674c8c/config/locales/en.yml#L2-L5

Keep the indentation the same! 👀 

See how it was indented under `application_controller:`? Copy that line in too, above your first one, but don't delete it from the original file. 

https://github.com/publiclab/plots2/blob/35449fd5da4975aa235a49beae9f7b248d674c8c/config/locales/controllers/application_controller/en.yml#L1-L4

In this case, you should end up with your root `en.yml` looking like:

```yml
...
en:
  application_controller:
    must_be_logged_in: "You must be logged in to do that."
  language: English
...
```

## Be sure to get all the "nested" lines

Many translations are nested a few levels deep, which means we have to copy each "level". Take this for example:

https://github.com/publiclab/plots2/blob/5c0d96274bd0be15feba69d8f30093082402ace5/config/locales/views/dashboard/_activity/en.yml#L3-L7

In this case, we want to copy those over when we move the translation, so it's still accessible by the string (in this example) `t('dashboard._activity.share_work')`. See this for an example of how these translations are used in code:

https://github.com/publiclab/plots2/blob/5c0d96274bd0be15feba69d8f30093082402ace5/app/views/dashboard/_activity.html.erb#L1

Don't worry, if you get this step wrong we can help you troubleshoot once you open a pull request! 🙌 

## See an example

The changes we've done in this guide have been compiled into a pull request so you can see them all together. Read through this for guidance!

https://github.com/publiclab/plots2/pull/3514/files

Good luck, and we're here to help! 👍 

Many thanks! ⚡️ ⚡️

Thekiddiejsandpython:
Partially fixes #3515.
With love for the environment,
Jazzonym - Google Code In (2018).
1st build: October 26 @ 5:07 Mountain Time
Travis CI: ❌
Dangerbot: ✔
Codeclimate: ✔
2nd build (Patch 1): October 28 @ 8:34 Mountain Time
Travis CI: ✔
Dangerbot: ✔
Codeclimate: ✔

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,11 +1,21 @@
-# See https://github.com/svenfuchs/rails-i18n/tree/master/rails%2Flocale for starting points.
-
+# See https://github.com/svenfuchs/rails-i18n/tree/master/rails%2Flocale for starting
 en:
   dashboard:
-    _activity:
+   _activity:
       share_work: "Share your work"
-  application_controller:
-    author_has_been_banned: "The author of that note has been banned."
+      activity: "Activity"
+      all_updates: "All updates"
+      dropdown:
+        all: "All"
+        research_notes: "Research notes"
+        questions: "Questions"
+        events: "Events"
+        comments: "Comments"
+        wiki: "Wiki"
+      from_other: "from other"
+      community_scientists: "community scientists"
+      past_week: "in the past week"
+      your_work: "Your work"
   language: English
   date:
     abbr_day_names:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -16,6 +16,8 @@ en:
       community_scientists: "community scientists"
       past_week: "in the past week"
       your_work: "Your work"
+  application_controller:
+    author_has_been_banned: "The author of that note has been banned."  
   language: English
   date:
     abbr_day_names:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,7 +1,7 @@
 # See https://github.com/svenfuchs/rails-i18n/tree/master/rails%2Flocale for starting
 en:
   dashboard:
-   _activity:
+    _activity:
       share_work: "Share your work"
       activity: "Activity"
       all_updates: "All updates"


config/locales/en.yml
```diff
@@ -1,11 +1,21 @@
-# See https://github.com/svenfuchs/rails-i18n/tree/master/rails%2Flocale for starting points.
-
+# See https://github.com/svenfuchs/rails-i18n/tree/master/rails%2Flocale for starting
 en:
   dashboard:
-    _activity:
+   _activity:
       share_work: "Share your work"
-  application_controller:
-    author_has_been_banned: "The author of that note has been banned."
```
aSquare14: Do not remove lines 7 and 8

config/locales/en.yml
```diff
@@ -1,11 +1,21 @@
-# See https://github.com/svenfuchs/rails-i18n/tree/master/rails%2Flocale for starting points.
-
+# See https://github.com/svenfuchs/rails-i18n/tree/master/rails%2Flocale for starting
```
aSquare14: Do not change the comment. Please keep it as it is.

config/locales/en.yml
```diff
@@ -1,11 +1,23 @@
...
 en:
   dashboard:
-    _activity:
+   _activity:
```
gauravano:
```suggestion
   _activity:
```
How about introducing a space ' ' before line 3 such that relative 2 space indentation is maintained between lines. @Thekiddiejsandpython everything else is  good. Thanks!

Thekiddiejsandpython: Thank you so much! 😃

jywarren:
Very cool! So, does this create new translations, but doesn't tackle the part of removing the old ones yet? Do you think you can try adding commits which remove the corresponding old translations? Thank you very much!!!

And thanks to @aSquare14 and @gauravano for helping move this along!

SidharthBansal:
@Thekiddiejsandpython are you stuck?
We can help you if you want

SidharthBansal:
cc @jywarren @gauravano

PR #3830, opened 2018-10-29
Extra scrubbing of email quoted text

### LINKED ISSUE #3218: Filter additional quoted text or signatures in email replies to comments
Many people use a signature in their emails, for example:

```

Hi there!

--

Auto-signed, 
Jeff
```

Let's try to add a new filter to hide signatures in reply-by-email. @namangupta01 just a heads up, I know you're wrapping up SoC but noting this so we don't forget!

Existing filters for Yahoo and Gmail are here: 

https://github.com/publiclab/plots2/blob/04d5bbca4f47066e9628c1dc78e9961ef32c20f7/app/models/comment.rb#L299

jywarren:
Fixes #3218

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -290,11 +290,15 @@ class Comment < ApplicationRecord
 
   def parse_quoted_text
     match = body.match(/(.+)(On .+<.+@.+> wrote:)(.+)/m)
-    {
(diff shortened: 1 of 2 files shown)


jywarren:
Working from these Regex expressions! http://rubular.com/r/c6GbMgDY7T

PR #3833, opened 2018-10-29
Update the file app/views/tag/show.html.erb

meinou:
GCI candidate first-time request


meinou:
Add sorting button options to tag show template #3462

gauravano:
This PR solves #3462

PR #3854, opened 2018-10-30
Handle case in which node doesn't exist

### LINKED ISSUE #3628: Question posting form broken at /questions/new with colon in tags
@Souravirus this looks like a routing issue and I know you're pretty good at finding those... or any @publiclab/reviewers actually. Can someone help out and figure this one out? We'd really appreciate it! Thank you!

https://publiclab.org/questions/new?tags=question:website&n=15534

shows "That page does not exist" which is a 404 error.

However, this does work:

https://publiclab.org/questions/new

So, i think the `:` colon character is messing up the routing. We will have to modify the routes.rb file to allow colons, i think... 

https://github.com/publiclab/plots2/blob/master/config/routes.rb

You might search "colon in routes Rails" on Stack Overflow for some tips!

Radhikadua123:
Fixes #3628 

It was throwing 404 because that Node didn't exist. I have added check for that.

### DIFF
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -30,8 +30,9 @@ class QuestionsController < ApplicationController
   # a form for new questions, at /questions/new
   def new
     # use another node body as a template
-    if params[:n] && !params[:body] && Node.exists?(params[:n])
-      node = Node.find(params[:n])
+    node_id = params[:n].to_i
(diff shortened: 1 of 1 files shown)


Radhikadua123:
@publiclab/reviewers Please review this PR.

jywarren:
Awesome catch, thank you!

PR #3879, opened 2018-11-01
[WIP] Ensure preservation of return_to through OAuth login flow

Radhikadua123:
Fixes https://github.com/publiclab/plots2/issues/3384 

@jywarren I didn't have push access to `oauth-return-to` branch. That's why I pushed to new one in my forked repo.

Reference: #3762

### DIFF
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -11,7 +11,7 @@ class UserSessionsController < ApplicationController
       handle_social_login_flow(auth)
     else
       # User clicked on login button on site
-      handle_site_login_flow()
+      handle_site_login_flow
     end
   end
 
@@ -109,7 +109,7 @@ class UserSessionsController < ApplicationController
       return
     end
 
-    if @user&.drupal_user&.status == 0
+    if @user&.drupal_user&.status.zero?
       flash[:error] = I18n.t('user_sessions_controller.user_has_been_banned', username: @user.username).html_safe
       redirect_to '/'
       return
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -86,73 +86,84 @@ class UserSessionsController < ApplicationController
       s = "This account doesn't have a password set. It may be logged in with " + hash[n] + " account, or you can set a new password via Forget password feature"
       flash[:error] = s
       redirect_to '/'
-    else
-      params[:user_session][:username] = params[:openid] if params[:openid] # second runthrough must preserve username
-      @user = User.find_by(username: username)
-      # try finding by email, if that exists
-      if @user.nil? && !User.where(email: username).empty?
-        @user = User.find_by(email: username)
-        params[:user_session][:username] = @user.username
-      end
-      if @user.nil?
-        flash[:warning] = "There is nobody in our system by that name, are you sure you have the right username?"
-        redirect_to '/login'
-      elsif params[:user_session].nil? || @user&.drupal_user&.status == 1
-        # an existing Rails user
-        if params[:user_session].nil? || @user
-          if @user&.crypted_password.nil? # the user has not created a pwd in the new site
-            params[:user_session][:openid_identifier] = 'https://old.publiclab.org/people/' + username + '/identity' if username
-            params[:user_session].delete(:password)
-            params[:user_session].delete(:username)
-            params[:openid] = username # pack up username for second runthrough
-          end
-          @user_session = UserSession.new(username: params[:user_session][:username],
-                                          password: params[:user_session][:password],
-                                          remember_me: params[:user_session][:remember_me])
-          saved = @user_session.save do |result|
-            if result
-              # replace this with temporarily saving pwd in session,
-              # and automatically saving it in the user record after login is completed
-              if current_user.crypted_password.nil? # the user has not created a pwd in the new site
-                flash[:warning] = I18n.t('user_sessions_controller.create_password_for_new_site')
-                redirect_to '/profile/edit'
-              else
-                flash[:notice] = I18n.t('user_sessions_controller.logged_in')
-                if session[:openid_return_to] # for openid login, redirects back to openid auth process
-                  return_to = session[:openid_return_to]
-                  session[:openid_return_to] = nil
-                  redirect_to return_to
-                elsif session[:return_to]
-                  return_to = session[:return_to]
-                  session[:return_to] = nil
-                  redirect_to return_to
-                elsif params[:return_to]
-                  redirect_to params[:return_to]
-                else
-                  redirect_to '/dashboard'
-                end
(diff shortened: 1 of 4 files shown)


Radhikadua123:
@jywarren I couldn't test the whole functionality because oauth is disabled for local setup. How can I test it out ?

jywarren:
Oh, this is cool! I'll add you as a contributor and you'll be able to push to unstable.publiclab.org where we do testing. It'll build for a bit (follow at https://jenkins.laboratoriopublico.org) and then you should be able to try this out. I'm not 100% sure OAuth will work the same, but it's worth a try.

jywarren:
Also Code Climate has made some suggestions about your code, we don't have to adopt all of them but they're at least interesting to look at: https://codeclimate.com/github/publiclab/plots2/pull/3879

I can approve this to get CodeClimate to return an OK, if you like, or adopt some but not all -- up to you!

jywarren:
Oh i'm sorry i already added you! So, to start testing on `unstable`, first just chime in on the chatroom to warn others (since it only supports one build at a time, maybe someone else is using it... best to ask!) then run `git push -f git@github.com:publiclab/plots2.git unstable` and it will begin to build.

Radhikadua123:
> you'll be able to push to unstable.publiclab.org where we do testing. It'll build for a bit (follow at https://jenkins.laboratoriopublico.org) and then you should be able to try this out. I'm not 100% sure OAuth will work the same, but it's worth a try.

I see.

> I can approve this to get CodeClimate to return an OK, if you like, or adopt some but not all -- up to you!

Sure, I'll check them out.

> Oh i'm sorry i already added you! So, to start testing on unstable, first just chime in on the chatroom to warn others (since it only supports one build at a time, maybe someone else is using it... best to ask!) then run git push -f git@github.com:publiclab/plots2.git unstable and it will begin to build.

Oh. Where's the chatroom ?

Radhikadua123:
@jywarren It's working fine now on https://unstable.publiclab.org

@publiclab/reviewers @SidharthBansal Please review the code.

Also it involves some UI related changes. I removed the "Login" dropdown from the navigation bar. And clicking on "Login" redirects to login page. This removes up duplicate code and also has better UX for mobile users. 

You can check it out at https://unstable.publiclab.org

app/controllers/user_sessions_controller.rb
```diff
@@ -7,140 +7,162 @@ def new
...
+
+    if @user&.drupal_user&.status.zero?
+      flash[:error] = I18n.t('user_sessions_controller.user_has_been_banned', username: @user.username).html_safe
+      redirect_to '/'
+      return
+    end
+
+    if @user&.drupal_user&.status == 1
+      # an existing Rails user
+      if @user
```
Radhikadua123: @jywarren Is there any difference in `@user` and `@user.nil` in this case ? If not, then we can remove if condition because we already have check for that in above code.

app/controllers/user_sessions_controller.rb
```diff
@@ -7,140 +7,162 @@ def new
...
+      return
+    end
+
+    if @user&.drupal_user&.status == 5
+      flash[:error] = I18n.t('user_sessions_controller.user_has_been_moderated', username: @user.username).html_safe
+      redirect_to '/'
+      return
+    end
+
+    if @user&.drupal_user&.status.zero?
```
Radhikadua123: Banned users have status code 0, right ? I got to know that from the tests. Just wanted to confirm this.

app/controllers/user_sessions_controller.rb
```diff
@@ -7,140 +7,162 @@ def new
...
+    params[:user_session][:username] = params[:openid] if params[:openid] # second runthrough must preserve username
+    @user = User.find_by(username: username)
+    # try finding by email, if that exists
+    if @user.nil? && !User.where(email: username).empty?
+      @user = User.find_by(email: username)
+      params[:user_session][:username] = @user.username
+    end
+
+    if @user.nil?
+      flash[:warning] = "There is nobody in our system by that name, are you sure you have the right username?"
```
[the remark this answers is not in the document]
jywarren: yeah, you can change it if you think that'd be best, good catch!

app/controllers/user_sessions_controller.rb
jywarren: I think you're right, here. Let's just be very careful and make changes incrementally since this is a pretty complex piece of code. It'll be easier to track any bugs we may introduce if we change fewer things at a time. I know this is not as satisfying as more thorough refactor here, but our login code is relatively critical, so we have to be careful, and not every thing can be protected by tests. Thanks!

app/controllers/user_sessions_controller.rb
```diff
@@ -7,140 +7,162 @@ def new
...
+    if u && u.password_checker != 0
+      n = u.password_checker
+      hash = { 1 => "Facebook", 2 => "Github", 3 => "Google", 4 => "Twitter"  }
+      s = "This account doesn't have a password set. It may be logged in with " + hash[n] + " account, or you can set a new password via Forget password feature"
+      flash[:error] = s
+      redirect_to '/'
+      return
+    end
+
+    params[:user_session][:username] = params[:openid] if params[:openid] # second runthrough must preserve username
```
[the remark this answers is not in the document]
jywarren: GREAT approach! As i mention below, we want to be conservative about refactoring the signup process.

jywarren:
OK, sounds like there are a couple spots you'd like to refine before merging this? Thanks so much!!!

Radhikadua123:
>  I know this is not as satisfying as more thorough refactor here, but our login code is relatively critical, so we have to be careful, and not every thing can be protected by tests.
> OK, sounds like there are a couple spots you'd like to refine before merging this? Thanks so much!!!

I can't agree more to this. I'll split this PR into multiple small patches by which it would be easier to catch bugs in whole process. But before doing that, just wanted to confirm if you agree with the following:

```
Also it involves some UI related changes. I removed the "Login" dropdown from the navigation bar. And clicking on "Login" redirects to login page. This removes up duplicate code and also has better UX for mobile users.

You can check it out at https://unstable.publiclab.org
```

jywarren:
Hmm, I think this is ok, though I do wonder about if the non mobile page is
slightly smoother with the drop-down... Are we able to have a separate page
just for mobile, even if it is still code duplication a bit?

Thank you for your consideration on this! I agree it'll make a big
difference for mobile users!

Radhikadua123:
> Hmm, I think this is ok, though I do wonder about if the non mobile page is slightly smoother with the drop-down... Are we able to have a separate page just for mobile, even if it is still code duplication a bit?

May be in future, we can create a modal for the login page ? So, on any page, clicking on "Login", will just open popup in which user will enter details and login. There won't be any separate login page.

Radhikadua123:
Deferring to #3883 and #3884

Issue #3848, opened 2018-10-30
Implement helper function to generate/validate unique token for user


Radhikadua123:
Subpart of #3798

You have to implement two helper function in this. 

1. Helper function will take two parameters, `user` and `expiry_time`. And it will return one token.
2. Helper function will take two parameters, `user` and `token`. And it will return whether token is valid or not, and within previously mentioned `expiry_time`.

PR #3801, opened 2018-10-26
Implement helper function for encrypt/decrypt data

Radhikadua123:
Part of #3798

In this, I have added helper module which can be used to encrypt and decrypt data.

Issue #3522, opened 2018-09-28
Email validation is not working on edit profile page


vishalka98:
### Please describe the problem (or idea)

> What happened just before the problem occurred? Or what problem could this idea solve?

> What did you expect to see that you didn't?
on updating profile page email validation is not working.

### Please show us where to look
![emil](https://user-images.githubusercontent.com/31175911/46201769-96c3de00-c332-11e8-8ed6-5c4a4db6d2de.PNG)

https://publiclab.org/profile/"username"/edit

### What's your PublicLab.org username?
caiptanamerica234

> This can help us diagnose the issue: 

### Browser, version, and operating system
windows 10-chrome
> Many bugs are related to these -- please help us track it down and reproduce what you're seeing!

****

## Thank you!

Your help makes Public Lab better! We *deeply* appreciate your helping refine and improve this site. 

To learn how to write really great issues, which increases the chances they'll be resolved, see:

https://publiclab.org/wiki/developers#Contributing+for+non-coders

igniteeng000:
Great catch. Thanks

igniteeng000:
@jywarren email can be changed to anything. Users can also login with the incorrect email also. Please take a look.

avsingh999:
@vishalka98  great

jywarren:
I think this would be a great project to plan out. What systems could we plan out to address this? 

Could we send an email validation when you try to change your email? Would we use a token? Would it expire? 

Thanks!

igniteeng000:
Yes, same as we did on sign up

igniteeng000:
@vishalka98 Add code links

vishalka98:
correct answer @igniteeng000 .

aashna27:
@vishalka98 @jywarren @igniteeng000 I would like to work on this . Please assign me this.

jywarren:
Hi! Assigning is not necessary -- please go ahead now that you've claimed
it, and THANK YOU!!!!! :-)

jywarren:
Hi everyone - this is a relatively complex multi-part issue which could use some planning and breaking up into smaller pieces. Please think about how we might take an initial step. We have a welcome email that's sent - could we require email validation using a link included there? How might we generate a unique secret token that'd be included in that validation link?

Thank you!

Radhikadua123:
Just wanted to give my views on this issue. :)

I can think of two ways to fix it:

### 1. Requires saving tokens in DB


[...]

PR #3885, opened 2018-11-03
Added is_verified column to users with default value false

### LINKED ISSUE #3849: Add is_verified to user model with default value False
Subpart of #3798

You have to add `is_verified` to the user model. That field will signify that user's email is not verified. This field will default to "False".

shubhscoder:
Fixes #3849

1. Added is_verified column to users to with default value false

Thanks!

### DIFF
--- a/db/migrate/20181103114645_add_is_verified_to_users.rb
+++ b/db/migrate/20181103114645_add_is_verified_to_users.rb
@@ -1,0 +1,5 @@
+class AddIsVerifiedToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :is_verified, :boolean, :default => false
+  end
+end


shubhscoder:
@Radhikadua123 I am deleting the previous PR and as you said I have made this PR for #3849 
Now I will start working on #3848 
I am thinking of using the following method:

1. Add a column to the users table to store the unique token that is generated for the user.
2. Implement a function generate_token(user,expiry_time) that will generate a randomized token for the user. This function also stores the generated token in the database.
3.  Implement a function validate_token(user,token) that will validate the token and also clear it from the database as the link would be a one time link. This function will also update the is_verified field for the user.

What I am not getting is how are we using the expiry time?

shubhscoder:
Also, I need some help on finding out why did the test fail and what can be done to fix it. Thanking you in advance.

Radhikadua123:
@shubhscoder The approach which you are following, needs saving token into the db. But it would be better if we don't save token in db at all. To know how to implement that up, you can check out approach 2 on this comment https://github.com/publiclab/plots2/issues/3522#issuecomment-433109387

Do let me know in case you have any doubts regarding that approach. 

About the failing test, @jywarren can you please help us out ? I'm not much aware of the environment, but shouldn't the travis CI delete db before running `rake db:setup` ?

shubhscoder:
Okay so we can use the second approach. It won't require the addition of one more column. We will have to use some gem to get the encryption and decryption done.

Radhikadua123:
@shubhscoder Encryption/Decryption was added in https://github.com/publiclab/plots2/pull/3801

Radhikadua123:
Deferred to https://github.com/publiclab/plots2/pull/3907

Radhikadua123:
I saw that you removed these changes from #3907 

That's a good practice to make atomic changes in the PR. 

So, I'm reopening this PR.

Radhikadua123:
@jywarren Since I'm new to rails, I'm inexperienced with db changes. Though the column addition definition seems fine to me.

Can you please review this PR ? 

Also will it cause down time during the time migration will be ran on the db ? Is that a concern for us ? Any idea ? Also may be @icarito knows about it.

jywarren:
Hi, no, it won't create significant downtime, no worries! So, now you have to update the `/db/schema.rb.example` file to reflect the new changes (just those related to your updates) and this should build properly. 

Thanks!!!

Issue #3850, opened 2018-10-30
Send verification email to the user on signup


Radhikadua123:
Subpart of #3798

We send the welcome email to the user. That email should contain the link by which user can confirm her account.

shubhscoder:
I would like to take this up. With whatever I have researched I think we can use the devise gem for the same. This is the link : https://rubygems.org/gems/devise/versions/4.2.0. Also it has a good number of downloads, hence can be quite reliable.

Radhikadua123:
I'm not sure about the devise gem, but we already send welcome email to the user. So, you might want to take a look at that. But anyway, if devise can make things easier, feel free to use it.

shubhscoder:
I need some more light on this. Should the email confirmation link be included in the welcome email itself ? Or we can send an another email for confirmation? Also do we need some other features like locking ?

jywarren:
I think sending just one email makes sense, so it could go in the welcome
email!

SidharthBansal:
> I think we can use the devise gem for the same.

@shubhscoder sorry, we don't use devise gem in plots2 repo. We are using a similar gem AuthLogic.

shubhscoder:
Yeah, i realized it. Thank you 👍

PR #3907, opened 2018-11-07
Add helper generate and validate token

### LINKED ISSUE #3848: Implement helper function to generate/validate unique token for user

shubhscoder:
Fixes #3848 
Additions : 
1. Added generate_token functions that generates a token for the user
2. Added validate_token function that verifies the generated token and makes sure that the token is generated within 24 hours.

Thanks!

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -418,15 +418,15 @@ class User < ActiveRecord::Base
   end
 
   def generate_token
-    user_id_and_time = [self.id,Time.now]
+    user_id_and_time = [id, Time.now]
     encrypt(user_id_and_time)
   end
 
   def validate_token(token)
     decrypted_data = decrypt(token)
-    if self.id!=decrypted_data[0]
+    if id != decrypted_data[0]
       return false
-    elsif (Time.now - decrypted_data[1])/1.hour>24.0
+    elsif (Time.now - decrypted_data[1]) / 1.hour > 24.0
       return false
     else
       return true
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -418,15 +418,15 @@ class User < ActiveRecord::Base
   end
 
   def generate_token
-    user_id_and_time = [id, Time.now]
+    user_id_and_time = { :id => id, :timestamp => Time.now }
     encrypt(user_id_and_time)
   end
 
   def validate_token(token)
     decrypted_data = decrypt(token)
-    if id != decrypted_data[0]
+    if id != decrypted_data[:id]
       return false
-    elsif (Time.now - decrypted_data[1]) / 1.hour > 24.0
+    elsif (Time.now - decrypted_data[:timestamp]) / 1.hour > 24.0
       return false
     else
       return true
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -7,6 +7,7 @@ class UniqueUsernameValidator < ActiveModel::Validator
 end
 
 class User < ActiveRecord::Base
+  include Utils
   self.table_name = 'rusers'
   alias_attribute :name, :username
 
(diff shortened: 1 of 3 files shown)


shubhscoder:
@Radhikadua123 , Does this seem fine?
@publiclab/reviewers

jywarren:
Aha, cool! Would you consider adding a simple test to `/test/unit/user_test.rb` to confirm these work? Thank you!

jywarren:
@Radhikadua123 this is looking good, huh? 😄

shubhscoder:
Yeah i ll do that 👍

shubhscoder:
@jywarren . I have added the tests. Thank you!

test/unit/user_test.rb
```diff
@@ -220,4 +220,14 @@ class UserTest < ActiveSupport::TestCase
     #as the username as "jeff" exists, hence username = "jeff" + 2 digit alphanumeric code will be created
     assert_not_equal jeffrey.username, "jeff"
   end
+
+  test 'generate token and validate token for user email verification' do
+    all_users = User.where("id<?",3)
+    for i in all_users
```
Radhikadua123: It's fine to run it for one user only. :)

Radhikadua123:
Hi @shubhscoder,

Thanks a lot for the PR! :) 

Can you please add more test for it to cover all the cases ?

e.g. 

1. Test to confirm that token generated 24hrs back from now should be treated as invalid.
2. Test to confirm that token for one user can't be used for any other user.
3. Test to confirm that invalid token raises respective exception

jywarren:
oh, this is looking awesome. Great work! And thanks Radhika for helping
this along!

shubhscoder:
Yeah I ll do that 👍

shubhscoder:
@jywarren @Radhikadua123 I have added the failing tests. Check if this looks good. Thank you!

test/unit/user_test.rb
```diff
@@ -220,4 +221,27 @@ class UserTest < ActiveSupport::TestCase
...
+
+    # Checking that a user should not be able to verify his email using someone elses token
+    if all_users.length>1
+     assert_not_equal all_users[1].validate_token(generated_token), true
+    end
+
+    generated_token = generated_token[2,generated_token.length]
+    begin 
+      assert_not_equal all_users[0].validate_token(generated_token), true
+    rescue => error
```
Radhikadua123: Please use `assert_raise` for testing this.

test/unit/user_test.rb
```diff
@@ -220,4 +221,27 @@ class UserTest < ActiveSupport::TestCase
     #as the username as "jeff" exists, hence username = "jeff" + 2 digit alphanumeric code will be created
     assert_not_equal jeffrey.username, "jeff"
   end
+
+  test 'generate token and validate token for user email verification' do
```
Radhikadua123: Almost there, just split these all these tests into multiple functions. 

In unit tests, you should try tests to be as small as possible so that they test one small functionality properly. When they fail, developer easily get to know which test failed.

app/models/user.rb
```diff
@@ -416,6 +417,22 @@ def customize_digest(type)
     end
   end
 
+  def generate_token
+    user_id_and_time = [id, Time.now]
```
Radhikadua123: May be we should use `hash` here instead of array ? 
`{ :id => id, :timestamp => Time.now }`

That would be more descriptive than using something like `decrypted_data[0]`.

shubhscoder:
@Radhikadua123 I thinked I have fixed them all! And thank you for guiding me through the process. If this goes fine, then I will start working on #3850 .

Radhikadua123:
@shubhscoder Thanks for fixing the issues!

@jywarren It looks good to me. Please review the changes.

jywarren:
Oh, lovely - thanks!!! Great tests, too!

PR #3930, opened 2018-11-09
Send verification link in email

### LINKED ISSUE #3850: Send verification email to the user on signup

shubhscoder:
Fixes #3850 
The welcome mail would now also contain the verification link for the user.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -345,6 +345,17 @@ class UsersController < ApplicationController
     end
   end
 
+  def verify_email
+    @user = User.find(params[:id])
+    if @user.validate_token(params[:token])
+      puts "Email successfully validated"
+      redirect_to "/login"
+    else
+      puts "Email validation failed"
+      redirect_to "/login"
+    end
+  end
+
   private
 
   def set_user
--- a/app/views/welcome_mailer/notify_newcomer.html.erb
+++ b/app/views/welcome_mailer/notify_newcomer.html.erb
@@ -19,6 +19,9 @@
     Looking forward to hearing from you soon,
 
     The Public Lab Community
+
+    You need to verify your email to confirm your registration. Please click the link below to confirm your email. Note that this link is valid for only 24 hours after its generation. Thank you. 
+    <%="localhost:3000/#{@user.id}/verify/#{@user.generate_token}" %>
 <% end %>
 
 <%= @footer %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -213,6 +213,7 @@ Plots2::Application.routes.draw do
   get 'profile/:id/likes' => 'users#likes'
   get 'feed/:author' => 'users#rss'
   get '/settings' => 'users#settings'
+  get ':id/verify/:token' => 'users#verify_email'
   post '/save_settings' => 'users#save_settings'
 
   post 'profile/tags/create/:id' => 'user_tags#create'
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,7 @@ gem 'grape-swagger-ui'
 gem 'http_accept_language'
 gem "i18n-js", ">= 3.0.0.rc11"
 gem 'impressionist'
-gem 'jbuilder', '~> 2.5'
+gem 'jbuilder', '~> 2.8'
 gem 'jquery-rails'
 gem 'less-rails', '~> 4.0'
 gem 'mocha', '~> 1.1'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -87,7 +87,7 @@ GEM
     commonjs (0.2.7)
     composite_primary_keys (11.0.3)
       activerecord (~> 5.2.0)
-    concurrent-ruby (1.0.5)
+    concurrent-ruby (1.1.1)
     connection_pool (2.2.2)
     coveralls (0.8.22)
       json (>= 1.8, < 3)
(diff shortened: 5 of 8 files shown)


Radhikadua123:
@shubhscoder Thanks for the PR!

Please merge/rebase latest master into this branch.

app/views/welcome_mailer/notify_newcomer.html.erb
```diff
@@ -19,6 +19,9 @@
     Looking forward to hearing from you soon,
 
     The Public Lab Community
+
+    You need to verify your email to confirm your registration. Please click the link below to confirm your email. Note that this link is valid for only 24 hours after its generation. Thank you. 
+    <%="localhost:3000/#{@user.id}/verify/#{@user.generate_token}" %>
```
Radhikadua123: Please don't hardcode `localhost:3000` here.

Radhikadua123: Please check password reset form from other websites and modify it accordingly. Like may be we would prefer a `<a>` tag here on which user can directly click and get account verified ? Also, along with `a` tag, may be there should be text link saying, if you can't click on that link, then copy paste it into the browser ?

config/routes.rb
```diff
@@ -213,6 +213,7 @@
   get 'profile/:id/likes' => 'users#likes'
   get 'feed/:author' => 'users#rss'
   get '/settings' => 'users#settings'
+  get ':id/verify/:token' => 'users#verify_email'
```
Radhikadua123: Do we need to include `id` in the endpoint here ?

app/controllers/users_controller.rb
```diff
@@ -345,6 +345,16 @@ def shortlink
     end
   end
 
+  def verify_email
+    @user = User.find(params[:id])
+    if @user.validate_token(params[:token])
+      puts "Email successfully validated"
```
Radhikadua123: It would be better to flash a message on the login page when user gets redirected instead of output in console.

Radhikadua123: When user gets validated successfully, you will have to update db too.

app/controllers/users_controller.rb
```diff
@@ -345,6 +345,16 @@ def shortlink
     end
   end
 
+  def verify_email
+    @user = User.find(params[:id])
```
Radhikadua123: Can you refactor your code a bit so that we first decrypt the data received and on the successful decryption, check the user id and time ?

shubhscoder:
@Radhikadua123 the PR is not complete. I was working on it. I requested someone to add the in progress label to this PR on gitter. But I think anyone did not see that message. Thank you 🙈

shubhscoder:
Oh sorry ! Sourav had added the label. Thank you @Souravirus

app/controllers/users_controller.rb
shubhscoder: But what would you compare your user id with?We have to make sure that the link generated for that user verifies his email. Would you mind providing a small example, that would make things clear. And if you don't want the user to to know his user_id , then we can use his username instead. Then I guess everything would work fine?

Radhikadua123: We won't really need user_id and username.

Let me give you an example:

URL would look like `publiclab.org/profile/verify/here_comes_token`

We will get that token and will call `validate_token` class method which will return user in case token is valid, else false in case token is invalid. 

How do you define validity of token ? We just check whether timestamp is not too old(24hrs) and data decryption is successful. 

Then from the decrypted data we get the user_id and then return the user object.

shubhscoder:
Yeah, the tests will also have to be changed then. That is not an issue
though. These are some minor changes that can be made anytime. However I am
stuck with an issue that I posted on gitter as well.

In schema.rb file we have 2 tables rusers and users. And somehow the User
model is mapped to the rusers table because User.columns matches exactly
with rusers table in sqlite3. I am trying to figure this out for a long
time, but not able to do that.
That is why , when I wrote a migration to add columns to the users table,
it got reflected in the schema of users table , but still I am getting a
Unknown attribute error. Thanking you in advance! This is what I have been
trying to figure out from yesterday but to no avail. Thank you !

Radhikadua123:
@shubhscoder I have added possible explanation over it on gitter.

app/controllers/users_controller.rb
shubhscoder: Okay so I have understood what you want to say, but in this case we ll have to make encrypt and decrypt methods that you implemented also class methods and not instance methods. So I can do that, not a problem. I am beginner with rails or in general with web development so I have this doubt: What is the problem in using user_id or username in the endpoint?

Radhikadua123: > encrypt and decrypt methods that you implemented also class methods and not instance methods

I think this won't be required.

> What is the problem in using user_id or username in the endpoint?

Good question! Actually it doesn't affect anything. It's totally fine to use them in the endpoints but since we don't need them, it's redundant. Though about using user_id, if you check any other website like facebook, they never show original user id in their endpoints. You will see string of random chars which represent an item uniquely on their system, but why ? May be they don't want anyone to know the original database id. It's just considered as bad practise to show internal db ids.

shubhscoder:
I have made the changes and soon make a PR for that. But one more question
, its fine that we don't want our user id to be revealed to the user, but
we can use username instead. Anyways the user is aware of his username

shubhscoder:
So we just want to avoid redundancy right?

PR #3953, opened 2018-11-12
Refactoring of helper functions

shubhscoder:
These are the refactoring changes made so that we dont need to use id in the endpoint.
Following changes are made
1. validate token is made a class method instead of instance method. This method would return the decrypted user_id upon successful decryption. If the decryption was not successful then it would return a idempotent value. 
2. The appropriate tests are also updated


shubhscoder:
@radhikadua123 this should work right?

PR #3964, opened 2018-11-14
Refactoring helper functions

shubhscoder:
In order to avoid using id/username in the endpoint some refactoring was needed in the helper functions
This PR does that refactoring

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -422,14 +422,17 @@ class User < ActiveRecord::Base
     encrypt(user_id_and_time)
   end
 
-  def validate_token(token)
-    decrypted_data = decrypt(token)
-    if id != decrypted_data[:id]
-      return false
-    elsif (Time.now - decrypted_data[:timestamp]) / 1.hour > 24.0
-      return false
+  def self.validate_token(token)
+    begin
+      decrypted_data = decrypt(token)      
+    rescue Exception => e
+      puts e.message
+      return 0
+    end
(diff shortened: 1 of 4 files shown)


shubhscoder:
@Radhikadua123 please review.

test/unit/utils_test.rb
```diff
@@ -2,7 +2,7 @@
 
 class UtilsTest < ActiveSupport::TestCase
   include Utils
-
+
```
Radhikadua123: We really need to setup overcommit in our project. I have filed an issue for it. https://github.com/publiclab/plots2/issues/3967

But never mind, it's fine to commit it like this for now.

app/models/user.rb
```diff
@@ -419,17 +419,20 @@ def customize_digest(type)
...
-  def validate_token(token)
-    decrypted_data = decrypt(token)
-    if id != decrypted_data[:id]
-      return false
-    elsif (Time.now - decrypted_data[:timestamp]) / 1.hour > 24.0
-      return false
+  def self.validate_token(token)
+    begin
+      decrypted_data = User.decrypt(token)      
+    rescue StandardError => e
```
Radhikadua123: Instead of StandardError, can you please check specific type of exception it throws and catch only that.

Radhikadua123:
Thanks for fixing it!

@jywarren Please review the changes.

shubhscoder:
@publiclab/reviewers Please review! Thank you!

jywarren:
This is fantastic. Great work, both of you!!!

Issue #3758, opened 2018-10-24
Combined wiki/note grids using [node:_____] syntax, for inline grids


jywarren:
This may be a bit more complex, but we'd love some help. 

We have inline grids on pages like this: https://publiclab.org/wiki/air-sensors

This means we have different ways to include inline tables of other pages, using syntax like:

`[notes:______]` or `[wikis:________]`

The templates for these are here:

https://github.com/publiclab/plots2/tree/master/app/views/grids

But sometimes you want to have both wikis and notes displayed. It'd be great to have a combined grid type for any node (see our /docs/ folder and the DATA MODEL) - with format:

`[node:_____]`

See how I've kind of done a broken version of this by just using one grid tag after the other -- showing 2 tables one after the other (a hack! Not very nice but... kind of works):

![image](https://user-images.githubusercontent.com/24359/47458076-ba9a0880-d7a6-11e8-84e5-1b0e4bebe576.png)

One issue is that notes and wikis have different columns - like the timestamp or author column. Let's try to find what overlaps and display those columns so it makes some sense -- but we can tweak this later.

KauNair:
I'll look into this issue. :+1:

jywarren:
Awesome, thanks!!!

SidharthBansal:
Published on GCI dashboard

SidharthBansal:
Yes

jywarren:
That's OK @KauNair we would still love help refining the grid to use Bootstrap styles as per https://github.com/publiclab/plots2/pull/3970#issuecomment-439167488 -- thanks!

ValentinaTironi:
@jywarren @SidharthBansal Hi! I have a question. Where I can see the notes and wikis grids? I looked of them but I couldn't find them. Can you give the route where the notes and wikis are displayed in grids? Thank you!

ValentinaTironi:
Another question: Where I have to show the grids with notes and wikis? And how many I should show? All of them? Sorry, I'm trying to understand the task to resolve it

ValentinaTironi:
In https://publiclab.org/wiki I see that: 
<img width="1280" alt="screen shot 2018-11-23 at 8 57 05 pm" src="https://user-images.githubusercontent.com/20969831/48962706-64b1af80-ef62-11e8-9334-9d3d14b8d847.png">

It's a grid. Everything is right here. But in https://publiclab.org/notes 
[...]

PR #3970, opened 2018-11-14
Show grid of notes with thumbnails

### LINKED ISSUE #1097: develop `notes:tagname` inline tag alternative to display a grid of notes, not a table

KauNair:
Fixes #1097 
I had to make minor changes in the main notes show html page since the grid was overflowing out of the container for some reason.
Please let me know if there are any changes to be made in the design. I'll add tests soon too.

Here's a screenshot:

plots2

### DIFF
--- a/app/views/notes/show.html.erb
+++ b/app/views/notes/show.html.erb
@@ -1,6 +1,6 @@
 <script src="/assets/notes.js" type="text/javascript"></script>
 
-<div class="col-md-9 note-show<% if @node.status == 4 || @node.status == 0 %> moderated<% end %>">
+<div class="col-md-12 note-show<% if @node.status == 4 || @node.status == 0 %> moderated<% end %>">
   <% if current_user && @node.tags.length == 0 %><div class="alert alert-warning"><%= raw t('notes.show.note_no_tags', url: 'javascript: $(".tag-input").focus();') %></div><% end %>
   <% if @node.has_power_tag('status:candidate') %><div class="alert alert-info">This was marked as a candidate activity -- <a href="https://publiclab.org/notes/warren/09-17-2016/what-makes-a-good-activity" >read about how to adapt its content</a> into a fully-fledged, step-by-step activity that can be easily replicated.</div><% end %>
   <% if @node.has_power_tag('replication') %><div class="alert alert-info"><%= raw t('notes.show.replication') %> <a href="/n/<%= @node.power_tag('replication') %>"><%= raw t('notes.show.replication_link') %></a>.</div><% end %>
@@ -75,7 +75,7 @@
 
   <%= render :partial => "notes/gallery" %>
 
-  <div id="content" class="pl-content note">
+  <div id="content" class="pl-content note col-md-9">
   <%= raw auto_link(insert_extras(@node.latest.render_body), :sanitize => false) %>
   </div>
 
--- a/app/views/notes/show.html.erb
+++ b/app/views/notes/show.html.erb
@@ -1,6 +1,6 @@
 <script src="/assets/notes.js" type="text/javascript"></script>
 
-<div class="col-md-12 note-show<% if @node.status == 4 || @node.status == 0 %> moderated<% end %>">
+<div class="col-md-9 note-show<% if @node.status == 4 || @node.status == 0 %> moderated<% end %>">
   <% if current_user && @node.tags.length == 0 %><div class="alert alert-warning"><%= raw t('notes.show.note_no_tags', url: 'javascript: $(".tag-input").focus();') %></div><% end %>
   <% if @node.has_power_tag('status:candidate') %><div class="alert alert-info">This was marked as a candidate activity -- <a href="https://publiclab.org/notes/warren/09-17-2016/what-makes-a-good-activity" >read about how to adapt its content</a> into a fully-fledged, step-by-step activity that can be easily replicated.</div><% end %>
   <% if @node.has_power_tag('replication') %><div class="alert alert-info"><%= raw t('notes.show.replication') %> <a href="/n/<%= @node.power_tag('replication') %>"><%= raw t('notes.show.replication_link') %></a>.</div><% end %>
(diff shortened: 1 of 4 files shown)


app/views/notes/show.html.erb
```diff
@@ -1,6 +1,6 @@
 <script src="/assets/notes.js" type="text/javascript"></script>
 
-<div class="col-md-9 note-show<% if @node.status == 4 || @node.status == 0 %> moderated<% end %>">
+<div class="col-md-12 note-show<% if @node.status == 4 || @node.status == 0 %> moderated<% end %>">
```
jywarren: Hm, ok! I see how this may help... however, do you think it could be potentially moving other page content around in a way that would cause trouble? I also wonder what might happen on other pages, such as wiki pages from `/app/views/wiki/show.html.erb` and `/app/views/question/show.html.erb` - other places this type of grid might be shown.

jywarren: er, i guess my note came out below 😄

KauNair:
Div tags are so difficult to work with. :disappointed: 
I changed the grid into a table, so now it works without having to change notes show view. :smile: 
Also what should I do for notes that do not have any main image?

The thumbnails are slightly bigger than before but I guess that's okay?
plots3

jywarren:
This looks great. Any adjustments can be done in follow-up! Thanks so much!!! Again, this was very sought after, so THANKS  A LOT!!! 👍 👍 👍

jywarren:
Would you be interested in a new issue, related to this one? Perhaps this one? https://github.com/publiclab/plots2/issues/3758

jywarren:
Also, let's think about what this looks like in mobile phones... can you try in a narrower window? See some about the class-based page layout system here: https://getbootstrap.com/docs/3.3/css/

KauNair:
This is how they look on mobile . Maybe we can try responsive tables so they have a scroll bar in them.
Smaller images make the grid irregular. We should scale the all the images to a fixed size so they look better.
Also this tag doesn't have test coverage.
All these should be easy to fix. Do you think we should make new issues for these or should I send a PR?


plots5

jywarren:
Hi, I wanted to note this is live now, and looks great! https://publiclab.org/wiki/micro#Assembly

image

Some of the followup steps will really make this shine. Great work!!!

ValentinaTironi:
Hi! I'm a little confused. I don't know how to render notes in the section so, if you can clear that for me would be amazing!  How I can show it?

ValentinaTironi:
Can someone tell me how render the notes like this? Because when I did `notes:tagname` it renders a grid, and I don't know how show them with images

jywarren:
Ah, take a look through the implementation as in the pr where I'd made
comments about bootstrap, to see how the HTML is generated. Does that help?

ValentinaTironi:
Yeah. This is merged in master branch, right? I did `git pull` but there aren't changes, I think that is something related to repository forked. I'm gonna google how to update a fork repository from original

jywarren:
if you are in a feature branch, you can do:

1. `git checkout master`
2. `git pull https://github.com/publiclab/plots2.git master`
3. `git checkout feature-branch`
4. `git merge master`

Looking for lines of code to help you out too now!

ValentinaTironi:
Ready! I could do it!

jywarren:
See how this assembles data in the node_shared file, and then displays them in a template? 

https://github.com/publiclab/plots2/pull/3970/files

Hope that helps!

ValentinaTironi:
Hey @jywarren I could do something like that
screen shot 2018-11-26 at 8 06 12 pm

What do you think?

jywarren:
That's lovely. I think that looks perfect; is it defaulting to a 2x3 grid?
Thanks!

And a link maybe that says "see more"?

ValentinaTironi:
Thank you! The maximum number of images per row is 3, if that is that you refer. 
Do you want that I add a link in every note?

jywarren:
ah, sorry, i guess i am thinking ahead -- what if there are too many
results? Should we limit it to 6 shown, and the "see more" would expose
more? If you look at a page like https://publiclab.org/w/lego-spectrometer,
you'll see some grids that overflowed and so they are limited to 10.

ValentinaTironi:
Oh, I understand. I'm trying to do it

ValentinaTironi:
Me again! Well, I did it
more
Also, I did a `show less` button but I didn't have time to recording because github just allow 10MB

PR #3992, opened 2018-11-16
Verification link in welcome mail

### LINKED ISSUE #3850: Send verification email to the user on signup

shubhscoder:
Fixes #3850

1. The welcome email would now contain, link for verification of email
2. Upon successful verification, the is_verified gets updated


thesparks:
```
if decrypted_user_id != 0
    User.find(decrypted_user_id).update_column(:is_verified, true)
    action_msg = "Successfully verified email"
end
```

If I'm right, this code will update the is_verified field even when the user has already activated an account. Can you change it?

app/views/welcome_mailer/notify_newcomer.html.erb
```diff
@@ -18,6 +18,13 @@
 
     Looking forward to hearing from you soon,
 
+    Please click on the link below to verify your email.
+
+    <% verifaction_link="https://publiclab.org/verify/#{@user.generate_token}" %>
```
jywarren: Here we could replace `publiclab.org` with `<%= ActionMailer::Base.default_url_options[:host] %>` so it's not host-specific!

shubhscoder: Yeah I fixed that.

shubhscoder:
Yes I ll do that asap. I am kindoff stuck with semester exams ! Will do
that in a day or two . Thank you so much !

jywarren:
good luck with your exams!

shubhscoder:
I tried writing the tests and encountered this error "undefined method `[]' for true:TrueClass
	from lib/authlogic_openid/authlogic_openid/acts_as_authentic.rb:74:in `save'
" on line "        user_obj.update(:is_verified => true)" in users_controller. Tried debugging it but to no avail. @jywarren do you know why this is happening? Thanking you in advance!

jywarren:
Is it in the Travis tests or on your local? Try rerunning them, I think
there is an intermittent issue with Travis...

Hope that helps!

shubhscoder:
Its not passing even on my local machine , would definitely fail on Travis !

jywarren:
Hmm, @SidharthBansal do you have any ideas on this?

SidharthBansal:
Can you @shubhscoder  please push in the failing tests so that we can help you?

shubhscoder:
@SidharthBansal I have pushed the failing tests. Sorry for the delay!

shubhscoder:
@SidharthBansal @jywarren I realised why the tests are failing. The users_controller_test is using the fixtures to get its object, where as the object gets updated in the users_controller, which updates it in the  sql database, that is why the is_verified of the  fixtures object never gets updated. I don't think there is a way to write tests for such a scenario. I have tested everything on the localhost I can make a video of the same to make sure that everything is working.

shubhscoder:
@SidharthBansal check if this looks good to you, or we need more changes. Thank you so much!

SidharthBansal:
@jywarren changes looks good to me. It is important PR so your careful review is must.

jywarren:
This is really nicely implemented, thanks! Just a few questions:

1. what currently happens when unverified?
2. relatedly, could we roll this out and then measure a bit whether people are successfully verifying even if we aren't yet requiring it? That'd give us a chance to build confidence in this workflow
3. what about people who join by OAuth, do we exempt them?

Thanks!

shubhscoder:
@jywarren 
1. Currently if the user is not verified then nothing happens (He is allowed to access everything) but after we merge #4493 the user will start getting a prompt on the dashboard page to verify his email( Again even after #4493 he is not restricted from doing anything.)

2. Yeah, that sounds a good idea, but in this PR the current users wont know that we have introduced a new email verification process. We would be able to check with only the people who join public labs after we merge this.

3. For the people joining with Oauth we have not made any provision yet, but yes we could exempt them if required
Thank you!

jywarren:
This is great! Thank you for the clear response. 

> 2. current users wont know that we have introduced a new email verification process. We would be able to check with only the people who join public labs after we merge this.

Great, this is fine. Let's make these follow-up steps in an issue and this can be merged!

app/views/welcome_mailer/notify_newcomer.html.erb
```diff
@@ -18,6 +18,13 @@
 
     Looking forward to hearing from you soon,
 
+    Please click on the link below to verify your email.
+
+    <% verifaction_link="#{ActionMailer::Base.default_url_options[:host]}/verify/#{@user.generate_token}" %>
+    <%= link_to "#{verifaction_link}", verifaction_link%>
+    
+    In case you are not able to click the link, try copy pasting the link in the browser.
+
```
gauravano: Hi @jywarren make sure too copy this change to the **welcome-email-body** feature whenever you push this PR to production as the changes made here are just on the fallback email body and the production would send the email using the welcome-email-body feature's content(if it's present). 

Thanks!

jywarren: Ah, then this is an issue - we can't execute code in the features system, so this will need to be moved outside the `<% end %>` on this line: https://github.com/publiclab/plots2/pull/3992/files/b74bbcbd3d85ff70b18bc7eae14bf79105270976#diff-a3214821bc5b12012e6fbb54ad82f05aR29

jywarren:
@shubhscoder do you see our comments above? I think this needs one more change, sorry!

app/views/welcome_mailer/notify_newcomer.html.erb
shubhscoder: @jywarren . Sorry but I am not getting what we are supposed to do?

shubhscoder: Oh I think I got it. We need to move the verification link and the corresponding instructions regarding the mail outside the if else block right?

gauravano:
Yes, Shubham!

PR #3998, opened 2018-11-18
Attempted fix for i18n fallback

### LINKED ISSUE #3961: Site-wide outage for german :de locale -- due to fallback fail
Simultaneously other people were able to access the site, strangely. 

> The page seems to be down, but https://downforeveryoneorjustme.com/publiclab.org says it’s up....
> PS: Windows 10 - Mozilla Firefox/ Edge/ Chrome; Google Pixel 2 - Chrome
> It showed: There was an error. Are you sure you have the right address?

This happened both on WiFi and cellular from the reporting person's comments. 

We should check the logs to see what the error was and more about the requests.

shubhscoder:
Fixes #3961 
I am not sure that this would work, but we can try this.


config/environments/production.rb
```diff
@@ -73,8 +73,8 @@
 
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation can not be found)
-  config.i18n.fallbacks = true
-
+  #config.i18n.fallbacks = true
+  config.i18n.fallbacks = {:en => [:en, :de], :de => [:de, :en]}
```
jywarren: Cool, did this work for you in your local copy?

shubhscoder: I tried setting up the system in production mode on my system and failed terribly. There were multiple failing migrations. I tried skipping them but because of these skipped migrations many later migrations also failed. Could we test it in staging?

jywarren:
Hmm, is it possible to seed your db using only "rake db:setup" instead of
running migrations, though?

jywarren:
I had to copy the development key from config/secrets.yml to the line for production, too...

jywarren:
Ah, hmm, so in production i'm seeing:

image

Plus all through the HTML: 

```html
<h3><span class="translation_missing" title="translation missing: de.tag.show.notes_tagged, tag: test">Notes Tagged</span></h3>
```

However these translations really are missing. But i'm not seeing the complete 500 error as in #3961...

jywarren:
So one way to test this would be to push this to the unstable branch, then go to https://unstable.publiclab.org/change_locale/de, then look at https://unstable.publiclab.org/questions/tag/parent:multispectral-imaging/author/nedhorning -- 

I'm pushing to unstable now. it'll take some time (10 minutes?) then this should be test-able.

jywarren:
I believe unstable is working again now, and we should be able to test this starting tomorrow if you like!

shubhscoder:
Yes sure we can test this on unstable! I am not sure of how we deploy on unstable, but think we have a different link along with a password to test things right @jywarren ? Thank you !

jywarren:
Hi, to test on unstable, we push to the unstable branch, which only some of us have access to. I just did, and will announce in the chatroom that I've done this so we don't conflict with others - there's only one test server to share. Then you can try it out at https://unstable.publiclab.org !

jywarren:
OK, it's on there until someone else uses unstable!

jywarren:
Hi @gauravano, should we close this? Thanks!!

jywarren:
@gauravano just checking if this could be closed, what do you think? Thanks, all!

cesswairimu:
Closing this. Feel free to open if we still need it.Thanks all

gauravano:
Hi @jywarren @cesswairimu sorry for missing out on your comments 😞. I went through the issue this PR aim to solve and looks like we had the closure there. 

So, maybe we can close the issue too - https://github.com/publiclab/plots2/issues/3961 . We can ask someone from PL team or from the community from Germany for testing it once more if you folks think it's required. 

Thanks!!

cesswairimu:
Thanks @gauravano :+1:  will ping Liz on the issue

PR #4013, opened 2018-11-20
Added insert_extra() properly into comments so they show inline shortcuts[resolves 3978]

### LINKED ISSUE #3978: use grids via "insert_extras()" in all types of comments
When using inline shortcodes like `[notes:balloon-mapping]` in wiki or note body text, they get converted using the `insert_extras()` method, like on these lines:

https://github.com/publiclab/plots2/search?utf8=%E2%9C%93&q=insert_extras&type=

See how it works in the sandbox: https://publiclab.org/wiki/sandbox-2

But this doesn't happen in comments -- see this example:

https://publiclab.org/notes/imvec/11-10-2018/community-microscope-laser-cut-file#c21267

It also doesn't work in wiki comments, question comments... yet some of those seem to have the `insert_extras()` code, as per the search results above. 

Let's figure this out!

conkytom:
Fixes #3978

[✔️] code is in uniquely-named feature branch and has no merge conflicts
[✔️] PR is descriptively titled
[✔️ ] ask `@publiclab/reviewers` for help, in a comment below

Replaced `filtered_comment_body` with `insert_extras` and now inline shortcuts like `notes:topic_tag` written in the comment will display a table of links.

### DIFF
--- a/app/views/notes/_comment.html.erb
+++ b/app/views/notes/_comment.html.erb
@@ -118,7 +118,7 @@
   <div style="border: 1px solid #e7e7e7;padding: 18px;" id="c<%= comment.cid %>show">
 
     <% comment_body = comment.render_body %>
-    <p id="comment-body-<%= comment.cid %>"><%= raw filtered_comment_body(comment_body) %></p>
+    <p id="comment-body-<%= comment.cid %>"><%= raw insert_extras(comment_body) %></p>
 
     <% if contain_trimmed_body?(comment_body) %>
       <span style="background-color: #DCDCDC;  border-radius: 7px; cursor: pointer;" ><a class="email-toggle" data-toggle="collapse" data-target="#comment-<%= comment.cid %>-extra-content" style='text-decoration: none;'>&nbsp;&ensp;&#x2022&#x2022&#x2022&nbsp;&ensp;</a></span>
--- a/app/views/notes/_comment.html.erb
+++ b/app/views/notes/_comment.html.erb
@@ -118,7 +118,7 @@
   <div style="border: 1px solid #e7e7e7;padding: 18px;" id="c<%= comment.cid %>show">
 
     <% comment_body = comment.render_body %>
-    <p id="comment-body-<%= comment.cid %>"><%= raw insert_extras(comment_body) %></p>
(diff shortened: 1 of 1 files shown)


conkytom:
Hello @publiclab/reviewers, could you please take a look at this when you get a chance.  It worked for me with comments in Wikis, Notes, Questions, Answers, and other Comments, not sure where else it could be tested.

app/views/notes/_comment.html.erb
```diff
@@ -118,7 +118,7 @@
   <div style="border: 1px solid #e7e7e7;padding: 18px;" id="c<%= comment.cid %>show">
 
     <% comment_body = comment.render_body %>
-    <p id="comment-body-<%= comment.cid %>"><%= raw filtered_comment_body(comment_body) %></p>
+    <p id="comment-body-<%= comment.cid %>"><%= raw insert_extras(comment_body) %></p>
```
jywarren: Awesome! This is indeed the method we need to add, but do you think you could add it within the `filtered_comment_body` method, and would that work? 

https://github.com/publiclab/plots2/blob/af82259df59b96fa043bbde4bb190278ee560347/app/helpers/application_helper.rb#L120-L125

Basically that method is used to trim out email signatures and such, so we kind of need to run both. Could you put the `insert_extras` method on maybe both lines 122 and 124 of the application_helper.rb file?

conkytom: I could, I'm not totally sure what you would prefer.  I could add `insert_extras` to the `filtered_comment_body` but that may may make uses of `filtered_comment_body` be able to have the the tagging abilities of `insert_extras` which may be overkill.  I just tried wrapping `insert_extras` around `filtered_comment_body` and it worked on my local host fine. 
```
    <p id="comment-body-<%= comment.cid %>"><%= raw insert_extras(filtered_comment_body(comment_body)) %></p>
```
Might be a more feasible limited approach.  If you like that I'll update and push again.

jywarren:
sure, that sounds great! Let's do it. Thanks!

jywarren:
Awesome!!! Great work. Many thanks!!!

PR #4042, opened 2018-11-25
Fixed the pagination problem

### LINKED ISSUE #2258: Pagination broken at /research
The pagination is in accordance to the research section always. It does not change to the events section pagination on clicking on the events.
image
Eventhough we have only 3 events it showed us n pages because here n pages are required for r research notes.

okonek:
Fixes #2258 (<=== Add issue number here)

Thanks!

pagination

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -1,14 +1,18 @@
+/* eslint-disable complexity */
+/* eslint-disable wrap-iife */
 
 
 (function() {
 
   var types = {
+    'all':      true,
     'note':     true, 
     'question': true, 
     'event':    true, 
-    'comment':  true, 
-    'wiki':     true
+    'comment':  true
   };
+
+  
 
   var viewport = function() {
     var e = window, a = 'inner';
@@ -19,96 +23,23 @@
     return { width : e[ a+'Width' ] , height : e[ a+'Height' ] };
   }
 
-  var setTypeVisibility = function(type, checked) {
-
-    if (type == "all") {
-
-      Object.keys(types).forEach(function(type, i) {
-
-        $('.node-type').prop('checked', checked);
-        setTypeVisibility(type, checked);
-
-      });
-
-    } else {
-
-
-      // record status
-      types[type] = (checked == true);
-
-
-      // record status in browser localStorage
-      if (localStorage) {
- 
-        localStorage.setItem('pl-dash-' + type, checked);
-        // match displayed state to localStorage saved state:
-        $('.node-type-' + type).prop('checked', checked);
- 
-      }
-
-
-      // if all checked?
-      var checked_array = $(".node-type").map(function(i, el) { return $(el).prop('checked'); });
-
-      // if contains some falses:
-      if (checked_array.toArray().indexOf(false) != -1) {
-
-        // if also contains some trues:
-        if (checked_array.toArray().indexOf(true) != -1) {
-
-          $('.node-type-all').prop('indeterminate', true);
-
-        } else {
-
-          $('.node-type-all').prop('checked', false);
-          $('.node-type-all').prop('indeterminate', false);
-
-        }
-
-      } else {
-
-        $('.node-type-all').prop('indeterminate', false);
-        $('.node-type-all').prop('checked', true);
-
-      }
-
-
-      // actually hide/show
-      if (checked && !(type == 'wiki' && viewport().width > 992)) {
- 
-        $('.note-container-' + type).show();
- 
-      } else {
- 
-        $('.note-container-' + type).hide();
- 
-      }
-
-
-      // use CSS clear:left to tidy columns 
-      $('.activity .col-md-6').css('clear', 'none');
-      $('.activity .col-md-6:visible:even').css('clear', 'left');
-
-
-      // change dropdown title 
-      if ($('.activity-dropdown input.node-type:checked').length < $('.activity-dropdown input.node-type').length) {
- 
-        $('.activity-dropdown .dropdown-toggle .node-type-filter').html(I18n.t('js.dashboard.selected_updates'));
- 
-      } else if ($('.activity-dropdown input.node-type:checked').length == 0) {
-
-        $('.activity-dropdown .dropdown-toggle .node-type-filter').html(I18n.t('js.dashboard.none'));
-
-      } else {
- 
-        $('.activity-dropdown .dropdown-toggle .node-type-filter').html(I18n.t('js.dashboard.all_updates'));
- 
-      }
- 
-    }
-
+  if(!viewport().width > 992) {
+    types.wiki = true;
   }
 
+  var setTypeVisibility = function(type, checked) {    
+    var filterTypes = [];
+    for(var i = 0; i < type.length; i++) {
+        if(checked[i]) filterTypes.push(type[i]);
+    }
+    var baseurl = window.location.href;
+    url = new URL(baseurl);
+
+    if(filterTypes.length > 0 || (baseurl.indexOf('types=') > -1)) {
+      url.searchParams.set("types", filterTypes);
+      window.location.href = url;
+    }
+  }
 
   // load any settings from browser storage
   var getLocalStorageActivity = function() {  
@@ -118,10 +49,9 @@
       Object.keys(types).forEach(function(key, i) {
  
         var type = types[key];
-        if (localStorage.getItem('pl-dash-' + key) == null) localStorage.setItem('pl-dash-' + key, true);
-        types[key] = localStorage.getItem('pl-dash-' + key) == "true",
-        setTypeVisibility(key, types[key]);
- 
+        if (localStorage.getItem('pl-dash-' + key) === null) localStorage.setItem('pl-dash-' + key, true);
+        types[key] = localStorage.getItem('pl-dash-' + key) === "true"
+        $('.activity-dropdown li.filter-checkbox').find('[data-type='+ key +']').prop('checked', types[key]); 
       });
  
     }
@@ -144,13 +74,89 @@
 
   });
 
-  $('.activity-dropdown li').click(function(e) {
+  $('.activity-dropdown a').click(function(e) {
+    e.preventDefault();
+    // use CSS clear:left to tidy columns 
+    $('.activity .col-md-6').css('clear', 'none');
+    $('.activity .col-md-6:visible:even').css('clear', 'left');
 
-    e.stopPropagation();
+    // change dropdown title 
+    if ($('.activity-dropdown input.node-type:checked').length < $('.activity-dropdown input.node-type').length) {
+ 
+      $('.activity-dropdown .dropdown-toggle .node-type-filter').html(I18n.t('js.dashboard.selected_updates'));
+ 
+    } else if ($('.activity-dropdown input.node-type:checked').length === 0) {
 
-    setTypeVisibility($(this).find('input').attr('data-type'), $(this).find('input').prop('checked'));
+      $('.activity-dropdown .dropdown-toggle .node-type-filter').html(I18n.t('js.dashboard.none'));
 
+    } else {
+ 
+      $('.activity-dropdown .dropdown-toggle .node-type-filter').html(I18n.t('js.dashboard.all_updates'));
+ 
+    }
   });
+
+  
+
+  function updateFilters() {
+    var types = [];
+    var checked = [];
+    $('.activity-dropdown li.filter-checkbox').each(function () {
+
+      var type = $(this).find('input').attr('data-type');
+      var ischecked = $(this).find('input').prop('checked');
+      if(!ischecked && types[0] === 'all') {
+        checked[0] = false;
+      }
+   
+      
+      if(type !== 'wiki') {
+        types.push(type);
+        checked.push(ischecked);
+      }
+      else if(!(viewport().width > 992)) {
+        types.push(type);
+        checked.push(ischecked);
+      }
+    });
+
+    types.forEach(function(element, index) {
+      var type = element;
+      var ischecked = checked[index];
+      if(index === 0 && checked.slice(1, checked.length).every(x => x)) {
+        ischecked = true;
+        checked[index] = true;
+      }
+      if (localStorage) {
+        localStorage.setItem('pl-dash-' + type, ischecked);
+        $('.node-type-' + type).prop('checked', ischecked);
+      }
+    });
+    updateDropdownTitle();
+    setTypeVisibility(types, checked);
+  }
+
+  $('.activity-dropdown li').click(function(e) {
+    e.stopPropagation();
+    if($(this).find('input').attr('data-type') === 'all') {
+      var all = this
+      $('.activity-dropdown li.filter-checkbox').each(function () {
+        $(this).find('input').prop('checked', $(all).find('input').prop('checked'));
+      });
+    }
+    updateFilters();
+    
+  });
+
+  function updateDropdownTitle() {
+    if ($('.activity-dropdown input.node-type:checked').length < $('.activity-dropdown input.node-type').length) {
+      $('.activity-dropdown .dropdown-toggle .node-type-filter').html(I18n.t('js.dashboard.selected_updates'));
+    } else if ($('.activity-dropdown input.node-type:checked').length === 0) {
+      $('.activity-dropdown .dropdown-toggle .node-type-filter').html(I18n.t('js.dashboard.none'));
+    } else {
+      $('.activity-dropdown .dropdown-toggle .node-type-filter').html(I18n.t('js.dashboard.all_updates'));
+    }
+  }
 
 
   $('.note-wiki, .wikis .wiki').each(function(wiki) {
@@ -230,7 +236,7 @@
       feed: 'https://publiclab.org/rssproxy/plots-dev/',
       //feed: '/home/fetch?url=https://groups.google.com/forum/feed/plots-dev/topics/rss.xml?num=15',
       url:   '/lists#publiclaboratory'
-    },
+    }
   };
 
   var show_list = function (list) {
@@ -265,7 +271,7 @@
         metaEl.find('.date').append(pubDate);
 
         if (author) metaEl.find('.date').append(' by <i>' + author + '</i>');
-        if (list != 'combined') metaEl.find('.date').append(' on <a href="' + lists[list].url + '">' + list + '</a>');
+        if (list !== 'combined') metaEl.find('.date').append(' on <a href="' + lists[list].url + '">' + list + '</a>');
 
       });
 
@@ -293,5 +299,17 @@
     window.location = '/search/' + $('.search-form-wiki input').val();
 
   })
+
+  $(function () {
+    var url_string = window.location.href;
+    var url = new URL(url_string);
+    var params = url.searchParams.get("types");
+    updateDropdownTitle();
+    
+    if(params !== null) {
+      params = params.split(",");
+    }
+
+  });
 
 })();
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -94,21 +94,38 @@ class HomeController < ApplicationController
       .where(type: 'note', status: 1)
       .where('term_data.name = (?)', 'hidden:response')
       .collect(&:nid)
-    notes = Node.where(type: 'note')
+
+    basenotes = Node.where(type: 'note')
       .where('node.nid NOT IN (?)', hidden_nids + [0]) # in case hidden_nids is empty
       .order('nid DESC')
-      .page(params[:page])
-    notes = notes.where('nid != (?)', blog.nid) if blog
+    basenotes = basenotes.where('nid != (?)', blog.nid) if blog
+    notes = basenotes
+
+    questions = basenotes.where(uid: '3')
+
+    events = Tag.find_nodes_by_type('event', 'note', 999999999999999)
+      .where('node.nid NOT IN (?)', hidden_nids + [0])
 
     if current_user && (current_user.role == 'moderator' || current_user.role == 'admin')
       notes = notes.where('(node.status = 1 OR node.status = 4 OR node.status = 3)')
+      events = events.where('(node.status = 1 OR node.status = 4 OR node.status = 3)')
+      questions = questions.where('(node.status = 1 OR node.status = 4 OR node.status = 3)')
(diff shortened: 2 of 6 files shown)


JonathanXu1:
Hey there, thanks for making this pull request! It's a good solution to the issue, but there are some issues that we should fix in the codeclimate so that the checks pass. I'll request some of them in the code.

okonek:
The mechanics of sorting are completely rewritten and the sort operation now happens in the backend, so with every filter change the website is reloaded.

okonek:
@publiclab/mentors Review please?

okonek:
@publiclab/mentors Review please?

okonek:
Ok, I'll try to solve them.

oorjitchowdhary:
I believe if the Travis Ci and Danger checks have passed, we may be able to ignore the CodeClimate issues.. 
@publiclab/reviewers Shall tell the exact steps..
@okonek Thanks for your PR.

okonek:
Now that the codeclimate tests pass I think there is no problem with the PR.

okonek:
@publiclab/reviewers

sashadev-sky:
@okonek I am having issues passing Travis CI tests and I saw in their pull requests bank that you had a similar error to mine which you managed to fix: 

`The command "if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then danger --verbose; fi" exited with 1`

I was wondering if you could tell me what was causing this error for you / how you fixed it please!

okonek:
Ok, just added another commit to this PR, because I just realised a small issue, that is now resolved. Could someone please review my task on GCI?

jywarren:
Wow this is a pretty involved PR! Thanks! Would anyone from @publiclab/reviewers or @publiclab/mentors be able to give it a close review? Thanks!!!!

okonek:
Yes, I worked pretty hard on that one. Please review. I'm waiting for two days.

Gemfile.lock
```diff
@@ -60,7 +60,6 @@ GEM
       scrypt (>= 1.2, < 4.0)
     authlogic-oid (1.0.4)
```
gauravano: Please remove Gemfile.lock from your PR, as we update it only when required. Thanks!

jywarren:
Hi, @okonek - apologies, it takes a bit to review a longer PR, and it looks like many mentors are tied up in exams at the moment. We appreciate your patience. 

@gauravano perhaps we should push this to unstable to test it out live, as well?

app/assets/javascripts/dashboard.js
```diff
@@ -1,15 +1,19 @@
+/* eslint-disable complexity */
+/* eslint-disable wrap-iife */
```
gauravano: I don't think these lines are needed.

gauravano:
Agreed @jywarren, I am pushing it on `unstable`.

jywarren:
Should be testable here: https://unstable.publiclab.org/dashboard

jywarren:
Oh, odd... @icarito did something more change in our subdomains? I can't seem to load unstable as it goes directly to Jenkins...

gauravano:
Let's see how it works after the build finishes .

gauravano:
Unstable is running as build finished.

jywarren:
oh oh oh WOW it's a new feature! Showing Jenkins until it's done!!! Wow!!!

gauravano:
I think I got your approach. Currently, we take all the notes, wikis, comments and then go to dashboard, and pagination show count totaling the sum. When user try to show notes only, or wikis only, other nodes get invisible :eyeglasses: but count remains same. For instance, suppose there are 1000 total notes displaying over 100 pages. And, there are 10 events in them, a event can be anywhere 8th page, 9th page.... Now, if we wish to show only events, other nodes(notes, wikis, questions get invisible) and only events show but on their same positions i.e., on 8th page, 9th page, etc and other pages appear empty with same count of pages because in reality other nodes are there only. 

So, you are reloading page in this PR and loading the content which is checked. Am I right?

jywarren:
What if we started showing the type only once someone begins clicking the
pagination... So normally at /dashboard it would show pagination for all,
but once you start going back to older pages like page=2, it begins to use
your filtering to show more accurate pagination by also appending
types=comments,questions, etc?

It also isn't perfect but it's a good compromise and also a good adaptation
of the code that's now been written?

gauravano:
I am testing the new dashboard on unstable and I feel there are some issues:
1. Everytime checkbox is clicked, the whole page loads
2. Checkbox having value ALL not working
3. I think loading is a issue

@jywarren do you think we should change layout of dashboard something like tabs on profile page - https://publiclab.org/profile/warren . Do you think it can be discussed in Openhour?

jywarren:
Hi @gauravano - yes, ok, i'm seeing the "page reloads on checkbox" -- what do you think of the short term solution of having the "types" parameter only appear when you click on a pagination number? This could be done by parsing the URL for the elements inside the `pagination` element?

```html
<ul class="pagination pagination"><li class="prev disabled"><span>&#8592; Previous</span></li> <li class="active"><span>1</span></li> <li><a rel="next" href="/research?page=2&amp;types=note%2Cevent%2Ccomment">2</a></li> <li><a href="/research?page=3&amp;types=note%2Cevent%2Ccomment">3</a></li> <li><a href="/research?page=4&amp;types=note%2Cevent%2Ccomment">4</a></li> <li><a href="/research?page=5&amp;types=note%2Cevent%2Ccomment">5</a></li> <li><a href="/research?page=6&amp;types=note%2Cevent%2Ccomment">6</a></li> <li><a href="/research?page=7&amp;types=note%2Cevent%2Ccomment">7</a></li> <li><a href="/research?page=8&amp;types=note%2Cevent%2Ccomment">8</a></li> <li><a href="/research?page=9&amp;types=note%2Cevent%2Ccomment">9</a></li> <li class="disabled"><span>&hellip;</span></li> <li><a href="/research?page=140&amp;types=note%2Cevent%2Ccomment">140</a></li> <li><a href="/research?page=141&amp;types=note%2Cevent%2Ccomment">141</a></li> <li class="next"><a rel="next" href="/research?page=2&amp;types=note%2Cevent%2Ccomment">Next &#8594;</a></li></ul>
```

Rather than having those params be added upon clicking a checkbox? @okonek what do you think of this solution too, to avoid the page refresh?

jywarren:
Just a note that I'm pushing another branch to unstable now!

sashadev-sky:
@okonek I resolved it by rebasing - it was because my feature branch was stale :see_no_evil:. I really appreciate the follow up  though! Thank you.

okonek:
I'm sorry but I dont get you. How would you like to assign the parameters on the checkbox without refresh.

okonek:
And yes gauravano is right about how I approached it in my PR. Thanks for lots of suggestions.

okonek:
Ok, I've removed the Gemfile.lock as @gauravano asked, but now there is a conflict. What should I do?

okonek:
What should I do with this Gemfile error? I removed the file as @gauravano  said.

gauravano:
I meant - Remove the Gemfile.lock from PR i.e, changes you done in the file. Copy the Gemfile.lock from master branch and make a commit so that git doesn't record any changes.

okonek:
Is it ok now for my task to be approved?

okonek:
> Hi @gauravano - yes, ok, i'm seeing the "page reloads on checkbox" -- what do you think of the short term solution of having the "types" parameter only appear when you click on a pagination number? This could be done by parsing the URL for the elements inside the `pagination` element?
> 
> ```
> <ul class="pagination pagination">
>   <li class="prev disabled"><span>&#8592; Previous</span></li>
>   <li class="active"><span>1</span></li>
>   <li><a rel="next" href="/research?page=2&amp;types=note%2Cevent%2Ccomment">2</a></li>
>   <li><a href="/dashboard?page=3&amp;types=note%2Cevent%2Ccomment">3</a></li>
>   <li><a href="/dashboard?page=4&amp;types=note%2Cevent%2Ccomment">4</a></li>
>   <li><a href="/dashboard?page=5&amp;types=note%2Cevent%2Ccomment">5</a></li>
>   <li><a href="/dashboard?page=6&amp;types=note%2Cevent%2Ccomment">6</a></li>
>   <li><a href="/dashboard?page=7&amp;types=note%2Cevent%2Ccomment">7</a></li>
>   <li><a href="/dashboard?page=8&amp;types=note%2Cevent%2Ccomment">8</a></li>
>   <li><a href="/dashboard?page=9&amp;types=note%2Cevent%2Ccomment">9</a></li>
>   <li class="disabled"><span>&hellip;</span></li>
>   <li><a href="/dashboard?page=140&amp;types=note%2Cevent%2Ccomment">140</a></li>
>   <li><a href="/dashboard?page=141&amp;types=note%2Cevent%2Ccomment">141</a></li>
>   <li class="next"><a rel="next" href="/dashboard?page=2&amp;types=note%2Cevent%2Ccomment">Next &#8594;</a></li>
> </ul>
> ```
> 
> Rather than having those params be added upon clicking a checkbox? @okonek what do you think of this solution too, to avoid the page refresh?

I understand what you are thinking of but then on the first page there could be no content if we would filter.

jywarren:
ah but what if we make it so that the default (if there is no parameter for
types) is to show all content, OR to preserve the client-side JavaScript
logic which was hiding some types of content? I guess the second option
would be preferable. That way all is actually generated, but JS/CSS will
hide the ones that are not checked. Then, we use the filtering on any
subsequent paginated pageloads.

How does this sound? This is a pretty tangled up issue, and we appreciate
your time on it!

jywarren:
Re: `How would you like to assign the parameters on the checkbox without refresh?` -- the JavaScript code had been hiding non-selected types. Under your new controller-based code this would only be necessary on the first pageload at /dashboard, when there are no parameters. But it still seems important, as taht's perhaps the most common page people will view, does that make sense?

okonek:
Yes, I implemented the thing, that when there is a blank /dashboard everything is shown without reloading.

okonek:
Doesn't it work for you?

okonek:
So, the only issue for you is that the page reloads on every checkbox check, right?
I really couldn't come up with any other idea.

jywarren:
Well but also that when you go back to /dashboard, it should remember what
checkboxes were checked from last time - that's the purpose of the
localstorage code. Your code is great! I just think we should apply the URL
parameters only to the pagination links, not upon clicking the checkboxes
(which we can stick with the JavaScript code for). That way, pagination
will work perfectly once you click the pagination links, because of your
code. What do you think?

okonek:
The localstorage works as it's supposed to. It remembers the checked state after closing the website.

okonek:
@jywarren I understand the idea of adding the parameters only on the pagination click and otherwise filtering in front, but there would still occur situation like that or worse, that there are only two or one or none nodes on the page.
jp2

okonek:
But if you think that's ok. I can do that.

okonek:
@jywarren What do you say?

jywarren:
I hear your concern, but think this is currently the case, and hasn't caused too much trouble -- it's something we could try to resolve in a separate PR.

The issue I'd like to bypass by re-using the JavaScript code is to avoid the current scenario where when you go to `/dashboard`, the page gets reloaded to `/dashboard?types=question%2Cevent%2Ccomment` immediately. I think we can make the most of the code you've written by having it not immediately reload, but instead use CSS/JS in the current page, but append the `types` parameter to the pagination links. 

Please go ahead with this solution and thanks a ton for working through this carefully with us!

okonek:
If you're sure about that, ok then.

okonek:
Well, I've done what you asked for. Normally the old JS code is filtering the nodes and when the pagination page is changed my controller-based filtering comes in. But I just realized a major issue that comes with this solution. 

Let's suppose, we have only two types of nodes on our page. Lots of events and only one research note which is the last, so it's on the third page. You are on the first page and filter to show only research notes. There is nothing left on the site (which is actually not a good solution). You go to the next page clicking the pagination link and even worse thing happens. The nodes are now filtered by the controller and since we have only one research note, only the first page of pagination exists and we request the second, so we end up not just with an empty page, but also not having a pagination bar at the bottom, because we requested the page that doesn't exist.

I think we should stick completely to mine solution or stay with the current because this hybrid is not a good idea. I also left a gif for the case I've described because I know that not everyone will have time to read it.

Let me know what do you think and what should I do to resolve this issue and solve this task so I can do more before the competition ends.
Thanks for all your help.
peek 2018-11-29 18-59

jywarren:
I see what you're saying. How about another potentially simple solution -- if we aren't showing any results, we display a notice that says 

"No results - try [broadening your filter]() to see more."

which would check "All" and reload if necessary?

Finding and documenting this edge case was very thoughtful. Thanks for being extremely thorough.

okonek:
I still urge to stay with my first solution. Why don't you like it? I mean it's better for the user to reload on every checkbox than to have this kind of errors.
I can also add a feature to my solution, that instead of reloading on every checkbox select you check how many checkboxes you like and then press a button which reloads and applies the filters. You know what I mean. Let me know what do you think about it.

jywarren:
Hi, @okonek - thanks for your patience. The main reason I prefer the solution you've finished (as [mentioned in this comment](https://github.com/publiclab/plots2/pull/4042#issuecomment-442932819), and adding the `"No results - try broadening your filter to see more."` message. First, it doesn't require a page refresh upon arriving at `/dashboard` if there are selections made (the javascript just displays the correct types). I think the page refreshing will be pretty disruptive to usability, i'm sorry to say! Would you please add the message above, and we can merge this? 

The only other solution I can think of is to store the settings in the session, or in the database, so that the server side can "know" to display the right types, and we wouldn't have to rely on JavaScript. But the JavaScript solution has been in place for a long time and people haven't had a major problem with it, so I think we might as well stick with what works rather than introduce a page refresh that happens a lot.

Thanks for working on this one, I think we can wrap it up now!

jywarren:
Yes, that's fine with me! Although I've love to see the latest code Jan
added. Thanks!

okonek:
@jywarren I think I resolved the issue perfectly now. With small changes page now fetches filtered data using AJAX, so there is no reloading. I also add a gif of how it works now. If there are no code errors, I think it is mergable.
Thanks for all your help.
ajax

okonek:
I also resolved the issue of filtering questions.

app/assets/javascripts/dashboard.js
```diff
@@ -19,109 +19,56 @@
...
-
-          $('.node-type-all').prop('checked', false);
-          $('.node-type-all').prop('indeterminate', false);
+  // Adding a wiki type if on mobile
+  if(!viewport().width > 992) {
+    types.wiki = true;
+  }
 
+  $("#notes").click(function(e) {
+    if(!$.trim($(this).find("#notes_container").html()).length) {
```
jywarren: Hmm, what is this section doing? If it's empty, then clicking on it will check the "all" box?

app/assets/javascripts/dashboard.js
```diff
@@ -130,28 +77,111 @@
...
   // allow native stylesheets to determine visibility on 
   $(window).on('resize', function() {
-
     $('.activity .col-md-6').css('display', '');
     getLocalStorageActivity();
-
   });
 
 
   $('.activity-dropdown .dropdown-toggle').click(function(e) {
```
jywarren: OK, this is now empty - shall we remove it?

app/controllers/home_controller.rb
```diff
@@ -94,21 +94,39 @@ def activity
...
+
+    basenotes = Node.where(type: 'note')
       .where('node.nid NOT IN (?)', hidden_nids + [0]) # in case hidden_nids is empty
       .order('nid DESC')
-      .page(params[:page])
-    notes = notes.where('nid != (?)', blog.nid) if blog
+    basenotes = basenotes.where('nid != (?)', blog.nid) if blog
+    notes = basenotes
+
+    questions = Tag.find_nodes_by_type('question:question', 'note', 999_999_999_999_999)
```
jywarren: Here you should be able to pass `nil` and the limit will be lifted. But unfortunately, this will query the database for all `question` nodes, and likewise below all `event` nodes and above all `notes` -- on the production server this will be thousands of records, so we need to optimize for it to work. The `.page()` method is quite sophisticated, and uses `limit()` and `offset()` to select just the right records to display on the given page. But the problem we're encountering here is that we are trying to subselect across different tables. I don't believe the `will_paginate` gem can do this. 

What we could do here is to start by fetching the paginated notes, by leaving in `.page()` for that query. For events and questions, which are also nodes, we could try to filter them in/out if they are not selected... this could be difficult but we could re-use some of the code in Tag.find_nodes_by_type, maybe... 

Then, we could use the timestamps from the nodes we fetched to select the comments which were made during the same period. These could be added in, and while the calculation of total # of pages would be correct, the number of items per page would vary since we would be calculating # of pages based only on nodes. 

I'm trying to think of how this could be done more simply but it was already a pretty complex thing to display in the first place... the critical part now is to not return the entire table's worth of each type on each page load. This works on a very small dataset but would almost certainly crash the server when 1000s of records are returned each time in production.

jywarren: https://stackoverflow.com/a/7959348 suggests synchronizing the `offset`, but then over time the comments and nodes would drift because they don't occur at the same rate over time.

jywarren: OK, here's an idea. I don't know if it's good, but just to think through it...

Let's simplify - let's put wiki edits permanently in the sidebar and display them below in mobile view, so we just don't worry about wikis as part of this. Then we're left with Notes, Events, Questions (all types of Node) and Comments.

The first three we could probably filter if we wanted based on what tags they have, using joins.  But Comments messes us up, because it's a different table (see my above comment). What if we, instead of showing pure comments, showed only Nodes in Activity, but we joined comments where they exist (let's forget Answer comments for now) and we sorted by `comment.timestamp` and then `node.created`: https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-order -- would that work? And then in the views, we detect if there is a comment attached and we display the `X commented on Y Z time ago` instead of the usual note template? 

This isn't air-tight, but seems plausible. I'm sorry I missed the change in pagination from the very first commit. I'm going to think on this more and happy to hear your thoughts too.

jywarren: In the above scenario, the right type of join will return multiple duplicate Node records, but one for each comment, and ordered by comment timestamp. I think that's fine, because each would really be a "stand in" for the associated comment record. But alternatively we could try to use a join type which would return only unique/discrete Node records, but choose the most recent associated comment record. Then we could show the Node template on the dashboard, and kind of "attach" the most recent comment to the bottom of it, so as to show why it's appearing in the timeline at that point. Just thinking through possibilities here.

jywarren: Finally, because this is already quite complex, we could stop offering to filter by Event and Question, simply grouping all Nodes together. That would simplify things a good bit, leaving only All, "Posts" (we'll call them?), and Comments. Trying to think of how to make this a reasonable path forward.

app/assets/javascripts/dashboard.js
okonek: Yes, that's what it's doing.

app/controllers/home_controller.rb
okonek: You know what? I have a simpler idea. I know it's not that great but it's better than fetching thousands. We could just fetch the first 37 from all of the tables, because it's the nodes limit for one page. You know what I mean? It's a temporary solution, but it's not that bad and it could be solved in another PR.

jywarren:
@publiclab/reviewers @publiclab/mentors we ran into a REALLY tough issue in @okonek's PR here... a true puzzle trying to develop some solutions for. @okonek has done some tremendous work on the problem but it's revealed an even larger challenge for database query optimization on the dashboard. If anyone likes that kind of problem, we'd be grateful for some ideas or input; see my comment at: https://github.com/publiclab/plots2/pull/4042/files#r238070635

app/assets/javascripts/dashboard.js
okonek: What do you mean? What should be removed?

app/controllers/home_controller.rb
jywarren: I think this is doable, and I agree with the principle of doing something now and following up in later PRs. But this would start to go out of time sync pretty fast because there are more comments per time period than nodes. Could we try the simpler solution but using timestamps instead of record count?

okonek:
@jywarren Could you accept limiting every query to 37 nodes? The maximum node count per page? Please, I'd like to be done with this issue.

app/controllers/home_controller.rb
jywarren: So you'd narrow your query with a `where()` and I think you can even pass a range to `where()` in activerecord... then it'd be like `.where(timestamp: basenotes.first.created..basenotes.last.created)` i think?

jywarren:
Check out my comment above -- limiting by timestamp instead -- i think that is pretty good. 

If you are really eager to move on to another issue, I totally understand. This one was extremely challenging, to tell the truth, especially with the added issue of database optimization. Your solution is super nice - the Ajax is really wonderful. It's just an issue which, if we could go back, perhaps should have been solved in smaller parts and architected more thoroughly.

okonek:
And could you tell me why limiting by 37 is bad? I mean, I can't find a case where it wouldn' work. Thanks

jywarren:
Yes I will but it'll have to be tonight as I have some work now. But it's
just like limiting to 37 but using a time boundary instead of a count. I
can offer some more code hints tonight however. Thanks, and especially
since we did not describe this particular challenge in the original issue,
don't worry, you'll definitely be getting credit for this PR!

Thanks again,
J

okonek:
I'm just thinking loud, but I think that neither my solution or yours would work, because this limiting would break the pagination if we would like to parse an old site like 4, 5 etc

okonek:
I really don't know what to do. I'm stuck with this issue for very long

jywarren:
Don't worry. I'll spend some time on this tonight.

okonek:
Thank you very much for a tremendous help on that one. I can't say I was working on this alone. This problem is a lot bigger than anyone could thought. Anyways thank you for all your help and I am sure I've learned a lot from this issue.
John

SidharthBansal:
@okonek as you have worked on this challenge for quite a long time. We want to give you credit for this challenge. You can claim the task, I will approve it.
Thanks for tremendous amount of work

SidharthBansal:
@jywarren has already given you points.

PR #4054, opened 2018-11-26
Combined wiki/note grids

### LINKED ISSUE #3758: Combined wiki/note grids using [node:_____] syntax, for inline grids

ValentinaTironi:
Fixes #3758 

Here is a gif showing how it works.
wiki_notes


jywarren:
This is amazing!!! Fantastic work. I'm merging this now! If you'd like to contribute a follow-up, it could be helpful to implement a test similar to one of these: 

https://github.com/publiclab/plots2/blob/master/test/unit/node_shared_test.rb

jywarren:
Awesome!!! 👍 👍 👍

ValentinaTironi:
@jywarren Thank you so much! I can try do the test!

jywarren:
awesome!

ValentinaTironi:
@jywarren Hi! I had complications when I did `git push`, I had to do a `git rebase` so, when I did `git push` again github pushed all commits again. So, I had to do a new pull request with test and code. I'm so sorry for inconveniences. Here is the new pr: https://github.com/publiclab/plots2/pull/4056/

PR #4056, opened 2018-11-26
Combined wiki/note grids

### LINKED ISSUE #3758: Combined wiki/note grids using [node:_____] syntax, for inline grids

ValentinaTironi:
Fixes #3758 

Here is a gif showing how it works.
wiki_notes

Test works:
screen shot 2018-11-26 at 5 23 00 pm


ValentinaTironi:
Here @jywarren

jywarren:
Fantastic work! And so fast!

jywarren:
I'm not sure if they're still working on it but there was an issue with thumbnail grid display -- formatting it with Bootstrap -- maybe you'd be interested in helping out? 

https://github.com/publiclab/plots2/pull/3970#issuecomment-439167488

Right now the table CSS they were using is actually disrupting other table display, so it'd be preferable to move that to a bootstrap-based layout as in the comment i left. If you're interested, we could definitely use your help!

jywarren:
if you'd like we can make a google code-in issue for it?

ValentinaTironi:
Omg, it would be really great! Thank u 💯

jywarren:
https://codein.withgoogle.com/dashboard/tasks/4957100062015488/ go for it! 👍

ValentinaTironi:
I can't claim that task without approve this in gci! (COMBINED WIKI/NOTE GRIDS USING [NODE:_____] SYNTAX, FOR INLINE GRIDS)

jywarren:
Approving now!

ValentinaTironi:
Thank you Jeff! I'm gonna work in that issue now!

PR #4085, opened 2018-11-30
Test for power tag embedded grids

jywarren:


jywarren:
This should be testable at: 

* https://publiclab.org/embed/grid/activities:coqui
* https://publiclab.org/embed/grid/nodes:purpleair
* https://publiclab.org/embed/grid/notes:grid:coqui

PR #4099, opened 2018-12-02
yarn upgrade test

jywarren:


jywarren:
Tried npm publishing LEL new version; we'll see if it gets published in plots2. If not, this will explicitly grab it.

Issue #547, opened 2016-05-27
photos with chinese characters don't upload as Main Image


ebarry:
![17](https://cloud.githubusercontent.com/assets/161439/15610329/af53573e-23f2-11e6-901a-c0a3d1395be6.jpg)

### Relevant URLs

https://publiclab.org/notes/Shizi/05-27-2016/kite-aerial-photography-in-china

### PublicLab.org username

(to help reproduce the issue)

### Browser, version, and operating system

For bug reports, fill out the above template; for feature requests, you can delete the template.

To learn how to write really great issues, which increases the chances they'll be resolved, see:

https://publiclab.org/wiki/developers#Contributing+for+non-coders

jywarren:
Can you post the original name of the image, since it's since been replaced on the linked-to research note? Image uploading uses the `paperclip` gem, and the first step is probably to figure out whether the image:

1. uploaded but does not display, or
2. failed on upload and never got to display. 

I think 1 is probably right, as I saw a "broken" image displayed when this note first published.

ebarry:
ah, i thought uploading it to github would preserve the orignal file name. i guess not. 

图片17.jpg

jywarren:
Hi, Liz, when you have a chance, can you add the steps needed to reproduce
this error? This will make it easier for people to understand and fix the
bug.

ananyo2012:
Yes the image is being uploaded but the name is displayed incorrectly as you can see in this test research note https://publiclab.org/notes/ananyo2012/08-19-2016/testing-out-a-note-within-chinese-characters-in-main-image. This is weird as the same image shows up when you upload it in the body.

ananyo2012:
This is the ASCII format of the image name: `%E5%9B%BE%E7%89%8717.jpg`. if you paste this name beside the image path i.e. https://i.publiclab.org/system/images/photos/000/017/679/original/%E5%9B%BE%E7%89%8717.jpg you will be able to see the uploaded image. But somehow the name is displayed incorrectly in the note. The name of the image is fetched by `@node.main_image.path(:original)` as you can see in this line: https://github.com/publiclab/plots2/blob/master/app/views/notes/show.html.erb#L7

ebarry:
Hi all, just bumping this issue up as image upload -- if the title has non-latin characters -- to the Main Image, as well as Drag-n-Drop into the Body of a research note, is failing: 
[...]

PR #4118, opened 2018-12-04
Unicode filenames

jywarren:
Attempting to write a test for #547 

Still needs a solution!


jywarren:
Oh, the error here is:

```
ERROR["test_image_creation_with_unicode_filename", #<Minitest::Reporters::Suite:0x00005654eb1d8e40 @name="ImagesControllerTest">, 63.142930936000084]
 test_image_creation_with_unicode_filename#ImagesControllerTest (63.14s)
RuntimeError:         RuntimeError: /app/test/fixtures/图片17.jpg file does not exist
            test/functional/images_controller_test.rb:47:in `block in <class:ImagesControllerTest>'
```

but we wanted an encoding error... not ready for a fix yet! but the fix would hopefully be as in https://github.com/publiclab/plots2/pull/3007/ --

```ruby
class ChangeImageTableToUtf8mb4 < ActiveRecord::Migration[5.2]
   def change
    # changing table that will store unicode execute:
    execute "ALTER TABLE images CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
    # changing string/text column with unicode content execute:
    execute "ALTER TABLE images MODIFY title VARCHAR(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
     # changing string/text column with unicode content execute:
    execute "ALTER TABLE images MODIFY photo_file_name VARCHAR(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
 end
end
```

jywarren:
Ah -- method `ImagesController#create as JSON

jywarren:
Hmm, still can't reproduce the error! ☹️

gauravano:
The image is stored in the db 

Screenshot from 2019-03-22 22-49-44

So, should we focus on giving a push at https://github.com/publiclab/plots2/pull/3007 @jywarren?

jywarren:
Hm, part of the problem was that we weren't able to reproduce the error we saw on production for any of these encoding bugs. I think we can go ahead with all the merges but we have no easy way to track if the solution was related to our fixes. But, we've been stuck on this for some time... so maybe?

What do you think, should we merge this?

gauravano:
Actually, the screenshot I copied here https://github.com/publiclab/plots2/pull/4118#issuecomment-475706615 was from master and not for this PR. 

What do you think about https://github.com/publiclab/plots2/pull/3007?

jywarren:
😢 yeah, we are stuck because although we can add the db encoding, we can't "catch" the error itself in tests at all. Not sure why this is the case! https://github.com/publiclab/plots2/pull/3007#issuecomment-461898604

Do you think we should just go ahead and do this encoding fix on all tables and just live with the fact that we can't protect it with a test?

jywarren:
Closing in favor of @icarito fixes...

PR #4136, opened 2018-12-06
/api/srch/profiles: search only by user tag 

### LINKED ISSUE #3506: Profiles search page: search by user tag

milaaraujo:
Fixes #3506

/api/srch/profiles: Adding the option to search for user tags.

`/api/srch/profiles?query=[tag_name]&field=tag`

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -43,9 +43,7 @@ All the endpoints have the optional parameter `limit` (10 by default) where you 
   `order_direction=[string]`: It accepts `ASC` or `DESC` (the latter is the default).
 
   `field=[string]`: Accepts the value `username` for searching profiles only by
-   the field `username`.
-
-   `tag=[string]`: The search can be refined by passing a tag field.
+   the field `username`. And accepts the value `tag` for searching profiles by tag.
 
 ### Notes
 
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -88,7 +88,7 @@ module Srch
                                            nickname: 'search_profiles'
 
       params do
-        use :common, :sorting, :ordering, :field, :additional
+        use :common, :sorting, :ordering, :field
       end
       get :profiles do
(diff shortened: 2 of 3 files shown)


doc/API.md
```diff
@@ -43,9 +43,7 @@ All the endpoints have the optional parameter `limit` (10 by default) where you
   `order_direction=[string]`: It accepts `ASC` or `DESC` (the latter is the default).
 
   `field=[string]`: Accepts the value `username` for searching profiles only by
-   the field `username`.
-
-   `tag=[string]`: The search can be refined by passing a tag field.
+   the field `username`. And accepts the value `tag` for searching profiles by tag.
```
jywarren: ah but the value `tag` is removed now, no? Or maybe I mis-read?

milaaraujo:
I removed the ‘tag’ option. But to search by tag it is necessary to send the *value* “tag” in the option ‘field’: ‘/api/srch/profiles?query=[tag_name]&field=tag’

In the same way that it is necessary to send  ‘/api/srch/profiles?query=[search_query]&field=username’ to search only by username (instead of username and bio).

Make sense?

jywarren:
Ah, yes, i see. Great. Merging!

jywarren:
And can we make sure that option is enabled in the profile search results page by default? Thank you!

milaaraujo:
>And can we make sure that option is enabled in the profile search results page by default? Thank you!

I think this is what @cesswairimu is working on (3911)!

jywarren:
oh awesome. This is all coming together!

PR #4149, opened 2018-12-06
Add Two Tag Unit Test For Search

### LINKED ISSUE #4125: Make a new Test to test for two-tag matches for search
This is a broken up part of this issue (#4121)
We are trying to test if the search gets matches that have two tags that equal the search.

https://github.com/publiclab/plots2/blob/26079dbbc10e13ce1e1874b50faac3de1e47f8ea/test/unit/node_test.rb#L52-L60

dinaelhanan:
Fixes #4125 (<=== Add issue number here)

### DIFF
--- a/test/fixtures/node_tags.yml
+++ b/test/fixtures/node_tags.yml
@@ -10,6 +10,12 @@ awesome:
   nid: 1
   date: <%= DateTime.now.to_i %>
 
+ cool:
+  tid: 2
+  uid: 2
+  nid: 1
+  date: <%= DateTime.now.to_i %>
+
 awesome2:
   tid: 2
   uid: 2
--- a/test/fixtures/node_tags.yml
+++ b/test/fixtures/node_tags.yml
@@ -10,7 +10,7 @@ awesome:
   nid: 1
   date: <%= DateTime.now.to_i %>
 
- cool:
+cool:
   tid: 21
   uid: 2
   nid: 1
--- a/test/fixtures/node_tags.yml
+++ b/test/fixtures/node_tags.yml
@@ -11,7 +11,7 @@ awesome:
   date: <%= DateTime.now.to_i %>
 
  cool:
-  tid: 2
+  tid: 21
   uid: 2
   nid: 1
   date: <%= DateTime.now.to_i %>
--- a/test/fixtures/tags.yml
+++ b/test/fixtures/tags.yml
@@ -80,6 +80,6 @@ unpublished-note-tag:
   tid: 20
   name: unpublished-note-tag
   
- cool:
+cool:
   tid: 21
   name: cool
--- a/test/fixtures/tags.yml
+++ b/test/fixtures/tags.yml
@@ -80,6 +80,6 @@ unpublished-note-tag:
   tid: 20
   name: unpublished-note-tag
   
- awesome_cool:
+ cool:
   tid: 21
-  name: awesome_cool
+  name: cool
--- a/test/fixtures/tags.yml
+++ b/test/fixtures/tags.yml
@@ -5,7 +5,7 @@ test:
 awesome:
   tid: 2
   name: awesome
-
+  
 spam:
   tid: 3
   name: spam
@@ -79,3 +79,7 @@ co-author:
 unpublished-note-tag:
   tid: 20
   name: unpublished-note-tag
+  
+ awesome_cool:
+  tid: 21
+  name: awesome-cool
--- a/test/fixtures/tags.yml
+++ b/test/fixtures/tags.yml
@@ -82,4 +82,4 @@ unpublished-note-tag:
   
  awesome_cool:
   tid: 21
-  name: awesome-cool
+  name: awesome_cool
--- a/test/unit/node_test.rb
+++ b/test/unit/node_test.rb
@@ -210,11 +210,9 @@ class NodeTest < ActiveSupport::TestCase
     assert !node.node_tags.empty?
     assert_not_nil node.tagnames
     assert node.tagnames.first.is_a?(String)
-    assert_equal 'test awesome spectrometer activity:spectrometer', node.tagnames.join(' ')
-    assert_equal 'test cool spectrometer activity:spectrometer', node.tagnames.join(' ')
+    assert_equal 'test awesome cool spectrometer activity:spectrometer', node.tagnames.join(' ')
     # used to generate CSS classes:
-    assert_equal 'tag-test tag-awesome tag-spectrometer tag-activity-spectrometer', node.tagnames_as_classes
-    assert_equal 'tag-test tag-cool tag-spectrometer tag-activity-spectrometer', node.tagnames_as_classes
+    assert_equal 'tag-test tag-awesome tag-cool tag-spectrometer tag-activity-spectrometer', node.tagnames_as_classes
   end
 
   test 'should have subscribers' do
--- a/test/unit/node_test.rb
+++ b/test/unit/node_test.rb
@@ -62,10 +62,10 @@ class NodeTest < ActiveSupport::TestCase
   test 'node mysql native fulltext search returning two tag-based matches' do
     assert Node.count > 0
     if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
-      nodes = Node.search(query: 'awesome',query: 'cool', limit: 1000)
+      nodes = Node.search(query: 'awesome cool' limit: 1000)
       assert_not_nil nodes
-      assert nodes.length > 0
-      assert_equal nodes.length, Tag.find_nodes_by_type('awesome','cool', ['note', 'page']).length
+      assert_equal 1, nodes.length
+      assert_equal nodes(:awesome-cool), nodes.first
     end
   end
 
--- a/test/unit/node_test.rb
+++ b/test/unit/node_test.rb
@@ -58,6 +58,16 @@ class NodeTest < ActiveSupport::TestCase
       assert_equal nodes.length, Tag.find_nodes_by_type('awesome', ['note', 'page']).length
     end
   end
+  
+  test 'node mysql native fulltext search returning tag-based matches' do
+    assert Node.count > 0
+    if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
+      nodes = Node.search(query: 'awesome',query: 'cool', limit: 1000)
+      assert_not_nil nodes
+      assert nodes.length > 0
+      assert_equal nodes.length, Tag.find_nodes_by_type('awesome','cool', ['note', 'page']).length
+    end
+  end
 
   test 'create a node' do
     # in testing, uid and id should be matched, although this is not yet true in production db
--- a/test/unit/node_test.rb
+++ b/test/unit/node_test.rb
@@ -65,7 +65,7 @@ class NodeTest < ActiveSupport::TestCase
       nodes = Node.search(query: 'awesome cool' limit: 1000)
       assert_not_nil nodes
       assert_equal 1, nodes.length
-      assert_equal nodes(:awesome-cool), nodes.first
+      assert_equal nodes(:awesome_cool), nodes.first
     end
   end
 
--- a/test/unit/node_test.rb
+++ b/test/unit/node_test.rb
@@ -59,7 +59,7 @@ class NodeTest < ActiveSupport::TestCase
     end
   end
   
-  test 'node mysql native fulltext search returning tag-based matches' do
+  test 'node mysql native fulltext search returning two tag-based matches' do
     assert Node.count > 0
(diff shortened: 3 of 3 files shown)


dinaelhanan:
@publiclab/reviewers please check this out!

dinaelhanan:
@SidharthBansal it looks like there's an error, can you help? thanks!

jywarren:
I think we'll have to resolve this bug - note Danger bot is highlighting
where it's happening -- but even once we debug these tests, remember that
they should be highlighting a feature that doesn't yet exist, right? So it
will still fail until the new feature is written. Familiar with Test-Driven
Development? (TDD)? Try googling it, it's a cool strategy for code
development!

dinaelhanan:
@jywarren oh that makes a lot more sense! So it's telling us that it didn't find anything (which we know it won't since that's the issue we are trying to solve, it not reading two tags) is that right? The more I read about it, the more it makes sense. Cool! I've never really been exposed to this side of things 'till now :)

dinaelhanan:
@SidharthBansal I have linked the PR and issue. Please check it out, thanks! 
https://codein.withgoogle.com/dashboard/task-instances/4824240835526656/

jywarren:
Awesome! Yes, it's a really cool design pattern because we can use a test
to exactly describe a problem we want to solve... then when the test
passes, it's both solved, and documented (with a real example -- the test)
-- and also protected against future code changes!

test/unit/node_test.rb
```diff
@@ -58,6 +58,16 @@ class NodeTest < ActiveSupport::TestCase
       assert_equal nodes.length, Tag.find_nodes_by_type('awesome', ['note', 'page']).length
     end
   end
+  
+  test 'node mysql native fulltext search returning two tag-based matches' do
+    assert Node.count > 0
+    if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
+      nodes = Node.search(query: 'awesome',query: 'cool', limit: 1000)
```
jywarren: Ah, hmm, i wonder if here one query would override the other. Perhaps its' best to just submit a single query for, say, `awesome cool`?

test/unit/node_test.rb
```diff
@@ -58,6 +58,16 @@ class NodeTest < ActiveSupport::TestCase
...
+      assert_not_nil nodes
+      assert nodes.length > 0
+      assert_equal nodes.length, Tag.find_nodes_by_type('awesome','cool', ['note', 'page']).length
```
jywarren: Likewise, i don't think `Tag.find_nodes_by_type` is set up to accept 2 separate strings. So this part of the test might need to be simpler -- just looking through the test data in `/test/fixtures/` to find a test node in the test fixtures that has 2 tags. Then the test would be expected to find just one node, and you could even do a comparison like:

```ruby
assert_equal 1, nodes.length
assert_equal nodes(:somenodefromfixtures), nodes.first
```

see how you can fetch a node from fixtures with the `nodes(:_____)` syntax. Fixtures are our test data and are described in flat `.yml` files and are loaded into the database when tests are run. Make sense? 

Great start here! Nicely written test! 👍

jywarren:
So after the above changes, we should probably get a test failure from Travis that says something like `expected 1 but found 0` for the assertion that `nodes.length` is 0. Then we'd know that our test is asserting the right thing -- that the correctly-formed search for a node using 2 different tags in a string did not actually return the node that has both those tags.

test/unit/node_test.rb
dinaelhanan: I think I understand... I tried to do this in the changes :)

test/fixtures/tags.yml
```diff
@@ -78,4 +78,8 @@ co-author:
 
 unpublished-note-tag:
   tid: 20
-  name: unpublished-note-tag
\ No newline at end of file
+  name: unpublished-note-tag
+  
+ awesome_cool:
```
jywarren: Ah but actually we want to have the test return if one node has two separate tags, "awesome" and "cool" -- given a single query "awesome cool" -- make sense?

dinaelhanan: I think so.. see my new changes :)

test/unit/node_test.rb
```diff
@@ -58,6 +58,16 @@ class NodeTest < ActiveSupport::TestCase
...
     end
   end
+  
+  test 'node mysql native fulltext search returning two tag-based matches' do
+    assert Node.count > 0
+    if ActiveRecord::Base.connection.adapter_name == 'Mysql2'
+      nodes = Node.search(query: 'awesome cool' limit: 1000)
+      assert_not_nil nodes
+      assert_equal 1, nodes.length
+      assert_equal nodes(:awesome_cool), nodes.first
```
jywarren: OK almost there! So now, you have to create an entry in the `fixtures/node_tags.yml` file to link your new `cool` tag to just one node that already has the `awesome` tag linked. Then you'll have just a single node with both tags linked. Make sense? Great progress here!

dinaelhanan: @jywarren not sure I understand. Do you mean like:
```
awesome_cool:
  tid:22
  name: awesome
  name: cool
```
or
```
 awesome_cool:
  tid: 22
  name: awesome, cool
```
or do I add parent: awesome_cool to each awesome and cool tag?

jywarren: Well I think theres already an awesome tag so you can just add cool. But in the other file, /test/fixtures/node_tags.yml, each entry links a tag to a node. So you'll need to add an entry there to ensure both distinct tags from this file are linked to a single node. Make sense?

dinaelhanan: @jywarren I think I get it! Check out my recent changes

dinaelhanan: @jywarren I don't know why I'm getting these errors.... :/

test/fixtures/node_tags.yml
```diff
@@ -10,6 +10,12 @@ awesome:
   nid: 1
   date: <%= DateTime.now.to_i %>
 
+ cool:
```
jywarren: yes, perfect!

test/unit/node_test.rb
jywarren: OK, your tags are great now. You've linked your `cool` tag to node with id=1. So on this line now, you'll want to assert that it's the same as this node: `nodes(:one)` -- that's this one: https://github.com/publiclab/plots2/blob/master/test/fixtures/nodes.yml#L1-L11

jywarren: Are you able to read the errors? If you click the little X next to your commit, you'll be able to read the full output. In this case, it looks like you may be missing a parenthesis?

```
/usr/local/bundle/gems/skylight-core-3.1.2/lib/skylight/core/probes.rb:119:in `require': /app/test/unit/node_test.rb:65: syntax error, unexpected tIDENTIFIER, expecting ')' (SyntaxError)
ch(query: 'awesome cool' limit: 1000)
```

https://travis-ci.org/publiclab/plots2/builds/465228995#L2953-L2954

dinaelhanan: I was missing a comma! That's why it expected the bracket to end after 'awesome cool'

dinaelhanan: Ok, @jywarren I think it's good now :)

SidharthBansal:
Reopening the pr as travis tests are failing and Jeff has approved the pr. Let's hope test cases will pass

dinaelhanan:
@SidharthBansal note that if the travis tests are failing and saying the below than that's correct. Just saying in case! :)
> So after the above changes, we should probably get a test failure from Travis that says something like `expected 1 but found 0` for the assertion that `nodes.length` is 0. Then we'd know that our test is asserting the right thing -- that the correctly-formed search for a node using 2 different tags in a string did not actually return the node that has both those tags.

dinaelhanan:
@SidharthBansal while I wait for jeff to approve this, can I submit a PR for another issue?

SidharthBansal:
Yeah sure. You should do it at the earliest.

dinaelhanan:
@SidharthBansal done :) please check it out here:
https://github.com/publiclab/plots2/pull/4198

dinaelhanan:
@sagarpreet-chadha hey! Maybe you can help out? I don't know why these tests are failing...

jywarren:
This looks like a YAML error in the logs:

```
ActiveRecord::Fixture::FormatError:         ActiveRecord::Fixture::FormatError: a YAML error occurred parsing /app/test/fixtures/tags.yml. Please note that YAML must be consistently indented using spaces. Tabs are not allowed. Please have a look at http://www.yaml.org/faq.html
        The exact error was:
          Psych::SyntaxError: (<unknown>): did not find expected key while parsing a block mapping at line 1 column 1
            
```

In YAML format, every space has significance -- i'll try to highlight where one is missing... hang on...


And just to re-iterate -- we can give approval for this GCI task even when tests do not pass -- they just have to not pass for the right reasons! 😄

test/fixtures/node_tags.yml
jywarren: I think this line needs zero spaces before it -- each double-space is counted as an indent and affects YAML formatting. Thanks!

dinaelhanan: @jywarren done ! :)

jywarren: Hmm, it's still showing YAML errors. Could there be more?

test/fixtures/tags.yml
```diff
@@ -78,4 +78,8 @@ co-author:
 
 unpublished-note-tag:
   tid: 20
-  name: unpublished-note-tag
\ No newline at end of file
+  name: unpublished-note-tag
+  
+ cool:
```
jywarren: Oh this has a leading space as well! Sorry, didn't catch it. Try this: 

```suggestion
cool:
```

dinaelhanan: woops, thanks for pointing that out!

jywarren:
Looks like we did it!

```
 FAIL["test_node_mysql_native_fulltext_search_returning_two_tag-based_matches", #<Minitest::Reporters::Suite:0x0000555af1eb7b60 @name="NodeTest">, 8.44883886499997]
 test_node_mysql_native_fulltext_search_returning_two_tag-based_matches#NodeTest (8.45s)
        Expected: 1
          Actual: 0
        test/unit/node_test.rb:67:in `block in <class:NodeTest>'
 FAIL["test_should_have_tags,_node_tags,_and_tagnames", #<Minitest::Reporters::Suite:0x0000555af2239f68 @name="NodeTest">, 8.829655708000018]
 test_should_have_tags,_node_tags,_and_tagnames#NodeTest (8.83s)
        --- expected
        +++ actual
        @@ -1 +1 @@
        -"test awesome spectrometer activity:spectrometer"
        +"test awesome cool spectrometer activity:spectrometer"
        test/unit/node_test.rb:213:in `block in <class:NodeTest>'
```

Fail 1 is the one we're looking for -- 👍 

Fail 2 is actually because another test, on line 213 of `test/unit/node_test.rb`, is checking what tags are associated with the node we modified. So we just have to update that test to also recognize the new `cool` tag. 

Then we're done!

dinaelhanan:
@jywarren Ok I've tried to make changes, I basically copy pasted what's doing for the awesome tag, since I thought that they should be similar enough to do that.

dinaelhanan:
@jywarren I'm still getting the same error, I don't get why though :/

test/unit/node_test.rb
```diff
@@ -201,8 +211,10 @@ class NodeTest < ActiveSupport::TestCase
     assert_not_nil node.tagnames
     assert node.tagnames.first.is_a?(String)
     assert_equal 'test awesome spectrometer activity:spectrometer', node.tagnames.join(' ')
+    assert_equal 'test cool spectrometer activity:spectrometer', node.tagnames.join(' ')
```
jywarren: ah ok -- so we should add cool after "awesome" on the above line instead of adding a new line. Same for the below version too, on 217! Thanks!!! 👍 😄

dinaelhanan: thanks! @jywarren

dinaelhanan:
@jywarren ok I've tried to fix it :)

jywarren:
Perfect! 

```
 test_node_mysql_native_fulltext_search_returning_two_tag-based_matches#NodeTest (102.33s)
        Expected: 1
          Actual: 0
        test/unit/node_test.rb:67:in `block in <class:NodeTest>'
```

This task is complete!!! Now we (or someone else wanting to take this on) need to actually get it to pass. Great work @dinaelhanan! This is TDD!

@milaaraujo - does this tag test make sense?

gauravano:
@jywarren @milaaraujo you want to create a new issue for this or want to add more details so others can take on this? Thanks!

SidharthBansal:

PR #4104, opened 2018-12-03
Keep grid image's original ascpect ratio

dinaelhanan:
Fixes #4082 

* [x] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [x] code is in uniquely-named feature branch and has no merge conflicts
* [x] PR is descriptively titled
* [x] ask `@publiclab/reviewers` for help, in a comment below

PR #4198, opened 2018-12-08
Fix Aspect Ratio For Images

### LINKED ISSUE #4082: Images distorted in thumbnail grids
We have some great new thumbnail grids as shown on this page: https://publiclab.org/wiki/micro#Assembly

image

But as you can see, some images are distorted. This is solved in the dashboard templates -- https://github.com/publiclab/plots2/blob/master/app/views/dashboard/_node_default.html.erb#L7 and the dashboard css:

https://github.com/publiclab/plots2/blob/e6608c89b3a7d1862d1e329e377ef8aea1e2ce90/app/assets/stylesheets/dashboard.css#L199-L204

Let's apply this to the new thumbnail grids too!

@ValentinaTironi would you mind linking to the thumbnail grid code? And would you be interested in making this into a first-timers-only issue to welcome someone else into the community? I'll paste in some guidance below:

****

dinaelhanan:
Fixes #4082  (<=== Add issue number here)

Before:
screenshot from 2018-12-07 20-56-53

Now:
screenshot from 2018-12-07 20-47-57
Yay! Thanks to all that helped in #4104

### DIFF
--- a/app/assets/stylesheets/style.css
+++ b/app/assets/stylesheets/style.css
@@ -366,7 +366,9 @@ div.note .meta {
   height:auto;
   overflow:auto;
 }
-
+/*start of img aspect css that will affect the 
+thumbnails which appears on the user's profile 
+under their notes*/
 #notes .note a.img {
   height: auto;
   display: inline-block;
@@ -389,6 +391,7 @@ div.note .meta {
   overflow: hidden;
   margin-bottom: 10px;
 }
+/*end of img aspect css*/
 
 #notes .note h3 a {
   color: #444;
--- a/app/assets/stylesheets/style.css
+++ b/app/assets/stylesheets/style.css
@@ -384,7 +384,10 @@ div.note .meta {
 
 .grid-image {
   width: auto;
-  height: 160px;
+  height: auto;
+  display: block; 
+  overflow: hidden;
+  margin-bottom: 10px;
 }
 
 #notes .note h3 a {
@@ -441,8 +444,11 @@ div.note.moderated h4 {
   }
 
   .grid-image {
-    width: auto;
-    height: 200px;
+  width: auto;
+  height: auto;
+  display: block; 
+  overflow: hidden;
+  margin-bottom: 10px;
   }
 }
 
--- a/app/assets/stylesheets/style.css
+++ b/app/assets/stylesheets/style.css
@@ -368,7 +368,7 @@ div.note .meta {
 }
 
 #notes .note a.img {
-  height:160px;
+  height: auto;
   display: inline-block;
   overflow: hidden;
   margin-bottom: 10px;
@@ -379,7 +379,7 @@ div.note .meta {
   overflow: hidden;
 }
 .note-image img {
-  width: 100%;
+  width: auto;
 }
 
 .grid-image {


SidharthBansal:
Let the tests pass first

dinaelhanan:
@SidharthBansal passed!

SidharthBansal:
This needs a review from @jywarren he knows better of image distortions.

dinaelhanan:
@SidharthBansal I think jeff is a little busy and with the deadline approaching, I would like to get this done soon so I can claim a final task.. Plus I am also waiting for his review on #4149. Can other reviews help maybe like @sagarpreet-chadha ?

SidharthBansal:
Yeah may be sagar can help.

jywarren:
Hmm, this looks good, but would you mind checking this solution on other pages where this style is used? Like perhaps /tag/test (but for notes with main images) If others are affected, we might need to restrain these changes to the thumbnail grid display only by making a narrower CSS selector. I'd be happy to help with this as we figure it out! Thanks!!!

dinaelhanan:
@jywarren Ok, I will look around for more and add it to tag :)

dinaelhanan:
@jywarren I don't get what you mean by for tag/test. I do understand that we could do this to notes with main images, but I thought that's what I did here in this pr (when I took the above gif, I made a note with a main image. Before, the main image was compressed but it's not anymore) thanks!

jywarren:
Sorry! I meant go to http://localhost:3000/tag/test to look for content tagged with the word `test` (or any other tag you know exists, or make a new one!) -- but actually i guess we should just look for anywhere else those styles are being used, and see if your changes have adverse effects on other layouts. Because we hadn't seen this distortion style affect anything before the new inline thumbnail grids, so i worry we're possibly messing with some other layout in the site. 

So: 

1. let's search for `grid-image` to see where else that CSS class might be used: https://github.com/publiclab/plots2/search?q=grid-image -- and same for `note-image`
2. let's check the "blame" view for those styles to see what projects or PRs they were associated with and where else they may be used: https://github.com/publiclab/plots2/blame/master/app/assets/stylesheets/style.css#L377-L388 (looks like we're OK -- it was Valentina's code from the original thumbnail grid implementation!
3. if we /are/ worried, we could try writing a more specific CSS query. But since it looks OK, maybe instead, we could add a pair of comments around the CSS in question to help future devs know exactly the scope of those CSS statements. That'll give "future us" more confidence to know exactly what we're changing. 

Sorry for the excess of caution! With a complex site, it's just increasingly worth ensuring we're not messing up some other carefully crafted code with our new code. I appreciate your care on this! Thanks!!!

dinaelhanan:
Thanks for the help! I understand that it's a big site, so being cautious is a great thing! I found different places where it does this. one of which is when you go to your profile and checkout your notes, and the thumbnails show up (some are cut off here because I uploaded some longer images, which should be OK since we don't want people uploading really long vertical images, I think the css handles this with a max-height property). 

screenshot from 2018-12-11 21-39-09

For some reason however, when I added the tag "water" to each note, and when to /tag/water it said no results found, even though I went back to my profile and it said that I had 3 notes with the tag water...

jywarren:
That is odd but it doesn't seem related to this code! I'm going to merge this now. Great work!!

dinaelhanan:
@jywarren hmm that's strange... does it have to do with the hidden overflow?  Or maybe because some of the classes I changed only have `height:auto;` but should also include `width:auto;` ?

jywarren:
I'd guess it's supposed to be with the hidden overflow. Did your fix include a container with fixed height but hidden overflow? Maybe use the developer inspector to see what styles are actually being applied on the live site, and see if some style is getting overridden?

Thank you!

dinaelhanan:
@jywarren so would I just upload the images as a blog post? (because from what I'm seeing on the macro page, the images are form events, how would I add that on the local version?) Also, here's the CSS affecting those images, looks like my changes were applied on the live site. I will try running it without the overflow being hidden and setting it to auto. Just wondering how I would access the events part of the page :)
screenshot from 2018-12-15 21-34-58
screenshot from 2018-12-15 21-35-23

dinaelhanan:
Weird..Even with the `overflow:hidden;` everything looks good on my end. I also made `overflow:auto;` and got the same result:
screenshot from 2018-12-16 09-19-19
screenshot from 2018-12-16 09-13-36

jywarren:
Could you check the live site in the inspector to see if it has exactly the
same styles applied? Thanks for doing some detective work! Css can be a bit
mysterious!

dinaelhanan:
@jywarren found it! It was this css property that affects the whole container (image and title) and not just the image. Yay!! My problem is that I can't find it anywhere in the stylesheet. Since 33.333 is such a specific number, I'm guessing something else created that.  see below for the property causing this:
screenshot 1
screenshot

jywarren:
OK, so what you can probably do is try applying your width styles to the image directly with `.note-image img` -- but if you want the container to be fixed, and the image within it to not be, you may need an additional container just for the image "box". I believe in the dashboard styling, we actually use the `a` tag to create this fixed box:

https://github.com/publiclab/plots2/blob/03812548111b0831e82052cd559aeea9a380098a/app/views/dashboard/_node_default.html.erb#L6-L10

Try inspecting an image on a regular note on the dashboard, where the image clipping is working, and see if we need to copy some of the CSS strategies there. I'm sure it's a little bit subtle in the implementation, so if you get stuck, I can give it a try (i wrote these styles maybe 2 years ago so there are probably a couple useful brain cells still to be tapped into 😄 )

But give it a try! I think you get the fundamentals here so it may take some experimentation.

dinaelhanan:
@jywarren this is the html for the images that we are having an issue with:
screenshot from 2018-12-18 15-09-28

I noticed that although it uses the same strategy of having an a container, that's not in the css the same way as the dashboard. (the a.img part. For the images we are having issues with, it just has .img) I tried to make those changes. On my local version, the same images are working correctly on /notes/recent. But I don't know how to access the images in a place like /micro (doesn't exist on local version) Hope I made the correct changes. Thanks!

jywarren:
Hi, @dinaelhanan - apologies that I missed this before the holiday. The grid on /micro is generated from this template:

https://github.com/publiclab/plots2/blob/master/app/views/grids/_thumbnail.html.erb

Shall we make a new issue for this so we have a consistent place to track the problem until we solve it? Thanks!!!

PR #4217, opened 2018-12-09
Create a comment_moderation mailer - Email to comment author part [WIP]

### LINKED ISSUE #4053: Create a comment_moderation mailer - Email to comment author part
**This issue is a part of project #3331** 

Create a new mailer as following:
1. First, the mailer would send the email to comment author notifying that comment has been approved.
2. Create a view of email by creating a new file in admin_mailer
3. Write functional tests to check whether flow is working as planned or not.

Name the method as **notify_author_of_comment_approval** and also use the same name for the file.

Refer the method https://github.com/publiclab/plots2/blob/master/app/mailers/admin_mailer.rb#L19-L26 .

**Hi, this issue has been opened as a task on GCI dashboard. If you are Google Code-In participant and want to solve this issue, please claim it from Google Code-In [website](https://codein.withgoogle.com/organizations/public-lab/) . Thanks!**

ValentinaTironi:
Fixes #4053 
email_author

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -162,7 +162,9 @@ class AdminController < ApplicationController
       else
         first_timer_comment = (@comment.status == 4)
         @comment.publish
-        @comment.author.unban
+        if @comment.author.banned?
+          @comment.author.unban
+        end
         if first_timer_comment
           AdminMailer.notify_author_of_comment_approval(@comment, current_user).deliver_now
         else
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -160,8 +160,14 @@ class AdminController < ApplicationController
       if @comment.status == 1
         flash[:notice] = 'Comment already published.'
       else
+        first_timer_comment = (@comment.status == 4)
         @comment.publish
-        flash[:notice] = 'Comment published.'
+        @comment.author.unban
+        if first_timer_comment
+          AdminMailer.notify_author_of_comment_approval(@comment, current_user).deliver_now
+        else
+          flash[:notice] = 'Comment published.'
+        end
       end
       @node = @comment.node
       redirect_to @node.path
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -38,6 +38,15 @@ class AdminMailer < ActionMailer::Base
     mail(to: @author.mail, subject: subject)
   end
 
+  def notify_author_of_comment_approval(comment, moderator)
+    subject = '[Public Lab] Your comment was approved!'
+    @author_mail = comment.author.mail
+    @moderator = moderator
+    @comment = comment
+    @footer = feature('email-footer')
+    mail(to: @author_mail, subject: subject)
+  end
+
   # Will this further bait spammers? If we don't,
   # will non-spammers whose posts were moderated get confused?
   # Should: show explanation/appeal process to authors who visit again
--- a/app/models/drupal_user.rb
+++ b/app/models/drupal_user.rb
@@ -73,6 +73,10 @@ class DrupalUser < ApplicationRecord
     self
   end
 
+  def banned?
+    self.status == 0
+  end
+
   def email
     mail
   end
--- a/app/models/drupal_user.rb
+++ b/app/models/drupal_user.rb
@@ -74,7 +74,7 @@ class DrupalUser < ApplicationRecord
   end
 
   def banned?
-    self.status == 0
+    status == 0
   end
 
   def email
--- a/app/models/drupal_user.rb
+++ b/app/models/drupal_user.rb
@@ -74,7 +74,7 @@ class DrupalUser < ApplicationRecord
   end
 
   def banned?
-    status == 0
+    status.zero?
   end
 
   def email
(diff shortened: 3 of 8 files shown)


ValentinaTironi:
The error that appears in test isn't mine, I ran the test from master and the error still there. It says is something with `SearchServiceTest`. What we have to do?

ValentinaTironi:
@SidharthBansal hey, can you check this? Thank you!

ValentinaTironi:
@SidharthBansal

Souravirus:
Hi @ValentinaTironi, the second error you are getting which is 
` FAIL["test_contributors_with_specific_tag_name", #<Minitest::Reporters::Suite:0x0000557baf2e7220 @name="TagTest">, 1.407246955000005]
 test_contributors_with_specific_tag_name#TagTest (1.41s)
        Expected: [1, 2, 5, 6]
          Actual: [1, 2, 5, 6, 19]
        test/unit/tag_test.rb:154:in `block in <class:TagTest>'`

is somewhat related to the `comment_status` you have added in comments.yml file. You can correct this by correcting the test `test_contributors_with_specific_tag_name`. 
Thank you!!

Souravirus:
Also the first error is also related to this change. So, you can correct the test `test_contributor_count_with_specific_tag_name` to remove this failure.

ValentinaTironi:
Thank you very much @Souravirus :) Works now!

app/controllers/admin_controller.rb
```diff
@@ -160,8 +160,14 @@ def publish_comment
       if @comment.status == 1
         flash[:notice] = 'Comment already published.'
       else
+        first_timer_comment = (@comment.status == 4)
         @comment.publish
-        flash[:notice] = 'Comment published.'
+        @comment.author.unban
```
gauravano: What would happen when a user is already a normal user and not banned? Let me check if we need this. Other changes looks good.

ValentinaTironi: I think it doesn't changes anything. If you want, I can check the user's status and unban only when it's necessary. What do you think?

gauravano: Looks like you found fixed a bug :smile: . The user should be unbanned whenever his comment gets published but an alert may be generated extra saying user is unbanned if user is already normal too, I think we wouldn't see that now because currently moderators can only mark as spam from email but once we add publish buttons on comment then it would show the extra alert and may confuse moderators.

gauravano: So, how about adding condition on alert?

gauravano: @ValentinaTironi add condition to get it approved. Thanks!

ValentinaTironi: Sorry, I don't understand. You say show an alert saying `user is unbanned if user is already normal too` when `user.stats == 1`?

ValentinaTironi: @gauravano

gauravano: > Sorry, I don't understand. You say show an alert saying user is unbanned if user is already normal too when user.stats == 1?

Let's start again. You agree that an alert would be shown saying "User is unbanned, when moderator will approve the comment" ?

gauravano: @ValentinaTironi

ValentinaTironi: Yup, but I think it has to be showed when user is banned

ValentinaTironi: @gauravano

gauravano: And, in case user is not banned? Do we still need to show that alert? @ValentinaTironi

ValentinaTironi: Nope

ValentinaTironi: So, I do the condition. `if user.status == 0` and if it's true `user.unban and show the alert`. Am I right?

ValentinaTironi: @gauravano

gauravano: But, according to current code 
snippet_admin_unban

alert would be shown.

gauravano: So, let's add condition to call unban only when needed. In case, a comment is approved from dashboard we don't need to call unban. Make sense?

ValentinaTironi: That method is from `admin_controller`, so isn't that using here:
screen shot 2018-12-10 at 12 49 27 pm
In the controller is using the `unban` method defines in `drupal_user` model:
screen shot 2018-12-10 at 12 50 17 pm
So, we need the alert how I showed in the first image

gauravano: ok, now I am confused.

gauravano: I have to see the code.

gauravano: But in first place, why are we calling the unban method if user is normal already?

ValentinaTironi: I don't know, I got confused I think

ValentinaTironi: Please, let me know if you are agree with me in this comment https://github.com/publiclab/plots2/pull/4217#discussion_r240265441 and I'm gonna do the pull request :) Hope we can finish this before claiming tasks time finish

gauravano: Let's add a condition something like this - 
If user is banned, unban the user. What do you think?

ValentinaTironi:
@gauravano Done! please if everything is ready approve my task! https://codein.withgoogle.com/dashboard/task-instances/5183539009028096/
Thank you so much 💯

ValentinaTironi:
@jywarren Hey, maybe you can merge this? Thank you :)

jywarren:
Wonderful!

PR #4256, opened 2018-12-10
Implement questions answered vs questions asked

### LINKED ISSUE #4137: Statistics on questions asked vs questions answered
Implement new statistics on the no. of questions answered versus questions asked filtered by a time range

cesswairimu:
Fixes #4137  

Thanks!

### DIFF
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -19,12 +19,23 @@ class QuestionsController < ApplicationController
   public
 
   def index
+    params[:period]
+    @asked = Node.questions.to_a.size
+    @answered = Answer.all.map(&:node).uniq.size
     @title = 'Questions and Answers'
+
     set_sidebar
     @questions = Node.questions
       .where(status: 1)
       .order('node.nid DESC')
       .paginate(page: params[:page], per_page: 24)
+    @week_asked = Node.questions.where('created >= ?', 1.week.ago.to_i).to_a.size
+    @month_asked = Node.questions.where('created >= ?', 1.month.ago.to_i).to_a.size
+    @year_asked = Node.questions.where('created >= ?', 1.year.ago.to_i).to_a.size
+    @week_answered = Answer.where("created_at >= ?", 1.week.ago).map(&:node).uniq.size
+    @month_answered = Answer.where("created_at >= ?", 1.month.ago).map(&:node).uniq.size
+    @year_answered = Answer.where("created_at >= ?", 1.year.ago).map(&:node).uniq.size
+    @period = [@week_asked, @month_asked, @year_asked]
   end
 
   # a form for new questions, at /questions/new
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -19,23 +19,13 @@ class QuestionsController < ApplicationController
   public
 
   def index
-    params[:period]
-    @asked = Node.questions.to_a.size
-    @answered = Answer.all.map(&:node).uniq.size
+    @stats = helpers.filtering(params[:period])
     @title = 'Questions and Answers'
-
     set_sidebar
     @questions = Node.questions
       .where(status: 1)
       .order('node.nid DESC')
       .paginate(page: params[:page], per_page: 24)
-    @week_asked = Node.questions.where('created >= ?', 1.week.ago.to_i).to_a.size
-    @month_asked = Node.questions.where('created >= ?', 1.month.ago.to_i).to_a.size
-    @year_asked = Node.questions.where('created >= ?', 1.year.ago.to_i).to_a.size
-    @week_answered = Answer.where("created_at >= ?", 1.week.ago).map(&:node).uniq.size
-    @month_answered = Answer.where("created_at >= ?", 1.month.ago).map(&:node).uniq.size
-    @year_answered = Answer.where("created_at >= ?", 1.year.ago).map(&:node).uniq.size
-    @period = [@week_asked, @month_asked, @year_asked]
   end
 
   # a form for new questions, at /questions/new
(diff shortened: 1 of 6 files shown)


jywarren:
Hi! Just a hint - we can ignore the CodeClimate things, but sometimes they're really helpful suggestions. You can read them in detail and how to resolve them here: https://codeclimate.com/github/publiclab/plots2/pull/4256

app/controllers/questions_controller.rb
```diff
@@ -16,15 +16,42 @@ def filter_questions_by_tag(questions, tagnames)
...
+    period.each do |key, value|
+      results = results.public_send(key, value) if value.present?
+    end
+    results
+  end
+
   public
 
   def index
+    if params[:period]
```
jywarren: This might be a good one to cache daily! I think esp. with collecting a year's worth of content each page load.

cesswairimu: great! thanks

cesswairimu:
past_year

cesswairimu:
Hey @gauravano  any suggestions on how I can best cache the stats here. Thanks.

gauravano:
As @jywarren mentioned we could cache for a day https://github.com/publiclab/plots2/pull/4256#discussion_r240290288 

we also cache home page data so you can see home_controller for reference. Let me know if you need help. Thanks!

cesswairimu:
Hey @jywarren I did some refactoring and removed the whole bunch of instance variables. Does the caching I did on the helper method make sense? Thanks

app/helpers/questions_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+      @answered = Answer.where("created_at >= ?", 1.send(period.downcase).ago).map(&:node).uniq.size
+      "#{@asked} questions asked and #{@answered} questions answered in the past #{period}"
+    end
+  end
+
+  def options
+    SORTING_OPTIONS
+  end
+
+  def cache_quiz_stats
```
jywarren: This works, but then people coding do have to remember to use the "cached version". Another way would be to build this inside the `filtering()` method itself, so all calls to that are filtered. If people really want the non-filtered version, they can manually flush it with... i forget but something like `Rails.cache.flush(...)`. Make sense? This is exciting!

cesswairimu: Yes it does thanks, on it

cesswairimu:
@publiclab/reviewers this is ready kindly review

app/helpers/questions_helper.rb
```diff
@@ -0,0 +1,25 @@
+module QuestionsHelper
+  SORTING_OPTIONS = %w(All Week Month Year).freeze
+
+  def filtering(period)
```
jywarren: This looks good but perhaps it could have a more descriptive name -- something describing how it's related to stats, or Q&A? Thanks!

app/helpers/questions_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+    return if period.nil?
+
+    if period == 'All'
+      Rails.cache.fetch("all_stats", expires_in: 1.days) do
+        @asked = Node.questions.to_a.size
```
jywarren: I think if you use `Node.questions.count` it should do a more efficient query, just a SQL SIZE query instead of fetching all the full records and counting them. Maybe same on line below! 👍

app/views/questions/index.html.erb
```diff
@@ -1,4 +1,14 @@
 <div class="col-md-12">
+    <div class ="row">
+        <div class="col-md-6" >
+            <h4><b> <%= @stats if params[:period].present?  %></b></h4>
+        </div>
+        <div class="col-md-4">
+            <%= form_tag request.url, :method => 'get'  do %>
+                <%= select_tag :period, options_for_select(options),onchange: "this.form.submit();", include_blank: "Filter stats", class: "form-control"%>
```
jywarren: Ooh this is cool, could you share a screenshot? Thanks!

app/helpers/questions_helper.rb
jywarren: Also, I'd love to see a simple test for these methods. Where would that live?

jywarren:
Regarding the question of where statistics code "home base" is - would it fit in a model, so it's testable via unit tests? But open to ideas and brainstorming. Maybe ask @gauravano or @SidharthBansal for input on this as well!

app/views/questions/index.html.erb
cesswairimu: @jywarren a screenshot is :point_up: above. Thanks for the feedback.

app/helpers/questions_helper.rb
cesswairimu: Let me try and figure out where the tests will go. Thanks

app/helpers/questions_helper.rb
cesswairimu: @jywarren tried `.count` was not working fine  because its  a hash returning  results like so `{82=>1, 85=>2, 86=>2, 87=>1}`. Using `.length` instead. hope that's a better query?

cesswairimu:
during
month


@jywarren I implemented the tests and changed the method name to a better name and improved the queries. Please take a look.

cesswairimu:
> Regarding the question of where statistics code "home base" is - would it fit in a model, so it's testable via unit tests? But open to ideas and brainstorming. Maybe ask @gauravano or @SidharthBansal for input on this as well!

@jywarren is this for this ticket #4317?

jywarren:
Actually it's more broad - that is, should we have a central location where we keep much of the statistics code, as if in a model, or should we be OK with some appearing in controllers, some in helpers... i'm just trying to think critically about how we're doing overall code organization. We don't have to come up with a final solution right now! Thanks!

jywarren:
This is great. Merging!!!

PR #3963, opened 2018-11-13
Trending tags showing duplicates

ValentinaTironi:
Fixes #3188

Issue #3188, opened 2018-08-02
Trending tags showing duplicates


jywarren:
**We are preparing to participate in Google Code-in, and have reserved this issue for participants in GCI - but we'd love to have your help with another one! Please check out https://code.publiclab.org to see more.**

I think we're missing a `group()` or something on the https://publiclab.org/dashboard's trending tags area - i'm seeing 5 of the same:

![screen shot 2018-08-02 at 12 29 19 pm](https://user-images.githubusercontent.com/24359/43597330-c6d11cee-964f-11e8-943d-dcbbcf8e62b6.png)

Here's the code - 

https://github.com/publiclab/plots2/blob/0663ac5f7ad7d047fc47e2b5b9ae59c943f354bc/app/models/tag.rb#L268-L277

And here's the test, where i'm not sure but maybe we should test to be sure they're unique? 

https://github.com/publiclab/plots2/blob/ceb7d05404b4619c7fa8cd8cae5e8215798cb02c/test/unit/tag_test.rb#L131-L137

Any help appreciated!

SidharthBansal:
This task is published on GCI dashboard. Thanks all.

ValentinaTironi:
I just have one question, How I send you the task ready? Do I have to do a pull request in repository or what? thanks!

gauravano:
Hi @ValentinaTironi , this task is reserved for Google Code-In students, have you claimed this from GCI website?

gauravano:
Awesome! You have to open the Pull request with your changes. Thanks!

ValentinaTironi:
Hi! I'm trying to do a `git push` to do a pull request here but it says that I don't have permission. What I have to do?

geekychaser:
@ValentinaTironi You need to do `git push` to your forked repository and then create a pull request.
Good luck!

jywarren:
so that would be something like `git push git@github.com:ValentinaTironi/plots2
feature-branch-name` or possibly `git push
https://github.com/ValentinaTironi/plots2 feature-branch-name`

ask if you need help, we're happy to support!

geekychaser:
@ValentinaTironi It seems like you have not forked the repository. So to get started follow these installation process `https://github.com/publiclab/plots2#installation`

ValentinaTironi:
@geekychaser Yes, I tried to do following these steps, but I can't do `git clone https://github.com/ValentinaTironi/plots2.git` like that said

jywarren:
[...]

PR #4263, opened 2018-12-11
Trending tags showing duplicates [WIP]

ValentinaTironi:
Fixes #3963 
I'm sure this works now! Check it out for you
screen shot 2018-12-11 at 1 30 05 am

I saw that the last time it didn't work so I would like try again :) Thank you!

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -271,10 +271,11 @@ class Tag < ApplicationRecord
        .where('node.status = ?', 1)
        .where('node.created > ?', start_date.to_i)
        .where('node.created <= ?', end_date.to_i)
-       .distinct
-       .group([:name, 'node.nid', 'term_data.tid', 'community_tags.nid', 'community_tags.uid', 'community_tags.date']) # ONLY_FULL_GROUP_BY, issue #3120
        .order('count DESC')
        .limit(limit)
+       .group([:name, 'node.nid', 'term_data.tid', 'community_tags.nid', 'community_tags.uid', 'community_tags.date'])
+       .distinct
+       .pluck(:name)
   end
 
   # select nodes by tagname and user_id
--- a/app/views/dashboard/_activity.html.erb
+++ b/app/views/dashboard/_activity.html.erb
@@ -51,8 +51,8 @@
   <small><a href="/tags" style="color: #888;">Trending topics:</a></small>
   <span style="font-size: 18px;">
    <% cache('trending-tags', expires_in: 24.hours, skip_digest: true) do %>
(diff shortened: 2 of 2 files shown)


ValentinaTironi:
@jywarren

milaaraujo:
@ValentinaTironi, I assigned #3188 to myself and leave a comment saying I would work on it 3 days ago. Don’t start working on something that another person is also working on without warning beforehand. This prevents us from wasting our time...

milaaraujo:
Anyway I can review this later, I already have a few examples ready on my computer to test the problem.

jywarren:
Hi, I just wanted to say this can happen, it's all right - and in this case, it's an interesting outcome because the solution proposed here was different. I think we have a potential path forward in #3188 but I am interested in whether the `distinct` being at the end is resulting in a different query that also works?

Thanks all for cooperating and your work is always valued here 👍 🙌

milaaraujo:
It is ok, @ValentinaTironi! It was not my intention to discourage anyone from collaborating with the project! ;) Any collaboration is welcome, but it is always good to try to coordinate our work in order to make the most of each one.

Your solution is also working properly:
```
SELECT  DISTINCT `term_data`.`name` FROM `term_data` 
INNER JOIN `community_tags` ON `community_tags`.`tid` = `term_data`.`tid` 
INNER JOIN `community_tags` `node_tags_term_data_join` ON `node_tags_term_data_join`.`tid` = `term_data`.`tid` 
INNER JOIN `node` ON `node`.`nid` = `node_tags_term_data_join`.`nid` 
WHERE (node.status = 1) AND (node.created > 1541995976) AND (node.created <= 1544587976) 
GROUP BY `term_data`.`name`, node.nid, term_data.tid, community_tags.nid, community_tags.uid, community_tags.date 
ORDER BY count DESC LIMIT 5;
+------------------------+
| name                   |
+------------------------+
| first-time-poster      |
| lat:74.92979837188261  |
| lat:35.67314305234031  |
| lon:21.714137445587394 |
| lon:45.76772749324683  |
+------------------------+
```
I just think the solution in #4266 is better because it is simpler and removes "group by" that does not make much sense here. What do you think?

ValentinaTironi:
I'm totally agree 💯 I left a comment in you PR already #4266

milaaraujo:
After merging #4266 we will have an extra step to test it with MySQL:

>"then let's actually open a PR where we change the database back to mysql (but version 5.7) to see if the error crops up there again. [...]"

You can help with this if you want!

ValentinaTironi:
That is a more efficient solution

PR #4300, opened 2018-12-14
Refactor Stats Controller

### LINKED ISSUE #4147: Refactor stats controller
Refactor https://github.com/publiclab/plots2/blob/master/app/controllers/stats_controller.rb

cesswairimu:
Fixes #4147 

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -95,9 +95,9 @@ class Node < ActiveRecord::Base
   validates_with UniqueUrlValidator, on: :create
 
   scope :published, -> { where(status: 1) }
-  scope :weekly, -> { published.where("created > ?", (Time.now - 7.days).to_i) }
-  scope :monthly, -> { published.where("created > ?", (Time.now - 1.months).to_i) }
-  scope :yearly, -> { published.where("created > ?", (Time.now - 1.years).to_i) }
+  scope :past_week, -> { published.where("created > ?", (Time.now - 7.days).to_i) }
+  scope :past_month, -> { published.where("created > ?", (Time.now - 1.months).to_i) }
+  scope :past_year, -> { published.where("created > ?", (Time.now - 1.years).to_i) }
 
   # making drupal and rails database conventions play nice;
   # 'changed' is a reserved word in rails
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -94,6 +94,11 @@ class Node < ActiveRecord::Base
   validates :title, presence: :true
   validates_with UniqueUrlValidator, on: :create
 
+  scope :published, -> { where(status: 1) }
+  scope :weekly, -> { published.where("created > ?", (Time.now - 7.days).to_i) }
+  scope :monthly, -> { published.where("created > ?", (Time.now - 1.months).to_i) }
+  scope :yearly, -> { published.where("created > ?", (Time.now - 1.years).to_i) }
+
   # making drupal and rails database conventions play nice;
   # 'changed' is a reserved word in rails
   class << self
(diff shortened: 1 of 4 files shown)


jywarren:
Ooh, is this ready for review? Feel free to tag them for review, or request a review from me or other Public Lab reviewers! Thanks Cess, you're doing a huge amount here, great work!

cesswairimu:
Still in progress will do when it ready. Thanks

cesswairimu:
@jywarren there is this query 
``
User.where(created_at: @start..@end)
      .joins('INNER JOIN users ON users.uid = rusers.id').where('users.status = 1')
``
Just wanted to be sure what the `joins` does. Does it look for users whose status is 1? Tried my new query on the webpage and console and it gives me the same results as the above query. But just wanted to make sure so that it doesn't break elsewhere. If that's the same then this is ready for review. Thanks

cesswairimu:
Investigating why the spec is test is failing

cesswairimu:
@publiclab/reviewers please review

app/models/node.rb
```diff
@@ -94,6 +94,11 @@ def updated_month
   validates :title, presence: :true
   validates_with UniqueUrlValidator, on: :create
 
+  scope :status_one, -> { where(status: 1) }
```
jywarren: what if we renamed this like "approved" or "published" since that's what it means?

cesswairimu:
@jywarren I solved the conflicts and rebased

app/models/user.rb
```diff
@@ -43,6 +43,10 @@ class User < ActiveRecord::Base
 
   before_save :set_token
 
+  scope :status_one, -> { where(status: 1) }
```
jywarren: Hmm, I think at this point because we've simplified the query a lot, maybe it makes sense to just use the simple .where method since it's not much more than .published anymore. What do you think? Then it's more standard rails too... It made total sense before with the longer query but maybe now not so much?

Thanks Cess!!!

cesswairimu: I am sorry, I am quite lost, are you suggesting we do away with all the scopes for the user_model, or just the `status_one` scope. Yeah I see the query is so much shorter now.

jywarren: Just this `status_one` scope since it's now not much shorter than the full `where()` method. Sorry i was typing from my phone!

cesswairimu: No worries, cool gotcha on it.

cesswairimu:
I rectified that @jywarren

app/models/revision.rb
```diff
@@ -19,6 +19,11 @@ class Revision < ApplicationRecord
   after_save :inline_hashtags
   before_create :setup
 
+  scope :published, -> { where(status: 1) }
+  scope :weekly, -> { where("timestamp > ?", (Time.now - 7.days).to_i) }
```
jywarren: Oh, i'm sorry i should've noticed this before. This is OK and I can merge if you like, but in sense, "weekly" seems to mean "by week" rather than "the past week". Maybe the scope should be called `past_week`, OR, it should be `weekly` but then accept a parameter like starting date? I dunno, what do you think? It's a small semantic thing but if people begin using `.weekly` in their code they might think it means something a little different than what it is. Can scopes accept params? Or should we call it `past_week` `past_year` and such? 

Thanks Cess! This is really cool.

cesswairimu: Aha! I see bad naming I did there..I am not really sure about the scope accepting params and it seems complicated so I will just add  `past`. Thanks for that.

jywarren:
Great!

jywarren:
Great work, Cess!

Issue #4307, opened 2018-12-15
Coveralls tests are not completing properly


jywarren:
Hi @Souravirus i wonder if you could look at this -- i think it's possible it was following the Rails 5.0 upgrade in https://github.com/publiclab/plots2/pull/2726

Basically, see how Coveralls tests are not working properly? Documented this here: https://github.com/lemurheavy/coveralls-public/issues/1235

I especially noticed they seem to run /before/ the main Rails tests, though after the Jasmine tests:

https://travis-ci.org/publiclab/plots2/builds/468228412#L2863

Perhaps this section needs to be reconfigured?

https://github.com/publiclab/plots2/blob/1d6f21e01729711fcb5a4cc2b55f4cee8d1157fa/Rakefile#L23-L31

****

Some potentially related bits of code:

https://github.com/publiclab/plots2/blame/7d04538d87ee021dc51e2df989bd49a308439af8/.travis.yml#L13

jywarren:
OK, tried re-enabling various apps to work on this. Let's look out at new builds to see how they do: 

https://travis-ci.org/publiclab/plots2

jywarren:
And maybe a clue - here is the page with the last successful coveralls build: https://coveralls.io/repos/143135/builds?page=220

It was: https://coveralls.io/builds/17560597

Looks like maybe this PR was the last one? https://github.com/publiclab/plots2/pull/2826/

jywarren:
But it does seem to be around the time of the Rails 5.0 upgrade.

jywarren:
@Souravirus would you mind trying to rearrange the Rakefile in a new PR to get Coveralls to run properly at the correct time?

jywarren:
#4309 worked!!! https://coveralls.io/builds/20648614

Souravirus:
Nice to see it solved @jywarren

PR #4308, opened 2018-12-15
Trying to fix Coveralls reporting

jywarren:
Attempting some changes for #4307 although not really sure what I'm doing...


jywarren:
Fixed another way!!

PR #4333, opened 2018-12-18
ensure the nearby tag inline grids properly make use of nearbyTag API 

### LINKED ISSUE #4275: ensure the nearby tag inline grids properly make use of nearbyTag API
We should ensure the nearby tag inline grids properly make use of this API, and perhaps also a test to demonstrate the grid is working? They /should/ show an inline map with markers on it for notes with the given tag.

https://publiclab.org/wiki/inline-maps#Tag-filtered+maps shows usage (currently broken! but there are correctly-tagged notes that should be appearing there) 

We should add a test, which could be modeled on one of these tests, but perhaps scan the response for the marker JavaScript line?

https://github.com/publiclab/plots2/blob/a09b6ff690c5e8dd72c401de7ef87b0e6b3baae0/test/unit/node_shared_test.rb#L197-L203

@sagarpreet-chadha @okonek @milaaraujo would one of you be interested in tackling this? Thank you!

sagarpreet-chadha:
Fixes  #4275 !

### DIFF
--- a/app/views/map/_leaflet.html.erb
+++ b/app/views/map/_leaflet.html.erb
@@ -29,21 +29,41 @@
 
     // setupFullScreen(map<= unique_id >  , map_lat , map_lon) ;
     setupLEL(map<%= unique_id %> ) ;
-    // var hash = new L.Hash(map< unique_id > );
 
     function onMapLoad(e){
-      var lat = <%= '%.2f' % lat %> ; // ensure format 0.00 (two decimal places)
-      var lon = <%= '%.2f' % lon %> ;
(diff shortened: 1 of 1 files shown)


sagarpreet-chadha:
The test is this : 

https://github.com/publiclab/plots2/blob/a09b6ff690c5e8dd72c401de7ef87b0e6b3baae0/test/unit/node_shared_test.rb#L111

sagarpreet-chadha:
Using` node_shared_test.rb `we can test if the map is rendered or not BUT we cannot check whether the markers are rendered on the map (because a string is returned from the functions in the `node_shared.rb` which contains the entire code AND not the html formed after that code is executed ) . 

We have separate tests written for testing API . Hence we are good to go 😄 !

PR #4379, opened 2018-12-24
Adding lat/lon values into node table

### LINKED ISSUE #4183: Develop latitude/longitude storage for tags

milaaraujo:
Fixes #4183

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -755,11 +755,11 @@ class Node < ActiveRecord::Base
           if tag.valid?
             if tag.name.split(':')[0] == 'lat'
               tagvalue = tag.name.split(':')[1]
-              self.update_attribute(:latitude, tagvalue)
-              self.update_attribute(:precision, decimals(tagvalue).to_s)
+              update_attribute(:latitude, tagvalue)
+              update_attribute(:precision, decimals(tagvalue).to_s)
             elsif tag.name.split(':')[0] == 'lon'
               tagvalue = tag.name.split(':')[1]
-              self.update_attribute(:longitude, tagvalue)
+              update_attribute(:longitude, tagvalue)
             end
           end
 
@@ -780,10 +780,10 @@ class Node < ActiveRecord::Base
   end
 
   def decimals(n)
-    if not n.to_s.include? '.'
-      return 0
+    if !n.to_s.include? '.'
+      0
     else
-      return n.to_s.split('.').last.size
+      n.to_s.split('.').last.size
     end
   end
 
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -750,6 +750,19 @@ class Node < ActiveRecord::Base
                                  uid: user.uid,
                                  date: DateTime.now.to_i,
                                  nid: id)
+
+          # Adding lat/lon values into node table
+          if tag.valid?
+            if tag.name.split(':')[0] == 'lat'
+              tagvalue = tag.name.split(':')[1]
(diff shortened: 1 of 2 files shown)


milaaraujo:
Hey @jywarren! When possible can you review this? I took the last days off, but I would like to get back to working on this as soon as possible!

jywarren:
I hope to tomorrow! Coming back online then! Happy new year!!!

jywarren:
OK, just a few CodeClimate changes to start with, should be pretty simple to resolve: https://codeclimate.com/github/publiclab/plots2/pull/4379

I'll do a close read in a bit!

app/models/node.rb
```diff
@@ -750,6 +750,19 @@ def add_tag(tagname, user)
                                  uid: user.uid,
                                  date: DateTime.now.to_i,
                                  nid: id)
+
+          # Adding lat/lon values into node table
+          if tag.valid?
+            if tag.name.split(':')[0] == 'lat'
+              tagvalue = tag.name.split(':')[1]
+              self.update_attribute(:latitude, tagvalue)
```
jywarren: OK this looks good but won't we need to change the type? to a decimal? 

Perhaps let's write a unit test to try this out - doing `node.add_tag('lat:10.0002', users(:bob))` and seeing if it adds the correct value and precision. If it fails due to the type not being correct, we can fix it after -- TDD!

app/models/node.rb
```diff
@@ -750,6 +750,19 @@ def add_tag(tagname, user)
...
+              self.update_attribute(:precision, decimals(tagvalue).to_s)
+            elsif tag.name.split(':')[0] == 'lon'
+              tagvalue = tag.name.split(':')[1]
+              self.update_attribute(:longitude, tagvalue)
```
jywarren: Here also we'll need to set precision.

app/models/node.rb
```diff
@@ -750,6 +750,19 @@ def add_tag(tagname, user)
                                  uid: user.uid,
                                  date: DateTime.now.to_i,
                                  nid: id)
+
+          # Adding lat/lon values into node table
+          if tag.valid?
+            if tag.name.split(':')[0] == 'lat'
+              tagvalue = tag.name.split(':')[1]
+              self.update_attribute(:latitude, tagvalue)
+              self.update_attribute(:precision, decimals(tagvalue).to_s)
```
jywarren: Ah, but is precision stored as a string? Shouldn't it be an int? Perhaps we should change it to an int; sorry if I missed this in the migration. Thanks!

jywarren:
Thanks!!! Ok! That sounds great. I still think we should write a test to
assure the values are copied in, and then we can merge this. Thanks a lot!

milaaraujo:
Not sure if we need a test if all we want to test is if we are not getting an error when adding the value. I'm testing here - creating a new note and adding the location - and the whole process seems to work just fine. I can attached here the screenshots of the process to show we are storing the correct values. Is it enough?

jywarren:
I'd just like to have a unit test to ensure the latitude and longitude are
properly copied into the node record. This will help defend your code
against future code changes, too! In case someone in the future doesn't
understand this dual system. Thank you!!

milaaraujo:
Ok, I can do it later then! 

Today I also want to start the PR to delete lat/lon. Where are we deleting the lat/lon tags? I didn’t find it.

milaaraujo:
And now we have tests!

jywarren:
Perfect!!!! Ok, merging this!

PR #4434, opened 2018-12-28
Fixes #4422 Login modal not working at /profile page

### LINKED ISSUE #4422: Login modal not working at /profile
The login modal is not working when clicked through the header on /profile.

profile login modal

https://publiclab.org/profile

oorjit_chowdhary

Chrome, Windows 10

****

## Thank you!

IshaGupta18:
Fixes #4422

On clicking on login button in navbar in /profile page (this page can be viewd to logged out user if he logged out from /profile page only) 

image

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -181,7 +181,7 @@
 <!-- Login/Signup Modal -->
 <div class="modal fade" id="loginModal" tabindex="-1" role="dialog" aria-hidden="true">
   <div class="modal-dialog" role="document">
-    <div class="modal-content">
+    <div class="modal-content" style="opacity: 1">
       <div class="modal-body">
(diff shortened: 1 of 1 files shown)


gauravano:
Hi @IshaGupta18 in your last PR you have changed the opacity of another modal and now login modal as it has to do something with page so I searched codebase and found this:

https://github.com/publiclab/plots2/blob/5a90cd8dd6366092f47712524d6b8fdbe8bc1791/app/views/users/_token.html.erb#L75-L77

i think removing these lines will solve all the issues and we will not be required to change opacity for each modal. Could you try all the modals after removing this and also `opacity : 1` added by you?
 Thanks!

IshaGupta18:
Moving to #4462

gauravano:
The Open Source work you are doing is voluntary so there is no need to feel sorry over a little delay. Reviewing #4462 . Thanks!

IshaGupta18:
Thanks a lot!

PR #4435, opened 2018-12-28
Move stats from user model to a concern

### LINKED ISSUE #4317: Move statistics methods out of user.rb into a model/concern file, add unit tests
Hi @cesswairimu this is just a suggestion but we have a range of different statistics related methods in user.rb like `weekly_note_tally` which really make the file super long. I was wondering if we might try to move them to a separate file somehow - maybe a concern, or some other type of sub-file, to help make the user.rb file more readable and short. What do you think of this idea?

cesswairimu:
Fixes #4317 

### DIFF
--- a/test/unit/statistics_test.rb
+++ b/test/unit/statistics_test.rb
@@ -1,0 +1,56 @@
+require 'test_helper'
+
+class StatisticsTest < ActiveSupport::TestCase
+  def setup
+    @user = users(:bob)
+  end
+
+  test 'daily_note_tally returns the correct type of hash' do
+    daily_notes = @user.daily_note_tally
+    assert_not_empty daily_notes
+    assert_equal daily_notes.count, 365
+    assert_equal Hash, daily_notes.class
+  end
+
+  test 'weekly_note_tally returns the correct type of hash' do
+    weekly_notes = @user.weekly_note_tally
+    assert_not_nil weekly_notes
+    assert_equal Hash, weekly_notes.class
+    assert_equal (0..52).count, weekly_notes.count
+  end
+
+  test 'weekly_comment_tally' do
+    weekly_comments = @user.weekly_comment_tally
+    assert_not_nil weekly_comments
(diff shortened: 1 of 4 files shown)


cesswairimu:
@jywarren the tests are covered on user_model_test, I could see some failing as I moved the methods...Should I move the tests to separate test file or have multiple tests for the same? Thanks

jywarren:
Ah, ok - yes, if we move the functions into a separate file, let's also
separate out the tests into a unit test specific to the concern. Sound
good? Thanks!

cesswairimu:
Yes it does, on it. Thanks.

cesswairimu:
@publiclab/reviewers I am looking for an effective way to best test the statistics module.  The tests I have do not really guarantee future safeguard :confused: and I couldn't find a better approach. Any ideas? Thanks

test/unit/statistics_test.rb
```diff
@@ -0,0 +1,41 @@
...
+test 'note_streak' do
+  assert_not_nil @user.note_streak.count
+end
+
+test 'wiki_edit streak' do
+  assert_not_nil @user.wiki_edit_streak
+end
+
+test 'weekly_comment_tally' do
+  assert_not_nil @user.weekly_comment_tally
```
jywarren: For things like this, we could confirm the /format/ of the returned items; we should leave the values untested because as people change our fixtures in the future, it'll cause side-effects by breaking these tests. But we can confirm that it returns, for example, 52 items, and maybe something about the parameter names of each item? Is there a Ruby equivalent of JS's `Object.keys()` which returns the key names of an object or associative array? 

Thanks, Cess! Great future-proofing!

cesswairimu: Thanks @jywarren :smiley: I will implement the count assertions and  research how to assert a hash format was returned or something.

cesswairimu:
@publiclab/reviewers This is ready. Please review

jywarren:
Fantastic work!!!

PR #4467, opened 2018-12-30
changing configurations, addition to gemfile

### LINKED ISSUE #4374: Solving the N+1 query problem

shubhscoder:
Fixes #4374
This is a sample of a javascript alert that occurs in case it detects that eager-loading can be used. 
bullet1

Also a log file is created where we can see the trace of the lines causing n+1 queries or anything of that sort. 
Thanks!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -91,6 +91,7 @@ group :development do
   gem 'listen', '~> 3.1.5'
   gem "nifty-generators"
   gem 'web-console', '>= 3.3.0'
+  gem 'bullet', '~> 5.7', '>= 5.7.5'
 end
 
 group :test, :development do
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -44,5 +44,12 @@ Plots2::Application.configure do
     host: 'localhost:3000'
   }
   config.action_controller.default_url_options = { host: 'localhost:3000' }
-
+  
+  config.after_initialize do
+    Bullet.enable = true
+    Bullet.alert = true
+    Bullet.bullet_logger = true
+    Bullet.console = true
+    Bullet.rails_logger = true
+end
 end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -1,3 +1,4 @@
+require 'bullet'
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb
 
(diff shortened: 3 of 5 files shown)


shubhscoder:
Also we can somehow integrate this with our testing enviroment. The documentation has details regarding integration with rpsec, but I am searching of how we can somehow integrate this with minitest. The advantage of this would be, that any n+1 query would fail the test itself and thus we can fix it, then and there. Thank you!

jywarren:
oh wow, this is cool! Is it possible to have a link to a page with more
information, perhaps saying "We've detected a potential for optimization.
Read more to learn how... etc... or open an issue to get this improved"

I'm wondering if an alert is the right way to do this - perhaps integration
with the testing environment and Travis is better?

shubhscoder:
So, there are multiple ways in which we can get alerts regarding potential optimization. Alerts is one of them. But we also have a log file that is generated in bullet/log which contains detail trace of the line of the code that caused the problem so we can use that instead of alerts. One of the other methods is checking logs in the console every time we load a page. Or one of the other options this gem offers is that we could directly log the warnings in rails logs. So we could use anyone and disable alerts.

Also it is possible to integrate this with our testing environment. However the official documentation (https://github.com/flyerhzm/bullet) guides on integrating this with rspec and not with minitest, however I have found a gist(https://github.com/flyerhzm/bullet) that guides on integrating bullet with minitest, however I am not sure of how reliable the gist is. I will make the changes on my local machine and check how it works then we can integrate this here.

shubhscoder:
@jywarren So I have integrated bullet with tests. This is how the output would like:
error_log_bullet. But now the problem is, travis would keep failing for all PR's until we fix all the unoptimized queries. So what we would do is for now we could refrain from failing tests if unoptimized queries are detected, instead we will just log the errors and then start fixing them one by one. So I need your suggestion here. Thank you so much!

jywarren:
Oh wow! Ok, we need some brainstorming. How many queries are there?

gauravano:
@shubhscoder you want to work on this or we can close this? Thanks!

shubhscoder:
Sorry gauravano for the delay! But I want to work on this. Will try to get
this running asap.

gauravano:
Ok! Thanks!

SidharthBansal:

gauravano:
As @shubhscoder mentioned he would be busy with GSoC, so closing this for now. And, we can take this up again when he is free? 

Thanks!!

PR #4493, opened 2019-01-01
addition of flash to verify email

### LINKED ISSUE #3851: Flash message and button to send email again if email not verified
Subpart of #3798

After #3850 is done, we need to flash a message which will contain button by which user can receive verification email.

shubhscoder:
Fixes #3851 
Also, Merge this after #3992 

So, this PR implements the feature of verifying email through a flash message. This feature is particularly useful when the initial token that we send in the welcome email expires. 

Thanks!

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -190,4 +190,12 @@ class ApplicationController < ActionController::Base
   def signed_in?
     !current_user.nil?
   end
+
+  def check_email_verified
+    unless current_user.nil?
+      unless current_user.is_verified
+        flash.now[:notice] = "Your email is not verified. Click <a href='../verify-email'>here</a> to verify email"
+      end
+    end
+  end
 end
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -1,5 +1,6 @@
 class HomeController < ApplicationController
   before_action :require_user, only: %i(subscriptions nearby)
+  before_action :check_email_verified, only: %i(dashboard)
 
   def home
     if current_user
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,7 +2,7 @@ class UsersController < ApplicationController
   before_action :require_no_user, :only => [:new]
   before_action :require_user, :only => %i(edit update save_settings)
   before_action :set_user, only: %i(info followed following followers)
-  
+
   def new
     @user = User.new
     @action = "create" # sets the form url
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,6 +2,7 @@ class UsersController < ApplicationController
   before_action :require_no_user, :only => [:new]
   before_action :require_user, :only => %i(edit update save_settings)
   before_action :set_user, only: %i(info followed following followers)
+  
   def new
     @user = User.new
     @action = "create" # sets the form url
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,7 +2,6 @@ class UsersController < ApplicationController
   before_action :require_no_user, :only => [:new]
   before_action :require_user, :only => %i(edit update save_settings)
   before_action :set_user, only: %i(info followed following followers)
-
   def new
     @user = User.new
     @action = "create" # sets the form url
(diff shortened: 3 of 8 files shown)


shubhscoder:
@publiclab/reviewers there is one problem. I need suggestions regarding where should we show the flash message to the user. If we show this message to the user everywhere then this can be quite annoying. So for now I had included this message only in the dashboard page or any function in the homes_controller. Also that is one of the reasons why travis is failing. 
Thank you so much!

shubhscoder:
Also I am not sure of why travis is failing. 
Thank you so much!

shubhscoder:
I am not sure of what happened but I just closed the PR and reopened it and that made travis pass. Seems weird :confused:

SidharthBansal:
There was some error sometimes in Travis. So when you close and reopen the PR the build starts again.

SidharthBansal:
I think at the dashboard its ok to show the message.

SidharthBansal:
There are conflicts please remove them.
Sorry to trouble you a lot here.

shubhscoder:
@SidharthBansal sorry! I think this branch had become stale, that is why the merge conflict arised. Anyways I have fixed this. But can you please review #3992 first ? We can merge this after #3992 is merged. There is some issue with tests for that PR. Thank you so much for helping me along!

shubhscoder:
@gauravano can you please review this, and suggest any required changes? Thanking you in advance !

app/controllers/home_controller.rb
```diff
@@ -1,5 +1,6 @@
 class HomeController < ApplicationController
   before_action :require_user, only: %i(subscriptions nearby)
+  before_action :check_email_verified, only: %i(dashboard)
```
gauravano: I guess, the notice will be shown at the dashboard only, so let's add a button somewhere on Profile page or maybe settings page to allow users to voluntarily go there and there they can "Request Confirmation Email" otherwise instead of button show a green tick with text as "Email Verified". How does this sound?

app/controllers/application_controller.rb
```diff
@@ -190,4 +190,12 @@ def redirect_old_urls
   def signed_in?
     !current_user.nil?
   end
+
+  def check_email_verified
+    unless current_user.nil?
+      unless current_user.is_verified
+        flash.now[:notice] = "Your email is not verified. Click <a href='../verify-email'>here</a> to verify email"
+      end
```
gauravano: Have you created any migration or is there a plan to create one to insert is-verified  as true for existing users of PL? Thanks!

shubhscoder: Okay so I need your suggestion on that. One thing we can do is, we would force only those users to verify email, who have registered after a certain date, Or we can write a migration to verify already registered users, Or we can make the current users to verify their email as well, this would help to get rid of spammers that are currently registered with fake email ids.

app/controllers/home_controller.rb
shubhscoder: I think this is a better idea. A button would be more user friendly. I ll add the button on the profiles page?

> Great work @shubhscoder  ! Just left some ideas, let me know what you think. Thanks!

Thank you so much ! I will the make the changes as soon as possible then we can probably merge this.

shubhscoder:
Also @gauravano, once we add the button to verify the email, should we remove the flash to verify email that appears on the dashboard ? Thank you!

app/controllers/application_controller.rb
gauravano: Actually, the spammers are banned as soon as they try to post something bad, so if we consider ideal case that there are no spammers out there for now, then we would be prompting only valid users. Also, is there any condition that unverified users can't post or something so as to prevent spam?

app/controllers/home_controller.rb
gauravano: Great!

gauravano:
I think we should let the notice there for some x number of days and after that user can verify from settings or profile page.

We can stop new users to post if they haven't verified email. That will help in controlling spam to some level?

shubhscoder:
yeah we can do that. I ll set X=10 as of now, we can change that if required!

app/controllers/application_controller.rb
shubhscoder: No, we have a different issue for that. So as of now there is no restriction on any user !

shubhscoder: Okay in case we are sure that there are no spammers as of now, then we can write a migration to make is_verified of users true

gauravano:
Let's make sure @jywarren is in sync here.

jywarren:
If we stop posting without verification, we should do our best to make it easy, seamless and friendly, but sure!!

SidharthBansal:

SidharthBansal:

PR #4498, opened 2019-01-02
LEL New layers added .

sagarpreet-chadha:
🔜


sagarpreet-chadha:
screenshot 2019-01-02 at 3 36 15 pm
screenshot 2019-01-02 at 3 35 48 pm

jywarren:
Sorry just catching up on this again... there's an awful lot of code in here that seems like it could be part of LEL itself, to simplify the integration. Does LEL have one-line configuration options to add these layers, rather than having to add them manually with esriLayer? Thanks!

jywarren:
Looking great, thanks! 

Now, one more request - could we use a partial to consolidate the external JS files we are including, into a single location? 

Thank you, Sagarpreet!

jywarren:
Awesome!!!

PR #2857, opened 2018-06-19
Deprecate drupal user model

Neidley:

PR #4514, opened 2019-01-03
Drop DrupalUsers "users" table

jywarren:
Follow-up to https://github.com/publiclab/plots2/pull/2857 once we confirm we have a backup and all is running smoothly.


gauravano:
@jywarren can we complete this now or we still need some more time? Thanks!

jywarren:
I think we can... it's been pretty good, no? What do you think?

gauravano:
@jywarren I agree on dropping table as we haven't faced any issue from a while.

jywarren:
OK, this should be good if it passes!

alonpeer:
@jywarren this PR causes a `db:migrate` error on master, because it's looking for a class named `RemoveDrupalUsers` (due to the file name), but the class is called `DropDrupalUsers`.
It can be fixed just by renaming the class.

gauravano:
@alonpeer interested in raising a PR?

alonpeer:
I would, but I'm pretty busy in the coming days so not sure I'll get around to doing that soon. If this can wait a bit, I'll be happy to do so.

gauravano:
Actually, I will fix it as this might trouble other contributors. Thanks @alonpeer for reporting :+1:

gauravano:
I've raised the fix at https://github.com/publiclab/plots2/pull/5330. Can you review it @alonpeer? Thanks!

PR #4526, opened 2019-01-05
Deleting lat/lon values from node table

### LINKED ISSUE #4183: Develop latitude/longitude storage for tags

milaaraujo:
Fixes #4183

Synchronize deletion between the two means of storing lat/lon

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -342,6 +342,15 @@ class TagController < ApplicationController
     # only admins, mods, and tag authors can delete other peoples' tags
     if node_tag.uid == current_user.uid || current_user.role == 'admin' || current_user.role == 'moderator' || node.uid == current_user.uid
 
+      tagname = Tag.joins(:node_tag)
+                   .select('term_data.name')
+                   .where(tid: params[:tid])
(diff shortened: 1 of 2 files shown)


app/controllers/tag_controller.rb
```diff
@@ -342,6 +342,15 @@ def delete
     # only admins, mods, and tag authors can delete other peoples' tags
     if node_tag.uid == current_user.uid || current_user.role == 'admin' || current_user.role == 'moderator' || node.uid == current_user.uid
 
+      tagname = Tag.joins(:node_tag)
+                   .select('term_data.name')
+                   .where(tid: params[:tid])
+                   .first
+
+      if (tagname.name.split(':')[0] == "lat") || (tagname.name.split(':')[0] == "lon")
+        node.delete_coord(tagname.name)
```
jywarren: Could we say tag instead of tagname as it's the whole record, and then also rename to delete_coord_attribute? Thanks!!!

milaaraujo: Done!

milaaraujo:
Hey @jywarren, can you merge this?

PR #4536, opened 2019-01-05
Code cleaning

sagarpreet-chadha:
Found these backslash in code . Seems useless after reading this : https://stackoverflow.com/questions/35042182/backslash-in-ruby-conditional-statement


sagarpreet-chadha:
Will need

PR #4538, opened 2019-01-05
Fixes #4311 optional profile picture at sign up

### LINKED ISSUE #4311: Optional Image while SignUp
@jywarren we have sidebar to add image at /profile/edit but we don't have any image upload option at the /signup.
Can we have an optional image upload in the signup?

IshaGupta18:
Fixes #4311 

(A little change: I had to change the default image, because of its presence in most places)

image

This is how it looks!

image

image

### DIFF
--- a/app/views/users/_create_form.html.erb
+++ b/app/views/users/_create_form.html.erb
@@ -17,10 +17,10 @@
 
     <div class="col-sm-6" style="display:flex; justify-content: center;">
       <div class="propic">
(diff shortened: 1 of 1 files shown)


jywarren:
Great work here. Really looks good. Thanks a lot @IshaGupta18 and @SidharthBansal !!!!

SidharthBansal:
I realised how efficiently we can do 10x work in 1x time if we have a supportive team. Giving right direction to 10 people can bring back 100 people which do 1k x work. Glad to be part of PL.  Thanks Isha

PR #4556, opened 2019-01-07
added password_placeholder

gurpreet9001:


gurpreet9001:
it is working but my installation has crashed so cant provide screenshot
it is a simple update therefore it is not hard to notice

gurpreet9001:
capture

this is the simple change

gurpreet9001:
i am not able to execute because my local copy is not working

SidharthBansal:
You can try out to debug it.

PR #4564, opened 2019-01-08
Fixing “last active” column on /people

### LINKED ISSUE #4559: Debug “last active” column on /people
@jywarren, what is wrong with this column?

milaaraujo:
Fixes #4559 

I'm changing the query to use revision timestamps instead of using node timestamps.

I also fixed the `Username` and `Joined` columns. `/people?sort=username` is trying to order the results by `name` instead of `username` (which generates an error, since there is no attribute with that name). `/people?sort=joined` is ordering users by the date his node was created instead of the user's creation date.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -92,11 +92,11 @@ class UsersController < ApplicationController
     end
 
     if sort_param == 'username'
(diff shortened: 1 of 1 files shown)


jywarren:
Great work!

PR #4572, opened 2019-01-08
UI fixes for range pages and stats page

### LINKED ISSUE #4139: UI fixes for range pages and stats page
* [x] Style `stats/range` page calendar section to be mobile responsive
* [x] Transform ` year, one month, or one week further back`  to fa arrow buttons
* [x]  Add cleaner links to range and subscriptions  pages on stats page
* [x] Re evaluate the choice of words used on the range page
* [x] Implement the new stats page design

cesswairimu:
Fixes #4139

### DIFF
--- a/app/views/stats/_stats_nav.html.erb
+++ b/app/views/stats/_stats_nav.html.erb
@@ -11,4 +11,3 @@
         <h4><a href="/stats/subscriptions">Subscription stats</a></h4>
     </nav>
 </div>
-
--- a/app/views/stats/_stats_nav.html.erb
+++ b/app/views/stats/_stats_nav.html.erb
@@ -1,13 +1,13 @@
-<div align="center" >
-    <nav class="navbar navbar-dark bg-primary">
-       <h4><a href="/stats"  style="color:white;">All Stats</a></h4>
-    </nav>
+<div align="center">
+<nav class="navbar navbar-light bg-light <%= current_page? stats_path%>">
+    <h5><a href="/stats"> All Stats</a></h5>
+</nav>
 
-    <nav class="navbar navbar-light" style="background-color:  #81d4fa;">
-        <h4><a href="/stats/range" style="color:#000080;">Range-based Stats</a></h4>
-    </nav>
+<nav class="navbar navbar-light bg-light <%= current_page? stats_range_path%>">
+    <h5> <a href="/stats/range">Range-based Stats</a></h5>
+</nav>
 
-    <nav class="navbar navbar-light" style="background-color: #e3f2fd;">
-        <h4><a href="/stats/subscriptions">Subscription stats</a></h4>
-    </nav>
+<nav class="navbar navbar-light bg-light <%= current_page? stats_subscriptions_path%>">
+    <h5><a href="/stats/subscriptions">Subscription stats</a></h5>
+</nav>
 </div>
--- a/app/views/stats/_stats_nav.html.erb
+++ b/app/views/stats/_stats_nav.html.erb
@@ -1,0 +1,14 @@
+<div align="center" >
+    <nav class="navbar navbar-dark bg-primary">
+       <h4><a href="/stats"  style="color:white;">All Stats</a></h4>
+    </nav>
+
+    <nav class="navbar navbar-light" style="background-color:  #81d4fa;">
+        <h4><a href="/stats/range" style="color:#000080;">Range-based Stats</a></h4>
+    </nav>
+
+    <nav class="navbar navbar-light" style="background-color: #e3f2fd;">
+        <h4><a href="/stats/subscriptions">Subscription stats</a></h4>
+    </nav>
+</div>
+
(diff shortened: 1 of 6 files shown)


cesswairimu:
barry
@ebarry from your feedback on the `/stats/range` page I changed and added some wordings. Here :point_up:  is a screenshot please take a look and tell me how it looks and any phrases I can change or if its cleare now. Thanks

gauravano:
Hi, @cesswairimu a design suggestion -  how about putting these starting and ending labels inside the input field only, at the extreme right of each field? 

We can keep this as a follow-up or a FTO too.

Thanks!

jywarren:
But a lot of the UI design of this can be addressed through the mockup @cesswairimu has posted in another issue - @cesswairimu did you want to link us there for reference?

jywarren:
Also in terms of consolidating, can we progressively move features from the basic stats pages to their "range" equivalents, then merge them?

cesswairimu:
@gavarano @jywarren thanks for the tips, I am fixing the labels and placeholders. 
I have just added the mockup image to the referenced issue above.
@jywarren I don't really follow  about the "moving features  from basic stats pages to range equivalents" kindly explain further. Thanks

jywarren:
Well, basically all the stats /could/ be presented for an arbitrary range
of dates. The 52 week stats, etc etc. could all be fit into a framework of
"pick a time period", you know?

So if we increasingly move everything into the range pages, we can start to
think of /all/ the stats functions as being part of the range pages. We may
need to be careful if people select really big ranges! But we could cache,
or have some "turn off" if the range is too large (and show a notice like
"pick a range shorter than X to see this graph" maybe?")

The most fundamental stats like "for all time" could be shown at the top,
above the range-specific region. Does that make sense?

cesswairimu:
Aha! Yes I now get it. Sounds great. Thanks

cesswairimu:
barry1

jywarren:
Hi, actually wanted to note maybe the Bootstrap form label styling -- not
placeholder -- might be important, since the start and end date may be
populated already in most views? So showing the small all-caps label as in
the Bootstrap form styles I'd linked to, just above the input, would be a
good solution for this. What do you think? Sorry, my screenshot had shown
placeholders TOO -- i keep sending ambiguous input! Apologies!!!

cesswairimu:
No worries @jywarren It was @gauravano suggestion to have the placeholders alone, but we can have both I guess. Thanks

cesswairimu:
new_range

new_stat

jywarren:
This is looking good - it's a little hard to see whether you're on the Range page or the general Stats page -- but if we merge the two, this won't matter. What do you think, Cess? Maybe we should merge the two first, then do a unified design? Or wait, let's try to do a unified design here and then that will guide the merging process?

jywarren:
This looks good - is it ready to be merged? Thanks!!!

jywarren:
Oops, left two different comments in 2 windows 😂 

I mean to say this looks good to merge, but we might want to do a bit more design work to achieve a unified design instead of 2 different types, and then we can proceed with implementing the unified design. 

Is it ready, then, and we can move forward like that? 

Thanks for your patience on my responding to this one!

cesswairimu:
Yeah @jywarren it is ready. I am working on the graphs but I guess that can come as a different PR. By merging the two do you mean we do away with the `/stats/range` page and move its  contents to `/stats`? Thanks

cesswairimu:
Maybe I can hide the link to the current page. So if one is at `/stats` the `All stats` link is hid. What do you think?

jywarren:
Sure, you could use the Bootstrap `disabled` class too. I'll wait to merge if you want to update that? Once it's there, we're good to merge here, thank you!!!

cesswairimu:
disable

cesswairimu:
@jywarren I pushed the code for the disabling active link. Thanks

app/views/stats/_stats_nav.html.erb
```diff
@@ -0,0 +1,13 @@
+<div align="center" >
+    <nav class="navbar navbar-dark bg-primary">
+        <h4><a href="/stats" class="<%= current_page? stats_path%>" style="color:white;">All Stats</a></h4>
```
jywarren: What if we removed the `h4` here and the one you're on could be `background-color: #e3f2fd` while the others would just have blank backgrounds? I think that could clarify a bit. Then we should be good!

cesswairimu: cool, on it

cesswairimu:
h5
@jywarren I removed the `h4` but  the links were looking weird. Using `h5` :point_up: how does it look?

jywarren:
that's strange -- how is it different from how they look in the Bootstrap
documentation? I think we can merge this now but would like to revisit and
see what we can do to get some of the Bootstrap stuff looking a little more
consistent. But great work!

jywarren:
Actually maybe you could ask for help from someone else who's done a lot of Bootstrap work... maybe someone in the weekly check-in? To team up with you on that follow-up?

Awesome work, Cess!!!

PR #4575, opened 2019-01-08
removed dynamic form (holding)

jywarren:
See original code here:

https://github.com/publiclab/plots2/commit/4f13238a4837356d3bbf4cfffd7d14a24e7e42dc

We may still use this plugin but shouldn't -- `error_messages` unfortunately occurs in many places! See:

https://github.com/publiclab/plots2/search?q=error_messages&unscoped_q=error_messages

Errors should instead be in format:

```ruby
  <% if @model.errors.any? %>
    <div id="error_explanation">
      <h2><%= pluralize(@model.errors.count, "error") %> prohibited this model from being saved:</h2>

      <ul>
      <% @model.errors.full_messages.each do |msg| %>
        <li><%= msg %></li>
      <% end %>
      </ul>
    </div>
  <% end %>
```

****

### Update

We'll refactor out all the old `dynamic_form` usages one by one as FTOs in #8545, then circle back here once that's all completed!


jywarren:
Unfortunately, we're going to have to refactor them in each of these places, OR make a partial to manage these, so we can call:

`error_messages(@node)` and it outputs them with the above template.

jywarren:
@gauravano - do you think we could craft an issue around this? Perhaps someone new would like to take it up?

gauravano:
@jywarren please update the status of PR. Thanks!

Manasa2850:
@jywarren @cesswairimu I think we can merge this now since #9237 is merged and all the GitHub Action tests are passing!
Thanks!

cesswairimu:
I will pull this and try to rebase,

cesswairimu (approved):
All green now :tada: 
Thanks @jywarren  and @Manasa2850

PR #4587, opened 2019-01-10
Follow tag asynchronously

### LINKED ISSUE #4565: Follow tags without redirecting and reloading
Currently following or unfollowing tags (https://publiclab.org/tag) redirects to a different link (https://publiclab.org/subscription) which is I assume not necessary, because a user wishing to follow multiple tags might be annoyed to these redirections. So, we can make follow, unfollow of tags asynchronous without any sort of reloading or redirection. That would I guess enrich the user experience!

Thank you!

shubhscoder:
Fixes #4565 
GIF for the feature
follow_tags

Thanks!


app/controllers/subscription_controller.rb
```diff
@@ -46,32 +46,23 @@ def add
           begin
             tag.save!
           rescue ActiveRecord::RecordInvalid
-            flash[:error] = tag.errors.full_messages
-            redirect_to "/subscriptions" + "?_=" + Time.now.to_i.to_s
+            render :json => { status: "500", message: "error" }
```
jywarren: Hi, this is great but I think we need to preserve the original non-JS version too, as this method is used in other places. Could we do that?

app/views/tag/index.html.erb
```diff
@@ -1,4 +1,7 @@
 <div class="col-md-12">
+  <div class="alert alert-success" id="follow-success" style="display: none;position: absolute;">
```
jywarren: Here, maybe we need to use Noty to display something using pure JavaScript, so we can consolidate all the JS for this into one place, rather than inserting new HTML... you can read about Noty here:

https://github.com/publiclab/plots2/search?utf8=%E2%9C%93&q=noty&type=

app/views/tag/index.html.erb
```diff
@@ -80,3 +83,21 @@
 </div>
 
 <div class="text-center"> <%= will_paginate @tags, :renderer => BootstrapPagination::Rails if @paginated %></div>
+
+<script type="text/javascript">
```
jywarren: Here, I'd like to propose that we make this into a separate JavaScript file, and a function which accepts a jQuery selector like:

```js
setupRemoteFollow($('follow-btn-remote'));
```

What it can do is read the `href` attribute of each `<a class="follow-btn-remote">` tag on the page, and parse it to set up the remote call. 

Then we include this JS file on all pages. Then all we have to do to implement the remote call is to give any link the `follow-btn-remote` class, and it gets hooked up automatically. And if JavaScript fails, it will revert to being a normal link that doesn't require JS.

How does that sound?

shubhscoder: Yes sure sounds good, I will do that !

app/controllers/subscription_controller.rb
shubhscoder: Yes, we can do that, just need to check if request.xhr return true right?

app/views/tag/index.html.erb
shubhscoder: Okay, that looks good I will change that!

app/views/tag/index.html.erb
shubhscoder: @jywarren I am trying to implement this. We are trying to generalize and use in this file in various remote requests right? But each one would have different requirements, like in this case we need to change the 'follow' button to 'following', etc. In this case how can we implement this? Thank you !

shubhscoder:
@jywarren I have made the required changes. Sorry for the delay. This is the updated GIF 
final_tags

config/initializers/assets.rb
```diff
@@ -48,5 +48,6 @@
   'user_tags.css.scss',
   'wiki.css',
   'horsey.js',
-  'atwho_autocomplete.js'
+  'atwho_autocomplete.js',
+  'remote_link_handler.js'
```
jywarren: Actually do you think we could include this only on that one page, using `<%= javascript_include_tag "remote_link_handler" %>` (i think that's the syntax, you may need to check on other pages to confirm this)? And perhaps a different name would be best, like `async_tag_subscriptions.js`?

shubhscoder:
@jywarren I m not sure how codeclimate can be fixed, because the nested ifs and concatenation of literals is kindoff necessary. Do you think there's a way around this?

gauravano:
@shubhscoder we can see about codeclimate fixes after handling the Travis tests errors - https://travis-ci.org/publiclab/plots2/builds/500247993#L3275-L3317

shubhscoder:
> @shubhscoder we can see about codeclimate fixes after handling the Travis tests errors - https://travis-ci.org/publiclab/plots2/builds/500247993#L3275-L3317

Yeah, travis errors are because in the production environment async_tag_subscriptions.js is not a precompiled asset. @gauravano do you have any idea of how can we just include the javascript file in a .html.erb with 
<%= javascript_include_tag "remote_link_handler" %>
without adding it to pre-compiled list of assets? Thanking you in advance!

app/assets/javascripts/async_tag_subscriptions.js
```diff
@@ -0,0 +1,34 @@
...
+
+      $('#tagform').submit(function(e) {
+        e.preventDefault();
+        window.location = "/tag/" + ($('#taginput').val()).replace(/\s/g, '-');
+      });
+      $('.index-follow-buttons').on('ajax:success', function(data, status, xhr){
+	    console.log("Hello");
+	    var data_recv = JSON.parse(JSON.stringify(status));
+	    notyNotification('relax', 3000, 'success', 'top', data_recv.message + 'Click <a href="../subscriptions"> here </a> to manage your subscriptions. ');
+	    var html_new = '<a rel="tooltip" title=Following' +' class="btn btn-default btn-sm active" href="'+'/unsubscribe/tag/'+ data_recv.tagname + '"> <i class="fa fa-user-plus" aria-hidden="true"></i>Following' +'</a>';
```
jywarren: Here I think CodeClimate is wondering why there are some string concatenations between strings -- could be simpler maybe? Also, i think `Following` needs quotation marks?

app/views/tag/index.html.erb
```diff
@@ -1,3 +1,4 @@
+<%= javascript_include_tag('async_tag_subscriptions.js') %>
```
jywarren: Not sure why it doesn't like this - same as here, no? 

https://github.com/publiclab/plots2/blob/ff7d91eb00fddabfd9169530257cb223e1846ac5/app/views/questions/show.html.erb#L1

jywarren: OH, you know what? Maybe it wants to precompile it in its own file, to minify it, and since we aren't referencing it in application.js, it's not getting merged with everything else. So you could follow the instructions and:

```
        Add `Rails.application.config.assets.precompile += %w( async_tag_subscriptions.js )` to `config/initializers/assets.rb` and restart your server
```

Sound right?

jywarren:
Looks great -- ready for merge?

shubhscoder:
@jywarren yes I think we can merge this, thank you so much for helping me along with this!

jywarren:
🎉

shubhscoder:
Thank you so much !

shubhscoder:
@jywarren , I will start with #4759  now. Thank you!

jywarren:
awesome!

PR #1799, opened 2017-11-26
Rails 4.1 upgrade

jywarren:
@ryzokuken @icarito  -- trying to tackle this here. Any PRs against this one welcome!

These are changes that are not compatible with Rails 3, unlike #1798.

Heavily based on #1160 by @sjones88

PR #4588, opened 2019-01-10
Spam revisions page fixed

### LINKED ISSUE #3082: spam/revisions page down
Hi!
Looks like the https://publiclab.org/spam/revisions page is down. Any ideas? 

Best,
Stevie

gauravano:
Fixes #3082 

Thanks!

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -71,7 +71,7 @@ class Node < ActiveRecord::Base
     updated_at.strftime('%B %Y')
   end
 
-  has_many :revision, foreign_key: 'nid' # , dependent: :destroy # re-enable in Rails 5
+  has_many :revision, foreign_key: 'nid', dependent: :destroy
   # wasn't working to tie it to .vid, manually defining below
   #  has_one :drupal_main_image, :foreign_key => 'vid', :dependent => :destroy
   #  has_many :drupal_content_field_image_gallery, :foreign_key => 'nid'
--- a/app/views/admin/_revisions.html.erb
+++ b/app/views/admin/_revisions.html.erb
@@ -9,7 +9,7 @@
   <% @revisions.each do |revision| %>
     <tr id="n<%= revision.vid %>">
       <td>
-        <i class="fa fa-file"></i> <a href="<%= revision.parent.path %>"><%= revision.title.truncate(40) %></a><br />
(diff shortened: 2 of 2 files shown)


app/models/node.rb
```diff
@@ -71,7 +71,7 @@ def updated_month
     updated_at.strftime('%B %Y')
   end
 
-  has_many :revision, foreign_key: 'nid' # , dependent: :destroy # re-enable in Rails 5
```
gauravano: Hi @jywarren, I just found this `dependent : : destroy` commented due to which if a node of spammed revision is deleted by author or  moderator, the revisions aren't deleted. 

And, ` # , dependent: :destroy # re-enable in Rails 5` is present at many places so we have collecting redundant data lately, which is still present. 

Should we enable all `dependent: :destroy`?

gauravano:
@jywarren fix is working! Verified from unstable. Please review before merging. Thanks!

app/models/node.rb
gauravano: Hey @Souravirus any idea about these comments? Thanks!

Souravirus: No, I didn't had any idea about these. Sorry about that.

gauravano: No issue!

gauravano: @jywarren these all comments are put by you at Rails 4.1 upgrade - https://github.com/publiclab/plots2/pull/1799/ 

Do you think we can remove these comments now? Also, should we delete the dependent data like revisions without node, etc?

jywarren: Hmm, i think we could open an issue for this - It was so long ago i don't remember. I think we have to be very very careful here, as incorrect `dependent: destroy` has caused data loss in the past. Let's do it separate from this PR! Thanks!

jywarren: image

app/models/node.rb
```diff
@@ -71,7 +71,7 @@ def updated_month
...
+  has_many :revision, foreign_key: 'nid', dependent: :destroy
```
gauravano: @jywarren but here :arrow_up: I have added this `dependent: :destroy`. Should I remove this?

jywarren: uh oh - oops, i missed this! Um, I think this is OK. I read the docs carefully. Thanks!

gauravano: I also peeked :mag: at this :arrow_right: https://guides.rubyonrails.org/association_basics.html while changing so this one should be fine.

PR #4590, opened 2019-01-10
add rake assets:precompile

jywarren:
will check `rake assets:precompile` in our Travis tests, and report failure if it doesn't complete (i think). But it may take some extra time to run! @icarito


jywarren:
@icarito looks like about 1 minute longer than ones like this: https://travis-ci.org/publiclab/plots2/builds/477931430

about 1-2m longer than the average

PR #4599, opened 2019-01-11
Username not editable in database

kunalvaswani123:
Fixes  #4367 (<=== Add issue number here)

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [x] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [x] code is in uniquely-named feature branch and has no merge conflicts 📁
* [x] PR is descriptively titled 📑
* [x] screenshots/GIFs are attached 📎 in case of UI updation
* [x] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software 

Thanks!

PR #4621, opened 2019-01-15
Test correction for PR #4599 which is failing travis builds

gauravano:
Fixes #4599  

The test added in PR #4599 was trying to create a user session without activating the Authlogic. The test passed through travis and so I merged the PR but after the merging of code, the builds were showing the below error, which should be shown in PR build #4599 but wasn't shown. So, this PR is attempt to correct the test added in that PR.
 
```ruby
You must activate the Authlogic::Session::Base.controller with a controller object before creating objects
```

Sorry for the inconvenience to all!


jywarren:
Great! Thanks!

Issue #4522, opened 2019-01-04
OAuth Login fail and Form Login fail for some users


kevinzluo:
![loginfail](https://user-images.githubusercontent.com/44309027/50713904-48faf800-103c-11e9-90af-b0e9a61bd84e.gif)

I already had an account created through Github with username kevinz_qu.

Chrome, Windows 10

jonathanzhang53:
i Tried logging in with GitHub / google and it gave me a success message but it still showed login/signup at the top right

geekychaser:
@kevinzluo i am facing same issue and i am also not able to login on my local server.
![image](https://user-images.githubusercontent.com/24489162/50721936-6b991980-10ed-11e9-90e9-acd50f1cad96.png)

thank you

SidharthBansal:
This is caused in DrupalUser PR most probably. We need to check that and fix it urgently.

SidharthBansal:
Google oauth working on Ubuntu + Chrome

SidharthBansal:
Github OAuth working on Ubuntu + Chrome on pl.org

SidharthBansal:
Is it working on the pl.org?

geekychaser:
no it is not working this side

geekychaser:
so the issue has arised with this commit id `258bf087753c78b239c796c9fd5b02d007e4264c` if you try to point head before this commit everything would work smoothly.
please correct me if i am wrong
thank you

gauravano:
@geekychaser the issue faced by you as described here - https://github.com/publiclab/plots2/issues/4522#issuecomment-451636135 is not an unexpected error. The alert shown in screenshot is rendered in case a moderator bans a user(in your case admin is banned). So, do you remember banning the admin or something like that? Thanks!

SidharthBansal:
It is also appearing on my localhost. It is appearing for user, moderator
and admin all.

gauravano:
> It is also appearing on my localhost. It is appearing for user, moderator
and admin all.

You mean user banned alert?

SidharthBansal:
Yeah

gauravano:
I am inspecting.

geekychaser:
[...]

PR #4623, opened 2019-01-15
Testing unresolved question in #4522 by trying to get test to fail

jywarren:
#4522


gauravano:
Please see #4621. Thanks!

jywarren:
This is potentially an additional issue - just that the tests I wrote were supposed to throw an error if `create_with_omniauth` did not explicitly set `user.status`, but actually they passed. So I'm trying to figure out why they didn't catch the issue which we have guessed caused the whole OAuth problem?

jywarren:
OK! This properly failed. Going to close it now!

PR #4625, opened 2019-01-15
set user status in seeds

### LINKED ISSUE #4609: Banned Accounts and probable API bug
Quoting the gitter chatroom, I mentioned

> Plots2 currently returns a "banned" message for https://publiclab.org/profile/rexagod
> Same results in case of https://publiclab.org/profile/dsxv007 and https://publiclab.org/profile/geekychaser
> However this doesn't happen in case of mods and admin. https://publiclab.org/profile/gauravano https://publiclab.org/profile/warren
> Hope I'm not vague. But I think this might be something that's worth looking into.

Edit: As of now, @gauravano has unbanned all of us.

Also another concern of mine was that @jywarren's profile wasn't returned in the output payload, is this a bug?

jywarren:
Fixes #4609

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -10,19 +10,28 @@
 everything_tag = Tag.create! "name"=>"everything", "description"=>"", "weight"=>0
 
 # set up 3 basic users: admin, mod, and user
-admin = User.create! "username"=>"admin", "email"=>"admin@example.com",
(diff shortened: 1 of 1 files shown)


jywarren:
Must've created this before merging @gauravano's fix in #4621

PR #4652, opened 2019-01-17
adjust last updated label in /people template

### LINKED ISSUE #4559: Debug “last active” column on /people

jywarren:
This could work, but if not: 

`t = user.revisions.order(timestamp: :desc).first.try(:created_at);time_ago_in_words(t) if (t)`

Fixes #4559


milaaraujo:
Working now (using your second approach)! :)

jywarren:
Amazing!!!

PR #4653, opened 2019-01-17
(WIP)Active topics

### LINKED ISSUE #4644: Subscription Prompt next to Topics on Dashboard
image

Tthis isn't like a normal view here, the view is being rendered through cache, from an entry in the database, here are the links:
https://github.com/publiclab/plots2/blob/9f7cfc47ded41ed3a7370c66a9c4f7af9704af59/app/views/sidebar/_dashboard.html.erb#L38-L41

https://github.com/publiclab/plots2/blob/9f7cfc47ded41ed3a7370c66a9c4f7af9704af59/app/helpers/application_helper.rb#L57-L64

So the view is rendered from the database's entry. Does this make sense?

@jywarren can you help us how we can know the tags passed in this section? We need to pass them in the Subscription button.

SidharthBansal:
Fixes #4644 (<=== Add issue number here)

Checklist
- [ ] which occur a lot on nodes they've posted
- [ ] which occur a lot on nodes on which they've commented
- [ ] tags user is not subscribed to
- [ ] user.active_topics(limit) -- which could return a max of limit tags
- [ ] we could limit this using time if we wanted to optimize! Or cache it.

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -329,4 +329,9 @@ class Tag < ApplicationRecord
          .limit(5)
     end
   end
+
+  def active_topics(user)
+    t = TagSelection.where('user_id = ? AND following = ?', user.id, true)
+                    .collect(&:tid)
+  end
 end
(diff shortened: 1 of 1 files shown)


SidharthBansal:
@jywarren working on this one.

PR #4654, opened 2019-01-18
Raw data from stats page

### LINKED ISSUE #963: Raw data from stats page
For the https://publiclab.org/stats page I'd like to either be able to download the raw data as a .csv (privacy considerations here?) and/or to potentially hover over a graph and be able to see the numbers per week.

cesswairimu:
Fixes #963 


cesswairimu:
@jywarren what else would we potentially like to download from the page. I am currently have `wikis` and `notes`

jywarren:
Hi @cesswairimu this is great! What does the output look like? Is it statistical output like the # per week, or per day, or is it the full raw data? I think @ebarry will be in on Monday and you can ask her what's most useful here. I think probably the count per period is more interesting than the full raw data of all the actual content...

I think for the time range, we should try to offer these for the range page periods, does that make sense? Calculating it for the whole site might be... stress on the system. 

For types of data, perhaps a count per period of:

* comments
* contributors
* questions asked
* questions answered
* subscriptions?

Also note that we may want to decide on a "bin" size to start with - so we can say `X comments per week over this period of weeks` -- right? But is "week" the right bin size, so to speak? This may be a question for Liz too. 

I think @milaaraujo will be doing some caching work soon and may be a good person to connect with as you think about optimizing some of these queries, and note this guide: https://guides.rubyonrails.org/caching_with_rails.html

Thanks!

cesswairimu:
The output is current the raw data I see that count would be more helpful. I change it to that as we wait for @ebarry to give us more direction for this. Will also checkout the caching blog and talk to @milaaraujo .Thanks @jywarren

cesswairimu:
screenshot from 2019-01-19 20-51-48
Reusing range from the range page as you suggested, maybe we can add these download buttons the data. I don't think another page would be necessary since we are showing the same count as this page. The download buttons would be visible to admins only download. If we use this the "bin" size will be month as that is the default range for the range page. I don't know how this sounds

ebarry:
ok i see what you mean about "binning" -- for reference, in the past, i have downloaded by one month, three month, and yearly timespans. I have never downloaded by day. However, to get the graphs that @skilfullycurled created about which days of the week are busiest (https://publiclab.org/evaluation#Online+analytics), that would require the full raw data.

skilfullycurled:
Hey everyone.  Wow!  So much data lately!!

If I'm understanding the questions (put very roughly):

1) **_Should the data be downloadable in a "raw" format where with one line per item and you let the analyst aggregate it into their own time periods (is this what you mean by bins?), or do you pre-aggregate it into those time periods and the analyst has the choice to download the daily, weekly, monthly, yearly etc._**

2) **_Which data counts do you include?_**

Full disclosure: this next part I write only with the understanding of what it's like to take a csv file and manipulate it, and therefore I recognize that not all of these ideas will be technically feasible due to system constraints or to amount of programming required.

On the surface, it seems like anyone using the data would want as much as possible and then later decide what periods they want to aggregate it into.  However, pre-aggregated periods of time would lower the bar for someone just getting started be it through programming or using a spreadsheet program, or needs a "quick and dirty" way to make a graph for some presentation.

For someone like myself, I want as much data as possible because I don't know what's interesting until I explore the data.  I wanted to explore things such as the distribution of counts per user per day.  So what I did was export the different tables as csv's and then join on keys as needed.  The schema.rb file was very helpful as was the mysql back end of the database when I installed it.  

Would it be possible to have the counts as you have suggested above but also a page that lists accessible tables (e.g. not the one's that have identifying data like email addresses) with csv download links for each table and let the analyst handle it from there?  The size of the wiki edits csv that I have for all edits from 2011 to summer of 2016 with the unix timestamp, nid, vid, uid, and title is only 1.2 MB.

skilfullycurled:
> and have another page for downloading all the information as you suggested above I hope that is what you meant with "lists for accessible tables"?

Yes, I think we're on the same (html) page with what I meant.  M overarching idea is to have a place where the most comprehensive dataset can be retrieved _**while requiring the least amount of strain on both the developers and servers**_.  Of course, some of the tables will have to be sanitized, and to the extent that it's not very labor intensive to exclude unnecessary things like session tokens, then that'd be great but in an ideal world, the developer would just export to csv and let person downloading sort through the rest. 

For example: Suppose someone wants to have a list of tags, their counts, and creation date.  The site provides (these tables may have changed) the community_tags and term_data tables as downloadable csv's, and it's up to them to do a join on "tid" to create a dataset with the tid, date, name, and count.  Happy to help with the documentation at some point.

With regard to which data is needed I could use some clarification on where the idea stands first.  Is the idea that you choose a specific start and end date or a start and end month and/or year?

cesswairimu:
Yes @skilfullycurled the idea is to choose a specific start date and end date and have an option of downloading the data created within that range. I would also like to hear @gauravano @jywarren @SidharthBansal point on this just to make sure I am not going out of scope code-wise

skilfullycurled:
Got it.  Thanks @cesswairimu.  I ask because if it's one specific date to the next then I think the question is, what type of user is the data geared towards?  Will that user know how to aggregate it by other means?  I think it'd be easy enough to have some documentation with how to do that in Excel.  

Also, I don't know if this would save you any computational resources, but I think you could just provide start and end windows by month and year.  I don't know how the server resources work but I can't think of a reason why you'd need to download exact dates like you're booking a flight.

cesswairimu:
@skilfullycurled kindly take a look at the new code here https://unstable.publiclab.org/stats/json and see if its anything close to what you had in mind... for now only `Download as Json` is working. Hopefully no one will push to unstable before you take a look :smile: Thanks

cesswairimu:
raw-range
@jywarren how does that position look?  Also any ideas on button color?

jywarren:
looks great! I think we could stick with the basic `btn-default` white bg
buttons, thanks!

cesswairimu:
Thanks..we restrict this download to admins alone?

cesswairimu:
raw-range

cesswairimu:
@jywarren I asked on slack but I am not sure if you understood my question.. On how I can  refactor the code climate issue above

jywarren:
Oh sorry! First, the buttons look beautiful! 

I just went ahead and approved it. CodeClimate recs are helpful but we needn't follow every single one. Thanks! Is this ready then?

cesswairimu:
Yeah its ready

PR #4655, opened 2019-01-18
[WIP] Adding user.watching_location() method

### LINKED ISSUE #4551: Prototype ways to subscribe to new node posts based on proximity
### Proximity subscriptions

>Basically this would adapt the existing subscription code but expand it so that you could subscribe to a location -- like your profile location (not sure how to select a location otherwise... could start with this but later generalize to anything with .lat and .lon methods?). When content gets posted, tagged within X distance of the subscribed location, you get a notification.
In theory we could poll "nearby content" via the API periodically and issue notifications. That could be easier. But we need some way to record that you are following a place. Hope that makes sense!
-@jywarren

The most basic idea is that:

1. someone -- person 1 -- could "watch" a location by subscribing to it;
we'll call it "location A"
2. if any content is posted near that location, say at "location B",
notifications would be sent to person 1
3. we may need to think of the "distance" or "radius" which would be
covered by the subscription
-------------------------------------------
Tasks:
* [ ] - [#4655] Implement `User.watchingLocation(nwlat, selat, nwlng, selng)` which should return a collection of User records

milaaraujo:
Fixes #4551 

>Start by developing a test fixture with a 2-coordinate subscription user tag, and then see if we can properly detect overlap using User.watchingLocation(lat, lon, lat2, lon2) which should return a collection of User records.

`user.watching_location(nwlat, selat, nwlng, selng)` returns all the users that are subscribed to an area **inside** (nwlat, selat, nwlng, selng)`.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -383,8 +383,7 @@ class User < ActiveRecord::Base
     self.username = registration['nickname'] if username.blank?
   end
 
-  def self.watching_location(nwlat,selat,nwlng,selng)
-
+  def self.watching_location( nwlat, selat, nwlng, selng)
     raise("Must contain all four coordinates") if nwlat.nil?
     raise("Must contain all four coordinates") if nwlng.nil?
     raise("Must contain all four coordinates") if selat.nil?
@@ -399,7 +398,6 @@ class User < ActiveRecord::Base
     uids = TagSelection.where('tag_selections.tid IN (?)', tids).collect(&:user_id).uniq || []
 
     User.where("id IN (?)", uids)
-
   end
 
   def self.find_by_username_case_insensitive(username)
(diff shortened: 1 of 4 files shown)


jywarren:
Wow this was fast! And impressive! 

I'll look at resolving the codeclimate notes, but i do wonder if we could compress those very repetitive lines at all? But not a big deal.

Could we add a negative test? Where we confirm it doesn't return a given result?

Also, would you mind changing the coordinate values to be below 180, which is typically a max value for longitude, and also introducing one that's negative, since we can have negative lat/lon values as well? 

One last thing - let's note the order of coordinates in a comment in the test, since some folks may use the test as an example of how to use the code. Sound good?

Then this should be good to merge, very cool! Thanks!

milaaraujo:
> would you mind changing the coordinate values to be below 180, which is typically a max value for longitude

As you can see I'm not very good at geography! :joy: 

I think I addressed all your points now. And we actually don't need the "Must contain all four coordinates" check... Calling the method using less than 4 parameters would failed anyway (`wrong number of arguments`).

milaaraujo:
Hey @jywarren, can you review this?

PR #4703, opened 2019-01-24
Adding `field` option to `/api/srch/nearbyPeople`

### LINKED ISSUE #4659: Consider ways to group people by topic focus
>Is there an API endpoint for people who follow a tag, and are in a bounding box? Would we be able to address the above through that? @jywarren 

* [x] - Adding `field` option to `/api/srch/nearbyPeople` (#4703)
* [x] - Add tests and clarification

milaaraujo:
Fixes #4659

`field=[string]`: Accepts the value `node_tag` for searching users following the node tag (topic).

`/api/srch/nearbyPeople?nwlat=100&selat=0&nwlng=0&selng=100&tag=water&field=node_tag`

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -139,7 +139,7 @@ class SearchService
 
   # Search nearby people with respect to given latitude, longitute and tags
   # and package up as a DocResult
-  def tagNearbyPeople(coordinates, tag, period = nil, sort_by = nil, order_direction = nil, limit = 10)
+  def tagNearbyPeople(coordinates, tag, field, period = nil, sort_by = nil, order_direction = nil, limit = 10)
     raise("Must contain all four coordinates") if coordinates["nwlat"].nil?
     raise("Must contain all four coordinates") if coordinates["nwlng"].nil?
     raise("Must contain all four coordinates") if coordinates["selat"].nil?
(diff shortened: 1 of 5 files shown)


app/services/search_service.rb
```diff
@@ -160,16 +160,23 @@ def tagNearbyPeople(coordinates, tag, period = nil, sort_by = nil, order_directi
...
     if tag.present?
-      user_locations = User.joins(:user_tags)
-                           .where('user_tags.value LIKE ?', tag)
-                           .where(id: user_locations.select("rusers.id"))
+      if field.present? && field == 'node_tag'
+        tids = Tag.where("term_data.name = ?", tag).collect(&:tid).uniq || []
+        uids = TagSelection.where('tag_selections.tid IN (?)', tids).collect(&:user_id).uniq || []
+      else
+        uids = User.joins(:user_tags)
+                   .where('user_tags.value LIKE ?', tag)
```
jywarren: Here is LIKE necessary? And if it is, should we use the % character? I'm not super familiar with LIKE outside the use of that character but just checking. Thanks!

milaaraujo: I think it would work the same way as `=` in this case (https://stackoverflow.com/questions/543580/equals-vs-like). But I changed it to `'user_tags.value = ?'`. And we don't want to use `%` character as we need the exact match.

doc/API.md
```diff
@@ -110,11 +110,13 @@ All the endpoints have the optional parameter `limit` (10 by default) where you
...
 
-  `nwlat=[northwest latitude],selat=[southeast latitude],nwlng=[northwest longitude],selng=[southeast longitude]`: search notes within the rectangle boundary made by these diagonally opposite coordinates.
+  `nwlat=[northwest latitude],selat=[southeast latitude],nwlng=[northwest longitude],selng=[southeast longitude]`: Search users within the rectangle boundary made by these diagonally opposite coordinates.
 
   **Optional:**
 
-  `tag=[string]`: the search can be refined by passing a tag field.
+  `tag=[string]`: The search can be refined by passing a user tag field.
+
+  `field=[string]`: Accepts the value `node_tag` for searching users following the node tag (topic).
```
jywarren: I'll go ahead and merge this. Maybe though we could be extra clear above that the default is user_tag and that this switches the use of the tag parameter to link node tags instead... Maybe just as a quick follow-up? Thanks!!

jywarren:
Merged, awesome!!

PR #4709, opened 2019-01-26
removing `/api/srch/peoplelocations` endpoint and adding tests to `/api/srch/nearbyPeople`

### LINKED ISSUE #4685: Remove `/api/srch/peoplelocations` endpoint
Hey @jywarren,

Since we have the option `sort_by=recent` in the `/api/srch/nearbyPeople`, do you think we can remove `/api/srch/peoplelocations` endpoint (search the the users with most recent activity that have coordinates - lat and lon values - on their user tags)? `nearbyPeople` search users within the rectangle boundary which is better than use the two points logic.

It looks like `/api/srch/peoplelocations` is not used anywhere.

milaaraujo:
Fixes #4685 and #4659


milaaraujo:
Hey @jywarren! Can you review this?

jywarren:
Hi, this looks good. I've asked @sagarpreet-chadha to look at the maps at https://publiclab.org/inline-maps, and develop tests around them. Can you confirm that none are failing due to API changes or errors? I'm just a bit concerned at the failure rate of our inline maps features and while I know you've checked this thoroughly, I'd like to just be extra sure as I think a few things may have gotten crossed I'm the awesome upgrades to the API that mutiple people have been coordinating around. 

In any case if you can give the go ahead one more time I'm happy to merge this.

jywarren:
Perfect, thanks both!!

Issue #4690, opened 2019-01-22
header wrapping issues at some window sizes (recurring issue)


jywarren:
We keep seeing issues with the header wrapping to 2 lines as items are added or removed from the header, and I'd like to think about how to address this - both the current issue and how to prevent this in the future. 

Right now I believe the new icons added have increased the width beyond the breakpoints in CSS. Read more about the issue here: 

https://github.com/publiclab/plots2/issues/4502

One open PR is being blocked by this issue as well. 

https://github.com/publiclab/plots2/pull/4461

Which is discouraging for the author, I'm sure!

Let's solve the immediate issue by removing the icons until we can fix the wrapping, and let's brainstorm a plan to either block edits to this code or to have a more flexible layout which can survive some changes without breaking. 

Thanks everyone for your patience, this is a delicate section of code and we appreciate your help!!

IshaGupta18:
Yes, removing the icons is solving this issue momentarily, we can brainstorm to make this code more flexible. But for now, do you want to remove the icons, I could make a PR for it?

jywarren:
That'd be great, thank you!

gauravano:
Hi @jywarren, the menu work is in progress at https://github.com/publiclab/plots2/pull/4517 , so you can add your comments there to help @Chrisyee22. 

@IshaGupta18 you can collaborate with @Chrisyee at #4517 so that issue can be resolved as soon as possible. Thanks!

IshaGupta18:
Yes absolutely @gauravano ! @jywarren #4713 is my PR for the temporary fix. Thanks a lot!

siddhishree:
@IshaGupta18 Can I work on this Issue with you?

IshaGupta18:
I am so sorry I missed your comment @siddhishree . Please feel free to work on this issue. We would love your help here!

jywarren:
I think we are OK on this, can someone try resizing the header to very small sizes and check that it doesn't fold to 2 lines when we do it? If we're OK, then we can close this for now!

IshaGupta18:
Checked! Works fine!

jywarren:
Thanks IshaGupta!

gauravano:
Great!

PR #4713, opened 2019-01-27
Removing icons from navbar tabs

IshaGupta18:
#4690 temporary solution by removing fa-icons from navbar tabs:

header


jywarren:
Thanks for this fast fix @IshaGupta18 - i'm going to merge this now!

PR #4740, opened 2019-01-31
"/comments" page with 2 sub-sections and buttons for moderating

### LINKED ISSUE #4616: Display comment status on https://publiclab.org/comments if comment is not approved
As a moderator, i see lots of spam comments on /comments, but they're not marked as such so I thought I had to moderate them. Let's have them appear moderated (labeled, greyed out?) if they are not currently published!

Follow-up: we could have a `https://publiclab.org/comments?status=1` route available to moderators so you don't have to see them, linked to from the top of the page. Regular users wouldn't need to see this link, but it's useful for mods!

gauravano:
Fixes #4616 

Thanks!

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -151,7 +151,7 @@ class AdminController < ApplicationController
     else
       flash[:error] = 'Only moderators can moderate comments.'
     end
-    redirect_to '/dashboard' + '?_=' + Time.now.to_i.to_s
+    redirect_back(fallback_location: root_path)
   end
 
   def publish_comment
--- a/app/views/comments/_comments.html.erb
+++ b/app/views/comments/_comments.html.erb
@@ -12,7 +12,7 @@
       <% end %>
     </tr>
 
-  <% @comments.each do |comment| %>
+  <% comments.each do |comment| %>
    <tr id="c<% comment.cid %>">
      <td>
       "<%= raw strip_tags(sanitize(RDiscount.new(comment.comment).to_html)).truncate(150) %>" | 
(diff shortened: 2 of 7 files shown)


gauravano:
Currently, /comments show all the comments mixed up so I made 2 divisions at that page All comments and Moderated comments. The later division would be visible only when a user is moderator and admin. I have also added buttons there so that it would be convenient to moderator to moderate comments. Also, we used to redirect to dashboard on marking a comment as spam/approve but I have removed that and added now page will reload so moderators don't have to open the page again and again. 

comment_div1


Moderator options view:
ezgif com-optimize 3


@jywarren @SidharthBansal please review. Thanks!

app/views/comments/_comments.html.erb
```diff
@@ -12,7 +12,7 @@
       <% end %>
     </tr>
 
-  <% @comments.each do |comment| %>
+  <% comments.each do |comment| %>
```
gauravano: @jywarren I adjusted user_controller.rb - https://github.com/publiclab/plots2/pull/4740/files#diff-4e05ad0d64e6100656b63ad1e78f32c5R287 to update with this change so working everything working. Do you remember, if we use this partial _comments anywhere else except profile page and /comments? Let me know, and I will make the change accordingly. Thanks!

jywarren: I don't believe we use it anywhere else! So then this should be ready to merge? 👍

gauravano: yes, the work is complete according to me. you can merge it, if you find all changes good in review. Also, should I test on `unstable` once? Thanks!

app/controllers/admin_controller.rb
```diff
@@ -151,7 +151,7 @@ def mark_comment_spam
     else
       flash[:error] = 'Only moderators can moderate comments.'
     end
-    redirect_to '/dashboard' + '?_=' + Time.now.to_i.to_s
+    redirect_back(fallback_location: root_path)
```
jywarren: Will this bust the cache the way the timestamp code does?

jywarren:
Just one clarification, but otherwise this is great! Feel free to merge it after addressing it. Thanks!!!!!

app/controllers/admin_controller.rb
gauravano: It will redirect to the page from where the request is generated so - If I approve a comment from a research note, then the page will just reload(redirecting to requesting page which is note only). As the behaviour is reload so we don't need to bust cache explicitly (I don't faced cache issue while testing). In case, redirect_back wasn't able to redirect to the page from where request is generated then it will redirect to the root_path which is dashboard

Fallback demo:
fallback_location

jywarren:
great, sounds good!

gauravano:
Hi @jywarren, merged this one :tada:. Thanks!

PR #4746, opened 2019-02-01
passing location tags through related

jywarren:


test/unit/node_test.rb
```diff
@@ -9,13 +9,28 @@ class NodeTest < ActiveSupport::TestCase
...
 
   test 'basic question attributes' do
     question = nodes(:question)
     assert question.answered
   end
 
+  test 'basic location attributes' do
+    map = nodes(:map)
+    assert_equal [], map.tags.collect(&:name)
+    assert map.has_power_tag('lat')
```
jywarren: Don't know why this is failing so I added an assertion to just dump what tagnames are actually in there... ☹️

jywarren:
Oh wow, `map` has no tags!??

PR #4763, opened 2019-02-05
Caching on heavily-used API calls: srch/api/profiles

### LINKED ISSUE #3149: Caching on heavily-used API calls
We had some outages this morning due to an old, unoptimized API endpoint being hit hard by the new `@` autocomplete... 

https://github.com/monterail/grape-rails-cache

Implementation looks pretty simple! @stefannibrasil @milaaraujo interested in this? As your new code gets relied on more, this seems like a good idea...

```ruby
modulemodul  MyApi < Grape::API
  format :json

  include Grape::Rails::Cache

  resources :posts do
    desc "Return a post"
    get ":id" do
      post = Post.find(params[:id])

      cache(key: "api:posts:#{post.id}", etag: post.updated_at, expires_in: 2.hours) do # <= 2 hour expiration... we can decide how much is appropriate!
        post # post.extend(PostRepresenter) etc, any code that renders response
      end

    end
  end
end
```

milaaraujo:
Fixes #3149

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -71,14 +71,14 @@ gem 'mailman', require: false
 gem 'reverse_markdown'
 
 # run with `bundle install --without production` or `bundle install --without mysql` to exclude this
-group :mysql, :production, :test, :development do
+group :mysql, :production do
   gem 'mysql2', '>= 0.4.4'
   # mysql 0.4.3+ causes a version mismatch, apparently, and demands 'activerecord-mysql2-adapter'
 end
 
-#group :sqlite, :development do
-#  gem 'sqlite3'
-#end
+group :sqlite, :development do
+  gem 'sqlite3'
+end
 
 gem 'coffee-rails', '~> 4.2.2'
 gem 'execjs' # See https://github.com/sstephenson/execjs#readme for more supported runtimes
--- a/Gemfile
+++ b/Gemfile
@@ -71,20 +71,21 @@ gem 'mailman', require: false
 gem 'reverse_markdown'
 
 # run with `bundle install --without production` or `bundle install --without mysql` to exclude this
-group :mysql, :production do
+group :mysql, :production, :test, :development do
   gem 'mysql2', '>= 0.4.4'
   # mysql 0.4.3+ causes a version mismatch, apparently, and demands 'activerecord-mysql2-adapter'
 end
 
-group :sqlite, :development do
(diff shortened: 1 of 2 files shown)


jywarren:
Great. I'll roll this out today and we can see how it does!

jywarren:
Strangely i don't see a lot of tracking in Skylight: https://oss.skylight.io/app/applications/GZDPChmcfm1Q/recent/6h/endpoints/Srch::Search%20%5BGET%5D%20srch%2Fprofiles

Not sure if this is because these calls aren't monitored?

milaaraujo:
Depending on how it does we can implement the caching for the other endpoints...

jywarren:
https://publiclab.org/api/srch/profiles?query=jywa&sort_by=recent&field=username&_=1549385453136 is about 1.9 seconds right now.

jywarren:
https://publiclab.org/api/srch/profiles?query=sidha&sort_by=recent&field=username&_=1549385453136 is 2.4 seconds

jywarren:
https://publiclab.org/api/srch/profiles?query=camila&sort_by=recent&field=username&_=1549385453136 1.8 seconds

https://publiclab.org/api/srch/profiles?query=mila&sort_by=recent&field=username&_=1549385453136 2.48 seconds

jywarren:
Wow, ok so i did a hard cache flush and reloaded this:

https://publiclab.org/api/srch/profiles?query=camila&sort_by=recent&field=username&_=1549385453136

Initially 6 seconds, but on repeating, 578ms

https://publiclab.org/api/srch/profiles?query=sidha&sort_by=recent&field=username

2 seconds, then on refresh 553 ms

https://publiclab.org/api/srch/profiles?query=jywa&sort_by=recent&field=username

 2 seconds then 80 ms

so, i think this worked pretty well. Let me try actually using it in a comment...

jywarren:
image

And a longer interactive session here:

image

You can see how typing `@warren` 3 times, the later ones show much much lower load times; the initial ones are at 2-4 seconds, the later at 68-631ms, although one was 2.4s

image

This is a big improvement, but I do want to note how many requests go out -- i think we can change the settings on the typeahead to refine this too. See a gif:


typeahead


Also note it's sending lots of requests where `query` is empty. We should adjust so it only sends when there are at least 2.

PR #4780, opened 2019-02-08
Trying to fixing width related oddities

### LINKED ISSUE #4732: Email digest follow-up changes list
Our daily email digest is now working! Now we have a few follow ups to do. 

* [x] Images seem broken
* [x] dates on posts don't seem right (https://github.com/publiclab/plots2/pull/4851)
* [x] daily digest is labeled weekly (https://github.com/publiclab/plots2/issues/4822) (solved in https://github.com/publiclab/plots2/pull/4955)
* [x] some page width oddities on a smartphone
* [x] images not loading https://github.com/publiclab/plots2/issues/5230
* [x] item content is in Markdown and needs to be parsed to HTML
* [x] looks like maybe updated wiki pages are included? For now let's just show notes (https://github.com/publiclab/plots2/pull/5568)
* [x] are we sure the same posts aren't being sent out multiple times? #5907
* [x] adding today's date to the heading title of the digest email https://github.com/publiclab/plots2/pull/5331/

Overall this is very exciting!

IshaGupta18:
Fixes #4732 

Code has been refactored to fix the width issues. Here's the gif:

email

### DIFF
--- a/app/views/subscription_mailer/send_digest.html.erb
+++ b/app/views/subscription_mailer/send_digest.html.erb
@@ -24,7 +24,7 @@
         </div>
         <% if n.main_image.present? %>
           <div style="padding-bottom: 30px;">
-            <img src="WhatsApp_Image_2019-01-29_at_10.44.10_PM.jpeg" style="width: 90px;height: 90px;border-radius: 5px;margin-left: 14px;">
+            <img src="<%= n.main_image.path(:default) %>" style="width: 90px;height: 90px;border-radius: 5px;margin-left: 14px;">
           </div>
          <% end %>
         <div colspan="2" style="padding-bottom: 32px;">
--- a/app/views/subscription_mailer/send_digest.html.erb
+++ b/app/views/subscription_mailer/send_digest.html.erb
@@ -1,43 +1,35 @@
 <header style="text-align: center;color:black">
-	<img src="https://avatars2.githubusercontent.com/u/4621650?s=200&v=4" style="width: 100px;height: 100px">
-	<p style="font-size: 20px">[Public Lab] Your Weekly Digest</p>
+  <img src="https://avatars2.githubusercontent.com/u/4621650?s=200&v=4" style="width: 100px;height: 100px">
+  <p style="font-size: 20px">[Public Lab] Your Weekly Digest</p>
 </header>
 
 <div style="width: 46vw;margin: 0 auto;margin-top: 54px;color: black;">
-	<p style="font-size: 1.2em; font-family: sans-serif;margin-bottom: 0;color: lightgray;">Top picks for you</p>
-	<hr id="first" style="border: none;height: 1px;background-color: #e2e2e2;margin-bottom: 32px;">
+  <p style="font-size: 1.2em; font-family: sans-serif;margin-bottom: 0;color: lightgray;">Top picks for you</p>
+  <hr id="first" style="border: none;height: 1px;background-color: #e2e2e2;margin-bottom: 32px;">
   <table style="font-family: sans-serif;">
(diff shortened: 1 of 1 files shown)


app/views/subscription_mailer/send_digest.html.erb
```diff
@@ -1,46 +1,38 @@
 <header style="text-align: center;color:black">
-	<img src="https://avatars2.githubusercontent.com/u/4621650?s=200&v=4" style="width: 100px;height: 100px">
-	<p style="font-size: 20px">[Public Lab] Your Weekly Digest</p>
+  <img src="https://avatars2.githubusercontent.com/u/4621650?s=200&v=4" style="width: 100px;height: 100px">
+  <p style="font-size: 20px">[Public Lab] Your Weekly Digest</p>
```
jywarren: Could you somehow check here if it's daily or weekly, and display accordingly? If it's too much, we can do it in a follow-up PR. This is great!

app/views/subscription_mailer/send_digest.html.erb
```diff
@@ -1,46 +1,38 @@
 <header style="text-align: center;color:black">
-	<img src="https://avatars2.githubusercontent.com/u/4621650?s=200&v=4" style="width: 100px;height: 100px">
-	<p style="font-size: 20px">[Public Lab] Your Weekly Digest</p>
+  <img src="https://avatars2.githubusercontent.com/u/4621650?s=200&v=4" style="width: 100px;height: 100px">
```
jywarren: This could use code like this: `<% if n.author.photo %><%= n.author.photo_path(:thumb) %><% end %>` but also could be done in a follow-up PR if you like?

IshaGupta18: I think you mean to change in line 16, ie Author's profile image right? This one is for the PL logo, if I am not wrong.

IshaGupta18: Okay, so I have done this, and it works!:

image

jywarren:
Oh good catch, thanks yes!

app/views/subscription_mailer/send_digest.html.erb
IshaGupta18: Okay so I am trying to do this, but I think that it is probably hard coded somewhere, and a variable for checking which type of digest is not being sent to the actions:

Like here, in the mailer code:
https://github.com/publiclab/plots2/blob/d3b122efaa42a9d6b06a70f53626334ad2cf085b/app/mailers/subscription_mailer.rb#L58

Here in the send_digest_email function, top picks are being selected for a week only:

https://github.com/publiclab/plots2/blob/d3b122efaa42a9d6b06a70f53626334ad2cf085b/app/models/user.rb#L325-L330

I think here, when send_digest_email function is being called, the frequency parameter should be passed somehow, and received everywhere, then we can correct the code in the send_digest.html.erb:

https://github.com/publiclab/plots2/blob/d3b122efaa42a9d6b06a70f53626334ad2cf085b/app/jobs/digest_mail_job.rb#L4-L14

Please correct me if I am wrong anywhere, then we can explore this better.

IshaGupta18: @jywarren any updates here?

jywarren: You're totally right. Let's leave it for now in this PR and if you could copy your notes into a new issue (linking from #4732 ) we can address this with some refactoring! Great sleuthing, @IshaGupta18 !!!

IshaGupta18: Absolutely @jywarren! Thanks a lot!

PR #4787, opened 2019-02-10
Addition of Translation prompts

### LINKED ISSUE #3262: Planning issue for translation/internationalization/localization systems
We have some basic internationalization/localization via the `i18n` gem, but need to think through how to improve it, and ALSO come up with a streamlined workflow to recruit and integrate new translations on an ongoing basis. This is a planning issue for this set of projects.

****

## Translation systems

Our translation files are spread over lots of files and folders in https://github.com/publiclab/plots2/tree/master/config/locales

I'm inclined to make them all one big file, or at least consolidate them some. One folder with key/value would be SO AWESOME! I wasn't sure why our implementation fragmented everything across so many folders. I think we could do this in Transifex.

Discussion of alternative platforms via @ebarry :

* LocalApp: https://www.localeapp.com/projects/9741/translations/5689967 
* https://crowdin.com/ (github integration)
* https://poeditor.com/ (github integration)

## Ongoing translation workflow

We need a workflow to regularly go from Transifex (or something similar) back into the translation files. 

Transifex has GitHub integration: https://docs.transifex.com/integrations/github/ and awesome ability to auto-sync in TravisCI: https://docs.transifex.com/integrations/github/#integrating-the-client-with-travis-ci

This would grab new translations each time we accepted a pull request, and register new texts that need translation!

### Outreach

For actually doing the translations, I'm hesitant to make a huge # of translation issues in the same repository, even if they are first-timer type issues. I think we could maybe get more contributors to translate outside the GitHub interface (which is most welcoming to coders). 

I think we'll need a good outreach game to actually recruit people to do translations! Maybe, we could tweet them out? We could show a random phrase at the bottom of the site each time the site is loaded, that needs translation?

****

If we happen to find alternatives (especially open source ones) that's great, but we can stick with Transifex if there's no big reason to switch. 

The ideal might be that with fewer files, we could copy in new changes from either another service like Transifex periodically (monthly?), or from files in another repository (although this doesn't make it more accessible by going outside the GitHub interface)

We might even try to automate the import and run it periodically. But it could be a regular task manually too. 

## Translation prompts

Anywhere we display translated text using the `t('some_string_here')` helper, we ought to display a little globe icon with a tooltip saying `Needs translation? Click to help translate this text.` or something (ideally in the browser-configured language!)

## Translating content

We'll need to think about how to translate both regular site content (wiki pages + notes) and ALSO features, which are used around the site for things like the static footer, explanatory text, etc.

Perhaps there should be a language type on revisions, and we could make it possible to track translated revisions alongside regular ones. Or perhaps a translated node, linked to the main node?

Compare https://publiclab.org/barnraising to https://publiclab.org/wiki/barnraising-guangzhou-china for an example of 2 versions of a page in 2 languages. 

### Features

Some parts of site content -- like the footer and front page -- are run in nodes, as database content. Let's think of a way to have features like `frontpage-en` vs `frontpage-zh` and `frontpage-es` 

## Helpers for translation

@siaw23 points out, and I agree, that translation is tough for newcomers submitting new code. The current format is something like:

```ruby
<%= t('some_string_here') %>
```

...where `some_string_here` is the key to some string of text. 

I propose a new helper, maybe called `translate()`, which looks for translations, but if none are present, displays the key text itself. That way, we can create new interfaces that include text, without first having to do translation for them -- they'll have the fallback text. So:

```ruby
<%= translate('Click here to begin') %>
```

which initially would have only the default english, but via the prompts (above) would encourage community translation efforts. 

## Scanning for translation opportunities

Finally, we might have a script which scans for uses of the new helper (from above) that don't have entries in the default translation file, and copies them in. 

****

This is a lot. Just kicking off some discussion and we can start to move things into separate issues soon! Love to hear peoples' input!

gauravano:
Fixes #3262  Part:2

Thanks!

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -136,4 +136,15 @@ module ApplicationHelper
   def current_page?(page)
     'is-active' if request.path == page
   end
+
+  def translation(key, options = {})
+    translated_string = t(key, options)
+    if translated_string.length > 3 && current_user &.has_tag('translation-helper')
+      %(<span>#{translated_string} <a href="https://www.transifex.com/publiclab/publiclaborg/translate/#de/$?q=text%3A#{translated_string}">
+          <i data-toggle="tooltip" data-placement="top" title="Needs translation? Click to help translate this text." style="position:relative; right:2px; color:#bbb;" class="fa fa-globe"></i></a>
+       </span>)
+    else
+      translated_string
+    end
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -139,9 +139,10 @@ module ApplicationHelper
 
   def translation(key, options = {})
     translated_string = t(key, options)
-    if translated_string.length > 3 && current_user &.has_tag('translation-helper')
+
+    if translated_string.length > 3 && current_user &.has_tag('translation-helper') && !translated_string.include?("translation missing")
       %(<span>#{translated_string} <a href="https://www.transifex.com/publiclab/publiclaborg/translate/#de/$?q=text%3A#{translated_string}">
-          <i data-toggle="tooltip" data-placement="top" title="Needs translation? Click to help translate this text." style="position:relative; right:2px; color:#bbb;" class="fa fa-globe"></i></a>
+          <i data-toggle='tooltip' data-placement='top' title='Needs translation? Click to help translate this text.' style='position:relative; right:2px; color:#bbb;' class='fa fa-globe'></i></a>
        </span>)
     else
       translated_string
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -142,7 +142,7 @@ module ApplicationHelper
 
     if translated_string.length > 3 && current_user &.has_tag('translation-helper') && !translated_string.include?("translation missing")
       %(<span>#{translated_string} <a href="https://www.transifex.com/publiclab/publiclaborg/translate/#de/$?q=text%3A#{translated_string}">
-          <i data-toggle='tooltip' data-placement='top' title='Needs translation? Click to help translate this text.' style='position:relative; right:2px; color:#bbb;' class='fa fa-globe'></i></a>
+          <i data-toggle='tooltip' data-placement='top' title='Needs translation? Click to help translate this text.' style='position:relative; right:2px; color:#bbb; font-size: 15px;' class='fa fa-globe'></i></a>
        </span>)
     else
       translated_string
--- a/app/views/dashboard/dashboard.html.erb
+++ b/app/views/dashboard/dashboard.html.erb
@@ -37,7 +37,7 @@
             <div class="blog-caption">
        
               <h4>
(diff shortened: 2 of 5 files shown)


gauravano:
Hey @jywarren, here's updated view as per your [suggestions](https://github.com/publiclab/plots2/issues/3262#issuecomment-454874880) 

screenshot from 2019-02-10 23-29-33

Please give your view on this - https://github.com/publiclab/plots2/issues/3262#issuecomment-461928807 . Thanks!

Thanks!

jywarren:
Can we tie this to a parameter like ?t=true so we can test this on
production a bit, share it with people for feedback and test the overall
workflow before publishing it for everyone?

Thanks Gaurav!

gauravano:
Hey @jywarren, we can surely do this, but currently I have only added the helper function on just 2 translations and we need to add the function on every translation or maybe some more to test it. 

I am waiting for your reply on this - https://github.com/publiclab/plots2/issues/3262#issuecomment-461928807 as here I have asked my query regarding the addition of value to translation helper instead of key and some more things. Whatever we decide on I will change the translation helper accordingly, so please answer those queries. Thanks!

gauravano:
> Can we tie this to a parameter like ?t=true so we can test this on production a bit, share it with people for feedback and test the overall workflow before publishing it for everyone?

We can do this, but the user has to add `?t=true` to every URL they navigate to for seeing prompts, how about a button which would add a cookie in the browser and disabling that button would delete the cookie?

Thanks!

gauravano:
@jywarren

jywarren:
Could we base it on a user tag? Thanks, and sorry i'm being slow
responding! I'd prefer to avoid cookies because they can create states that
are hard to debug. So, `user.has_tag('translation-helpers')` maybe?

gauravano:
User_tag seems good. Ok, will make changes to the helper, tomorrow. Thanks!

jywarren:
awesome, thank you!!!

gauravano:
Hey @jywarren, I have added the check for user tag **translation-helper** and I have added the helper method to "/blog" and "/subscriptions" pages so we can check. Thanks!

gauravano:
Here's the screenshot of Subscription page with prompts:

subscriptions_page_prompt

@jywarren pushing this to unstable, let me know if there is any design change required. Also, I have accommodated all previous ideas like we are not showing the prompt icon if translated string is less than 3 char in length and if there is no translation present.

jywarren:
This is great -- however, shall we only show prompts when a translation doesn't exist? Or are you showing this already? Thanks!!!

jywarren:
Oh i see from your last comment this is right. Perfect then! Ready to merge?

jywarren:
If so, go ahead!!! 👍 🎉 This is awesome!

gauravano:
hey @jywarren, you saw build failed on stable due to missing debounce/index.js - https://jenkins.laboratoriopublico.org/job/Plots-Stable/944/

jywarren:
Yep, I think Sebastian has a PR open for reordering those steps but needs
to test on unstable. Thanks!

Issue #3571, opened 2018-10-02
Suggestion / request: use Dependabot to keep dependencies up-to-date (reopening for Yarn)


greysteil:
First of all, thanks for PublicLab!

I've got a suggestion / request: would you be up for using [Dependabot](https://dependabot.com) to automatically create dependency update PRs for this repo? I ran it against my fork and it generated [these PRs](https://github.com/greysteil/plots2/pulls). I'll port the `uglifier` one across to this repo so it's easier for you to check it out.

I built Dependabot, but I'm honestly only suggesting it because I hope it can save you some time. I'd love any feedback, and obviously having open source repos using Dependabot helps boost its profile, but if it's not helpful to you then it's not really worth anything.

You can install it from [here](https://github.com/apps/dependabot) or [here](https://github.com/marketplace/dependabot) if you decide to give it a try. It's been through GitHub's security testing (to be allowed in the GitHub Marketplace) and is used by a few thousand organisations, and the source code is [here](https://github.com/dependabot/dependabot-core).

:octocat:

jywarren:
Wow, cool! @publiclab/reviewers what do you think? 

Is there a way to have it do the PRs one at a time? Our PR queue is overflowing a bit. 

Also, will to be able to deal with any conflicts if it opens multiple at once and we start merging? 

Thanks!

greysteil:
> Is there a way to have it do the PRs one at a time? Our PR queue is overflowing a bit.

I could set it to only have one PR open at a time, but maybe it would be better for me to help out getting the initial PRs merged? I can review (Dependabot itself is written in Ruby) and get each PR to a point where a maintainer just needs to hit merge. Then, once you're up-to-date you should expect 1-2 PRs a week.

> Also, will to be able to deal with any conflicts if it opens multiple at once and we start merging?

Yep! Automatically detects conflicts and fixes them. I've been working on it full time for 18 months, so it's pretty good at what it does. 😎

jywarren:
Thank you! How many PRs will Dependabot open, do you think? Can we do them in batches? Sorry, we're just a bit cautious about opening more than 5 at a time, but even if it's 10, we can just get ready for it so we're not overwhelmed 😅

Another thing we can do is run this on a less active Public Lab project for starters -- like MapKnitter. Although we would have to warn you that some of our less active projects are far less tested and up to date, so maybe that is not a great idea...

[...]

Issue #1215, opened 2017-01-25
Switch from Bower to Yarn


ryzokuken:
**NOTE: Quite a bit of the content in here is directly taken from different pages on the internet, but it's done just to make this proposal as self-contained as possible**

[Rails 5.1 ships with yarn](https://twitter.com/dhh/status/808348184481124352) also, it's now possible to use webpack on Rails using yarn, using which we could use newer JS features like classes, modules and the overall cleaner and concise syntax of ES2015. 

# Why would Yarn work for us?

1. Firstly, yarn does not use an independent package index. Rather, it fetches packages from both the `npm` and `bower` registries, thus making sure all our current packages will work just fine. 

2. Because it supports the npm registry in addition to the normal bower registry we've been utilizing, it certainly increases the amount of packages we have available to us for a certain task.

3. By making use of certain network and package hacks (like local caching of packages), yarn speeds up the downloading, installation and in general the workflow.

4. As @ananyo2012 suggested, lockfiles are helpful in projects as significant as ours, literally keeping all the developers as well as the CI and production "on the same page".

5. Because installing packages is faster with yarn, it will speed up the testing process, making the CI take significantly less time.

6. Numerous tutorials and articles on the internet regarding usage of yarn as a replacement for bower and the asset pipeline make it evident that not only is it more efficient, but also that we're not the only ones considering to use it in a rails environment.

7. Yarn was created by Facebook in a collaboration with Exponent, Google, and Tilde. It is maintained and supported by all these giants and the Javascript community in general. This was done keeping in mind that none of the present package managers were equally effective.

*I would try to keep adding more stuff here as and when they come to my mind*

# What makes Yarn cool?

## Ultra Fast

Yarn caches every package it downloads so it never needs to download it again. It also parallelizes operations to maximize resource utilization so install times are faster than ever.

## Mega Secure

Yarn uses checksums to verify the integrity of every installed package before its code is executed.

## Super Reliable

Using a detailed, but concise, lockfile format, and a deterministic algorithm for installs, Yarn is able to guarantee that an install that worked on one system will work exactly the same way on any other system.

## Offline Mode

If you've installed a package before, you can install it again without any internet connection.

## Deterministic

The same dependencies will be installed the same exact way across every machine regardless of install order.

## Network Performance

Yarn efficiently queues up requests and avoids request waterfalls in order to maximize network utilization.

## Same Packages

Install any package from npm and keep your package workflow the same.

## Network Resilience

[...]

PR #4751, opened 2019-02-03
Translation: Put single/double quotes around values in config/locales/controllers files

akshaygupta533:
Fixes #4729 #4702

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [x] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [x] code is in uniquely-named feature branch and has no merge conflicts 📁
* [x] PR is descriptively titled 📑
* [x] screenshots/GIFs are attached 📎 in case of UI updation
* [x] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software 

Thanks!

PR #4804, opened 2019-02-13
update yarn sources

jywarren:
Related to #1215 #3571 potentially https://github.com/publiclab/plots2/pull/4099/files

We may be able to just remove a number of these based on this analysis:

https://github.com/publiclab/plots2/blame/3f0bdf2bb12a89d72a05a3d52738f2d8d41d3aeb/bower.json


jywarren:
Major reductions!  Can prob remove flot soon too -- it's only used on https://publiclab.org/contributors now.

jywarren:
Hi @gauravano i'm working to get a bunch of yarn deps upgraded and one was `i18n-js` - do you think you can try `3.x.x` to see if it works as you've been using it? There are actually no available versions at `2.x.x` unless we point directly at the GitHub repository and I'd prefer just to come up to date. Thanks!

https://github.com/fnando/i18n-js

gauravano:
Hey @jywarren, thanks for mentioning. I definitely want to test newer version. Can we do it after testing and merging #4751 ? Thanks!

package.json
```diff
@@ -21,51 +21,43 @@
...
+    "bootstrap-datepicker": "^1.6.0",
+    "bootstrap-tokenfield": "^0.12.0",
+    "chart.js": "^2.7.0",
+    "cytoscape": "^3.3.2",
+    "file-saver": "^2.0.0",
+    "flot": "^0.8.3",
+    "font-awesome": "^4.5.0",
+    "horsey": "^3.0.0",
+    "i18n-js": "^3.2.1",
+    "ics": "nwcell/ics.js^0.2.0",
```
jywarren: Tried changing this to `ics` via npm but that's actually a different project https://github.com/adamgibbons/ics (maybe better maintained)

whereas we're using: https://github.com/nwcell/ics.js

jywarren:
This seems to be running smoothly at unstable. @gauravano if you think the i18js works, you can go ahead and merge this!

PR #4848, opened 2019-02-18
Add more relevant search results

### LINKED ISSUE #4823: Add extra search results for related terms
There might be multiple ways to do this, and there are a few goals:

1. return search results for non-hyphenated terms when entering hyphenated terms, like `purpleair` when searching for `purple-air` (see https://github.com/publiclab/plots2/issues/3677#issuecomment-432246848)
2. the reverse; returning results for hyphenated terms when entering non-hyphenated ones, so adding `purple-air` results when searching for `purpleair` (harder because there is no simple algorithmic way to add hyphens in, like `maneatingcabbage` => `man-eating-cabbage` or `maneating-cabbage`)
3. we could also simply add in search results from associated terms based on a key-value listing, so when searching for `discussion lists` we could return `mailing lists` if we make a list of these term pairings. This could address the challenge of (2) above, but not sure what the implementation looks like. 

Could be related to stemming/lemmatizing as in https://github.com/publiclab/plots2/issues/3666 and solution https://github.com/publiclab/plots2/pull/4533 by @shubhscoder. 

Lemmatizer uses these dictionaries: https://github.com/yohasebe/lemmatizer/tree/master/lib/dict and I've opened an issue to ask about supplying custom dictionaries (this may also be helpful for different languages): https://github.com/yohasebe/lemmatizer/issues/5

However we may not want to "reduce" a word to a common core, rather, we may want to add additional search terms -- so for example, we could simply supply a YAML file of pairs like:

```yaml
purple-air: purpleair
purpleair: purple-air
h2s: hydrogen-sulfide
```

maybe in `/config/initializers/matchwords.yml`? - and we could match search terms against this in the `transform()` function here, adding them in:

https://github.com/publiclab/plots2/blob/e3cf2112469c8be284e541579d2bc02b62d39e7a/app/services/search_criteria.rb#L44-L48

So that a search for `h2s` would become a search for the query `h2s hydrogen-sulfide`

@shubhscoder this seems like it wouldn't be too difficult. Do you have any interest in implementing this?

shubhscoder:
Fixes #4823  

Thank you so much!


shubhscoder:
@jywarren need your reviews on this ! I was thinking of adding a few more tests for individual nodes like tags, wikis ,etc. Rest everything looks fine to me. What do you think? Thank you so much!

app/controllers/search_controller.rb
```diff
@@ -45,6 +55,24 @@ def all_content
...
+        @additional_search_querys << SearchCriteria.new(params)
+      end
+      params[:query] = results_with_probable_hyphens(@search_criteria.query)
+      @additional_search_querys << SearchCriteria.new(params)
+    end
+  end
+
+  def add_extra_results_for_transformed_queries(type)
+    search_type_object = ExecuteSearch.new.by(type, @search_criteria)
+    @additional_search_querys.each do |added_search_criteria|
```
jywarren: This is cool! Could we pass this into the above functions by returning the value of `additional_search_querys` from this method, instead of using an instance variable? Thanks!

lib/text_search.rb
```diff
@@ -5,4 +5,13 @@ def lemmatize(word)
     lem = Lemmatizer.new
     lem.lemma(word)
   end
+
+  def non_hyphenate_query(word)
+  	word.gsub('-','')
```
jywarren: Would you mind using two-space indents here? It helps us keep the code tidy!

test/functional/search_controller_test.rb
```diff
@@ -77,4 +77,19 @@ class SearchControllerTest < ActionController::TestCase
     assert_equal nodes_with_trawl, nodes_with_trawls
     assert_response :success
   end
+
+  test "search for purple-air also includes results for purpleair and vice versa" do
```
jywarren: Let's also add "for hyphenated searches" just so it's easy to find the word `hyphen` when searching the codebase!

lib/hyphenated_words.dict.txt
```diff
@@ -0,0 +1 @@
+noun purpleair purple-air
```
jywarren: Do you think we should just strip /all/ hyphens and provide them as related terms? We could do that with Regex! (https://rubular.org) and `'string'.gsub(/.../)`

We might rename this and use it for more complex additional terms like `h2s: hydrogen-sulfide` as in the original issue.

app/controllers/search_controller.rb
shubhscoder: I thought we could use it in the view saying "Also showing results for : values in additional search querys" that is why I made it a instance variable. No problem though I ll change it now!

lib/hyphenated_words.dict.txt
shubhscoder: @jywarren sorry But I am not getting what you are trying to say. Also the link you provided is not loading 

:sweat_smile:
--

lib/text_search.rb
shubhscoder: Did that. Thank you!

app/controllers/search_controller.rb
jywarren: Oh! Well, could we pass it into there too by parameter? Let me look in a bit, thanks!

lib/hyphenated_words.dict.txt
jywarren: Oops, http://rubular.com/ ! I meant we could do `'string'.gsub('-', '')` to remove hyphens pretty easily. I clarified in #4823 in a longer comment about the 3 cases, i hope that helps!

jywarren: Oops, http://rubular.com/ ! I meant we could do `'string'.gsub('-', '')` to remove hyphens pretty easily. I clarified in #4823 in a longer comment about the 3 cases, i hope that helps!

jywarren: Oops, http://rubular.com/ ! I meant we could do `'string'.gsub('-', '')` to remove hyphens pretty easily. I clarified in #4823 in a longer comment about the 3 cases, i hope that helps!

shubhscoder:
@jywarren  sorry for the delay, I have modified the PR according to our discussion in #4823 . Please let me know if we need some more changes. Thank you so much!

lib/hyphenated_words.dict.txt
jywarren: OK, but so can we do one more thing and provide a second dict that's for non-hyphen-related transforms like in my comment:

`h2s: hydrogen-sulfide`

Otherwise this PR is good to merge!

jywarren: I mean that we can either have one file specifically for "adding hyphens in" and another for "related terms" (even though both work similarly) OR we should simply consider "adding hyphens" to be a subset of "related terms" and have a single file. I'm fine with either, just don't want to have to ask people to put hyphen-unrelated word pairs into a file that's labelled with relation to hyphens, make sense? 

Great!!!!

shubhscoder:
@jywarren

> I mean that we can either have one file specifically for "adding hyphens in" and another for "related terms" (even though both work similarly) OR we should simply consider "adding hyphens" to be a subset of "related terms" and have a single file. I'm fine with either, just don't want to have to ask people to put hyphen-unrelated word pairs into a file that's labelled with relation to hyphens, make sense?

We can have one file? We can name it related_terms and that takes care of hyphenated terms as well as related terms, whats your suggestion on this?

jywarren:
Sure one file sounds good. Ready then? This is exciting!!!!!!

shubhscoder:
Yes @jywarren , I think this is ready! Thank you!

jywarren:
Congratulations! Great work!!!! 🎉 🎉

jywarren:
Shall we post a "call for related terms" now, in a new issue, where we can collect related term pairs?

shubhscoder:
Yes @jywarren . I will open the issue now, you can suggest some changes there if required. 
Thank you so much!

jywarren:
Hi @shubhscoder - i also noticed that in Travis each word is being output into the logs. Is there a way to turn that off? Just curious!

jywarren:
https://travis-ci.org/publiclab/plots2/builds/503876232

shubhscoder:
@jywarren , I am not sure of this, I will open a issue in the gem repository, maybe we could get some help there. Thank you!

PR #4850, opened 2019-02-18
Inline map changes 

### LINKED ISSUE #4722: Make Tests for inline maps .
This page :  https://publiclab.org/wiki/inline-maps  shows the documentation of all the inline maps that we have . 

All Inline maps use the same partial `_leaflet.html.erb`  except people-inline-map which uses 
`_peopleleaflet.html.erb` .

- [ ]  Remove HASH from URL when using people-inline-map -> No sense of using URL HASH when having more than one inline map on a page .

- [ ] Multiple dependencies being loaded when more than 1 inline map used .

- [ ] Documentation needs to be updated on the page -  https://publiclab.org/wiki/inline-maps . 

- [ ]  Write test :  Leaflet Map is loaded --- check if `<div class="leaflet-map"...` is present or not . When 2 inline maps are used then 2 such div's should be present . Add fixture data which calls `self.notes_map_by_tag` function in `nodes_shared.rb` to demonstrate the test . 

- [ ] Another test : Check if the marker is being loaded on the map ! This can be done by checking if 
` name of map instance uses number in range [1 ,100] , more map on 1 page causes names to clash !

sagarpreet-chadha:
Fixes #4722 !

### DIFF
--- a/app/views/map/_leaflet.html.erb
+++ b/app/views/map/_leaflet.html.erb
@@ -21,9 +21,7 @@
     var map_lon = <%= lon %> ;
 
      window.setTimeout(function(){
-      
-        // var user_latlng = L.latLng(map_lat, map_lon) ;
-        // map<%= unique_id %>.setView(user_latlng ,  <%= lat.to_s.length.to_i %> + 6) ;
+
         map<%= unique_id %>.setZoom(<%= lat.to_s.length.to_i %> + 6) ; 
     
      }, 3500);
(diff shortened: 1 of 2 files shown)


jywarren:
Awesome Sagarpreet!!

jywarren:
Published and better but still a little odd... what do you think?

jywarren:
Looks like: 

```
VM297:29409 Uncaught TypeError: Cannot read property 'featureLayer' of undefined
    at wisconsinLayer
```

Issue #4849, opened 2019-02-18
getting "Sprockets::FileNotFound in Home#dashboard" error suddenly on localhost


shubhscoder:
Suddenly I dont know what happened but I am unable to load any page on localhost because I am getting the above error.I have attached a screeshot of the error. Thank you so much!

![error_load](https://user-images.githubusercontent.com/18489285/52936234-8f54ae00-3381-11e9-8ef5-db9d44858385.png)

## Update

See below for `yarn install` and `yarn update` to resolve this

sashadev-sky:
@shubhscoder its the bootstrap file it specifies I haven't figured out how to fix it though. For me I can run my localhost but all my pushes to unstable fail

gauravano:
Copying @cesswairimu's message from Gitter channel :
> After pulling master yesterday I have been getting this error Sprockets::FileNotFound in uncommenting jquery in app/assets/javascripts/application.js seems to solve it. Is there anyone else experiencing this?

sashadev-sky:
it worked for me locally (localhost) when I ran yarn install and yarn upgrade but I don't really understand why it would work locally but not with a push to unstable

gauravano:
I ran the same commands some days back for resolving this issue. We can see `unstable`'s build log to find if the yarn install is running or not?

sashadev-sky:
https://jenkins.laboratoriopublico.org/job/Plots-Unstable/497/console
it says it runs but is using cache?

gauravano:
That's the issue!

sashadev-sky:
I cleared cache a bunch of times is there a way to set it so that jenkins stops using cache

gauravano:
I found this - https://stackoverflow.com/questions/35594987/how-to-force-docker-for-a-clean-build-of-an-image and let's see if it work or not

@cesswairimu and @shubhscoder you can resolve this issue locally by running commands as mentioned by sasha -https://github.com/publiclab/plots2/issues/4849#issuecomment-464647875 . 

@Souravirus @icarito any hint about this issue? 

Thanks!

shubhscoder:
Thank you so much @gauravano ! 
1. yarn install
2. yarn upgrade 
worked on the local.

gauravano:
@sashadev-sky unstable is back to normal.

gauravano:
I just force pushed the master branch on unstable and it's back to normal. You tried the force push, Sasha?

gauravano:
[...]

PR #4858, opened 2019-02-20
Try not to mount app directory only /public/system

### LINKED ISSUE #3261: Avoid running bower twice for testing
https://gitlab.com/publiclab/plots2/commit/2292036555b77f901256387eaa15e436874a538a

I tried the above patch but it somehow broke Jenkins. We should try again.
Bower is configured (thru `./.bowerrc`) to put libs in `./public/lib` but then we also mount the current dir, so we *shadow* the files with the current directory. That's why we've needed to run bower twice, but we will save some time if we manage to solve this! Maybe putting bower libs outside `/app`?

https://publiclab.org/ ...

> Many bugs are related to these -- please help us track it down and reproduce what you're seeing!

****

## Thank you!

icarito:
Intends to fix #3261 and also possibly #4849 and other issues arising from mountint the app dir as a volume in container.

Thanks!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ RUN bundle install --jobs=4
 
 ADD . /app
 WORKDIR /app
-RUN rake assets:precompile && rake tmp:cache:clear
 
 RUN yarn install && yarn upgrade
 RUN passenger-config compile-nginx-engine --connect-timeout 60 --idle-timeout 60
+RUN rake assets:precompile && rake tmp:cache:clear
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,6 @@ RUN bundle install --jobs=4
 
 ADD . /app
 WORKDIR /app
-RUN rake assets:precompile && rake tmp:cache:clear
 
 RUN yarn install && yarn upgrade
 RUN passenger-config compile-nginx-engine --connect-timeout 60 --idle-timeout 60
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,7 @@ RUN bundle install --jobs=4
 
 ADD . /app
 WORKDIR /app
+RUN rake assets:precompile && rake tmp:cache:clear
 
 RUN yarn install && yarn upgrade
 RUN passenger-config compile-nginx-engine --connect-timeout 60 --idle-timeout 60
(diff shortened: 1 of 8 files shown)


icarito:
This is passing now! I'll test it out in unstable, @jywarren please review if https://github.com/icarito/plots2/blob/not_mount_app_volume/containers/docker-compose-production.yml#L40-L45 sounds like all directories needed at runtime for _writing_.

icarito:
Tried this in _unstable_ and found 404 for `public/system` files, investigating...

icarito:
Issue was with assets precompilation. This is running in unstable now please try it out!

gauravano:
Hi @icarito, please update the status of the issue. Thanks!

Dockerfile
```diff
@@ -38,3 +38,4 @@ WORKDIR /app
 
 RUN yarn install && yarn upgrade
 RUN passenger-config compile-nginx-engine --connect-timeout 60 --idle-timeout 60
+RUN rake assets:precompile && rake tmp:cache:clear
```
icarito: This is the main insight to put assets inside the Docker container. This way we can push to production without running bundler / yarn in production.

icarito:
Hi! I _believe_ this is ready for merging - it does change fundamentally how we treat assets and in particular how we access local files (specific needed directories only are "mounted") - all else will be inside the container.
We should test this out in unstable - to confirm stuff like uploading files works in all instances.
This is a step in the direction of deploying pre-built containers to production instead of building them on the production server. Also it should fix issues with Gemfile.lock and Yarn.lock when deploying, that @jywarren reported.

icarito:
_Except_ travis just failed on it, checking again!

icarito:
Wow it looks like since I made this, we upgraded to a version of rails that _doesn't use_ `initialize_on_precompile=false` which is what I was doing to be able to compile assets without a database.
I need to find a workaround.

icarito:
There is a workaround for this detailed [here](https://iprog.com/posting/2013/07/errors-when-precompiling-assets-in-rails-4-0):

```
passing a dummy database config while precompiling:

bundle exec rake RAILS_ENV=staging DATABASE_URL=mysql2://none:none@127.0.0.1/none assets:precompile
```

I think we need https://github.com/publiclab/plots2/issues/5422 in order to have a workaround for this. So this is stuck for the moment.

PR #4886, opened 2019-02-27
Refactoring User model

siaw23:


cesswairimu:
Hi @siaw23 thanks for opening the PR could you please add issue number in the description?

siaw23:
Hi @cesswairimu I'm not sure if it's worth filing an issue here. This is more like randomly contributing a tiny small cleaner code.

cesswairimu:
Aha gotcha, I think you should rename the PR to "refactoring user model" or a name that suggests its a code cleaning implementation to make the title is more descriptive. The changes look good but I will leave this for another reviewer to take a look not sure if we accept PRs with no related issues. If you are interested in more refactoring we can created an issue with a checklist of places that need refactoring. Thanks

app/models/user.rb
```diff
@@ -12,15 +12,14 @@ class User < ActiveRecord::Base
   self.table_name = 'rusers'
   alias_attribute :name, :username
 
-  NORMAL = 1 # Usage: User::NORMAL
-  BANNED = 0 # Usage: User::BANNED
-  MODERATED = 5 # Usage: User::MODERATED
```
jywarren: Here i'd like to preserve these to show people how to use these states, if that's OK? Thanks!!

app/models/user.rb
```diff
@@ -50,8 +49,8 @@ class User < ActiveRecord::Base
 
   before_save :set_token
 
-  scope :past_week, -> { where("created_at > ?", Time.now - 7.days) }
-  scope :past_month, -> { where("created_at > ?", Time.now - 1.months) }
+  scope :past_week, -> { where("created_at > ?", 7.days.ago) }
```
jywarren: ha awesome

app/models/user.rb
```diff
@@ -78,9 +77,7 @@ def nodes
   end
 
   def notes
-    Node.where(uid: uid)
-      .where(type: 'note')
-      .order('created DESC')
+    Node.where(uid: uid).where(type: 'note').order('created DESC')
```
jywarren: Actually i prefer the multi-line version because sometimes we change only one element and then it shows up more clearly in the blame -- would you mind changing back? Thank you so much @siaw23 !!!

app/models/user.rb
```diff
@@ -92,12 +89,10 @@ def coauthored_notes
   end
 
   def generate_reset_key
-    # invent a key and save it
-    key = ''
-    20.times do
-      key += [*'a'..'z'].sample
-    end
+    key = [*'a'..'z'].sample(20).join
```
jywarren: haha so nice and compact!!!

siaw23:
Looks like changing the mirror fixes test failures on Travis. @jywarren

jywarren:
Thanks!!!

PR #4890, opened 2019-02-27
Remove flot js library

cesswairimu:
#4884 
Removing flot plotting library and dependent pages since its not being used anymore.


jywarren:
Whoa, really great! 

Was `app/views/tag/contributors.html.erb ` totally unused??? 😂

cesswairimu:
@jywarren I couldn't find another place that it was being uses expect for the '/contributors' page...But I have restored it back just in case

jywarren:
OK - well, maybe we should mark it for deletion anyways... we can do it in
another PR though. But yeah, i think it was the result of some confusion
and actually may not be used anymore!

cesswairimu:
okay cool

jywarren:
🎉

PR #4918, opened 2019-03-05
Clean up

siaw23:


app/views/wiki/_header.html.erb
```diff
@@ -103,7 +103,7 @@
 
       <li class="<%= 'active' if (controller.action_name == 'comments' )%>">
         <a href="/wiki/<%= @node.slug_from_path %>/comments" rel="tooltip" title="Read Comments regarding this wiki page" data-placement="top">
-          <span class="hidden-xs"> <%= @node.comments.count %></span>
+          <span class="hidden-xs"> <%= @node.comments.size %></span>
```
jywarren: Does size still generate a SQL count, or does it can full records and then count them? If the former them this PR is ready to merge!

siaw23: `size` here will use values from `counter_cache` value on `Comment` so it doesn't hit the DB needlessly, this way we have a faster response for the `@node.comments` count.

jywarren: Ok perfect!!

jywarren:
Thank you @siaw23!!!

PR #4933, opened 2019-03-06
Tiny fix

siaw23:


app/models/comment.rb
```diff
@@ -379,62 +379,61 @@ def self.receive_tweet_without_using_since
...
-              replied_tweet_text = replied_tweet_text.gsub('@','')
-              comment = node.add_comment(uid: user.uid, body: replied_tweet_text, comment_via: 2, tweet_id: tweet.id)
-              comment.notify user
-            end
-          end
-        end
+      next unless tweet.reply?
+
+      in_reply_to_tweet_id = tweet.in_reply_to_tweet_id
+      next unless in_reply_to_tweet_id.class == Integer
```
jywarren: Oh did you mean these to be part of this?

siaw23: Yes, `Fixnum` is deprecated.

jywarren: no i meant the whole file refactor - it's fine, i just thought maybe you'd accidentally added in something from another PR!

Issue #4891, opened 2019-02-28
Unable to add name in GSoC's mentors list


IgorWilbert:
Hi all! I am having a bit of trouble while trying to add my name on the list of potential mentors for GSoC, as can be seen below. The same message shows up whether I am logged in or not. If it happens to be relevant, I don't have anything related to plots2 installed in my computer, just JavaScript projects.
![gif](https://user-images.githubusercontent.com/42987854/53560538-5cea4400-3b2b-11e9-8901-3e8b88ee2797.gif)

gauravano:
Hey @IgorWilbert, have you started working on solving this issue or you want others to try?  Thanks for opening the issue.

IgorWilbert:
Hi @gauravano ! I would like others to try it, since currently I haven't plots2 installed. Sorry for not saying it right away.

gauravano:
No issue :+1:

jywarren:
Oh, i see a "lock" icon on the wiki page - i believe that due to wiki spam, we started blocking wiki edits until people have made... i can't remember if it's one comment or one post. So that is now blocking the inline input prompts! Great catch. 

We could... exempt the inline prompts from that restriction, and that's probably a good idea - -the route is https://publiclab.org/wiki/replace/<node_id> and it's the wiki_controller "replace" action.

However, it looks like we may have another issue - i'm getting a 404 for that even as user "warren" -- maybe something's broken? 

The test doesn't look like it's been touched for 9 months: https://github.com/publiclab/plots2/blame/master/test/functional/wiki_controller_test.rb#L441-L449

The request as sent from the browser was:

```
before: [prompt:text:Your name here]
after: * Liz Barry (@liz)

[prompt:text:Your name here]
```

Not sure why that didn't work!

jywarren:
So, in summary, 2 potential overlapping issues - the worst kind of bug!

jywarren:
Actually i think the 'replace' action is already exempted from the first-time-poster limitation, because that's only enforced in the `update` action: https://github.com/publiclab/plots2/blob/master/app/controllers/wiki_controller.rb#L192

jywarren:
Can someone test out:

1. adding the line `[prompt:text:Your name here]` in the middle of some wiki page
2. trying to enter something in the input and press `Add`
3. copy in the error we find to this issue?

jywarren:
Huh, just noting that I tried this and it worked without an error... will try again but with a copy of the `gsoc-ideas` page!

[...]

PR #4934, opened 2019-03-07
Test fix

siaw23:
Test fix for https://github.com/publiclab/plots2/issues/4891

### DIFF
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -1,7 +1,7 @@
 require 'rss'
 
 class WikiController < ApplicationController
-  before_action :require_user, only: %i(new create edit update delete replace)
+  before_action :require_user, only: %i(new create edit update delete replace show)
 
   def subdomain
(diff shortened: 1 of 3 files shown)


siaw23:
Hi @jywarren can we test this on unstable? I have deliberately commented out tests if it works then at least we know where to look next to have this fixed for good.

siaw23:
So the idea is to merge this to unstable, and check if we can add mentors without it asking us to login.

jywarren:
Yes, you should be able to `git push -f git@github.com:publiclab/plots2.git unstable` !

siaw23:
That's not working. Is that correct?

jywarren:
i have to head home now, but others in the chatroom use this a lot and can
help! that works for me, but I wonder if you are set up with the push
privileges?

siaw23:
I guess not. I can't push to it.

jywarren:
can you try now, i added you!

siaw23:
No luck.

PR #4944, opened 2019-03-07
sentry

### LINKED ISSUE #4903: Tracking errors in production
As far as I can see we don't have any means to track errors on production. Some users will report errors here, most won't but we can automate these reports. I think error tracking is necessary. What do you guys think?

I'd like to suggest [Rollbar](https://rollbar.com/). They have a free plan that we can use.

So from time to time we'd look in Rollbar and fix any error that occurs in production.

jywarren:
Fixes #4903

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -599,4 +599,4 @@ RUBY VERSION
    ruby 2.4.4p296
 
 BUNDLED WITH
-   2.0.1
+   1.17.3
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -149,8 +149,6 @@ GEM
       railties (>= 3.1)
     hashie (3.6.0)
     highline (2.0.0)
-    hoe (3.17.0)
-      rake (>= 0.8, < 13.0)
     http (3.3.0)
       addressable (~> 2.3)
       http-cookie (~> 1.0)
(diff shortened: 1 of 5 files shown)


siaw23:
Sweet. Good start! 

I'm wondering why you're bundling with Bundler 2 though 🤔

Gemfile.lock
```diff
@@ -598,4 +599,4 @@ RUBY VERSION
    ruby 2.4.4p296
 
 BUNDLED WITH
-   1.17.3
+   2.0.1
```
jywarren: i dunno i must've updated recently. is it a problem?

jywarren: ugh, `You must use Bundler 2 or greater with this lockfile.`

Will try again downgrading bundler...

siaw23:
Looks good to me. If they approve maybe you could put a link of where we can track the errors? Maybe in the README or wherever you find appropriate?

jywarren:
Hmm, its coming through (can't invite yet, but will once we're approved) but there's a sprockets error: 

https://sentry.io/share/issue/9b54ec2922c0426683b689eea965ce87/ @siaw23 @icarito

jywarren:
ok we are approved, inviting ppl now

siaw23:
OK please send me and invite too. I'd love to look around. I think this will help us a lot in improving our application! Thanks for doing this Jeff!!

icarito:
Friends I'm loving this! I really wanted to have a window into errors _as they were happening_!

PR #4955, opened 2019-03-08
Fixing #4822 daily vs weekly subject

### LINKED ISSUE #4822: Refactoring daily vs weekly subject title in Email Digests
Okay so I am trying to do this, but I think that it is probably hard coded somewhere, and a variable for checking which type of digest is not being sent to the actions:

Like here, in the mailer code:
https://github.com/publiclab/plots2/blob/d3b122efaa42a9d6b06a70f53626334ad2cf085b/app/mailers/subscription_mailer.rb#L58

Here in the send_digest_email function, top picks are being selected for a week only:
https://github.com/publiclab/plots2/blob/d3b122efaa42a9d6b06a70f53626334ad2cf085b/app/models/user.rb#L325-L330

I think here, when send_digest_email function is being called, the frequency parameter should be passed somehow, and received everywhere, then we can correct the code in the send_digest.html.erb:

https://github.com/publiclab/plots2/blob/d3b122efaa42a9d6b06a70f53626334ad2cf085b/app/jobs/digest_mail_job.rb#L4-L14

IshaGupta18:
Fixes #4822 

So it works locally now on changing the value of k(parameter for daily v/s weekly):

weekly:
image

daily:

image

### DIFF
--- a/app/mailers/subscription_mailer.rb
+++ b/app/mailers/subscription_mailer.rb
@@ -61,7 +61,7 @@ class SubscriptionMailer < ActionMailer::Base
       @subject = "Your daily digest"
     end
     @user = User.find(user_id)
-    @top_picks = nodes
+    @nodes = nodes
     mail(to: @user.email, subject: @subject)
   end
 end
--- a/app/mailers/subscription_mailer.rb
+++ b/app/mailers/subscription_mailer.rb
@@ -54,10 +54,10 @@ class SubscriptionMailer < ActionMailer::Base
     )
    end
 
-  def send_digest(user_id, top_picks,k)
-    if k==0
+  def send_digest(user_id, top_picks, k)
+    if k == 0
       @subject = "Your weekly digest"
-    elsif k==1
+    elsif k == 1
       @subject = "Your daily digest"
     end
     @user = User.find(user_id)
--- a/app/mailers/subscription_mailer.rb
+++ b/app/mailers/subscription_mailer.rb
@@ -54,14 +54,14 @@ class SubscriptionMailer < ActionMailer::Base
     )
    end
 
-  def send_digest(user_id, top_picks, k)
-    if k == 0
+  def send_digest(user_id, nodes, freq)
+    if freq == 0
       @subject = "Your weekly digest"
-    elsif k == 1
+    elsif freq == 1
       @subject = "Your daily digest"
     end
     @user = User.find(user_id)
-    @top_picks = top_picks
+    @top_picks = nodes
     mail(to: @user.email, subject: @subject)
   end
 end
--- a/app/mailers/subscription_mailer.rb
+++ b/app/mailers/subscription_mailer.rb
@@ -54,10 +54,14 @@ class SubscriptionMailer < ActionMailer::Base
     )
    end
 
-  def send_digest(user_id, top_picks)
-    subject = "Your weekly digest"
+  def send_digest(user_id, top_picks,k)
+    if k==0
+      @subject = "Your weekly digest"
+    elsif k==1
+      @subject = "Your daily digest"
+    end
     @user = User.find(user_id)
     @top_picks = top_picks
-    mail(to: @user.email, subject: subject)
+    mail(to: @user.email, subject: @subject)
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -316,17 +316,17 @@ class User < ActiveRecord::Base
   end
 
   def send_digest_email
-    top_picks = []
+    nodes = []
     freq = 1
     if self.has_tag('digest:daily')
-      top_picks = content_followed_in_period(1.day.ago)
+      nodes = content_followed_in_period(1.day.ago)
       freq = 0
-    elsif self.has_tag('digest:weekly')
-      top_picks = content_followed_in_period(Time.now - 1.week, Time.now)
+    else
+      nodes = content_followed_in_period(Time.now - 1.week, Time.now)
       freq = 1
     end
-    if !top_picks.nil? && top_picks.count > 0
-      SubscriptionMailer.send_digest(id, top_picks, freq).deliver_now
+    if nodes.count > 0
+      SubscriptionMailer.send_digest(id, nodes, freq).deliver_now
     end
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -316,8 +316,8 @@ class User < ActiveRecord::Base
   end
 
   def send_digest_email
-    top_picks=[]
-    freq=1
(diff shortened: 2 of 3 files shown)


IshaGupta18:
Hey @jywarren, I think the code climate issues are very trivial and can be ignored. However, I fail to understand the travis build fail. Could you please explain what happened and what can be done to fix it?

app/mailers/subscription_mailer.rb
```diff
@@ -54,10 +54,14 @@ def notify_tag_added(node, tag, tagging_user)
     )
    end
 
-  def send_digest(user_id, top_picks)
-    subject = "Your weekly digest"
+  def send_digest(user_id, top_picks, k)
```
jywarren: Can we give this parameter a more informative name, so people understand when reading the code? Thanks!

jywarren: Also, maybe top picks could just become "nodes", you know?

app/models/user.rb
```diff
@@ -316,9 +316,15 @@ def liked_pages
   end
 
   def send_digest_email
-    top_picks = content_followed_in_period(1.week.ago, Time.now)
-    if top_picks.count.positive?
-      SubscriptionMailer.send_digest(id, top_picks).deliver_now
+    if self.has_tag('digest:daily')
```
jywarren: nice!

app/models/user.rb
```diff
@@ -316,9 +316,15 @@ def liked_pages
...
+      top_picks = content_followed_in_period(Time.now - 1.day, Time.now)
```
jywarren: Not sure, but I think `@siaw23` was able to use `1.day.ago` instead of subtracting... maybe worth a try cause it's short?

app/models/user.rb
```diff
@@ -316,9 +316,15 @@ def liked_pages
...
+      k = 0
+    elsif self.has_tag('digest:weekly')
+      top_picks = content_followed_in_period(Time.now - 1.week, Time.now)
+      k = 1
+    end
+    if top_picks.count > 0
```
jywarren: This is where the error points to. 

```

ERROR["test_send_digest_email", #<Minitest::Reporters::Suite:0x00005572aa9d18a0 @name="UserTest">, 24.959422129000018]
 test_send_digest_email#UserTest (24.96s)
NoMethodError:         NoMethodError: undefined method `count' for nil:NilClass
        Did you mean?  concat
            app/models/user.rb:326:in `send_digest_email'
            test/unit/user_test.rb:225:in `block in <class:UserTest>'
```

app/mailers/subscription_mailer.rb
IshaGupta18: Yes absolutely, and I am gonna rename top_picks parameter to nodes then?

app/models/user.rb
IshaGupta18: Oh yes, so probably because the daily top_picks were empty, that's why it couldn't trigger the condition, should I put a nil check there?

app/models/user.rb
```diff
@@ -316,9 +316,15 @@ def liked_pages
   end
 
   def send_digest_email
-    top_picks = content_followed_in_period(1.week.ago, Time.now)
-    if top_picks.count.positive?
-      SubscriptionMailer.send_digest(id, top_picks).deliver_now
+    if self.has_tag('digest:daily')
+      top_picks = content_followed_in_period(Time.now - 1.day, Time.now)
+      k = 0
+    elsif self.has_tag('digest:weekly')
```
jywarren: Aha - so if a user has /neither/ tag set, we are still trying to call top_picks. We should set it to `[]` by default? We could define it above as `top_picks = []` before these IFs?

IshaGupta18: Yes I thought of doing the same, but I didn't. So sorry!

app/models/user.rb
IshaGupta18: So I tried to a check like `!top_picks.nil? && top_picks.count > 0` ,  however another error showed up in travis.

jywarren: Sure or as i said above, you can start by defining it as `[]`. Great!

IshaGupta18: I have tried both, but still, the new error prevails, could you please tell me why?

jywarren:
Hmm, i see in Travis:

```
 FAIL["test_send_digest_email", #<Minitest::Reporters::Suite:0x00005629dcddcc30 @name="UserTest">, 41.071117536999964]
 test_send_digest_email#UserTest (41.07s)
        Expected nil to be truthy.
        test/unit/user_test.rb:225:in `block in <class:UserTest>'
```

So do you see how to look up that line where the error occurred? See how it points here: https://github.com/publiclab/plots2/blob/master/test/unit/user_test.rb#L225

Are we sure that `send_digest_email` still returns something that's not nil?

app/models/user.rb
```diff
@@ -316,9 +316,17 @@ def liked_pages
...
+    freq = 1
+    if self.has_tag('digest:daily')
+      top_picks = content_followed_in_period(1.day.ago)
+      freq = 0
+    elsif self.has_tag('digest:weekly')
+      top_picks = content_followed_in_period(Time.now - 1.week, Time.now)
+      freq = 1
+    end
+    if !top_picks.nil? && top_picks.count > 0
+      SubscriptionMailer.send_digest(id, top_picks, freq).deliver_now
```
jywarren: aha, so could it be that the test was set up such that the test user doesn't actually have these tags, and so is now not getting the digest email? 

Also, if you wouldn't mind changing `top_picks` to `nodes` or `subscribed_content` perhaps, that'd be nice, thanks!

app/models/user.rb
jywarren: no worries!

jywarren:
Isha this is going really well. You're getting better and better at tracing errors through the logs, figuring out what they mean and what caused them, and identifying the spot in the code where you need to make a change. Travis tests can really help guide your debugging, so it's great to build that skill set!

app/models/user.rb
IshaGupta18: Oh it could be that! Yeah sure, I will change it here as well, as I did in the mailer!

IshaGupta18:
Thanks a lot! Yes, I think that's very important too. I do know how to look up for the errors and find the files that caused them. Let me try another hand at debugging this one!

IshaGupta18:
@jywarren so the checks have passed, I think I caught the error somewhere in user.rb :tada: . Kindly review and let me know if its ready to be merged! Thanks a lot!

jywarren:
Fantastic work, @IshaGupta18 !!!

PR #4967, opened 2019-03-09
make mail delivery asynchronous - attempt, deliver_later

### LINKED ISSUE #4782: When mails fail to send, queue them for later.
Mail systems can sometimes have interruptions (e.g. we have a throttle from google for example).

Could we queue up failed outgoing emails?

There are subtle bugs like https://github.com/publiclab/plots2/issues/4774#issuecomment-461646291 that are confusing and if we had a proper mail queue cache we would avoid them.

One non-coding way to do this is to implement an MTA container and configure it for such retry queue.
Or it can be done at the application level. MTAs are designed and for this task, but on the other hand this feature would only be pre-configured when running with Docker if we went the container path.

jywarren:
Fixes #4782 

This should be tested out carefully! @icarito


jywarren:
I think we need to figure out how to adjust these tests for asynchronous mail delivery!

jywarren:
This shows an error when checking deliveries:

https://github.com/publiclab/plots2/blob/29ffa07484f5eaacf5653478a14e8d1fd42880a0/test/functional/admin_controller_test.rb#L171

Could we trigger all the asynchronous mails before checking this or alternatively, check the queue instead of deliveries?

jywarren:
Yes!!!

`expect(enqueued_jobs.size).to eq(1)`

And looks like there's this although not sure if it's only rspec:

```
perform_enqueued_jobs do
      example.run
    end
```

jywarren:
Looks like a general test helper!
https://apidock.com/rails/v4.2.1/ActiveJob/TestHelper/perform_enqueued_jobs

jywarren:
Hmm, this isn't working... https://api.rubyonrails.org/classes/ActiveJob/TestHelper.html#method-i-perform_enqueued_jobs

jywarren:
Working from https://stackoverflow.com/questions/27647749/how-to-test-actionmailer-deliver-later-with-rspec and https://stackoverflow.com/questions/39677612/deliver-later-not-working-in-the-test-environment-in-rails-5

jywarren:
If this doesn't work, we might wrap the whole thing in:

```ruby
perform_enqueued_jobs do
  # do everything
```

as in https://stackoverflow.com/a/46555068

jywarren:
:-) @icarito this seems to have worked!!! We could extend this refactoring to all controllers. Anything else we want to do to test it out?

icarito:
Awesome! I did look at it but couldn't figure out the tests part.

I just double checked how we setup mail from staging: It's all siphoned
into the `testing` user account which is accessible from Tycho server
with a shell account and a mail user agent. One can use Mutt to read the
staging catch-all email : `(tycho)$ sudo -u testing mutt`.

Then we'll know if it sends!

jywarren:
I'll "spam" something on stable as soon as it builds and we can check there too.

jywarren:
http://stable.publiclab.org/people

jywarren:
OK - i "banned" https://stable.publiclab.org/profile/bkleist, but not sure that sends an email...

OK, now I spammed https://stable.publiclab.org/n/14087 -- by bkleist, it loaded fine, so we should look for the email now. @icarito would you be able to check it? Thanks!

jywarren:
@icarito we could move to a new issue if it helps us track!

PR #5011, opened 2019-03-11
lower search per type limit from 10 to 5

jywarren:


CleverFool77:
I'm curious, @jywarren. What is this PR about. :thought_balloon:

jywarren:
:-) I noticed in the navbar search, we don't display all the results, just
like 10 or so. So if there are too many username results, like 10, we see
only usernames. That's not good! The rest get cut off. So, short-term, i'm
reducing the # per type of search to 5, so we see multiple types. If you're
interested in this, i bet it could use a more sophisticated fix too!

CleverFool77:
Thanks for your answer @jywarren. I was just curious about the change. I'll try to do a better fix for it.

CleverFool77:
Why is this PR taking 3 days for checks. :thinking: 
I'm really curious. :sweat_smile:

jywarren:
Me too! I don't know; i'm going to try to restart it by making a small commit.

PR #5040, opened 2019-03-12
Check that node is not nil

siaw23:
Fixes [Sentry issue](https://sentry.io/organizations/publiclab/issues/924962616/?project=1410626&query=is%3Aunresolved&statsPeriod=14d&utc=false)

### DIFF
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -55,7 +55,7 @@ class QuestionsController < ApplicationController
       @node = Node.find params[:id]
     end
 
-    redirect_to @node.path unless @node.has_power_tag('question')
+    redirect_to @node.path unless @node&.has_power_tag('question')
 
     alert_and_redirect_moderated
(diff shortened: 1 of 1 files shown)


jywarren:
Awesome! Should we think of writing tests against Sentry issues? Gosh, i wonder if there's even a way for Sentry to write a test for us... lol i mean technically it probably has enough information to, right?

siaw23:
It would be nice if Sentry could do that. I haven't heard of that one before. What's surprised me so far is though Coveralls claims a high coverage it appears a lot of things haven't be tested. My guess would be there's not enough unit, functional tests. Right now I can't look at all these at once but would love to  inspect the tests some time soon.

jywarren:
yeah, i think we have good coverage but not necessarily thorough parameter
coverage, like we test in only the expected ways but there are lots of ways
you could pass in parameters in unexpected ways. Thanks!

PR #5075, opened 2019-03-14
Remove inexistent method call on User model

siaw23:
Fixes https://sentry.io/share/issue/5efcb5d68d5a4f7cbcfe30d83c06ba8a/


jywarren:
Thanks, both!!!

PR #5116, opened 2019-03-16
Corrected Signup form validation

### LINKED ISSUE #3340: Sign-up Form Validation message correction

ananyaarun:
Fixes #3340 

signup1

Thanks!


ananyaarun:
@gauravano , Form doesn't accept null values anymore during signup. Also i have made only username , email. password and repeat password compulsory. Since image and Bio is up to user if they want to add or not.

ananyaarun:
@gauravano , The thing as of now I have used the same method of form validation in Both login and signup. But signup uses a rotating wheel fontawesome icon. ie upon one click the icon appears and signup option disappears as you said. Upon removing this (like how it isnt there for login) this is how it looks.

SIGNUP

So now even if there is an error while filling there is a prompt to refill but the icon doesnt appear.
Thoughts ?

gauravano:
Can't we show that progress icon only when JavaScript validation is passed and in other cases we can just focus on the fields?

ananyaarun:
@gauravano , Hmm I just looked into this. So the method of using `required: true` is basically form validation in ruby. Whereas this button is a JS element which always executes whenever the signup button is clicked. 
I'm unsure of how to integrate the ruby validation along with JS spinning button script. I tried adding if conditions but the ruby form elements arent accessible outside the form template where the script is defined.
A solution would be to implement this just like login which doesnt have a spinning button at the moment. 
I'll look more into this though and get back to you if something works ?

ananyaarun:
@gauravano , So far i can get 2 solutions to this issue.

- The signup from works and ensures it gets submitted only when all fields are filled. But the spinning wheel never appears.(Just like how login is implemented presently)
- When i try to integrate the spinning wheel icon with ruby form validation, it appears on click no matter which fields are unfilled/filled.

Any thoughts on what can be done ?

gauravano:
How about this - "User fills the form partially and goes on to submit the form by pressing sign-up and we can show a warning on hovering over the sign-up button only or maybe disable it too?"

ananyaarun:
So basically there are 4 compulsory fields a user needs to fill. The user may not fill this in any permutation or combination.
Right now what happens according to the changes i made is that whichever field isnt filled a "pls fill this " msg appears below it when the user clicks signup. So in a way it is disabled until all fields are filled by the user.

gauravano:
But, user need to refresh the screen right?

ananyaarun:
No not any more, I made a slight change. I'll attach a gif in a minute.

ananyaarun:
@gauravano  Here is the gif, User need not refresh after each "Pls fill this field" msg. Basically until all fields are filled this msg comes and form is submitted on one click after that.

signup2

app/views/users/_create_form.html.erb
```diff
@@ -23,13 +23,17 @@
       <div class="row col-sm-6">
      
         <div class="form-group col-sm-12" id="username_div">
+          
           <label for="username"><%= t('user_sessions.new.username') %></label>
-          <%= f.text_field :username, { tabindex: 1, placeholder: "Username", class: 'form-control', id: 'username-signup' } %>
+          <%= f.text_field :username, { tabindex: 1, placeholder: "Username", class: 'form-control', id: 'username-signup', required:true } %>
+
```
gauravano:
```suggestion
```
Let's remove this empty space

app/views/users/_create_form.html.erb
```diff
@@ -23,13 +23,17 @@
...
+          <%= f.text_field :username, { tabindex: 1, placeholder: "Username", class: 'form-control', id: 'username-signup', required:true } %>
+        
         </div>
      
         <div class="form-group col-sm-12">
+          
           <label for="email"><%= t('users._form.email') %></label>
-          <%= f.text_field :email, { tabindex: 3, placeholder: "you@email.com", class: 'form-control', id: 'email' } %>
+          <%= f.text_field :email, { tabindex: 3, placeholder: "you@email.com", class: 'form-control', id: 'email', required: true } %>
+
```
gauravano:
```suggestion
```

app/views/users/_create_form.html.erb
```diff
@@ -106,21 +116,15 @@
...
         </a>
 
-      <script>
-        (function() {
-          $('.btn-save').click(function onClick(e) {
-            $(this).addClass("disabled") // disable the button after it is clicked
-                   .html("<i class='fa fa-spinner fa-spin'></i>"); // make a spinner that spins when clicked
-          })
-        })();
-      </script>
```
gauravano: Instead of removing this,  we can disable the button only when all the required fields are filled i.e., spinner will show only when all required fields are filled otherwise focus on the fields?

ananyaarun: Yes that would be ideal. Do you have any idea how this can be done ?. Because the problem is since button is a JS element and form validation happens in ruby, i wasn't able to find a way to check if all fields are filled before displaying the button. The fields i try to access show "Cant be accessed error on console". I'll keep trying other ways to do this in the meantime maybe i was wrong, Thanks @gauravano  :)

gauravano: The form/input fields we create in rails also result in the generation of HTML form only and so can be assessed just like that, open the sign-up form of publiclab.org and try `$('#username-signup').val()` in your console and you'll get the value of username field.

ananyaarun: Thanks, I'll try this :)

ananyaarun:
@gauravano , I got an idea how to implement this keeping the spinner button, Sorry have been a little busy with assignments, exams and writing my proposal lately. Will work on this as soon as i am done with these.

gauravano:
No issue! And, thanks for working on it.

SidharthBansal:
There are some conflicts in your pr. It will be great if you can rebase your pr.
Thanks for the work done here at PL

gauravano:
Hi @ananyaarun, please update this PR too whenever you get time. Thanks!

ananyaarun:
Sure, I have resolved the conflicts.
Still trying ways to keep the spinner. As of now validation works but I have removed the font awesome spinner icon.

jywarren:
OK, just because this being signup, it's a very critical system, would you mind sharing a gif of the current latest functionality just to confirm, @ananyaarun ? Thank you!!!!

ananyaarun:
@jywarren here
Basically won't allow a user signup without entering username and password.
Also other fields are optional (ie not 'required') as of now, but not entering those gives a spam detection msg as expected.
signup1 1

jywarren:
Ok, would you mind testing this on unstable to be sure the recaptcha
version of anti spam system works? I believe there's a PR open with new
documentation on unstable server usage?

Thanks!!!

ananyaarun:
@gauravano @SidharthBansal  Can you test this on unstable ? Let me know if more changes are required.
Thanks!

ananyaarun:
@gauravano @jywarren I dont have permission to push to unstable to test.
Can someone who has permission test and let me know if there are further changes to be made ?
Thanks !!

SidharthBansal:
I agree!
For authentication issues please add my name also.
Please feel free to close this issue and open a new issue as suggested
above.

ananyaarun:
Sure at @sashadev-sky !
Closing this issue would make sense. This has been open for quite some time, We don't want to confuse newcomers / GCI students with stale/ invalid issues.

Will u be opening another issue for this ?

cesswairimu:
Hey @ananyaarun , seems the issue you were fixed in this PR was closed..Closing this. Thanks

PR #5120, opened 2019-03-17
Create visual difference between notes/wiki

### LINKED ISSUE #152: increase visual/design difference between wikis and research notes
Consider contrasting the difference between a wiki and a research note.  Currently the way I tell them apart mainly is by looking at the URL.  I would suggest something like changing the background color or pattern to differentiate between the two.

CleverFool77:
Fixes #152 First part
## Description 
The design change is not major rather a minor one.
This is the first part of the changes towards the difference in between both pages. 
The image section was taken out of div of col-md-9 which had the data. So that instead of being inside grid column, the image would be outside it.
And the image width was set to 100%, so that it would take whole screen-with.

## Earlier 
Screenshot from 2019-03-17 11-33-41 (1)
Screenshot from 2019-03-17 11-52-35

## After changes
Screenshot from 2019-03-17 11-33-41 (1)
Screenshot from 2019-03-17 11-33-03

Thanks!

### DIFF
--- a/app/views/wiki/_header.html.erb
+++ b/app/views/wiki/_header.html.erb
@@ -2,12 +2,6 @@
   <div class="alert alert-warning"><%= raw t('wiki.show.page_has_no_tags', url: 'javascript: $(".tag-input").focus();') %></div>
 <% end %>
 
-<% if @node.main_image && !@presentation %>
-  <a style="margin-bottom:10px;" href="<%= @node.main_image.path(:original) %>">
-    <img style="max-width:100%;max-height:600px;margin-bottom:10px;" class="img-rounded" src="<%= @node.main_image.path(:large) %>" />
-  </a>
-<% end %>
-
 <div style="margin-top:10px;" class="hidden-print">
   <%= render partial: "like/like", locals: {node: @node, tagnames: @tagnames } %>
 </div>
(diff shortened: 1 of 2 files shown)


CleverFool77:
@jywarren @gauravano should I divide this issue too into parts like other issues?

gauravano:
Isn't both after changes images for notes only?

CleverFool77:
It shows note but it was wiki, My bad.. @gauravano I'll change it

CleverFool77:
@gauravano Thanks for mentioning that out. I've changed the image.

gauravano:
Ok, so we would be having same width for note but full screen main image for wiki. right?

CleverFool77:
Yes, because I thought wiki is like a blog which can be edited by everyone. We can start off like this and then do further changes. Maybe we can have image as absolute in background. And the data below relative and half of it would be above the image. :thinking:

jywarren:
Hi! This is looking good; do you think it could be an issue though on pages with big images? Consider for example https://publiclab.org/coqui -- perhaps this would push all content down too far? Do you want to try it to see how it works there?

CleverFool77:
Hi @jywarren . Should I fix the height of image or crop it accordingly ? like Facebook does for its cover photos. It won't make the content pushed down far below then. :thinking:

SidharthBansal:
There are some conflicts in your pr. It will be great if you can rebase your pr.
Thanks for the work done here at PL

gauravano:
Hi @CleverFool77, please resolve the merge conflicts here. Thanks!

PR #5134, opened 2019-03-17
Added unit test to find email using twitter username

### LINKED ISSUE #5020: Write unit tests for parts of Reply By Tweet feature
There's a new Reply by Tweet feature, but it needs some unit tests! @namangupta01 maybe you could help people get started on this? 

https://github.com/publiclab/plots2/blob/f4d965c94de289ef3244bc7eb3d555ef4edcfbe0/app/models/comment.rb#L347-L441

Any of these methods could use a unit test, in this file: https://github.com/publiclab/plots2/blob/master/test/unit/comment_test.rb

Here's an example of a unit test! https://github.com/publiclab/plots2/pull/4571/files

We'd love help with this!

divyabaid16:
Fixes #5020 

Thanks!

### DIFF
--- a/test/unit/comment_test.rb
+++ b/test/unit/comment_test.rb
@@ -308,4 +308,12 @@ class CommentTest < ActiveSupport::TestCase
     assert_not_nil graph
     assert graph.class, Hash
   end
+
+  test 'find email using twitter user name' do
+    require 'yaml'
+    config = YAML.load(File.read('test/fixtures/user_tags.yml'))
+    username = config["twitter3"]["data"]["info"]["nickname"]
+    email = Comment.find_email(username)
+    assert_equal email, "01namangupta@gmail.com"
+  end
 end


divyabaid16:
@namangupta01 can you please review this?

test/unit/comment_test.rb
```diff
@@ -308,4 +308,12 @@ class CommentTest < ActiveSupport::TestCase
     assert_not_nil graph
     assert graph.class, Hash
   end
+
+  test 'find email using twitter user name' do
+    require 'yaml'
```
siaw23: I don't think you need this line...

test/unit/comment_test.rb
```diff
@@ -308,4 +308,12 @@ class CommentTest < ActiveSupport::TestCase
...
+    config = YAML.load(File.read('test/fixtures/user_tags.yml'))
+    username = config["twitter3"]["data"]["info"]["nickname"]
```
siaw23: Also why don't you just use the string here directly? I think it's unnecessary to load the file and look for keys. You're only testing `find_email` so the lines below should suffice. When you replace `username` with the value you're you're looking for.

test/unit/comment_test.rb
divyabaid16: @siaw23 Actually I looked at few other tests in the file like
```
  test 'should have gmail quote' do
    require 'mail'
    require 'nokogiri'
    mail = Mail.read('test/fixtures/incoming_test_emails/gmail/incoming_gmail_email.eml') 
    mail_doc = Nokogiri::HTML(mail.html_part.body.decoded) # To parse the mail to extract comment content and reply content 
    gmail_quote = Comment.gmail_quote_present?(mail_doc)
    assert_equal gmail_quote, true
  end
```
So taking inspiration from these, I had used this way.
So do I really need to hardcode the username?

divyabaid16:
@siaw23 Actually I looked at few other tests in the file like
```
  test 'should have gmail quote' do
    require 'mail'
    require 'nokogiri'
    mail = Mail.read('test/fixtures/incoming_test_emails/gmail/incoming_gmail_email.eml') 
    mail_doc = Nokogiri::HTML(mail.html_part.body.decoded) # To parse the mail to extract comment content and reply content 
    gmail_quote = Comment.gmail_quote_present?(mail_doc)
    assert_equal gmail_quote, true
  end
```
So taking inspiration from these, I had used this way.
So do I really need to hardcode the username?

siaw23:
Now that I'm looking at those other tests I think it'll be too much work to make those your responsibility unless you want to... but for this PR trying remove the `require` and see if it works out. You don't want to require items you don't need...

divyabaid16:
When I delete `require` line, it is showing error in my terminal because the YAML file is loaded using that way.

divyabaid16:
@siaw23 finally figured out the way to do it. 
Thank you so much!

jywarren:
Thanks so much!!!

PR #5141, opened 2019-03-17
Fixed loading a tags page when there are no "questions" in DB

### LINKED ISSUE #5140: Missing questions in dev seeds break tags notes page
The seeds for local development do not contain any nodes which are attached to a "question" tag (`Node.questions == []`).
This has a side-effect of breaking a page such as http://localhost:3000/tag/five, which one would expect to return the "Blog Post" node, but instead nothing is returned.
The reason is that in `tag_controller` [the list of questions is fetched](https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L97), then [used to filter](https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L135) the list of nodes to only fetch notes. Since `qids == []`, this adds `WHERE node.nid NOT IN (NULL)` to the query, resulting in an empty set.

The solution IMO should be either to add at least one question to the local dev seeds, or avoid adding that query condition. I lean towards the latter, as I find it cleaner, since it doesn't assume the existence of specific data in the system.

alonpeer:
Fixes #5140

Thanks!

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -94,8 +94,6 @@ class TagController < ApplicationController
                   'contributor'
                 end
 
-    qids = Node.questions.where(status: 1).collect(&:nid)
-
     if params[:id][-1..-1] == '*' # wildcard tags
       @wildcard = true
       @tags = Tag.where('name LIKE (?)', params[:id][0..-2] + '%')
@@ -132,8 +130,15 @@ class TagController < ApplicationController
     end
     nodes = nodes.where(created: @start.to_i..@end.to_i) if @start && @end
 
-    @notes = nodes.where('node.nid NOT IN (?)', qids) if @node_type == 'note'
-    @questions = nodes.where('node.nid IN (?)', qids) if @node_type == 'questions'
+    qids = Node.questions.where(status: 1).collect(&:nid)
+    if qids.empty?
+      @notes = nodes
+      @questions = []
+    else
+      @notes = nodes.where('node.nid NOT IN (?)', qids) if @node_type == 'note'
+      @questions = nodes.where('node.nid IN (?)', qids) if @node_type == 'questions'
+    end
+
     @answered_questions = []
     @questions&.each { |question| @answered_questions << question if question.answers.any?(&:accepted) }
     @wikis = nodes if @node_type == 'wiki'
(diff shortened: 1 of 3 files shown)


jywarren:
Ah, thank you - but will this show notes instead of questions if there are no question? Could we instead, in the view template, say `there are no questions for the topic ____, but click here to see posts on the topic` or something?

Thank you!

alonpeer:
@jywarren I'm not sure I understand your question.
The bug is on the Research Notes page (e.g. http://localhost:3000/tag/five when running the app locally), and not the Questions page (e.g. http://localhost:3000/questions/tag/five).
When you visit the Notes page, if there are no questions at all in the DB, even if there ARE notes for this tag, they won't display.

Sure, I can also add that sentence you suggested to the **questions** view, but this is unrelated.

You can load the links I suggested before and after this fix, and see what I mean.

app/controllers/tag_controller.rb
```diff
@@ -132,8 +132,14 @@ def show
     end
     nodes = nodes.where(created: @start.to_i..@end.to_i) if @start && @end
 
-    @notes = nodes.where('node.nid NOT IN (?)', qids) if @node_type == 'note'
-    @questions = nodes.where('node.nid IN (?)', qids) if @node_type == 'questions'
+    if qids.empty?
```
jywarren: Hm, could we move this line so it's just above here, and a bit easier to read? https://github.com/publiclab/plots2/pull/5141/files#diff-a8eb2425ebeb0814d010864bb014e815R97

app/controllers/tag_controller.rb
```diff
@@ -132,8 +132,14 @@ def show
...
+      @notes = nodes
+      @questions = nodes
```
jywarren: Here, if there are no questions, could we just assign `[]` to `@questions`? We wouldn't want to pass in non-question nodes just because there are no questions, perhaps?

app/controllers/tag_controller.rb
alonpeer: good idea

alonpeer: ah actually, that's not possible, because `@notes` and `@questions` depend on `nodes` so these have to be set after `nodes` is set. Instead I moved the creation of `qids` down here.

app/controllers/tag_controller.rb
alonpeer: Yeah, I guess you're right 👍

alonpeer:
@jywarren please take another look! :)

jywarren:
Awesome, thank you! Merging this now. Great work! 👍

PR #5149, opened 2019-03-18
Bugfix empty pages on search results #4871 [WIP]

### LINKED ISSUE #4871: Too many page numbers on search results
When searching for something using tags, the results show more page numbers than necessary.
Most pages in the end are empty
plots

swiatek7:
Fixes #4871 

**BEFORE**
empty-pages-before
empty-pages-before

**AFTER**
empty-pages-after
empty-pages-after

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -67,8 +67,9 @@ class SearchService
     tlist = Tag.where(name: sterms)
       .joins(:node_tag, :node)
       .where('node.status = 1')
-      .select('DISTINCT node.nid,node.title,node.path')
+      .select('node.nid,node.title,node.path')
       .limit(limit)
+      .distinct
   end
 
   # Search question entries for matching text


swiatek7:
Hi @publiclab/reviewers :)

swiatek7:
@publiclab/reviewers 

I guess SearchServiceTest is failing now, because there is right number of search results...

swiatek7:
@gauravano
I need some feedback and help on this. Do you know who could provide such?

sashadev-sky:
@swiatek7 have you tried pushing to unstable? You'll be able to see there easily if your solution is correct

app/services/search_service.rb
```diff
@@ -67,8 +67,9 @@ def search_tags(query, limit = 10)
     tlist = Tag.where(name: sterms)
       .joins(:node_tag, :node)
       .where('node.status = 1')
```
sashadev-sky: just a small note to sanitize the value in the where statement

swiatek7: Thanks for feedback.

First of all this is not part of my change, it was already like this. 
If you think it should be done differently, you can propose the solution.

Second of all passing hash to where method is already safe and user input get sanitized by rails.
You can read up here: https://guides.rubyonrails.org/security.html#injection

Third of all I think node.status values is never set by user and it is not user's input.

Cheers
Mateusz

gauravano:
Would circle back here shortly.

gauravano:
I retriggered the Travis build but got this error before hitting tests :point_down: 

Screenshot from 2019-03-24 03-29-40

https://travis-ci.org/publiclab/plots2/builds/507805452

gauravano:
I pushed this PR to `unstable` after commenting the test which was throwing the error and rebasing the branch and so it passed.

The changes are live at https://unstable.publiclab.org/ and working fine.

app/services/search_service.rb
sashadev-sky: All valid points! Sorry thought that the other queries around it were sanitizing the strings passed to where too so I suggested it for consistency. But looking again that is not the case so taking this back!!

swiatek7:
> I pushed this PR to `unstable` after commenting the test which was throwing the error and rebasing the branch and so it passed.
> 
> The changes are live at https://unstable.publiclab.org/ and working fine.

What are the next steps here? Should I look more closely on the test and modify it?

gauravano:
Hey @swiatek7, yes, we need to inspect the reason of test failure and modify the test or our solution accordingly. 

@milaaraujo @stefannibrasil might help you with the tests.  Thanks!

stefannibrasil:
I'll take a look with more time tomorrow morning :+1:

stefannibrasil:
It's been a while I don't see those files. Is it possible to send a print of the output inspecting this

https://github.com/publiclab/plots2/blob/4c6ecb2325b326f451b2a5cc90f9feaedb7475a7/test/unit/api/search_service_test.rb#L42

`result` before and after this commit?

My guess is, if this change is indeed selecting distinct values, then it should return indeed only those two nodes:

https://github.com/publiclab/plots2/blob/4c6ecb2325b326f451b2a5cc90f9feaedb7475a7/test/fixtures/node_tags.yml#L7-L10

and 

https://github.com/publiclab/plots2/blob/4c6ecb2325b326f451b2a5cc90f9feaedb7475a7/test/fixtures/node_tags.yml#L13-L16

because they *have* different nids.

But I just wanted to see how it was before to make sure. I don't know if this is much of a help, but I need to refresh my memory :/ thanks!

stefannibrasil:
Also, how different are the results by placing the `distinct` after or before the `limit`?

SidharthBansal:
Say database has the following array,
arr[ ] = [1,1,2,3,2,4,4,5]
arr.limit(5).distinct() will give [1,1,2,3,2].disctinct() that is {1,2,3}
whereas arr.distict().limit(5) will give [1,2,3,4,5].limit(5) that is {1,2,3,4,5}
Does this makes sense?

stefannibrasil:
yes, it does, thanks @SidharthBansal. But I should've added a better question. My concern was more related to what approach do we want to follow here. For me, it would be better to place the distinct before limiting the results, for example.

gauravano:
@jywarren can you help here? Thanks!

Issue #5015, opened 2019-03-11
/api/srch/notes/... API query is slowing down site with 1-2 minute long responses


jywarren:
Looks related to #3147 in some ways but for the title "related notes" suggestions shown at publiclab.org/post (to test, see https://stable.publiclab.org/post, our testing site, so as not to make the issue worse)

Or, https://stable.publiclab.org/api/srch/notes?query=coqui should trigger it directly on the stable test server. Interestingly, this isn't very slow, even though `stable.publiclab.org` is much lower power machine with the same sized database.

Skylight analysis:

https://oss.skylight.io/app/applications/GZDPChmcfm1Q/recent/6h/endpoints/Srch::Search%20%5BGET%5D%20srch%2Fnotes

Oddly, it seems not to be a SQL slowdown??? 

![image](https://user-images.githubusercontent.com/24359/54155554-c109e380-441a-11e9-8713-feaaef7b2af7.png)

Could this actually be slowing down due to something else?

jywarren:
Yeah, see how it got worse and worse:

![image](https://user-images.githubusercontent.com/24359/54155810-4c837480-441b-11e9-90f5-b97a0d886f60.png)

I also see a few requests for profiles, so it's possible that was a root cause (since the users table is so giant) but not sure...

jywarren:
It's just not very convincing that the notes api endpoint is the cause - it's a simple and limited query:

https://github.com/publiclab/plots2/blob/1eb99a97f1c0a78e3448cbb0995cedebc30b44d2/app/services/search_service.rb#L48-L51

I'm more inclined to think it could've been caused by another query and we're just seeing the effects but why would the loadtime affect this query and not other controllers?

jywarren:
@icarito i think we need to do a deeper look at the long-running mysql queries?

icarito:
Didn't find anything upon arriving
![imagen](https://user-images.githubusercontent.com/199755/54158074-b4d05700-4417-11e9-9eef-d51c34997477.png)

icarito:
I'll run the requests_log_analyzer on the mysql_slow log like in the old days

icarito:
I'm seeing a process related to the comment by tweet I think:
```
                                                                                                                                                              
 PID USER      PRI  NI  VIRT   RES   SHR S CPU% MEM%   TIME+  Command                                                                                         
[...]

PR #5184, opened 2019-03-19
Add debounce to questions and wiki forms, extend to 1 second

jywarren:
Re: #5015


jywarren:
This can be merged as soon as it passes!!!

PR #5224, opened 2019-03-21
group subscription tags with number of subscribers

### LINKED ISSUE #4603: Explore how to present subscription stats
Currently the stats page  https://publiclab.org/stats/subscriptions is being displayed as key-value pairs

* [x] Listing all tags and the number of subscribers in the right-hand list by @GettyOrawo 
* [ ] Interactive display of the page

GettyOrawo:
Fixes #4603 

publiclab-subscription-stats

**Display After Refactors**

new-stats


GettyOrawo:
@publiclab/reviewers I need a review on this one please

app/controllers/stats_controller.rb
```diff
@@ -5,7 +5,7 @@ def subscriptions
       @tags[tag.tagname] = @tags[tag.tagname] || 0
       @tags[tag.tagname] += 1
     end
-    render plain: @tags.inspect, status: 200
+    @tags = @tags.group_by { |_k, v| v }.map { |k, v| { k => v.map { |x| x.join("-") } } }
```
jywarren: So let's take this for example:

```ruby
> tags
=> {"boagd"=>3, "boa"=>34, "boag"=>39, "b"=>66}
```

see how if group by v/10 then 34 becomes 3, 66 becomes 6, so ones within 10 of each other get added to the same group?

```ruby
> tags.group_by { |k, v| v/10 }
=> {0=>[["boagd", 3]], 3=>[["boa", 34], ["boag", 39]], 6=>[["b", 66]]}
```

I think the second `map` is something we can do in the template though, since it relates to how we want to display the data. So i'll suggest:

```suggestion
    @tags = @tags.group_by { |k, v| v/10 }
    @tags = @tags.reverse # because it's naturally sorted "ascending", so let's flip it!
```

jywarren: nice that the group is automatically sorted, too!

app/views/stats/subscriptions.html.erb
```diff
@@ -4,4 +4,28 @@
...
     <h2>Subscriptions</h2>
-</div>
+ 	<table class="table inline-grid">
+		<tr>
+			<thead>
+				<th>Subscriptions</th>
+				<th>Tags</th>
+			</thead>
+		</tr>
+		<% @tags.each do |map| %>
```
jywarren: Here we could rename `map` to `group`, since each item in `@tags` represents a group of tags.

app/views/stats/subscriptions.html.erb
```diff
@@ -4,4 +4,28 @@
...
+		<% map.each do |subscriptions,tags| %>
```
jywarren: Here we wouldn't have to unpack things, we could skip this line!

app/views/stats/subscriptions.html.erb
jywarren:
```suggestion
		<% @tags.each do |range, tags| %>
```

We could do this because the key will represent the range of numbers, and the value will contain the tags themselves, for example `['tagname', 46]`.

app/views/stats/subscriptions.html.erb
```diff
@@ -4,4 +4,28 @@
...
+		<tr>
+			<tbody>
+				<td><%=subscriptions%></td>
```
jywarren: Here we could put this, because we'd want to expand back out to the full ranges (since we divided by 10 earlier) -- try this out:

```suggestion
				<td><% (range - 1) * 10 - <%= range*10 %></td>
```

app/views/stats/subscriptions.html.erb
```diff
@@ -4,4 +4,28 @@
...
+				<td>
+					<select style="width: 50%; height: 50px;">
+						<%tags.each do |tag| %>					      
+							<option><%=tag%></option>
```
jywarren: Then here, I actually think we could just display a list of tags, one per line in plain text, not in a select, so just the tag's name, a colon, and the true count next to it -- remember each of these is in the format `['tagname', 888]` :

```ruby
<p><%= tag[0] %>: <%= tag[1] %></p>
```

app/controllers/stats_controller.rb
GettyOrawo: Yes actually this works better and is way cleaner. The ruby hash doesn't have the reverse method though I think it's safe to use `sort`?

app/views/stats/subscriptions.html.erb
GettyOrawo: oh alright. so I'll do away with the joining tags to the count with hyphens (-)

app/views/stats/subscriptions.html.erb
GettyOrawo: Yes makes perfect sense. I figured this `<%=range*10%> - <%=range*10 + 9%>` displays from the minimum range to the range + 9 just so we have it kind of like this:

count

is this okay?

app/views/stats/subscriptions.html.erb
GettyOrawo: Yes this works okay, the display is however a little off.

new-stats

Is there something I can do about this or should i just leave it at this?

app/views/stats/subscriptions.html.erb
GettyOrawo: Okay I'll just use the tags map as it is.

app/views/stats/subscriptions.html.erb
jywarren: Huh, that is so weird. Let me look at the code, but I'm not sure why the `<p>` tags aren't making newlines. We could try `<br />` instead, between each line?

app/views/stats/subscriptions.html.erb
```diff
@@ -2,6 +2,25 @@
     <h4> About this page </h4>
     <p> People subscribe to tags so that they can be updated when a research is posted with the tag they are subscribed to. This page shows the number of subscriptions and subscription rates on different tags. </p>
 </div>
+
 <div class="col-md-8 col-md-offset-1">
     <h2>Subscriptions</h2>
-</div>
+ 	<table class="table inline-grid">
+		<tr>
```
jywarren: Not a huge deal, but would you mind using just 2 spaces for indents? We've tried to standardize a bit on this to help keep our code tidier. Thanks, and sorry for the minute details! 😄

app/views/stats/subscriptions.html.erb
```diff
@@ -2,6 +2,25 @@
...
+				<th>Subscriptions</th>
+				<th>Tags</th>
+			</thead>
+		</tr>
+		<% @tags.each do |range,tags| %>
+		<tr>
+			<tbody>
+				<td><%= range * 10 %> - <%= range * 10 + 9 %></td>
+				<% tags.each do |tag| %>
+				<td><%= tag[0]%>:<%=tag[1] %></td>
```
jywarren: Oh, aha - yes, so instead of the `<td>` we need to also use `<p>` tags - and the `<td>` tags should go outside the `each` loop. That way the whole collection will be in a single table cell, but each tag will be in its own paragraph (on its own line). Make sense?

GettyOrawo: Yes though long  it's clear. and more readable too. 

ptags

app/views/stats/subscriptions.html.erb
GettyOrawo: No  no. I hadn't used` <p>` tags here.

app/views/stats/subscriptions.html.erb
GettyOrawo: yes of course. sorry I switched to a new editor that I'm still finding my way around. I'll fix that.

GettyOrawo:
Hey @jywarren I cleared the suggested reviews. Any more fixes I need to do on this one?

app/views/stats/subscriptions.html.erb
jywarren: Great! Actually let's tackle a few more things in a follow-up PR:

1. let's sort this DESCENDING or `desc` so we start with the most-used tags
2. let's hide anything more than 25 results and show a "View all" link to show them?
3. let's make the tag names into links that go to `/tag/TAGNAME`

and we can do more! This is a great start!

jywarren:
Fantastic work!!! I'll open a new issue with the follow-ups. https://github.com/publiclab/plots2/issues/5260

Thanks so much!!!!

PR #5236, opened 2019-03-23
Fixing heading display in search results

### LINKED ISSUE #5156: Issue in search results of two or more words
We search with the input as two or more words, the first result seems to be fine but when I click the same or another button again such `All content` or `Notes` etc, the search bar and the results seems to include a plus sign `+` between the input words.

**Current behaviour:**
plissue2

**Expected behaviour:**

- [ ] The search bar and the `Results for` section should not include the `+` sign between the words on button clicks.

IshaGupta18:
Fixes #5156 

The heading is now space separated instead of "+" separated for all content tab on both first and multiple results (ie selecting the tab after selecting another tab, say notes, will now yield the same heading)

image

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -39,7 +39,7 @@ class SearchController < ApplicationController
     @profiles = @nodes[:profiles]
     @questions = @nodes[:questions]
     @tags = @nodes[:tags]
-    @heading =search_heading
+    @heading = search_heading
   end
 
   private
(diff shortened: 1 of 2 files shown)


IshaGupta18:
@jywarren please review and let me know what else can be done!

IshaGupta18:
Okay, I will look into your suggestion and let you know. I deliberately wrote a function just for heading generation (`search_heading`) so that no harm is done to the query param and a clean string is returned.

IshaGupta18:
@gauravano what do you think of this?

gauravano:
Hey @IshaGupta18, I agree with @vansjyo, in this case using `url_encode`  is better as encodes our query as URL which is required in this case. Thanks!

IshaGupta18:
Moving to #5349

PR #5248, opened 2019-03-24
Restore using node.views for unique impressions count

### LINKED ISSUE #1196: Impressionist counter_cache not working

We've installed the [impressionist gem](https://github.com/charlotte-ruby/impressionist) and are successfully counting unique impressions on pages now. But the "counter cache" setup (which would keep a column on the `node` table updated with a calculated tally of views) is not working, though that's in the code:

https://github.com/publiclab/plots2/blob/master/app/models/node.rb#L181

We've temporarily based `node.totalcount` on the sum of `node.legacy_views + node.impressions_count`:

https://github.com/publiclab/plots2/blob/master/app/models/node.rb#L183-L187

However, if we can get this line of the tests passing, we can re-institute the cached value, which will improve performance for node page views (notes and wikis):

https://github.com/publiclab/plots2/blob/master/test/functional/notes_controller_test.rb#L74

alonpeer:
Fixes #1196

This should improve performance of various page loads, as it avoids
running COUNT() queries on the impressions table, and instead uses the
views value, which the impressions gem keeps up-to-date.

Notice that I update the tests to `.reload` the node before reading values,
because ActiveRecord caches those, which is why tests fail without it.

---

Thanks!


alonpeer:
@publiclab/reviewers please review. @jywarren I saw you were involved in the early adoption of the impressions gem, I'm curious what your take on this is.

alonpeer:
I actually just realized there's a caveat with this fix.
Right now, `node.views` is always increased for **any** impression. After this change, it'll be increased only for unique IP addresses. This means that `node.totalviews` before this change is accurate (only unique views are returned), while after this change it'll increase a lot, due to duplicates per IP address. 

In order to fix this, along with applying this change, a migration is needed: the `node.views` values for all rows should be updated with the count of uniques according to the data in the `impressions` table.

I'm not sure exactly what the procedure is in PublicLab for these kind of changes, but I'm happy to pair with someone to get this done.

jywarren:
Hi @alonpeer ! Thank you so much; we could roll a migration into this PR if you think that makes sense? I think we are not /super/ worried about unique vs. non-unique, as this metric is just for getting a general idea anyways -- for a while, bots were triggering it. So, as long as we make a reasonable call, i think we can be OK with it. Thanks so much!

alonpeer:
@jywarren how many rows are there in `node` in production? I'm happy to add a migration here, but would like to avoid crashing the deploy process, or OOM it, which is why I suggested migrating this manually.
But if the number of rows is a few thousands, or 10k, I think that'll be just fine.

alonpeer:
I actually see other migrations written recently, which seem to do similar amount of work, so I assume that it's just fine 😃

jywarren:
Maybe run it only for nodes with status=1? But yeah, good call! Thanks!!

alonpeer:
Hmm, can't nodes with `status=0` get published again?

alonpeer:
Anyway, migration file added.

alonpeer:
Removed the [DO NOT MERGE] instruction 😄

gauravano:
Hey @alonpeer, I think you need to regenerate timestamp by running migration again as someone else has changed the file before you.

jywarren:
You can also manually forward-date it and it should work :-) Thank you!

PR #5251, opened 2019-03-25
add codeclimate test coverage

jywarren:
https://docs.codeclimate.com/docs/travis-ci-test-coverage


jywarren:
Goodness, why would this have any effect on a specific test like this?

```

 FAIL["test_contribution_graph_making", #<Minitest::Reporters::Suite:0x0000563e4480dda0 @name="NodeTest">, 12.947790247999933]
 test_contribution_graph_making#NodeTest (12.95s)
        Expected: 8
          Actual: 11
        test/unit/node_test.rb:440:in `block in <class:NodeTest>'
```
😕

jywarren:
Resolved in another PR, so this might need a rebase to get past it...

jywarren:
Hmm, don't see coverage, but it's not broken, so I'll try merging and see if CodeClimate starts reporting.

jywarren:
Hmm, actually it should be showing up on CodeClimate, even in a branch, but it isn't... probably needs more tweaking?

SidharthBansal:
@jywarren I think we can include this in GCI too
cc\ @VladimirMikulic @Uzay-G

jywarren:
I'm not sure, because codeclimate integration may need admin privileges?
But i hope I'm wrong!

VladimirMikulic:
@SidharthBansal to set up this, I would need admin privileges.

Uzay-G:
Yeah, this would be an interesting issue to work on, but me and @VladimirMikulic cannot complete it without admin privileges

SidharthBansal:
@jywarren I don't have admin privileges. Please ping him. Thanks

jywarren:
I'm afraid we haven't yet figured out how to extend privileges on CodeClimate. But if this is sure to work, you could screenshot the changes we need to make in the admin dashboard and we can merge this and make the changes for you? Could that work?

VladimirMikulic:
@jywarren I think that admins are able to this with only one click. You should have a button that says something like "Setup/Add test coverage" and when you click it, it should automatically modify your repo to support test coverage.

jywarren:
It looks like this only works with SimpleCov, and only < v0.18--

https://docs.codeclimate.com/docs/configuring-test-coverage#supported-languages-and-formats

https://github.com/codeclimate/test-reporter/issues/413

So, let's abandon for now, thanks!

PR #5246, opened 2019-03-24
Fixing Multi Tag Subscription Button

IshaGupta18:
Fixes #5245 
The multi-tag subscription now works!
image

PR #5314, opened 2019-03-30
Removing condition for multi-tag subscribe button

### LINKED ISSUE #5313: Removing Like condition from multi-tag subscribe button
Just a little change to #5246 .
Removing the condition whether the user has liked a node or not and displaying the multi-tag subscribe button irrespective of this.

IshaGupta18:
Fixes #5313 

Multi-tag subscribe button can now be visible to all users!

### DIFF
--- a/app/views/tag/_tagging.html.erb
+++ b/app/views/tag/_tagging.html.erb
@@ -3,7 +3,7 @@
 <!-- This is the sidebar tagging display, also renders Subscribe button for multiple subscription -->
 <div>
   <h1 style="display: inline-block; margin-right: 20px;">Tags</h1>
-  <% if @node && !@node.node_tags.empty? && ((current_user && @node.liked_by(current_user.uid)) || !current_user) %>
+  <% if @node && !@node.node_tags.empty? %>
     <%= render partial: 'tag/subscribe_button', locals:{tags: @node.node_tags} %>
   <% end %>
 </div>


IshaGupta18:
@jywarren @gauravano here's my PR for the follow up on #5246. Let me know what else can be done here. Thanks a lot!

IshaGupta18:
Also, I feel that we don't need a follow button on tooltips anymore, so we can get rid of that. What do you think?
Screenshot from 2019-03-31 01-55-32

I could make an FTO for this if it's okay!

gauravano:
It's your call @IshaGupta18, whether you want to open an FTO for this or not. I think work is complete here, so whatever it would be let me know. I am ready to merge this too. Thanks!

IshaGupta18:
Actually, I realized the FTO thing a little late, so that's why I suggested
that I could make another FTO in it's place, for the follow button thing.
As for this one, since the work is complete here, I think this can be
merged. Would that be okay?

gauravano:
Yes :+1:

gauravano:
Merged :tada: . Thanks @IshaGupta18!

IshaGupta18:
Thanks a lot 🎉

jywarren:
Hi IshaGupta18! Thanks! Merging this now, but I think we should try keeping
the tooltips a bit longer to see how this does. We can also do some user
studies to see what people prefer, where they get stuck, using
https://lookback.io -- what do you think?

IshaGupta18:
Hey, I think it makes sense. If we do this, we can also improve on the
button's design to make it more usable. We can keep the tooltip and explore
on how to make the multi-tag subscribe button more accessible. Thanks!

PR #5338, opened 2019-04-01
Use node views cache

### LINKED ISSUE #1196: Impressionist counter_cache not working

alonpeer:
Fixes #1196

This should improve performance of various page loads, as it avoids
running COUNT() queries on the impressions table, and instead uses the
views value, which the impressions gem keeps up-to-date.

Notice that I update the tests to `.reload` the node before reading values,
because ActiveRecord caches those, which is why tests fail without it.

---

Thanks!

### DIFF
--- a/test/functional/notes_controller_test.rb
+++ b/test/functional/notes_controller_test.rb
@@ -81,7 +81,7 @@ class NotesControllerTest < ActionController::TestCase
     assert_equal '0.0.0.0', Impression.last.ip_address
     Impression.last.update_attribute('ip_address', '0.0.0.1')
 
-    assert_difference 'note.totalviews', 1 do
+    assert_difference 'note.reload.totalviews', 1 do
       get :show,
           params: {
           author: note.author.name,
@@ -90,10 +90,10 @@ class NotesControllerTest < ActionController::TestCase
           }
     end
 
-    assert_equal 2, note.totalviews
+    assert_equal 2, note.reload.totalviews
 
     # same IP won't add to views twice
-    assert_difference 'note.totalviews', 0 do
+    assert_difference 'note.reload.totalviews', 0 do
       get :show,
           params: {
           author: note.author.name,
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -166,12 +166,10 @@ class Node < ActiveRecord::Base
 
   public
 
-  is_impressionable counter_cache: true, column_name: :views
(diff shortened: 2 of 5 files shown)


alonpeer:
@publiclab/reviewers @gauravano @jywarren I accidentally deleted my branch and closed PR #5248. This PR replaces it and fixes the conflict.

test/functional/notes_controller_test.rb
```diff
@@ -81,7 +81,7 @@ def teardown
     assert_equal '0.0.0.0', Impression.last.ip_address
     Impression.last.update_attribute('ip_address', '0.0.0.1')
 
-    assert_difference 'note.totalviews', 1 do
+    assert_difference 'note.reload.totalviews', 1 do
```
jywarren: Oh interesting. Can we add a bit of documentation about this usage? I'm not familiar with it! Thanks!

jywarren: Ah! I see your note above:

> Notice that I update the tests to .reload the node before reading values,
because ActiveRecord caches those, which is why tests fail without it.

Does it cache it just within the context of the test? Any other situations where we'd need to flush the cache? Thank you!!!

alonpeer: From my experience with Rails, this is a tests-specific hack. The cache is on request level, so it shouldn't affect any production paths. 
In this specific test, we're reading the note, then changing the impressions, then trying to read a value from the note. Since the note is cached (i.e. already read), we have to force it to reload from DB, so that we get the most recent value of `totalviews`.

jywarren:
All right -- this is great then! Merging now!!! 🎉 Thanks for your patience!!!

PR #5349, opened 2019-04-02
Fixing query string in search results

### LINKED ISSUE #5156: Issue in search results of two or more words

IshaGupta18:
Fixes #5156 

Closing #5236 and moving to this one. 

Using url_encode to achieve the required string for displaying search results.

image

### DIFF
--- a/app/views/search/all_content.html.erb
+++ b/app/views/search/all_content.html.erb
@@ -3,12 +3,12 @@
 <div class="container">
   <% if  @nodes.all?{|k,v| v.blank?} %>
     <h4>
-      No Content found for <%= url_encode(params[:query]).gsub!('%2B',' ') %> :-(
+      No Content found for <%= CGI.unescape(params[:query]) %> :-(
       </br></br>
       Try searching <a href="/search">for another topic</a>
     </h4>
   <% else %>
-    <h2> Results for <%= url_encode(params[:query]).gsub!('%2B',' ') %></h2>
+    <h2> Results for <%= CGI.unescape(params[:query]) %></h2>
     <% unless @notes.blank? %>
       <h3> NOTES</h3>
       <table class="table inline-grid">
--- a/app/views/search/all_content.html.erb
+++ b/app/views/search/all_content.html.erb
@@ -3,12 +3,12 @@
 <div class="container">
   <% if  @nodes.all?{|k,v| v.blank?} %>
     <h4>
-      No Content found for <%= params[:query] %> :-(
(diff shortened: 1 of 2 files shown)


IshaGupta18:
@jywarren  @gauravano kindly review this PR if it's good to go. If yes, I would create follow up first-timer issues for the same. Thanks a lot!

IshaGupta18:
Okay, so I see that the latest test written in `search_test.rb` is failing. @jywarren could you shed some light on it, please?

gauravano:
Hey @IshaGupta18, this would just solve the issue for single tab. Instead of replacing it here, you can replace at https://github.com/publiclab/plots2/blob/master/app/views/search/_nav.html.erb#L3 as suggested here https://github.com/publiclab/plots2/pull/5236#pullrequestreview-218072935. Thanks!

app/views/search/all_content.html.erb
```diff
@@ -3,12 +3,12 @@
 <div class="container">
   <% if  @nodes.all?{|k,v| v.blank?} %>
     <h4>
-      No Content found for <%= params[:query] %> :-(
+      No Content found for <%= CGI.unescape(params[:query]) %> :-(
```
IshaGupta18: @gauravano this is what has been done to fix the string. In every tab. `params[:query]` is displayed. To make it universal for all tabs, we can make a function in the controller and return `CGI.unescape(params[:query])` from it, kind of like what has been done in the review, What do you think?

gauravano: this line https://github.com/publiclab/plots2/blob/master/app/views/search/_nav.html.erb#L3 is the one which is calling the helper method generating all the nav links so just changing here from CGI.unescape to url_encode will do all the work. Try it out. T

IshaGupta18: Does this work now?

gauravano: Yes!

IshaGupta18: Is it good to be merged?

jywarren:
Thank you so much!!!! Sorry i wasn't able to help earlier @IshaGupta18 but you did a great job. Thanks to @gauravano as well!!!

PR #5368, opened 2019-04-04
Add validation for form fields

### LINKED ISSUE #5361: Form field validation and password visibility
## Description
While signup and editing the profile, there is need to write the password and confirm it by writing it again.
There is no way to check whether the password written in both of the field is same unless the submit button is pressed.
And while writing in form, we have no way to know that the data we wrote in field is in correct format.

**Instead we can have these two features in password** - 
* Password matching while filling up the form.
* Eye-icon similar to login form available in signup to toggle the visibility. or maybe we can have show password.

**And we can have this feature for form**
* Add Form validation so that the user gets notified about field errors as soon as the user tabs to next field.

Screenshot from 2019-04-03 22-28-03

I made a similar issue  and PR earlier for one of the part of this issue, which was for password visibility but it was for login form. 
https://github.com/publiclab/plots2/issues/5201

CleverFool77:
Fixes #5361 (<=== Add issue number here)

## Description
Currently, there is no way to validate the form fields.
This PR adds realtime validation to form fields and notifies the user about the same on field focus change.
I've added the form validation for both create and edit profile forms.
* Edit profile
gg

* Create profile
dd

Thanks!


CleverFool77:
Hi @jywarren @gauravano 
Is this method correct ?

CleverFool77:
Please Review.
Thanks !!!

gauravano:
Hi @CleverFool77, sorry for delay here. Could you please remove the conflicts? Thanks!

app/views/users/_edit_form.html.erb
```diff
@@ -19,14 +19,14 @@
...
         <label for="username"><%= t('users._form.username') %></label>
-        <%= f.text_field :username, { readonly: true, tabindex: 1, placeholder: "Username", class: 'form-control', id: 'username' , title: 'Username field cannot be changed', disabled: true} %>
+        <%= f.text_field :username, { readonly: true, tabindex: 1, placeholder: "Username", class: 'form-control', id: 'username' , name: 'username', title: 'Username field cannot be changed', disabled: true} %>
         <input type="hidden" name="id" value="<%= @user.username %>" />
       </div>
 
       <div class="form-group col-sm-6">
 
         <label for="email"><%= t('users._form.email') %></label>
-        <%= f.text_field :email, { tabindex: 3, required: true, pattern: '[^@]+@[^@]+\.[a-zA-Z]{2,6}', placeholder: "you@email.com", class: 'form-control', id: 'email' } %>
```
CleverFool77: Hi @gauravano 
I removed this validation of email using as this has many failed cases.
eg - 
* It doesn't allow lekhika@lekhika which is considered valid email.
https://stackoverflow.com/questions/20573488/why-does-html5-form-validation-allow-emails-without-a-dot
* It doesn't allow the lekhika@hotmail.engineer because the regex limited the length of domain name to max 6 characters which isn't correct..
* Also, I've used jquery validation library which automatically performs reatlime email validation.

gauravano:
Hi @CleverFool77, I tried your PR:
I used:

Username: you
Email: you@email.com
password: password 

And got these errors:
Email should look like an email address.
Email is invalid
Username is too short (minimum is 3 characters)
Username should use only letters, numbers, spaces, and .-_@+ please.
Username is invalid
Password is too short (minimum is 8 characters)
Password confirmation is too short (minimum is 8 characters)

app/views/users/_edit_form.html.erb
gauravano: We can remove but please check if rails allow it or not.

CleverFool77:
> Hi @CleverFool77, I tried your PR:
> I used:
> 
> Username: you
> Email: [you@email.com](mailto:you@email.com)
> password: password
> 
> And got these errors:
> Email should look like an email address.
> Email is invalid
> Username is too short (minimum is 3 characters)
> Username should use only letters, numbers, spaces, and .-_@+ please.
> Username is invalid
> Password is too short (minimum is 8 characters)
> Password confirmation is too short (minimum is 8 characters)

HI @gauravano As it's a realtime validation. So it shows these changes while the focus is on field but the condition is not fulfilled. As soon as the condition is fulfilled. It validated them by not showing any error anymore.
It shows one more error of password not matching. If the password written is not same.
Regarding username and password minimum length. Is this condition should not be there ?

gauravano:
First of all, the above errors I mentioned, I got those after I clicked signup button which means either JS validations should stop me for signup or warn me but nothing like this happened. 

If we want to validate our form using JS then we should try to put same conditions which rails imposes otherwise user will get no error while filling form but get errors after submitting like the above case.

Thanks!

CleverFool77:
Hi @gauravano Sorry for the delay. I'm unable to do some updates on PR till a week due to exams.
I'll update them soon.
Thanks !!!

CleverFool77:
Hi @gauravano I see that many different PRs for username and email and other stuff is already added in different PRs either using regex and other stuff. 
So I'll close this one and open a new PR which will be having just the password part as this part is not added yet.

gauravano:
Cool. Thank you!

PR #5402, opened 2019-04-07
fix the bug of traceback error of server on webpage

### LINKED ISSUE #5378: traceback of error from server is shown in webpage
## Description
While working on the issue https://github.com/publiclab/plots2/issues/5327
I found this. 
It showed the traceback of error from server on the webpage. And after I refreshed it, the tag was added.
The gif shows the steps to reproduce it - 

ff

CleverFool77:
Fixes #5378 https://github.com/publiclab/plots2/issues/5404(<=== Add issue number here)

The changes - 
* If the user adds with tag like `with`,then  there won't be a traceback error like earlier
* If the user adds `with:` or `barnstars:` tag, then it won't show the traceback error rather it would show that the power tag can't be empty.
* The tagname is stripped of whitepaces in ends. As for admin, the code doesn't check condition to find username rather adds it directly. So In such cases the whitespace might be added. 

**Before changes**
error

**After changes**
ff

Thanks!

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -291,7 +291,7 @@ class TagController < ApplicationController
       if Tag.exists?(tagname, nid)
         @output[:errors] << I18n.t('tag_controller.tag_already_exists')
 
-      elsif tagname.include?(":") && tagname.split(':').length < 2 # cant tag empty power tags
+      elsif tagname.include?(":") && tagname.split(':').length < 2
         if tagname.split(':')[0] == "barnstar" || tagname.split(':')[0] == "with"
           @output[:errors] << I18n.t('tag_controller.cant_be_empty')
         end
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -287,25 +287,33 @@ class TagController < ApplicationController
     node = Node.find nid
     tagnames.each do |tagname|
       # this should all be done in the model:
-
+      tagname = tagname.strip
       if Tag.exists?(tagname, nid)
         @output[:errors] << I18n.t('tag_controller.tag_already_exists')
+
+      elsif tagname.include?(":") && tagname.split(':').length < 2 # cant tag empty power tags
+        if tagname.split(':')[0] == "barnstar" || tagname.split(':')[0] == "with"
+          @output[:errors] << I18n.t('tag_controller.cant_be_empty')
+        end
+
       elsif node.can_tag(tagname, current_user) === true || current_user.role == 'admin' # || current_user.role == "moderator"
         saved, tag = node.add_tag(tagname.strip, current_user)
-        if tagname.split(':')[0] == "barnstar"
-          CommentMailer.notify_barnstar(current_user, node)
-          barnstar_info_link = '<a href="//' + request.host.to_s + '/wiki/barnstars">barnstar</a>'
-          node.add_comment(subject: 'barnstar',
-                           uid: current_user.uid,
-                           body: "@#{current_user.username} awards a #{barnstar_info_link} to #{node.user.name} for their awesome contribution!")
+        if tagname.include?(":") && tagname.split(':').length == 2
+          if tagname.split(':')[0] == "barnstar"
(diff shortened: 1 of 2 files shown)


CleverFool77:
Hi @jywarren @gauravano ,
while working on this issue.
I found some bugs, for which I've opened a new issue https://github.com/publiclab/plots2/issues/5404

app/controllers/tag_controller.rb
```diff
@@ -300,7 +300,7 @@ def create
                            body: "@#{current_user.username} awards a #{barnstar_info_link} to #{node.user.name} for their awesome contribution!")
 
         elsif tagname.split(':')[0] == "with"
-          user = User.find_by_username_case_insensitive(tagname.split(':')[1])
+          user = User.find_by_username_case_insensitive(current_user.username)
```
gauravano: Power-tag `with:user1` adds user with username `user1` as co-author of the note. So, we want to get their username from tag and so this change is wrong

CleverFool77: Ahh!!! I missed out this. Sorry for the wrong change. I'll commit with correct one soon.
Thanks @gauravano

CleverFool77:
Hi @gauravano I've updated the PR as requested. :smile:
Thanks !!!!!

CleverFool77:
@publiclab/plots2-reviewers 
Please Review.
Thanks !!

app/controllers/tag_controller.rb
```diff
@@ -287,25 +287,33 @@ def create
     node = Node.find nid
     tagnames.each do |tagname|
       # this should all be done in the model:
-
+      tagname = tagname.strip
       if Tag.exists?(tagname, nid)
         @output[:errors] << I18n.t('tag_controller.tag_already_exists')
+
+      elsif tagname.include?(":") && tagname.split(':').length < 2 # cant tag empty power tags
```
gauravano: You mean `Can't add empty power tags` :sweat_smile: ?

app/controllers/tag_controller.rb
```diff
@@ -287,25 +287,33 @@ def create
...
+        if tagname.split(':')[0] == "barnstar" || tagname.split(':')[0] == "with"
```
gauravano: hey @CleverFool77, I think we can implement this for all power tags, we have many - https://publiclab.org/wiki/power-tags. What do you say?

gauravano:
Great work here! @CleverFool77 :100:

app/controllers/tag_controller.rb
CleverFool77: Hi @gauravano 
What I mean was we can add `with` but not `with:` ? :sweat_smile: 
Is this not supposed to be like this ?

app/controllers/tag_controller.rb
CleverFool77: Hi @gauravano In other power tags, If I'll add some of the other power tags like `build:`, `alert:` ,`notes:` etc then neither it shows traceback error nor its gets added, which is correct.

Only `with:` and `barnstars:` which were implemented in different way were showing traceback error and That's why I added only both of them here.
Thanks !!!

app/controllers/tag_controller.rb
gauravano: I was referring to comment not code :sweat_smile:

app/controllers/tag_controller.rb
gauravano: Ok! I think, we can discuss it after merging this that whether we want to show alert if someone try to use incomplete power tag, even in case where no error is thrown. Thanks!

gauravano:
I think, we are almost done here. Just correct the comment or you can also remove it. Thanks!

app/controllers/tag_controller.rb
CleverFool77: Ahh!! Forgot about that. I'll fix it.

CleverFool77:
Hi @gauravano 
I've updated the PR.
Thanks !!

gauravano:
:tada: :100: :+1:

PR #5411, opened 2019-04-08
Skip large gif post-processing

### LINKED ISSUE #1159: Avoid to convert huge animations
oglwmu
Big gif animation was uploaded.

### Relevant URLs
(I've attached the culprit image)

### PublicLab.org username
icarito

(to help reproduce the issue)

Any

For bug reports, fill out the above template; for feature requests, you can delete the template.

jywarren:
Fixes #1159

### DIFF
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -18,10 +18,14 @@ class Image < ApplicationRecord
 
   before_validation :download_remote_image, if: :remote_url_provided?
   validates :remote_url, presence: true, if: :remote_url_provided? # , :message => "is invalid or inaccessible" # this message thing is old-style rails 2.3.x
-  before_post_process :is_image?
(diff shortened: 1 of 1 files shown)


jywarren:
We could add a test like this:

https://github.com/publiclab/mapknitter/pull/503/files#diff-5ac214a3841eff7c16d7408995977adaR11

```ruby
    @uploaded_data = ActionDispatch::Http::UploadedFile.new(tempfile: @file, filename: File.basename(@file), type: "image/png")
    post :create, map_id: @map.id, uploaded_data: @uploaded_data
````

using a big gif... although we'd want to fetch it from somewhere outside the repo...

jywarren:
@icarito is this urgent enough to merge now, or can we take an extra step and write a test for it?

Issue #5427, opened 2019-04-09
Add comment if the user is removed as co-author from note


CleverFool77:
## Description
I noticed that we add a comment if the person is added as co-author for a note.
But there is no way to show that the user is deleted as co-author from note unless looked into the tags.
So we can add a comment showing that the `user is removed as co-author from note` from just like it's done while adding a user to note as co-author.

CleverFool77:
@publiclab/plots2-reviewers What's your opinion regarding this issue.
Thanks !!

PR #5438, opened 2019-04-10
Add comment if user is removed as co-author from note

CleverFool77:
Fixes https://github.com/publiclab/plots2/issues/5427 (<=== Add issue number here)

comment

Thanks!

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -348,6 +348,14 @@ class TagController < ApplicationController
         node.delete_coord_attribute(tag.name)
       end
 
+      if tag.name.split(':')[0] == "with" && tag.name.split(':').length == 2
+        user = User.find_by_username_case_insensitive(tag.name.split(':')[1])
+        CommentMailer.notify_coauthor(user, node)
+        node.add_comment(subject: 'co-author',
+                         uid: current_user.uid,
+                         body: " @#{current_user.username} has removed @#{tag.name.split(':')[1]} as a co-author. ")
+      end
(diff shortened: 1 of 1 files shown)


jywarren:
Hi, thank you @cleverfool77! I think this one may need some input
from @publiclab/community-reps -- Thanks!!!

IshaGupta18:
Hey @jywarren @CleverFool77 what I think of this issue is that we c don't need to put it as a comment, we could send a notification kind of email to the co-author removed so that it doesn't look too bad as a comment. Get what I mean?

CleverFool77:
Hi @IshaGupta18 If we are adding a comment when a user is tagged, then why not same for removal of user as -co-author.
And I don't think it would look bad. I didn't get it. Even on GitHub, when we unassigned a issue or close a issue , we show it directly. we do `unassigned`. That's not bad I guess.
Here in the code, it not just adds comment but the user is also getting notified about comments.

And if we just want a notification email for removal, then it should be same for addition too. The procedure for both methods should be same, considering they are the part of same event that is addition or deletion of `with` power tag.

@jywarren Sure. It would be great to have some inputs and discussion.

> Hey @jywarren @CleverFool77 what I think of this issue is that we c don't need to put it as a comment, we could send a notification kind of email to the co-author removed so that it doesn't look too bad as a comment. Get what I mean?

jywarren:
Hi all, I was thinking more about this and think that for the time being, we dont' need to leave such a comment. But I appreciate your help! Thank you!!!

PR #5261, opened 2019-03-25
[WIP] Add 'accepted' emoji for verified comments which are previously accepted answers

ViditChitkara:
Fixes #5203

Issue #4094, opened 2018-12-02
Planning issue: merging Answers into Comments


jywarren:
There is a long list of issues with Answers and Comments, not only because they're easy to confuse (#884) but also because having both requires double maintenance and lots of duplicated code (#3535, #4034, we'd need to develop a moderation system for answers in addition to questions, etc etc). 

Let's make a few changes to the system, and consider merging the two types. We would have to:

1. make it possible to comment in response to another comment (perhaps using the comment.aid field to refer to a comment.cid it's in response to, instead of the answer.aid it's in response to)
2. convert all Answers into Comments, preserving timestamps, and changing all attached comments to the new Comment record, in a big database migration
3. display comments that respond to other comments [in the threaded way that Answers are currently shown](https://publiclab.org/questions/warren/02-21-2018/what-do-you-think-about-this-new-design-sketch-for-a-welcome-page)
4. make a "reply" button on comments that inserts the `cid` into the form so that the response goes to that comment, OR build in a commenting form to each comment... ?
5. migrate `answer_selection` (i.e. [likes on answers](https://github.com/publiclab/plots2/blob/master/app/models/answer.rb#L22)) to the differently-plumbed [comment likes](https://github.com/publiclab/plots2/blob/7740bee42c96cfdc32f188e4f69c98bd360a8064/app/models/comment.rb#L181-L207) (which is actually a nicer system with emoji and such)
6. figure out how the `question.answers` features will work, because answers can be "accepted" -- perhaps this also becomes part of the comment liking system, with a "checkbox" emoji? 
7. ensure statistics for Answers (http://publiclab.org/stats) are preserved somehow or merged into comments?

This is a big, complex project. But the outcome is that the codebase would be MUCH simpler and less redundant. There wouldn't be two different mixed types of responses to a post (comments + answers), but just one. We'd no longer have to have systems for turning comments into answers, and discussions wouldn't get fragmented across the two systems, as in https://publiclab.org/questions/warren/02-21-2018/what-do-you-think-about-this-new-design-sketch-for-a-welcome-page

This is a large project that will take careful work! Most of it may have to happen in the same pull request! 

We could conceivably make the comment-in-reply-to-comment system first, test it, then begin merging the old Answers into that system, to do this in 2 parts. But no matter what, it'll take some hard focused work to achieve and not lose any content.

ViditChitkara:
[...]

PR #5478, opened 2019-04-14
refactored questions/answered page

### LINKED ISSUE #5203: listing answered questions after deprecating Answers
Hi @ViditChitkara -- just want to open an issue to tackle how this will work:

https://publiclab.org/questions/answered/

Both on a) how we'd list them on this page, and also b) how answered comments are represented when viewing a list of comments in a question page?

I know you've thought this through, i just want to try to sum it up in one place as we take the next space. 

Thanks!

ViditChitkara:
Fixes #5203

### DIFF
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -71,12 +71,12 @@ class QuestionsController < ApplicationController
   end
 
   def answered
-    @title = 'Recently answered'
+    @title = 'Recently Commented'
     @questions = Node.questions
       .where(status: 1)
     @questions = filter_questions_by_tag(@questions, params[:tagnames])
-      .joins(:answers)
-      .order('answers.created_at DESC')
+      .joins(:comments)
+      .order('comments.timestamp DESC')
       .group('node.nid')
       .paginate(page: params[:page], per_page: 24)
     @wikis = Node.limit(10)
--- a/app/views/dashboard/_node_question.html.erb
+++ b/app/views/dashboard/_node_question.html.erb
@@ -14,12 +14,6 @@
       <% end %>
   
       <h4><a href="<%= node.path(:question) %>"><%= node.title %></a></h4>
-  
-      <% if node.answers.collect(&:accepted).include?(true) %>
-        <p><span class="badge accepted">Answered</span></p>
-      <% end %>
-      
-      <a href="<%= node.path(:question) %>#post-answer" class="btn btn-outline-secondary btn-sm float-right respond answer"><%= t('dashboard._node_question.post_answer') %></a>
 
       <p class="meta"><%= t('dashboard._node_question.question') %> <%= render partial: "dashboard/node_meta", locals: { node: node } %></p>
 
--- a/app/views/questions/_questions.html.erb
+++ b/app/views/questions/_questions.html.erb
@@ -11,18 +11,11 @@
 
         <h4><a <% if @widget %>target="_blank"<% end %> href="<%= node.path(:question) %>"><%= node.title %></a></h4>
 
-        <% if node.answers.collect(&:accepted).include?(true) %>
-          <p><span class="label accepted">Answered</span></p>
-        <% end %>
-
-
-
         <span id="tags">
         <% node.tags.limit(4).each do |tag| %>
           <span id="tag_<%= tag.id %>" ><a class="badge badge-primary" href="/questions/tag/<%= tag.name %>"><%= tag.name %></a></span>
         <% end %>
         </span>
-        <a href="<%= node.path(:question) %>#post-answer" class="btn btn-outline-secondary btn-sm">Post an answer</a></p>
         <p class="meta">asked <%= render partial: "dashboard/node_meta", locals: { node: node } %></p>
       </div>
     </div>
--- a/app/views/questions/index.html.erb
+++ b/app/views/questions/index.html.erb
@@ -1,7 +1,7 @@
 <div class="col-lg-12">
 
   <% if params[:action] == 'answered' %>
-  <h2>Recently Answered <small class="d-md-none d-lg-inline">Recently answered questions</small></h2>
+  <h2>Recently Commented <small class="d-md-none d-lg-inline">Recently commented questions</small></h2>
   <% elsif params[:action] == 'popular' %>
   <h2>Popular Questions <small class="d-md-none d-lg-inline">Most viewed questions</small></h2>
   <% elsif params[:action] == 'unanswered' %>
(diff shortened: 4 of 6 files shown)


ViditChitkara:
Screenshot 2019-04-16 at 10 43 40 PM

Changed the questions/answered page

jywarren:
This looks great!!! Thanks! Would you mind resolving conflicts? It could've been the Bootstrap 4 upgrade?

ViditChitkara:
Resolved. @jywarren

jywarren:
Great!! Just checking if we wanted to follow up by looking for checkmark responses as a metric for "answering"?

jywarren:
@ebarry may want to chime in here? @skilfullycurled? Thanks all!

jywarren:
The context is that we're wrapping up the merge of "answers" as a distinct type into just comments and comment replies.

skilfullycurled:
I wrote a longer piece about this [here](https://github.com/publiclab/plots2/issues/4094#issuecomment-497064249), but on thing to consider is the nature of the questions and help that people need.  Would it be better to have a indicator of "helpfulness" like a "thumbs up" metric?  

Trying to get a spectrometer to take a clearer picture could have lots of causes which is unlike traditional stack overflow where an answer can be tested and verified leaving the only matter to be which is the best answer which is indicated via upvotes.

Someone might say something helpful which leads to an answer but it might not be the answer itself.  I've been fortunate that my questions have been answered in the classical sense, but overall, I'm not sure how many Q&A's we get that haven't technically achieved the binary status of having been answered, but did lead to a helpful resolution of the problem.

ViditChitkara:
> Great!! Just checking if we wanted to follow up by looking for checkmark responses as a metric for "answering"?

Hi @jywarren , @skilfullycurled 
So, I started with the checkmark emoji part. However, it didn't seem to work on unstable. Wasn't able to figure out why this was happening. Any ideas here would be highly appreciated #5261

@skilfullycurled according to me, the ability to accept an answer shall be given to moderators or admins who might analyse the number of "thumbs ups" or other emojis and accordingly accept the answers. Or we may set a threshold number of "thumbs up" after which the comments automatically get accepted. What do you think? In #5261 we are just creating 'accepted' likes for old answers which were converted to comments.

Thanks

skilfullycurled:
@ViditChitkara, thanks for including me in the conversation!  The first idea requires some feedback from moderators.  If it were me, that would feel like a lot of work to pay attention to all the questions that are coming in, even if you were able to filter them somehow.  If their acceptance is going to be based on a threshold of indicators, then we might as well choose the threshold and do it automatically (your second idea).

I've been asking myself, "Okay, why would we want a check mark or a thumbs up?"  What's the goal?  For the check mark, I can think of two goals: say someone is having a problem with balloon mapping.  There's going to be a lot of questions on that, so they might want to look at all of the questions that have answers first.  And, when they go to the question, they might want to jump straight to a working answer.  Answer-ers might want to know where their time would be best spent and if they're just stopping by, quickly see if their is anything to add given the accepted answer.  So right now, I'm seeing it as a sort of an information filtering mechanism.

What I'm grappling with is how to balance such an indication with the fact that while some questions might have single answers, more likely, the conversation will lead to many potential answers.  So maybe it would be better to think of check marks as "problem solved" in general (the number of people who say this question contains information that resolved their issue) and the "thumbs up" as the parts of the conversation thread that were useful in arriving at a resolution.  That's a long way to say, I'm moving away from the desire to have single answers that are accepted.  

It'd be great to hear from folks who were around when the feature was designed to learn more about it's intention, and also people who answer a lot of questions to hear about the ways in which people are using it now that the feature has been released.

ViditChitkara:
> It'd be great to hear from folks who were around when the feature was designed to learn more about it's intention, and also people who answer a lot of questions to hear about the ways in which people are using it now that the feature has been released.

Coming back here. Ideas anyone @jywarren @skilfullycurled ?

ViditChitkara:
Perhaps that's the way comments on github work too...Comments with more thumbs ups are more useful than any others according to me. Maybe tick mark would add as a metric to a formal answer to a question. The problem however is how to decide which comment receives this checkmark. A couple of ideas we discussed were:-
1) Shall be in hands of moderators. Again it will increase a lot of overhead and there might be cases that the accepted appears ambiguous to groups of people (which may appear right to the moderators). Also there is a chance that a questions is missed by chance.

 2) Decide a threshold number of thumbs up or any set of emoji after which the comment would be automatically tick marked. For eg: 5 👍 4 ❤️ may be enough for a ✅.(The number is just an assumption).

skilfullycurled:
Thanks for bumping up the conversation again @ViditChitkara.  Here are my thoughts since we last discussed:

a) I like your threshold idea, it's really novel, however, if we went with it, I think it would be good for a different emoji.  Perhaps it's something like the medal or trophy emoji?  In the aggregate that would connote two things about the question/answer pairing: 1) it contains answers that were very useful and 
2) since it's based on a threshold, that in some sense, the question itself was a good one.

b)  I'm no longer sure that marking a single answer to a question with a check mark is useful even for a place like Stack Overflow.  Often, what is accepted at that time is later superseded by a better answer, and despite an answer being the best by some objective measure (say, speed and complexity) a different answer might be better for someone who seeks simplicity and doesn't need to scale a program.  Therefore, I think both Stack Overflow and Public Lab, would do best to allow multiple "accepted answers" as an indication that the probability of one of the answers working will be very high.  That would allow one to prioritize looking at questions with 5 answers and 30 check marks over 1 answer and 4 check marks.

I think this would be even more applicable to Public Lab where the questions lend themselves to different answers all of which are correct, but just not the answer in a specific instance.  There's a bit more troubleshooting going on.  Additionally, since Public Lab allows open ended questions, there will be some that by definition can't have a single answer

jywarren:
Yeah i think our original idea was also not to have a binary `answered/unanswered`, but to have `unanswered/answered/answer-accepted`, AND to have these certainly allow for multiple answers, either accepted or not. The just-deprecated system has allowed such multiple answers with 3 states. Now that we are on a simpler system we can rethink, but I'm inclined to aim for an even simpler system for answering.

I think that instead of marking completion, we should think as @skilfullycurled has posited:

> a indicator of "helpfulness" like a "thumbs up" metric?

and

> they might want to look at all of the questions that have answers first. And, when they go to the question, they might want to jump straight to a working answer. Answer-ers might want to know where their time would be best spent and if they're just stopping by, quickly see if their is anything to add given the accepted answer. So right now, I'm seeing it as a sort of an information filtering mechanism.

I think the most important function here is to make it possible /in list views/ to see which questions have NOT gotten answers. 

Currently some questions are marked with a grey checkmark if they have been answered, and a green one if an answer has been accepted. This distinction doesn't seem that important, but in skimming a list, it is great to see when:

1. a question has no answers at all (which we could replace with just "no comments at all" OR "no comments that have been positively responded too, however we choose to define that)
2. a question has been responded to a lot (which at least indicates that it led to a lot more discussion and information exchange

So really i think we can just narrow this to something like:

1. any question that has comments on it should be visibly "responded to" and 
2. any question that has a comment with a ✅  (OR a 👍 ?) should be visibly "answered at least once" without a strong implication that one answer is the end state, or better than any other answer. So, maybe the technical metric would be, visibly "commented on in such a way that SOMEONE feels satisfied with at least one comment" 😄 ✅ 

What say you?

ViditChitkara:
These all seem like good ideas. I'll make an issue for further discussing this. Although we have much information. Will compile it up there.

PR #5496, opened 2019-04-16
remove answer notification code

ViditChitkara:
Part of #4094


jywarren:
Looks good; we resolved an upstream test failure today so perhaps this just needs a rebase? Not sure, though! Thanks!!!!

ViditChitkara:
@publiclab/reviewers , @jywarren I think this is ready to merge.

app/controllers/answer_like_controller.rb
```diff
@@ -11,8 +11,7 @@ def likes
       AnswerSelection.set_likes(current_user.uid, @answer.id, false)
     else
       AnswerSelection.set_likes(current_user.uid, @answer.id, true)
-      user = User.find(current_user.uid)
-      AnswerMailer.notify_answer_like(user, @answer).deliver_later
+      User.find(current_user.uid)
```
jainaman224: I think this is not required. We already have user as `current_user`.
😄

ViditChitkara: Yeah, I think that line didn't make any sense. Thanks for pointing out!!

ViditChitkara:
> Looks good but is this supposed to be there? Just checking!

Woah, I think that was added by mistake. Removed it.

jywarren:
Looks good!!!! Thank you!!!

PR #5566, opened 2019-04-21
Login subscription flow rectified

### LINKED ISSUE #4760: ability to go directly to `/subscribe/multiple/tag/tag1,tag2` and correctly enter and login
Follow-up to #4739, we should make sure you can go directly to `/subscribe/multiple/tag/tag1,tag2` and correctly enter and login, and get subscribed to those two tags. 

It should be working:

https://github.com/publiclab/plots2/blob/5cde763bbdc3bfb9b91c26d68389285050b5ba8c/app/controllers/subscription_controller.rb#L178-L179

https://github.com/publiclab/plots2/blob/5cde763bbdc3bfb9b91c26d68389285050b5ba8c/app/controllers/user_sessions_controller.rb#L142

but it isn't on unstable. Let's debug this, potentially with an integration test?

SidharthBansal:
Fixes #4760 
Problem was the redirection to the 
image
after the user is logged in. Also, presence of the session[:return_to] and params[:return_to] both in this case which is not handled by current controller code. Now everything works fine.
Now it is solved as shown in the screenshot.
image


app/controllers/user_sessions_controller.rb
```diff
@@ -143,7 +143,7 @@ def handle_site_login_flow
                   session[:return_to] = nil
                   redirect_to return_to + hash_params
                 elsif params[:return_to]
-                  redirect_to params[:return_to] + hash_params
```
SidharthBansal: Update:
 By removing `hash_params` the routing is correct now. It is now moving to /subscriptions but tags are not subscribed. Checking how to fix it.

SidharthBansal: I receive flash[:notice] 'Successfully logged in.' now

SidharthBansal:
@jywarren I have added integration tests here as per your suggestion in the issue. 
@jywarren @gauravano please review and merge this.

SidharthBansal:
We can ignore codeclimate issues I think.

test/integration/login_flow_test.rb
```diff
@@ -104,4 +104,11 @@ class LoginFlowTest < ActionDispatch::IntegrationTest
     assert_not_nil request.env['omniauth.auth']
   end
 
+  test 'redirect to multiple subscription route if user is not logged in and tries to subscribe to multiple tags' do
```
jywarren: awesoooommmeeeeeee

SidharthBansal:
Oh thanks.

PR #5568, opened 2019-04-21
narrow content_followed_in_period to notes by default

jywarren:
Re: #4732

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -242,7 +242,7 @@ class User < ActiveRecord::Base
     Node.questions.where(status: 1, uid: id)
   end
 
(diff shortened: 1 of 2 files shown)


jywarren:
Good, we're filtering out one item! 

```
 FAIL["test_returns_nodes_created_in_given_period_of_time", #<Minitest::Reporters::Suite:0x0000558f8ab97980 @name="UserTest">, 150.46908899099992]
 test_returns_nodes_created_in_given_period_of_time#UserTest (150.47s)
        Expected: 5
          Actual: 4
        test/unit/user_test.rb:124:in `block in <class:UserTest>'
```

PR #5559, opened 2019-04-20
Preserve `return_to` param when switching from /login to /signup

SidharthBansal:
Fixes #4893 (<=== Add issue number here)

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [ ] PR is descriptively titled 📑 and links the original issue above 🔗
* [ ] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [ ] code is in uniquely-named feature branch and has no merge conflicts 📁
* [ ] screenshots/GIFs are attached 📎 in case of UI updation
* [ ] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software 

Thanks!

PR #5591, opened 2019-04-23
System test for Preserve return_to param when switching from /login to /signup

### LINKED ISSUE #5585: System test for Preserve `return_to` param when switching from /login to /signup
Write System test for Preserve `return_to` param when switching from /login to /signup

SidharthBansal:
Fixes #5585

### DIFF
--- a/test/system/users_test.rb
+++ b/test/system/users_test.rb
@@ -1,0 +1,9 @@
+require "application_system_test_case"
+
+class UsersTest < ApplicationSystemTestCase
+  # test "visiting the index" do
+  #   visit users_url
+  #
+  #   assert_selector "h1", text: "Users"
+  # end
+end
--- a/test/system/users_test.rb
+++ b/test/system/users_test.rb
@@ -9,6 +9,8 @@ class UsersTest < ApplicationSystemTestCase
   test "preserve return_to while switching from login to signup" do
     visit '/login?return_to=/subscribe/multiple/tag/tag1,tag2'
     click_on "sign up"
-    assert_redirected_to '/signup?return_to=/subscribe/multiple/tag/tag1,tag2'
+    assert_text "Create a password"
+    assert_text "Confirmation"
+    assert_text "Bio"
   end
 end
--- a/test/system/users_test.rb
+++ b/test/system/users_test.rb
@@ -4,6 +4,7 @@ class UsersTest < ApplicationSystemTestCase
   test "preserve return_to while switching from login to signup" do
     visit '/login?return_to=/subscribe/multiple/tag/tag1,tag2'
     click_on "sign up"
+    assert_redirected_to '/signup?return_to=/subscribe/multiple/tag/tag1,tag2'
     assert_text "Create a password"
     assert_text "Confirmation"
     assert_text "Bio"
--- a/test/system/users_test.rb
+++ b/test/system/users_test.rb
@@ -1,11 +1,6 @@
 require "application_system_test_case"
 
 class UsersTest < ApplicationSystemTestCase
-  # test "visiting the index" do
-  #   visit users_url
-  #
-  #   assert_selector "h1", text: "Users"
-  # end
   test "preserve return_to while switching from login to signup" do
     visit '/login?return_to=/subscribe/multiple/tag/tag1,tag2'
     click_on "sign up"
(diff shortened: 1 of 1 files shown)


SidharthBansal:
@jywarren another PR completed. @gauravano @sashadev-sky @jywarren please review and suggest changes if any.
Thanks

test/system/users_test.rb
```diff
@@ -0,0 +1,16 @@
+require "application_system_test_case"
+
+class UsersTest < ApplicationSystemTestCase
+  # test "visiting the index" do
+  #   visit users_url
+  #
+  #   assert_selector "h1", text: "Users"
+  # end
```
siaw23: Lets' remove commented out code.

test/system/users_test.rb
```diff
@@ -0,0 +1,16 @@
...
+  test "preserve return_to while switching from login to signup" do
+    visit '/login?return_to=/subscribe/multiple/tag/tag1,tag2'
+    click_on "sign up"
+    assert_text "Create a password"
+    assert_text "Confirmation"
+    assert_text "Bio"
```
siaw23: I don't think this test does what you mean or what I understand from the title. You want to test that a user is returned to a URL after they try to login and it appears they don't have an account, so after they sign up, they are returned to the URL they were trying to access?

SidharthBansal: I think you have not understood the functionality properly. This is the continued work of #5559.
The test ensures that when the user clicks on the ``sign up `` button on the login page with url `/login?return_to=x` ,then they get redirected to the signup page with the url '/signup?return_to=x`.

I am sorry for the inconvenience.

SidharthBansal: Thanks for the input. I have modified the tests according to your suggestions.

test/system/users_test.rb
```diff
@@ -0,0 +1,12 @@
+require "application_system_test_case"
+
+class UsersTest < ApplicationSystemTestCase
+  test "preserve return_to while switching from login to signup" do
+    visit '/login?return_to=/subscribe/multiple/tag/tag1,tag2'
+    click_on "sign up"
+    assert_redirected_to '/signup?return_to=/subscribe/multiple/tag/tag1,tag2'
```
SidharthBansal: @siaw23 @gauravano @jywarren can you please help me why this line shows error?

SidharthBansal: line 7 i mean

siaw23: Line 7 is an assertion. Which error do you mean @SidharthBansal? Can you please paste it here?

SidharthBansal:
Please see the travis
It is failing

test/system/users_test.rb
siaw23: Right. So you don't do assertion here like this. When you `click_on "sign up"` you have to assert the presence of some text on that page after clicking. So try manually clicking on "sign up" then assert the presence of some text there like so `assert_text 'Some Text'` where "Some Text" is some unique text that you can only see after clicking "sign up". Let's try that if that works.

SidharthBansal:
Hi, I have already written statements that confirm the presence of text
which appears on the signup page
I need to confirm that I am routed to
/signup?return_to=subscribe/multiple/tag/tag1,tag2 page
How shall I do it?
Any help much appreciated.

siaw23:
Now I see the assertion `assert_text "Create a password`. That's enough. That assertion passing is your confirmation that you've been redirected. To test you're redirected again back to the initial page you visited, something has to happen right? So on the "creating password" page another button has to be clicked, THEN you can make a new text assertion. Don't use `assert_redirected_to` here. Click buttons and assert the presence of texts.

SidharthBansal:
Let me try this. Thanks

test/system/users_test.rb
SidharthBansal: Test will fail as we don't have access to Recaptcha feature in the testing environment.
So, we will not be able to submit the form by clicking on the  "Sign Up" button.

SidharthBansal: Can we try unit test for this? We just want three steps.
1. User clicks on '/login?return_to=X'
2. User clicks on "Signup"
3. User is now at '/signup?return_to=X'
But I don't know how to do statement 2 in unit tests.

jywarren:
Noting that we may have some similarities in debugging in https://github.com/publiclab/plots2/pull/5605 -- 🙌 

System tests are hard to get used to but they're coming along!!!

SidharthBansal:
back on it after a long time

PR #5620, opened 2019-04-25
Adding requireLogin class to subscribe button popover

### LINKED ISSUE #5614: Pop up subscription modal /after/ opening multi-tag subscription modal
@SidharthBansal just looking to wrap this up; am I right that we can remove the login modal part of this segment:

https://github.com/publiclab/plots2/blob/7744d91e9bf03544da2f38e7e66056b6f20162e0/app/views/tag/_subscribe_button.html.erb#L12

and add the `requireLogin` class to this line?

https://github.com/publiclab/plots2/blob/7744d91e9bf03544da2f38e7e66056b6f20162e0/app/views/tag/_tags_popover.html.erb#L11

If so, let's do this, and then write a system test to protect it! Thank you!!!

IshaGupta18:
Fixes #5614 

requireLogin class added to Subscribe button for multiple tag subscription.

(when the user isn't logged in, on clicking on the subscribe+ button, requireLogin class is activated)
image

(when the user is logged in, normal flow resumes)

image


IshaGupta18:
@jywarren @SidharthBansal kindly review this PR and let me know what else can be done here! Thank you!

jainaman224:
Hey @IshaGupta18,
You can use `git rebase master` command to get rid of merge commit.

IshaGupta18:
Oh okay, will do from next time on if it's okay for now! Thanks!

IshaGupta18:
Is it okay for now?

jywarren:
It should be fine, as long as it doesn't say there are conflicts! Thanks
this looks amazing!

jywarren:
This looks good, but if there is no current user maybe we should check them
all? And, on what line is the modal finally triggered? Thanks!!!

app/views/tag/_tags_popover.html.erb
```diff
@@ -4,7 +4,7 @@
     <% tags.each do |tag| %>
     <div class="form-check">
       <label class="form-check-label">
-        <input type="checkbox" id="<%= tag.id %>" name="tagnames[]" value="<%= tag.name %>" <% if current_user.following(tag.name) %> checked <% end %> ><span class="<% if tag.name.include?(':') %> badge badge-primary <% else %> badge badge-secondary <% end %>" style="margin-left: 3px; margin-right: 10px; display: inline-block;"> <%= tag.name %></span>
+        <input type="checkbox" id="<%= tag.id %>" name="tagnames[]" value="<%= tag.name %>" <% if current_user && current_user.following(tag.name) %> checked <% end %> ><span class="<% if tag.name.include?(':') %> badge badge-primary <% else %> badge badge-secondary <% end %>" style="margin-left: 3px; margin-right: 10px; display: inline-block;"> <%= tag.name %></span>
```
IshaGupta18: I have checked if there is no current user here.

app/views/tag/_subscribe_button.html.erb
```diff
@@ -1,19 +1,13 @@
 <!-- This button is used to display tags in a popover and subscribe to multiple tags using a single button -->
-<% if current_user %>
-  <a id="popover_button" class="btn btn-outline-secondary btn-sm sub" >Subscribe 
-    <i class="fa fa-plus"></i> 
-  </a>
+<a id="popover_button" class="btn btn-default btn-xs sub requireLogin" > Subscribe <i class="fa fa-plus"></i> </a>
```
IshaGupta18: If the user is not logged in, this line will trigger the modal

IshaGupta18: If the user is not logged in, the modal is triggered here.

app/views/tag/_tags_popover.html.erb
IshaGupta18: I have checked if there is no current user here.

jywarren:
Oh, I meant that maybe all tag checkboxes should be checked if there is no current user?

IshaGupta18:
> Oh, I meant that maybe all tag checkboxes should be checked if there is no current user?

Actually what is happening here is: When the user is not logged in, he is shown the login modal and not the multiple tag subscription popover. Is that okay?

SidharthBansal:
Great work. Can you please write tests to ensure that this stuff doesn't break?
Thanks

IshaGupta18:
Yes I will incorporate that, I just want @jywarren to clarify if this is done right or not, just to make sure if we are on the same page.

SidharthBansal:
Cool

IshaGupta18:
@jywarren are we on the same page here? Thanks!

IshaGupta18:
Hey @jywarren @gauravano what do you think of this?

jywarren:
I am SO SORRY I missed this notification!!!! It's very late here now, but I
will address this first thing in the morning. Please accept my apologies
for being so slow on this!!!

IshaGupta18:
Hey, it's not a problem at all! Please don't apologize so much for such a trivial thing. You can address it whenever it's suitable for you! Thanks a lot!

app/views/tag/_subscribe_button.html.erb
jywarren: OK, so, this is what I'd like to modify - if there's no logged-in user, we should still wait until after the popover is open to prompt a login modal, and we should have all the checkboxes checked (we can later modify this so it's only the top 3, maybe? But... later!)

That way, if someone comes to the site for the first time, clicks Subscribe, they can in as few steps as possible, get registered AND subscribed to the most relevant tags on their starting page, rather than having to register and then come back to subscribe. Does this make sense? 

So, a non-logged-in user would still see:

image


And they'd be prompted to log in only when pressing the blue `Subscribe` button. This would establish a `return_to=/subscribe/multiple/tag/arduino,games`, in the login/signup modal. Make sense?

jywarren:
Thanks Isha for understanding! I'm extra apologetic because this is going to be such an improvement for newcomers, I'm eager to see it implemented!

Also -- will any changes have to be made for the new Bootstrap 4 code that's now on the master branch? Perhaps a rebase will expose some incompatibilities? Here's the docs for v4: https://getbootstrap.com/docs/4.3/components/popovers/

Thanks, Isha! I left a clarification above.

jywarren:
And as to testing, if you'd like, we can make a system test similar to this one for posting a note: https://github.com/publiclab/plots2/blob/master/test/system/post_test.rb

You can use this CapyCorder chrome extension to "record" a set of interactions and it produces a system test in Ruby from the browser! https://github.com/polarblau/capycorder/

It's not perfect but it's a good starting point for developing a system test for this interaction. Give it a try!

app/views/tag/_subscribe_button.html.erb
IshaGupta18: Oh okay, I think I get you, however it might be a little problematic since `requireLogin` was not working on this, but lemme take a stab at it again and get back to you! Thanks for explaining so clearly!

IshaGupta18:
No problem at all! I also want this to be implemented because multi-tag subscription is a creative feature created by us, and it should be worked on to make it as fruitful as possible! I would definitely have a look at the testing procedure. Thanks a ton again for such clarity!

app/views/tag/_subscribe_button.html.erb
jywarren: Ah yes - so, I think the reason it wasn't working is that the popover is constructed /after/ this code is run, which is the code which makes `requireLogin` work:

https://github.com/publiclab/plots2/blob/e190eae1ce7bf215b99b6efe7f828e17deb3213e/app/views/user_sessions/_form.html.erb#L74-L81

So, we have to do something similar, re-running it, or re-running it with a narrower scope just for the popover, you know?

IshaGupta18: Yes, I think I get you somewhat, but I'll have to try it hands on to exactly get it!

IshaGupta18: Hey, I am facing a bit of a problem in understanding the flow here, could you please help me out @jywarren, to identify what is going on and what should be done so that the `requireLogin`class activates the modal after the popover is rendered. Should we add the class dynamically? Or should we build it from scratch for this case?

jywarren:
One thing you could do is to extract the function which makes requirelogin
work - name it something like setupRequireLogin, and make it accessible to
the global page scope. Then the place we originally connect it up (that I
linked to) can reference it by name like
$('.requireLogin').click(setupRequireLogin);
but, you can also use it yourself in your popover function, like
$('.multiSubscribeBtn').click(setupRequireLogin);

What do you think of this?

IshaGupta18:
I'll have to try it out to see how it goes. Thanks a ton!

IshaGupta18:
Okay so I rewrote the function to trigger the requireLogin's functioning and it works. Although, there is a problem:
image

I think this is because we are accessing the `href` parameter of the submit button of the checkbox form, which is probably null or undefined. What do you think?

jywarren:
Ah, indeed. It's null because we assemble the request in JavaScript, right? I suppose we should just look at what the `requireLogin` function does, and do it manually in this special case, instead of relying on the class and the `href`; what do you think?

So, effectively, after we build the popover, we connect this function to the click event for the button:

```js
$('submitMultipleSubscribePopover').click(function() {
  require_login_and_redirect_to("//subscribe/multiple/tag/" + tags.join(',')); 
});
```

How about that?

IshaGupta18:
Yes, sounds good I think!

jywarren:
Hi @IshaGupta18 ! Not a rush at all, but wondering if this is going well? Thank you!

IshaGupta18:
Hey @jywarren I am really sorry I kinda forgot about this because of the work going on graph-plotter. I will get back at this really soon. Thanks for your patience!

jywarren:
no problem, thanks!

jywarren:
Hi Isha! Did you want to give this another try? I'm trying to reduce the number of open PRs! Thanks!

IshaGupta18:
Actually this issue has been pushed on so many times now, I did take
another try on it but it didn't work so well. Maybe someone else would like
to give it a shot? I would definitely work with them on it. What do you
think?

jywarren:
No problem at all! I think I've got it in https://github.com/publiclab/plots2/pull/5928 -- thanks for all your work on this! It's such a subtle issue! And we can write a system test for it now :-)

PR #5654, opened 2019-05-03
remove authlogic startup from system test

jywarren:
this should.... work?


jywarren:
Nope, this didn't work! Because apparently now we aren't logged in: 

```
        expected to find visible css ".row.header > h1" with text "Dashboard" but there were no matches. Also found "Community research", which matched the selector but not all filters. 
        test/system/dashboard_test.rb:19:in `block in <class:DashboardTest>'
```

PR #5660, opened 2019-05-05
Issue 5259: Switched to logged_in_as(roles) method to simplify authentication logic

### LINKED ISSUE #5259: Switch to logged_in_as(roles) to simplify auth logic!
the `User` model could use some simpler helpers, which would dramatically simplify some template code for showing things only to moderators or admins, like this:

https://github.com/publiclab/plots2/blob/aba49c4378fd94ab3f37de1d398243b75cf90365/app/views/users/list.html.erb#L8-L11

Instead, let's make a user model method like:

```ruby
user.is_role(:admin)
```

Then that code could become:

```ruby
if user.is_role(:admin, :moderator)
```

But we need to be able to specify if it's AND or OR -- any ideas?

Maybe:

```ruby
user.is_one_of_roles(:admin, :moderator)
```

We could also simplify by making an application_helper method like:

```ruby
user_has_role(:admin, :moderator)
```

where we also check if there is a `current_user`, further simplifying the code!

We'd love some help with this, and ideally the new methods would be tested as well! Thanks!

SrinandanPai:
Fixes #5259 

Reviewers
----

Please review.

- @publiclab/reviewers

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -7,7 +7,7 @@ class AdminController < ApplicationController
   def promote_admin
     @user = User.find params[:id]
     unless @user.nil?
-      if current_user && current_user.role == 'admin'
+      if logged_in_as(['admin'])
         @user.role = 'admin'
         @user.save
         flash[:notice] = "User '<a href='/profile/" + @user.username + "'>" + @user.username + "</a>' is now an admin."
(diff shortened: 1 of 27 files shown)


SrinandanPai:
@publiclab/reviewers - this is my first pull request, I'd love any feedback you could give me. Thanks!

jywarren:
This is super!!! Thanks so much!!!

PR #5715, opened 2019-05-13
sort related tags by co-occurence

### LINKED ISSUE #5714: Tag counts aren't updating - and reworking Related Tags query
I think we're prioritizing related tags on the `count` field, but don't actually

https://github.com/publiclab/plots2/blob/fe2ffd6d71b9c5fc7c4ff09bb9d770ce30f4fe78/app/models/tag.rb#L362-L375

We have a `run_count` method here, but it's never called:

https://github.com/publiclab/plots2/blob/fe2ffd6d71b9c5fc7c4ff09bb9d770ce30f4fe78/app/models/tag.rb#L36-L39

Let's do 2 things 

- [ ] fix `run_count` by running it from time to time, but also:
- [ ] tally/histogram `nids` in the first query above, which will allow us to sort them by how many times they co-occurred with the original tag, and fetch only the top 5. 

Also note: previously we were trying to collect the top 5 by overall high tag usage, but we were failing due to `run_count` not being updated. The result was that `related_tags` was just 5 randomly selected co-occurring tags, but not the 5 most commonly co-occurring.

Now, we'll be measuring by how OFTEN they co-occur.

We should be able to run:

```
data = ["a", "b", "a", "b", "a", "c", "t"]
data.group_by{ |v| v }.map{ |k, v| [k, v.size] }
# => [["a", 3], ["b", 2], ["c", 1], ["t", 1]]
```

jywarren:
Fixes #5714

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -367,7 +367,8 @@ class Tag < ApplicationRecord
 
       # sort them by how often they co-occur:
       nids = nids.group_by{ |v| v }.map{ |k, v| [k, v.size] }
-      nids = nids.collect(&:first)[0..4] # take top 5
+      nids = nids.collect(&:first)[0..4]
+                 .collect(&:nid) # take top 5
 
       Tag.joins(:node_tag)
          .where(NodeTag.table_name => { nid: nids })
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -365,6 +365,10 @@ class Tag < ApplicationRecord
                      .where(Tag.table_name => { name: tag_name })
                      .select(:nid)
 
+      # sort them by how often they co-occur:
+      nids = nids.group_by{ |v| v }.map{ |k, v| [k, v.size] }
+      nids = nids.collect(&:first)[0..4] # take top 5
+
       Tag.joins(:node_tag)
          .where(NodeTag.table_name => { nid: nids })
          .where.not(name: tag_name)


jywarren:
```

ERROR["test_graph_data_for_cytoscape", #<Minitest::Reporters::Suite:0x00007fec83938d80 @name="TagTest">, 363.121115693]
 test_graph_data_for_cytoscape#TagTest (363.12s)
TypeError:         TypeError: can't quote CompositePrimaryKeys::CompositeKeys
            app/models/tag.rb:398:in `block (2 levels) in graph_data'
            app/models/tag.rb:397:in `each'
            app/models/tag.rb:397:in `block in graph_data'
            app/models/tag.rb:383:in `graph_data'
            test/unit/tag_test.rb:204:in `block in <class:TagTest>'
ERROR["test_related_tags", #<Minitest::Reporters::Suite:0x00007fec83926748 @name="TagTest">, 363.13011305600014]
 test_related_tags#TagTest (363.13s)
TypeError:         TypeError: can't quote CompositePrimaryKeys::CompositeKeys
            test/unit/tag_test.rb:26:in `block in <class:TagTest>'
```

jywarren:
Just a heads up @skilfullycurled ! This identified an error in how we calculated importance of related tags, and we also changed how it's calculated (apart from fixing the original bug). This will affect the tag graph and related tags across the site.

We'll see this publish shortly at https://stable.publiclab.org and expect 🤞 this to publish to the live site tonight.

PR #5727, opened 2019-05-15
Remove spam comments

### LINKED ISSUE #5726: Comment stats counting spam comments?
I don't believe comment stats count spam, but looking at this line in the range action I'm not sure, @cesswairimu  maybe we should take a look? 

https://github.com/publiclab/plots2/blob/b8011954823c2e76cd53e2113353c5dcd5bd46cc/app/controllers/stats_controller.rb#L28

Also this line?

https://github.com/publiclab/plots2/blob/b8011954823c2e76cd53e2113353c5dcd5bd46cc/app/models/comment.rb#L44-L47

Should we put status=1 in all these lines?

cesswairimu:
Fixes #5726 

- Add status filter and change to size from count
- Size is a little faster query than count

### DIFF
--- a/app/controllers/stats_controller.rb
+++ b/app/controllers/stats_controller.rb
@@ -17,19 +17,19 @@ class StatsController < ApplicationController
     @end = fin
     @notes = Node.published.select(%i(created type))
       .where(type: 'note', created: @start.to_i..@end.to_i)
-      .count(:all)
-    @wikis = Revision.select(:timestamp)
+      .size
+    @wikis = Revision.published.select(:timestamp)
       .where(timestamp: @start.to_i..@end.to_i)
-      .count - @notes # because notes each have one revision
+      .size - @notes # because notes each have one revision
     @people = User.where(created_at: @start..@end).where(status: 1)
-      .count
+      .size
     @answers = Answer.where(created_at: @start..@end)
-      .count
-    @comments = Comment.select(:timestamp)
-      .where(timestamp: @start.to_i..@end.to_i)
-      .count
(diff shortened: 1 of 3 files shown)


cesswairimu:
Tested in unstable before and after the fix
Screenshot from 2019-05-15 21-26-50

new-comm

cesswairimu:
Highest before 140, after 135.

Screenshot from 2019-05-15 21-26-50


new graph

cesswairimu:
@jywarren tested in unstable...there is a little change :point_up: but maybe there aint as much spams in unstable what do you think?

jywarren:
It's possible! Do you think this is ready for merge? We can flush the
caches and it'll regenerate.

cesswairimu:
Yes @jywarren its ready to merge

jywarren:
Heads up to @ebarry and @skilfullycurled about this modification!

jywarren:
Thanks Cess, this is great!!! 🎉

jywarren:
And just to be clear, the change is in the height of the peaks, not the activity before/after Nov 2016 -- that's just because the Stable server has a db export from that moment in time. 📈

cesswairimu:
Aha, noted. Thanks Jeff

jywarren:
Yeah, i did a double take! But for example on the highest peak the graph
label shows 140 before, 136 after (i think) so that's like 4 spam comments
that had been counted. Not so bad.

skilfullycurled:
Ahhhhhh...good to know about unstable!

PR #5738, opened 2019-05-18
Translation prompt improvement

### LINKED ISSUE #5651: Translation prompt improvement
Currently, we show Translation prompts on /subscriptions page, some on /dashboard and /blog page. 

Conditions for translation prompt:
1. translated_string.length > 3 - If translated string has length greater than 3 
2. current_user &.has_tag('translation-helper') - User has tag `translation-helper` on their profile
3. !translated_string.include?("translation missing") - There shouldn't be "translation missing" in the translated_string. This effectively means we show translation prompt only in case where translation is present.

The condition 3 is buggy or I would say totally wrong. Why? Because, previously we were running only 3 langs and in those while testing I got a span with translation missing, something like `"<span class=\"translation_missing\" title=\"translation missing: ru.home.subscriptions.notes, fallback: false\">Notes</span>"` but this would be returned when the translation key is not found in any locale yes, not in en.yml too but in all other cases we would get something i.e., either a translated value in the Russian. japan, etc or a English string because we fallback to English.

Let's say, we want to show prompt only in case a translation is missing in the language we are browsing publiclab.org 

then, it would be something like:

Subscription page in English:

eng_as_lang
 
Subscription page in Russian:

russian_as_a_lang

So, you can see so many prompts because we have less Russian translations. So, what should be the conditions in which prompt should be visible?

Bugs to solve: currently, we directly wrap the returned translation in the span(prompt is made using span element), but we don't get string always in translated value some times it is code too and in those cases, :boom: :point_down: :sweat_smile: 

bug

Let's collect ideas and issues about improving translation prompts here. Thanks!

gauravano:
Fixes #5651 #5749 

Thanks!

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -40,12 +40,12 @@ module ApplicationHelper
   def emoji_info
     emoji_names = ["thumbs-up", "thumbs-down", "laugh", "hooray", "confused", "heart"]
     emoji_image_map = {
-        "thumbs-up" => "https://github.githubassets.com/images/icons/emoji/unicode/1f44d.png",
-        "thumbs-down" => "https://github.githubassets.com/images/icons/emoji/unicode/1f44e.png",
-        "laugh" => "https://github.githubassets.com/images/icons/emoji/unicode/1f604.png",
-        "hooray" => "https://github.githubassets.com/images/icons/emoji/unicode/1f389.png",
-        "confused" => "https://github.githubassets.com/images/icons/emoji/unicode/1f615.png",
-        "heart" => "https://github.githubassets.com/images/icons/emoji/unicode/2764.png"
+      "thumbs-up" => "https://github.githubassets.com/images/icons/emoji/unicode/1f44d.png",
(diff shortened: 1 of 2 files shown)


gauravano:
If code is returned from translation, prompt not shown as per https://github.com/publiclab/plots2/issues/5651#issuecomment-488789810

Screenshot from 2019-05-19 02-50-19

Page in Spanish -- prompts shown as translations missing
Screenshot from 2019-05-19 02-51-47

Page in English -- Prompts not shown as translation are there 
Screenshot from 2019-05-19 02-51-33

jywarren:
Aha you're way ahead of me! Awesome. Will try to get the test to pass.

gauravano:
Hey @jywarren, you just restarted the test and it passed? But, why it errored in initial builds? :confused:

jywarren:
This is related to the issues in #5683 -- but should be good now? Let's merge!

gauravano:
Done!!

gauravano:
@jywarren let me know when you push it to production as after that we can add more languages to the footer :smiley:

PR #5648, opened 2019-05-02
Arranging tags in descending order & Adding link to tags

IshaGupta18:
Fixes #5260 

Arranging the stats in desc order for the most used to appear first and adding links to tags:

image

On clicking on a tag, redirected to tag's page:

image

PR #5769, opened 2019-05-22
Alphabetizing the tags on stats subscription page

### LINKED ISSUE #5260: Refinements to new tag subscription statistics page
Following up on #5224, we can tackle a few more things in this follow-up issue, now that @GettyOrawo has done a great job getting this page created!

* [ ] let's sort this DESCENDING or `desc` so we start with the most-used tags
* [ ] let's hide anything more than 25 results and show a "View all" link to show them?
* [ ] let's make the tag names into links that go to `/tag/TAGNAME`

and we can do more too! This is a great start!

IshaGupta18:
Fixes #5260 

Moving from #5648 

@jywarren after alphabetizing the tags, this is how it looks:

image

### DIFF
--- a/app/views/stats/subscriptions.html.erb
+++ b/app/views/stats/subscriptions.html.erb
@@ -17,6 +17,7 @@
       <%= range * 10 %> - <%= range * 10 + 9 %>
     </td>
     <td>
+      <% tags = tags.sort %>
       <% tags.each do |tag| %>
       <a class="badge badge-primary" href="/tag/<%= tag[0] %>">
         <%= tag[0]%> 


IshaGupta18:
@jywarren is this a good way to reduce the ambiguity with the tags in the rows?

jywarren:
Oh I love it, maybe ask @ebarry in the chatroom? She uses it most!

IshaGupta18:
The sorting technique?

ebarry:
This will be a big help @IshaGupta18. 
Definitely +1 to merge. 

I have a separate insight on the category of "0-9" tags: 
- first, there are no tags that exist that are used 0 times. Because of this the range should be listed as 1-9. 
- secondly, i advise to break "used 1 time" into its own category because it is so large. I personally do a lot of tag standardization by working on tags that are used one time. 

thanks for your consideration!

ebarry:
When you have a moment, please double-check that the tag counting is bring done properly. See in this gif below, that `barataria` shows up with a count of 1, but there are more than 16 notes tagged `barataria`

tag-count-error

IshaGupta18:
Sure! Thanks a lot! We can definitely create a "used 1 time" category. For the counting, I will have to see into the model and controller code, because sometimes this does happen, in the count stuff. Thank you again!

jywarren:
Thanks all, we'd best open a new issue for the "barataria" count issue, so
it doesn't get forgotten. I would suspect that the tag tallying in the
tag.rb 'counting' code is at fault - it's supposed to maintain a field with
the count, but perhaps it's not re-counting when it's supposed to be?

jywarren:
Would we like to merge this and break out new issues for the rest? Thanks!

app/views/stats/subscriptions.html.erb
```diff
@@ -17,6 +17,7 @@
       <%= range * 10 %> - <%= range * 10 + 9 %>
     </td>
     <td>
+      <% tags = tags.sort %>
```
siaw23: Hi @IshaGupta18, nice job but maybe you could do `<% tags.sort.each do |tag| %>` instead? So we can remove this redundant line? What do you think?

IshaGupta18:
Sure definitely! I'll push these changes along with the one one @ebarry has
asked for! Thank you!

siaw23:
Also on line 14 there's a `reverse_each`, maybe we could check if we need that. If we don't then you can change that from `reverse_each` to just `each` and perform the actual sorting in the controller as it's in the controller that this should be done in fact and not in the view. @IshaGupta18

jywarren:
I'm going to merge this and reference the new ideas in the original issue!

SidharthBansal:
@jywarren I think you forget to merge. You commented that you are going to merge it

jywarren:
Indeed. Doing so!!!

PR #5801, opened 2019-05-28
Tags page (/tags) powertags visible only to admins (#5258)

### LINKED ISSUE #5258: Show powertags to admins only on /tags page
We want to display powertags to only admins and not normal users. So, this can be achieved by filtering out tags which consist `:` in them.

here's the code for the /tags page - https://github.com/publiclab/plots2/blob/master/app/controllers/tag_controller.rb#L5-L67

Refer #179 for more details.

simonastefan:
Fixes #5258 

Thanks!

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -8,7 +8,7 @@ class TagController < ApplicationController
     @title = I18n.t('tag_controller.tags')
     @paginated = true
     @order_type = params[:order] == "desc" ? "asc" : "desc"
-    powertag_clause = current_user&.admin? ? '' : ['name NOT LIKE ?', '%:%']
+    powertag_clause = params[:powertags] == 'true' ? '' : ['name NOT LIKE ?', '%:%']
 
     if params[:search]
       keyword = params[:search]
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -8,6 +8,7 @@ class TagController < ApplicationController
     @title = I18n.t('tag_controller.tags')
     @paginated = true
     @order_type = params[:order] == "desc" ? "asc" : "desc"
+    powertag_clause = current_user&.admin? ? '' : ['name NOT LIKE ?', '%:%']
 
     if params[:search]
       keyword = params[:search]
@@ -16,6 +17,7 @@ class TagController < ApplicationController
         .where('node.status = ?', 1)
         .where('community_tags.date > ?', (DateTime.now - 1.month).to_i)
         .where("name LIKE :keyword", keyword: "%#{keyword}%")
+        .where(powertag_clause)
         .group(:name)
         .order(order_string)
         .paginate(page: params[:page], per_page: 24)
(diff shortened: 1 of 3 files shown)


simonastefan:
I need a bit of help. I don't understand why [Travis fails](https://travis-ci.org/publiclab/plots2/jobs/538337203) in `test/system/dashboard_test.rb`. Locally, I put a breakpoint in the code I modified and while the tag_controller tests stop, this test doesn't. Also, on my machine I got it to fail only twice. System tests pass the majority of time. Makes it difficult to find the cause. If anyone could give some pointers it would be great.

Regarding CodeClimate, it might affect legacy code that doesn't seem to be well covered by tests. Should I fix them here or maybe a different PR?

Thank you!

test/functional/tag_controller_test.rb
```diff
@@ -685,4 +685,23 @@ def setup
     get :graph
     assert_response :success
   end
+
+  test 'index should not have powertags for non admin' do
```
jywarren: Love this test! 👍

jywarren:
Hi! I think it's not your fault, but some oddities with our new system tests; see how one of the test runs ran fine, but not the other:

image

I've also bypassed the recommendations from CodeClimate, and this looks perfect; ready for a merge once we get that test run to pass!

Thanks a million, i know @ebarry will be really happy with this.

app/controllers/tag_controller.rb
```diff
@@ -8,6 +8,7 @@ def index
     @title = I18n.t('tag_controller.tags')
     @paginated = true
     @order_type = params[:order] == "desc" ? "asc" : "desc"
+    powertag_clause = current_user&.admin? ? '' : ['name NOT LIKE ?', '%:%']
```
jywarren: If you want to take one more step, we could make this be dependent on a parameter instead, like `params[:powertags]` -- that way, anyone could use it, but only by using the URL `/tags?powertags=true` -- what do you think?

simonastefan: No problem, switched from a user role check to a param check. Let me know if I should have kept the admin one as well.
As expected, pagination is ok but `powertags=true` is lost if tags sort criteria changes and has to be put into the URL again. Should I fix it in [tag view](https://github.com/publiclab/plots2/blob/8ca4b786fdc895e6c2c3e06529a394c38b959a8d/app/views/tag/index.html.erb#L23-L25) or is a UI powertags button planned/preferred? 
Thanks!

gauravano: Hi @simonastefan, I think it would be good to fix in the Tag view only as we don't want to have separate buttons for power tags. @jywarren what do you think? Thanks!

simonastefan: Updated with view fix

simonastefan:
Sorry for the delay! Instead of admin role used URL param approach (see resolved conv); ready for review @publiclab/reviewers . Thanks!

jywarren:
Looks great!!

jywarren:
Thanks so much, @ebarry will be pleased!!!

Issue #5733, opened 2019-05-17
UI Improvements PLanning Issue : Questions Page


CleverFool77:
cc : @gautamig54 

![PublicLab_mockups](https://user-images.githubusercontent.com/26685258/57947963-f8ef4480-78fd-11e9-9724-a30711ca27ae.jpg)

#### Checklists

- [x] Create shadow questions page @CleverFool77 
- [x] first design block with some relevant information @CleverFool77 
- [x] card of text area for questions with buttons and post options @gautamig54 
- [x] featured questions cards   @CleverFool77 
- [ ] Show recent questions on some topic - use same card as used in tags section @gautamig54 

Viewable at:

http://stable.publiclab.org/shadow_question

http://publiclab.org/shadow_question

CleverFool77:
Hi @jywarren @gautamig54 
Is division of checklists for questions section correct ?

CleverFool77:
Hi @gautamig54 I'll start off with assignment.

gauravano:
Great start here!! I can see some planning issues for your project. Is there any central planning issue where links to all the sub-planning issues can be found or we can discuss general project direction, goal priority, etc? Thanks!

gautamig54:
@CleverFool77 division of work is perfect. Let's get started on with it!

gautamig54:
@gauravano We can refer to #4922 as central planning issue.
Should I reopen it or make a new one?
Thanks!

gauravano:
please open a new one. You can copy the goals from #4922 . Thanks!

CleverFool77:
Hi @jywarren @gauravano 
As most of the portions in Design issues are dependent. So I'm taking two issues sideways. As some parts or portions of some issues are dependent on each other and it might make us to stop for a while as we are unable to move further. 
I just opened one PR for beginning the work for People section #5740 . And till it would get reviewed and merged. I wanted to begin the work in Questions Page.

So mainly, I wanted to ask that What's the purpose of image and info in First section. 
![Screenshot from 2019-05-19 18-47-05](https://user-images.githubusercontent.com/26685258/57982700-83eb4e80-7a66-11e9-88cd-fab6939087bc.png)

So That I can start off with work.
Thanks !!!

CleverFool77:
HI @jywarren @gauravano 
I wanted to ask that Is first section of questions page not available in plots2 code ?

[...]

PR #5861, opened 2019-06-08
Add featured question section to shadow question page

CleverFool77:
Fixes https://github.com/publiclab/plots2/issues/5733 part

Screenshot from 2019-06-08 22-57-13


CleverFool77:
Hi @jywarren @gautamig54 Just added the featured question section.
Thanks !!

app/controllers/questions_controller.rb
```diff
@@ -27,6 +27,15 @@ def index
       .paginate(page: params[:page], per_page: 24)
   end
 
+  def index_shadow
+    @title = 'Questions and Answers'
+    set_sidebar
```
jywarren: Is this line maybe not needed?

CleverFool77: Yes. I'll fix it.

CleverFool77:
For now, I have followed the same method for shadow pages as earlier controllers for real pages.

CleverFool77:
I tweaked frontend a bit and different cells being called from `/questions` and `/dashboard` were changed as per the info needed.

gautamig54:
Great work @CleverFool77 !!
It will be good if you can decrease the font size of the question in the cards as it will look absurd in case of long questions. 
Also, I think we need to add ellipsis here as well to avoid overflow of questions.

app/controllers/questions_controller.rb
CleverFool77: Hi @jywarren Did you mean we don't need index_shadow in controller ?
Thanks !!

CleverFool77:
> Great work @CleverFool77 !!
> It will be good if you can decrease the font size of the question in the cards as it will look absurd in case of long questions.
> The heading and the cards are not aligned, can you do that as well.
> Also, I think we need to add ellipsis here as well to avoid overflow of questions.

Hi @gautamig54 I guess in design The question went to second line in card.

CleverFool77:
Screenshot from 2019-06-10 17-09-09

app/controllers/questions_controller.rb
CleverFool77: But here that would be the method doing whole stuff in this PR.

CleverFool77:
Hi @jywarren I just pushed the latest changes.
Thanks !!

app/controllers/questions_controller.rb
jywarren: just the `set_sidebar` part, actually, sorry for ambiguity!

jywarren:
Looking good, added clarification to my inline question... @gautamig54 can you approve this PR too and both ping me (or another maintainer) when it's ready? Great work!!!

And, just for quick reference, can you point me at the URL where we'll be able to test this on stable when the time comes? Thank you!!!

CleverFool77:
hi @jywarren @gautamig54  Just pushed the latest changes.
regarding the test, did you mean the main url `/question_shadow` ?

gautamig54:
@CleverFool77 Can add a screenshot with testing question to be greater that two lines? Say 15 words?

CleverFool77:
Hi @gautamig54
Its there in flask notice while Filling up the form for posting qiestion that when title is exceeded more than 8-9 words. It gives message that `way too long titile`.

CleverFool77:
Hi @gautamig54 what do you say about my thought above? 
Maybe we can forward the progress of this PR then.
Hi @jywarren 
Can we have this PR merged first. 
Regarding question size,
Do we need overflow fixing for the question size in frontend? Or a restriction in backend? 
If yes, I'll fix this in next PR.
If no, then I guess it's usually understood for question to be not too long? 
And importantly how many words does the questions usually consist of?
As If fix is needed, then I might work on size of question based on words.

CleverFool77:
Seems like good idea sagarpreet. So maybe truncating the title of question to two-three lines with ellipsis in end?

sagarpreet-chadha:
Thank you and yes sounds great!

gautamig54:
Restricting question length to just 8-9 words may be less. We should restrict it to atleast 20 words? I don't think 20 words is like a too long for a question. 
Also, adding ellipsis(...) for longer question is the best suitable solution.

CleverFool77:
Sure gautami. Lets have this merged first. Till then I'm looking over the matter of overflow.
Thanks !!

CleverFool77:
The screenshot feature looks super cool :open_mouth: :open_mouth: . Damnnnn :heart_eyes: !!!

jywarren:
Do you wanna add the questions page? https://github.com/publiclab/plots2/tree/master/test/system/screenshots_test.rb

CleverFool77:
Sure I'll add this for shadow route for now.

CleverFool77:
Hi @jywarren just pushed the latest changes with the addition of test screenshot for question and shadow route.
Thanks.

jywarren:
Oooooh very cool!!!

app/views/questions/_questions_shadow.html.erb
```diff
@@ -0,0 +1,84 @@
...
+      <% end %>
+
+      <div class="col-md-2">
+      </div>
+  </div>
+</div>
+
+<style>
+
+.card {
```
jywarren: Should we start to consolidate these styles and re-use them across cards? What do you think?

CleverFool77: Though the cards looks same but still there are many differences. The positioning and size of cards in different pages varies.

gautamig54:
This looks great @CleverFool77! Let's get this one merged and move ahed with the recent questions section.

app/views/questions/_questions_shadow.html.erb
jywarren: OK, I still am thinking perhaps we should make a planning item for moving all the re-usable card layouts to one template folder so they're all next to one another. They are hard to find! Like `/app/views/items` or `/app/views/cards` maybe?

jywarren: But doesn't have to be in this PR

CleverFool77: Sure. Should I do this in next PR ?

CleverFool77: Or Le'ts wait a bit for all the cards and then shift them altogether to `/app/views/cards`

jywarren:
OK, either option for /views/cards is fine, just want us to be thinking
ahead. Thanks!!!

Issue #5793, opened 2019-05-25
Add timestamps to tag_selection and node_tags tables


jywarren:
We need a new migration to add `created_at` and `updated_at` to these two tables; this will help solve https://github.com/publiclab/plots2/issues/4925, but also will help us add ordering to our pinned nodes (https://github.com/publiclab/plots2/issues/766), so that they are pinned in order -- first pinned is first displayed:

https://github.com/publiclab/plots2/blob/35e10a2eeabdf5afce7bb67704548435997dc58a/app/models/concerns/node_shared.rb#L415-L419

We'd love help with this!

> Active Record automatically timestamps create and update operations if the table has fields named created_at/created_on or updated_at/updated_on.

siaw23:
What are the two tables?

jywarren:
I believe it's tag_selection and term_node!

skilfullycurled:
@jywarren, this is great, will this actually be a return to the previous way that tags were entered in the database?  At least prior to 2016, tags did have timestamps for the first time they were created, and I believe a time stamp for when they were added to a node.  I'll have to look to see if this would change anything to the tags download feature which right now downloads all of the term_nodes because they do not have timestamps anymore.

jywarren:
Huh - i wonder how that got changed -- i think we simplified the tables but
perhaps didn't preserve these features when we should've! But yes, this
would be super.

skilfullycurled:
Yeah, there used to be a table called "tags" which was [removed in a commit](https://github.com/publiclab/plots2/commit/eff8a442a6b51a7d00fc462d5e039c82e1b4ea52#diff-4ef680bc0c8c6ec87925705fae6c0a1cL360) October 8, 2016 (I think).

skilfullycurled:
@jywarren, I've been researching migrations to see if this was something I could help on.

It seems like there is an internet consensus (n=3) that in Rails 5+ adding created_at and updated_at is done in the following [manner](https://stackoverflow.com/a/46521142/6023530) (also see [this](https://medium.com/@ebrugulec/add-timestamps-to-existing-table-in-db-rails-5-9e185cfab37e) and [this](https://makandracards.com/makandra/19325-your-database-tables-should-always-have-timestamps) which are slightly different but I think less clear versions of the same). 

```
class AddTimestampsToTags < ActiveRecord::Migration[5.2]
  def change
    # add new column but allow null values
    add_timestamps :term_node, null: true 
    add_timestamps :tag_selections, null: true 

[...]

PR #5875, opened 2019-06-12
Subscribers over time

### LINKED ISSUE #4925: Subscribers over time
Display the # of subscribers over time on tag stats pages like this: https://publiclab.org/tag/ui/stats

cesswairimu:
Fixes #4925

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -511,6 +511,8 @@ class TagController < ApplicationController
     @tag_wikis = @tags.first.contribution_graph_making('page', @start, @end)
     @tag_questions = @tags.first.quiz_graph(@start, @end)
     @tag_comments = @tags.first.comment_graph(@start, @end)
+    @subscriptions = @tags.first.subscription_graph(@start, @end)
+    @all_subscriptions = TagSelection.graph(@start, @end)
   end
 
   private
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -417,6 +417,21 @@ class Tag < ApplicationRecord
     end
   end
 
+  def subscription_graph(start = DateTime.now - 1.year, fin = DateTime.now)
+    date_hash = {}
+    week = start.to_date.step(fin.to_date, 7).count
+
+    while week >= 1
+      month = (fin - (week * 7 - 1).days)
+      range = (fin - week.weeks)..(fin - (week - 1).weeks)
+      weekly_subs = subscriptions.where(created_at: range)
+                                 .size
+      date_hash[month.to_f * 1000] = weekly_subs
+      week -= 1
+    end
+    date_hash
+  end
+
   private
 
   def tids
--- a/app/models/tag_selection.rb
+++ b/app/models/tag_selection.rb
@@ -36,6 +36,6 @@ class TagSelection < ApplicationRecord
   end
 
   def self.start_tracking
-    TagSelection.first.created_at.to_date.to_formatted_s(:long_ordinal)
+    TagSelection.first.updated_at.to_date.to_formatted_s(:long_ordinal)
   end
 end
--- a/app/models/tag_selection.rb
+++ b/app/models/tag_selection.rb
@@ -18,4 +18,20 @@ class TagSelection < ApplicationRecord
   def tagname
     tag.name
   end
+
+  def self.graph(start = DateTime.now - 1.year, fin = DateTime.now)
+    date_hash = {}
+    week = start.to_date.step(fin.to_date, 7).count
+
+    while week >= 1
+      month = (fin - (week * 7 - 1).days)
+      range = (fin - week.weeks)..(fin - (week - 1).weeks)
+      weekly_subs = TagSelection.select(:following, :created_at)
+                                .where(following: true, created_at: range)
+                                .size
+      date_hash[month.to_f * 1000] = weekly_subs
+      week -= 1
+    end
+    date_hash
+  end
 end
--- a/app/models/tag_selection.rb
+++ b/app/models/tag_selection.rb
@@ -34,4 +34,8 @@ class TagSelection < ApplicationRecord
     end
     date_hash
   end
+
+  def self.start_tracking
+    TagSelection.first.created_at.to_date.to_formatted_s(:long_ordinal)
+  end
 end
--- a/app/views/tag/_graph.html.erb
+++ b/app/views/tag/_graph.html.erb
@@ -1,7 +1,18 @@
 <div id="chart"> </div>
 <i> The graphs above are stacked </i>
+    <br>
+ <% if @tags.first.count.nil? %>
+        <i> There are no posts for this tag yet </i>
+    <% else %>
+        <b>Total Tag count : <%= @tags.first.count %></b>
+    <% end %>
 <br>
 <br> <br>
+
+<h5><u> We've started tracking Subscriptions from <%= TagSelection.start_tracking %></u> </h5>
+<br> <br>
+
+<div id="subscriptions"> </div>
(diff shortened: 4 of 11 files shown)


jywarren:
Oh wow! @skilfullycurled just a heads up on this. I wasn't sure if you and @cesswairimu were synced. This will be great! Also, if you rebase, you can add a screenshot test get automated screenshots!

cesswairimu:
Will do, thanks @jywarren. Am sorry @skilfullycurled I had no idea you are working on this, please let me know how far along you are on this and I can close this or work on top of what you have thanks.

skilfullycurled:
Oh, no no, @cesswairimu.  That's very thoughtful of you but I think @jywarren was giving me a heads up because a) I like stats! and b) I'm working on filtering out past spam from users.  I don't think that will effect this since it'll just change the figures but it will be nice to see places where the user count is unusually high.  Great work!

skilfullycurled:
Oh, wait, @cesswairimu, he may be referring to #5793.  Does this PR mean we do not need the migration anymore?

cesswairimu:
Oh, yeah, @skilfullycurled I had to add the migration before I could continue....did you get to write the migration yet?

skilfullycurled:
No problem, @cesswairimu, it seems like it should not have been made a sepperate issue since if I'm understanding you correctly, this issue couldn't be closed without doing it.  

And the good news is that it confirms what I learned about how to add it which was one of my questions so thanks for the help! ; )

However, I still think this is worth discussing in #5793 for two reasons.

1) @jywarren thought there would have to be migrations for both tag_selections and term_node (I think?  There were two.)

2) Backfill options of which I wrote about a few of varying complexity.  Enhancing the backfill could become it's own issue.  Wouldn't time.now (or I forget the variable) make everything subscribed to today?

Anyway, would be great if you could chime in.

cesswairimu:
Oh yeah thanks for bringing that up @skilfullycurled ...I updated all the existing records timestamps to have `Time.now`..what do you think about this? also quite not sure how we should display on the graph...maybe we can have a note that `we started tracking from....and only show the graph as from the next day  from that day. :thinking: but would love to your thoughts.

jywarren:
https://github.com/publiclab/plots2/blob/master/doc/DATA_MODEL.md may help some -- and I can't remember completely but I think it's possible I asked for timestamps on 2 models because we will need them for ordering pins as well, not just for subscriptions, so apologies if I added complexity; i just tend to think fewer migrations is a good thing. 👍 

I like "we've started tracking from"!

jywarren:
And to clarify - I wanted to add timestamps to the `NodeTag` MODEL, not table name, which is `community_tags` - so that we can know when a node was pinned (which uses a NodeTag record). Sorry for the confusion there: https://github.com/publiclab/plots2/blob/master/app/models/node_tag.rb

jywarren:
Thanks all!!!!

cesswairimu:
Thanks Jeff. I understand `community_tags` has a `date` column do we want to remove this or should I just add the timestamps and leave the `date` column

jywarren:
Oh! well, is it a timestamp? Sometimes the former Drupal tables have a timestamp but its not in a Rails native format, so we use a converter method. Could that work? If `community_tags` has valid date data, maybe we can use that?

skilfullycurled:
@cesswairimu, regarding the `time.now` question, here are a few considerations:

For the sake of clarity, I am going to refer to `time.now` as `time.merge` since that will eventually be the "now".

1. Depends on the limitations put on the visualization on the page.  If the page doesn't allow you to query information from the day after whenever `time.merge` is run, then it's fine because in that technical sense, that day will be of the same value as 4000 BC.

2) However, there might be future situations where we want to have both the tag data as well as some other data that exists prior to `time.merge`. This is where I think it would be safer to use a date prior to all of the data we have such as 1/1/2000.  This way, no one can search for it, and it will never cause a load on the server.  That's a pretty small change code wise I think and it would simplify any potential future queries.

3) We could log a sepperate enhancement issue to actually give these tags dates using the following query.  Assuming that each `tag_selection` is attached to some content node, then we could use the `created_at` or `updated_at` of the attached content to approximate the `created_at` of the `tag_selection`. It's jut an approximation because they could have created the content with two tags and than added another later but it would be a good educated guess.

skilfullycurled:
Oh, wait, @cesswairimu and @jywarren, before we merge the [community_tags commit](https://github.com/publiclab/plots2/pull/5875/commits/9e9d5688105a199dbfde7742b7ed624eb7bf2d2d), can check to see if the `date` field on `community_tags` actually has a date in it or if it's just empty?  If there is a date, then can we convert that to a timestamp instead of updating it to time.now?  This way we can have timestamps but we won't loose any fidelity on the data we already have.

cesswairimu:
Aha! good idea... I just checcked the date data are nil(could be because they are invalid) but others have data not sure if the `date` column is updated during creation but I can look more into it this could be great if it is. Thanks...

> Oh, wait, @cesswairimu and @jywarren, before we merge the [community_tags commit](https://github.com/publiclab/plots2/pull/5875/commits/9e9d5688105a199dbfde7742b7ed624eb7bf2d2d), can check to see if the `date` field on `community_tags` actually has a date in it or if it's just empty? If there is a date, then can we convert that to a timestamp instead of updating it to time.now? This way we can have timestamps but we won't loose any fidelity on the data we already have.

cesswairimu:
> @cesswairimu, regarding the `time.now` question, here are a few considerations:
> 
> For the sake of clarity, I am going to refer to `time.now` as `time.merge` since that will eventually be the "now".
> 
> 1. Depends on the limitations put on the visualization on the page.  If the page doesn't allow you to query information from the day after whenever `time.merge` is run, then it's fine because in that technical sense, that day will be of the same value as 4000 BC.
> 
> 1. However, there might be future situations where we want to have both the tag data as well as some other data that exists prior to `time.merge`. This is where I think it would be safer to use a date prior to all of the data we have such as 1/1/2000.  This way, no one can search for it, and it will never cause a load on the server.  That's a pretty small change code wise I think and it would simplify any potential future queries.
> 2. We could log a sepperate enhancement issue to actually give these tags dates using the following query.  Assuming that each `tag_selection` is attached to some content node, then we could use the `created_at` or `updated_at` of the attached content to approximate the `created_at` of the `tag_selection`. It's jut an approximation because they could have created the content with two tags and than added another later but it would be a good educated guess.

Awesome outline will implement this logic. Thanks so much @skilfullycurled

cesswairimu:
@skilfullycurled the date column is not populated during create with a timestamp, by default it is `0`

cesswairimu:
@skilfullycurled 
- For `tag_selections` I have changed the `created_at` of all of the existing records to 1/1/2000.
- For `community_tags` where `date` column has a value I have used that as the `created_at` value. Those without a date value I have used  `1/1/2000`.(Let me know what you think about this) Thanks

cesswairimu:
If there are no subscriptions tracked during the period selected.
notrack

If there its a range that has subscriptions
track

Full page "tag/stats"
2019-06-14-14-47-unstable publiclab org

skilfullycurled:
@cesswairimu, that sounds perfect.  Thanks for being open to that, I hope I didn't make life that much more difficult.  In line with 4000 BC I will be referring to 2000 as "10 BPL".  Anyway...great commit, looking forward to seeing the new features it will enable!

cesswairimu:
@publiclab/reviewers kindly review

jywarren:
Hooray!!! Merging, and we should be able to test it out on stable once it publishes. Thanks for a tremendous job!!!!! 🎉

jywarren:
cc @ebarry once you get back!

PR #5868, opened 2019-06-11
screenshots in system tests

jywarren:
Re #5320 but squashed!

PR #5320, opened 2019-03-31
Take screenshots during system tests, upload to Google Cloud storage, and append to Dangerbot output

jywarren:
Re #5316 screenshots of tests

Also we might try: `bin/rails test:system test` to get the coverage to not run twice (we went down from 91-44% but that's because we're running tests twice and 44 is just for system tests i guess?)

https://api.rubyonrails.org/classes/ActionDispatch/SystemTestCase.html

Attempting to generate inline screenshots with https://iterm2.com/documentation-images.html

But fallback could be to generate them as files, then use Dangerfile to leave them as comments, but collasped in a `<details>` el:

```
<details>
    <summary>Screenshots 📸 (click to expand)</summary>
    <h3>Front page:</h3>
    <p><img src="dataurl" /></p>
</details>
```

PR #5878, opened 2019-06-12
add screenshots for grids

jywarren:
Re #5320 #5868 #5316


sagarpreet-chadha:
This screenshot feature is super cool !!!

jywarren:
Yeah!!! We do need to either lengthen that inline grids screenshot or split it into 2 pages.

PR #5881, opened 2019-06-12
grids max-height

jywarren:
follow-up to #5872


jywarren:
Strange. Maybe the headless browser has no `max-height` CSS ability? I'll try merging this and testing on Stable.

PR #5897, opened 2019-06-13
Added Fix For Reply By Email message_id check

### LINKED ISSUE #5867: Reply-by-email emails sent repeatedly in loop
Related to #5817 we noticed that some comments were getting posted over and over; we're not sure why, but we should look closely at this section of code; we've had to turn off `mailman` to stop this from going on forever (100s of repeated emails):

https://github.com/publiclab/plots2/blame/master/app/models/comment.rb#L262-L292

@namangupta01 can you poke around a little to see if you can see if anything has changed? 

We've been messing with this area of the code to try to resolve the memory issues, but it doesn't look like anything we changed would cause an infinite loop of email receiving:

https://github.com/publiclab/plots2/blob/ca1011840ac47940f4295a7a8679280cf74828b9/script/mailman_server#L22-L39

namangupta01:
Fixes #5867


namangupta01:
@jywarren This is good to merge.

jywarren:
Awesome!!!

PR #5905, opened 2019-06-14
Create screenshots_mobile_test.rb at 375px wide, and increase height of reg screenshots to 2000px

jywarren:
Re #5316


jywarren:
Aha, just a name collision for the test: `test_front_page_with_navbar_search_autocomplete is already defined in ScreenshotsTest`

jywarren:
This is running but the screensize is not changing...

PR #5395, opened 2019-04-07
Exception / for textarea

CleverFool77:
Fixes #5387 (<=== Add issue number here)

In edit profile - 
ex

In comment section - 
hh

In editor section -
jj

* [x] PR is descriptively titled 📑 and links the original issue above 🔗
* [x] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [x] code is in uniquely-named feature branch and has no merge conflicts 📁
* [x] screenshots/GIFs are attached 📎 in case of UI updation
* [x] ask `@publiclab/reviewers` for help, in a comment below

Thanks!

PR #5922, opened 2019-06-18
Update keybindings.js to exempt rich text prompt areas from / searchform focus

gauravano:
Re #5395 thank you @steviepubliclab!!


gauravano:
here's where we were seeing this effect:

Screenshot from 2019-06-18 15-54-52

Issue #5890, opened 2019-06-13
UI Improvements Planning Issue : Individual tags page


gautamig54:
Next we start with the implementation of individual tags page. 

<img width="497" alt="Capture" src="https://user-images.githubusercontent.com/35326753/59403744-8e78d980-8dc1-11e9-8590-79db2b0c62ac.PNG">

**Checklist**

- [x] Divide the page into three sections(Main image and the introduction card, sidebar with map, activity cards) @gautamig54 
- [x] Adjust the main image and add the number of people following on to the right. @CleverFool77 
- [x] make an introduction card for the tag containing the subscribe and learn more options. @gautamig54 
- [x] Activity cards with new design of style guide @CleverFool77 
- [ ] ~Show map on the sidebar and people near options @gautamig54~
- [ ] Sort by name, users and people following  @CleverFool77  **Wait a bit**
- [ ] ~Post a question option with dropdown @gautamig54~
- [x] Implement type dropdown containing options for all, questions, wikis and notes related to the tag. @CleverFool77 
- [x] Add related tags @gautamig54 

cc : @CleverFool77

gautamig54:
I had a few doubts regarding the design. 
Not all tags have a wiki page due to which there will be no background image. What should be the design in that case?
Where should we add the related tags section?
Also, I would like to suggest to add the list of sub-pages for a parent tag here on its main page itself.

gautamig54:
@jywarren @CleverFool77 @gauravano @SidharthBansal 
Let's get started on this. Thanks!

CleverFool77:
Ohk. Seems like the Issues ave been divided here. So I'll close the earlier issue that I made then.
Thanks !!

CleverFool77:
The new style guide design looks good but there are many features that we are yet to place in it Like gautami mentioned in the comment.


> I had a few doubts regarding the design.
> Not all tags have a wiki page due to which there will be no background image. What should be the design in that case?
> Where should we add the related tags section?
> Also, I would like to suggest to add the list of sub-pages for a parent tag here on its main page itself.
> Do we need to make this as a shadow page as well??

gautamig54:
@CleverFool77 I have assigned the tasks to us. You can change it according to your preference. Thanks. 
And let's wait for @jywarren comment on the doubts we have.

CleverFool77:
Hi @gautamig54 do the assignment alternativily. So that it's easier to work and it's faster.

gautamig54:
[...]

PR #5902, opened 2019-06-14
Partitioning the individual tags page

gautamig54:
Fixes #5890  (<=== Add issue number here)

Screen Shot 2019-06-14 at 9 57 59 AM

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [x] PR is descriptively titled 📑 and links the original issue above 🔗
* [x] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [x] code is in uniquely-named feature branch and has no merge conflicts 📁
* [x] screenshots/GIFs are attached 📎 in case of UI updation
* [ ] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software 

Thanks!

PR #5925, opened 2019-06-20
Individual tagpage: Add dropdown to distinct btw info and add test

CleverFool77:
Fixes https://github.com/publiclab/plots2/issues/5890 part
The position of dropdown button looks floated to rightmost.
The position will get fixed with the help of one  https://github.com/publiclab/plots2/pull/5902 where the grid system is being applied.

Screenshot from 2019-06-20 15-01-22

Screenshot from 2019-06-20 15-00-39

### DIFF
--- a/app/views/tag/show/_nav_tabs.html.erb
+++ b/app/views/tag/show/_nav_tabs.html.erb
@@ -1,44 +1,45 @@
-<ul class="nav nav-tabs" style="margin-top:10px;">
+<div class="dropdown" style="margin-top:10px; float:right;">
+  <button class="btn btn-outline-secondary dropdown-toggle" id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">By Type</button>
+  
   <!-- when will it not be show...? -->
-  <% if params[:action] == "show" %>
-      <% unless params[:id].match("question:") %><li class="nav-item"><a class="nav-link <% if @node_type == "note" %> active<% end %>" href="/tag/<%= params[:id] %>"><i class="fa fa-file"></i> <span class="d-none d-lg-inline"><%= raw t('tag.show.research_notes') %></span></a></li><% end %>
-    <!-- can we create a generic node type hash, and give it href and class and content classes? -->
-    <!-- this should be an if/else, because @node_type should only be one value -->
-    <li class="nav-item"><a class="nav-link <% if @node_type == "questions" %> active<% end %>" href="/questions/tag/<%= params[:id] %>"><i class="fa fa-question-circle"></i> <span class="d-none d-lg-inline"><%= t('tag.show.questions') %></span></a></li>
-    <li class="nav-item"><a class="nav-link <% if @node_type=="wiki" %> active <% end %>" href="/wiki/tag/<%= params[:id] %>"><i class="fa fa-book"></i> <span class="d-none d-lg-inline"><%= raw t('tag.show.wiki_pages') %></span></a></li>
-    <li class="nav-item"><a class="nav-link <% if @node_type == "maps" %> active <% end %>" href="/maps/tag/<%= params[:id] %>"><i class="fa fa-map-marker"></i> <span class="d-none d-lg-inline"><%= t('tag.show.maps') %></span></a></li>
-
-    <!-- show if we can see the list of contributors
-    again, can we have a generic list of contributors that is
-    calculated in the controller? -->
-    <% if !@wildcard %>
-        <li class="nav-item"><a class="nav-link <% if @node_type == "contributors" %> active <% end %>" href="/contributors/<%= params[:id] %>"><i class="fa fa-user"></i> <span class="d-none d-lg-inline"><%= raw t('tag.show.contributors') %></span><span class="badge"><%=@length %></span></a></li>
+  <div class="dropdown-menu dropdown-menu-right mt-1" aria-labelledby="dropdownMenuButton">
+    <% if params[:action] == "show" %>
+        <% unless params[:id].match("question:") %><a class="dropdown-item <% if @node_type == "note" %> active<% end %>" href="/tag/<%= params[:id] %>"><i class="fa fa-file"></i> <span class="d-none d-lg-inline"><%= raw t('tag.show.research_notes') %></span></a><% end %>
+      <!-- can we create a generic node type hash, and give it href and class and content classes? -->
+      <!-- this should be an if/else, because @node_type should only be one value -->
+      <a class="dropdown-item <% if @node_type == "questions" %> active<% end %>" href="/questions/tag/<%= params[:id] %>"><i class="fa fa-question-circle"></i> <span class="d-none d-lg-inline"><%= t('tag.show.questions') %></span></a>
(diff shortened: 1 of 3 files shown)


gautamig54:
Looking great! 
@CleverFool77 How do I generate related tags for a particular tag in my local system?

CleverFool77:
Do you still need help gautami?

CleverFool77:
Hi @jywarren just pushed one the parts of individual tagoage issue here.

CleverFool77:
Hi @jywarren I wanted to ask that
```
  test 'should have active question tab for question' do
    tag = tags(:question)

    get :show, params: { id: tag.name }
    selector = css_select "a[href = '/questions/tag/question:spectrometer']"
    assert_equal selector.size, 1
    selector = css_select '#questions.active'
    assert_equal selector.size, 1
  end
```
Here We are doing testing by using `question.active` in tab. But in new design we have dropdown. And unless dropdown menu is being shown, despite of being at questions page, that class won't be active for dropdown-item leading to failure of test. SO should I add condition  where unless dropdown-menu is having show class, then only the test for question dropdown-item should have active class should be tested ?

gautamig54:
@CleverFool77 Do you think it's really required? Like, initially(by default) some tab say, wiki will be active and it has nothing to do with the dropdown. In that case, it might fail tests. You then have to add wikis to the tests separately. 
What do you think?

CleverFool77:
But here it's not getting the drop down item as active at all. Because the drop down menu is closed. I provided toggle for drop down, so it closes as we chose one of the items. So i thought maybe we need to check whether dropdown menu is open, and then only check the active class for item.

Or should I remove the `toggle for drop down menu` . So the menu doesnt get closed itself. Then I can direct check the active class for `drop down item `

gautamig54:
I am not sure if we need to do this. I am not good with writing tests so lets wait for @jywarren's view on this one.

jywarren:
> Here We are doing testing by using question.active in tab. But in new design we have dropdown. And unless dropdown menu is being shown, despite of being at questions page, that class won't be active for dropdown-item leading to failure of test. So should I add condition where unless dropdown-menu is having show class, then only the test for question dropdown-item should have active class should be tested ?

Hmm, so isn't the issue perhaps that there is no fixed URL for showing questions, and perhaps there should be? What if the dropdown loads the route `https://publiclab.org/questions/tag/pi-camera` and then we can ensure that what's being shown is questions? this also most minimally disrupts the current routing setup and tests, even if it does require a page refresh still. 

Am I understanding your question correctly?

CleverFool77:
> > Here We are doing testing by using question.active in tab. But in new design we have dropdown. And unless dropdown menu is being shown, despite of being at questions page, that class won't be active for dropdown-item leading to failure of test. So should I add condition where unless dropdown-menu is having show class, then only the test for question dropdown-item should have active class should be tested ?
> 
> Hmm, so isn't the issue perhaps that there is no fixed URL for showing questions, and perhaps there should be? What if the dropdown loads the route `https://publiclab.org/questions/tag/pi-camera` and then we can ensure that what's being shown is questions? this also most minimally disrupts the current routing setup and tests, even if it does require a page refresh still.
> 
> Am I understanding your question correctly?

Hi @jywarren I didn't understand the point clearly though.
But I thought of a method. I'll remove the test written for tab and write new tests for dropdown directly.

jywarren:
That sounds great! Thank you!

CleverFool77:
HI Jeff,
I was going through the tests and I need some help. 
I'm not unable to understand how to check something which is conditional. 
As in this, the tab earlier were direct, and so was the test. But now in dropdown, The dropdown-menu dissappears and toggles to button after clicking on one of the items. So If I have to test the item, its necessary to have the dropdown-menu to be shown.
I tried some methods like checking dropdown menu has show class and etc. But they doesn't seem to work. :thinking:

app/views/tag/show/_nav_tabs.html.erb
```diff
@@ -1,44 +1,46 @@
...
-    <li class="nav-item"><a class="nav-link <% if @node_type == "maps" %> active <% end %>" href="/maps/tag/<%= params[:id] %>"><i class="fa fa-map-marker"></i> <span class="d-none d-lg-inline"><%= t('tag.show.maps') %></span></a></li>
-
-    <!-- show if we can see the list of contributors
-    again, can we have a generic list of contributors that is
-    calculated in the controller? -->
-    <% if !@wildcard %>
-        <li class="nav-item"><a class="nav-link <% if @node_type == "contributors" %> active <% end %>" href="/contributors/<%= params[:id] %>"><i class="fa fa-user"></i> <span class="d-none d-lg-inline"><%= raw t('tag.show.contributors') %></span><span class="badge"><%=@length %></span></a></li>
+  <div class="dropdown-menu dropdown-menu-right mt-1" aria-labelledby="dropdownMenuButton">
+    <% if params[:action] == "show" %>
+        <% unless params[:id].match("question:") %><a class="dropdown-item <% if @node_type == "note" %> active<% end %>" href="/tag/<%= params[:id] %>"><i class="fa fa-file"></i> <span class="d-none d-lg-inline"><%= raw t('tag.show.research_notes') %></span></a><% end %>
```
jywarren: Can u use the dev console elements inspector to see what the code looks like? I mean, here, it seems like we should be able to test against it with `css_select`. Maybe `css_select` docs say that the element has to be shown?

If you can't get it to work, you could write a system test for it, where you literally send a click event to open the dropdown. I've done that for some system tests, so it's a backup plan!

CleverFool77: But Doesn't `css_select` work only when the element is being shown not when it's hidden like dropdown menu here ? 
I did system test though, to have a screenshot when menu is open and closed.

CleverFool77:
```
FAIL["test_viewing_the_dropdown_menu", #<Minitest::Reporters::Suite:0x00005653ac649ca8 @name="TagTest">, 67.22464900100005]
 test_viewing_the_dropdown_menu#TagTest (67.22s)
        expected to find css ".dropdown-menu > dropdown-item" but there were no matches
        test/system/tags_controller_test.rb:15:in `block in <class:TagTest>'
[Screenshot]: tmp/screenshots/test_viewing_the_dashboard.png
[Screenshot]: tmp/screenshots/test_viewing_the_dashboard.png
 FAIL["test_viewing_the_dashboard", #<Minitest::Reporters::Suite:0x00005653ac855f38 @name="DashboardTest">, 129.53698157000002]
 test_viewing_the_dashboard#DashboardTest (129.54s)
        expected to find visible css ".row.header > h1" with text "Dashboard" but there were no matches. Also found "Community research", which matched the selector but not all filters. 
        test/system/dashboard_test.rb:19:in `block in <class:DashboardTest>'

```
Uhmm !! :thinking: The second test seems to be failing and same goes for first one. The reason for both of them is same that there is no matches :thinking:

CleverFool77:
All checks passed :tada: :tada:

test/system/tags_controller_test.rb
```diff
@@ -0,0 +1,17 @@
+require "application_system_test_case"
+
+class TagTest < ApplicationSystemTestCase
+  Capybara.default_max_wait_time = 60
+
+  test 'viewing the dropdown menu' do
```
jywarren: Very awesome!

CleverFool77:
Sure Jeff !!!
Thanks !!

jywarren:
OK, merged https://github.com/publiclab/plots2/pull/5948 and this. Now to https://github.com/publiclab/plots2/pull/5938 !!!

CleverFool77:
Cool !!

PR #5928, opened 2019-06-20
multi-subscribe login sequence!

### LINKED ISSUE #5614: Pop up subscription modal /after/ opening multi-tag subscription modal

jywarren:
Fixes #5614, supercedes #5620

### DIFF
--- a/app/views/tag/_subscribe_button.html.erb
+++ b/app/views/tag/_subscribe_button.html.erb
@@ -1,18 +1,11 @@
 <!-- This button is used to display tags in a popover and subscribe to multiple tags using a single button -->
-<% if current_user %>
-  <a id="popover_button" class="btn btn-outline-secondary btn-sm sub" >Subscribe 
-    <i class="fa fa-plus"></i> 
-  </a>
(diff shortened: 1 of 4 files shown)


jywarren:
```
 FAIL["test_should_list_notes_and_questions_in_user_profile", #<Minitest::Reporters::Suite:0x000055c768574830 @name="UsersControllerTest">, 29.149527155999976]
 test_should_list_notes_and_questions_in_user_profile#UsersControllerTest (29.15s)
        Expected nil to not be nil.
        test/functional/users_controller_test.rb:155:in `block in <class:UsersControllerTest>'
```

PR #5929, opened 2019-06-20
Make test screenshots only upload from system test process in Travis

jywarren:
We're seeing 5 updates to the Dangerbot comment from our 5 parallelized test VMs (see docs at https://docs.travis-ci.com/user/speeding-up-the-build/#parallelizing-your-builds-across-virtual-machines). But this means the screenshots are often overwritten by one or more of the non-screenshot test processes. 

This tries to use an env variable in each job so we can only post the comment from the system test job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,11 @@ env:
   global:
     - COMPOSE_FILE=./containers/docker-compose-testing.yml
   matrix:
-    - TASK="rake test:all"
-    - TASK="rails test test/unit"
-    - TASK="rails test test/functional"
-    - TASK="rails test test/integration"
-    - TASK="rails test:system"
+    - TASK="TEST_TYPE=mocha rake test:all"
+    - TASK="TEST_TYPE=unit rails test test/unit"
+    - TASK="TEST_TYPE=functional rails test test/functional"
+    - TASK="TEST_TYPE=integration rails test test/integration"
+    - TASK="TEST_TYPE=system rails test:system"
 
 before_install:
   - echo -e "repo_token:\n  $COVERALLS_REPO_TOKEN" >> ./.coveralls.yml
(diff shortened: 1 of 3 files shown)


jywarren:
Welp... that didn't work. No screenshots at all!

Dangerfile
```diff
@@ -1,79 +1,82 @@
-message "@#{github.pr_author} Thank you for your pull request! I'm here to help with some tips and recommendations. Please take a look at the list provided and help us review and accept your contribution! And **don't be discouraged if you see errors** -- we're here to help."
+# run only on system tests? This will break interactions with junit test reports but fix automated screenshot posting 
+if ENV['TEST_TYPE'] == "system"
```
jywarren: I'm guessing this is not working.

jywarren: Ah ok actually the Ruby code is working; the insertion of the ENV variable is not. Maybe we have to turn that on in Travis...

jywarren:
OK, so we have: `“TEST_TYPE”=>””,` -- not sure why!

.travis.yml
```diff
@@ -12,11 +12,11 @@ env:
   global:
     - COMPOSE_FILE=./containers/docker-compose-testing.yml
   matrix:
-    - TASK="rake test:all"
-    - TASK="rails test test/unit"
-    - TASK="rails test test/functional"
-    - TASK="rails test test/integration"
-    - TASK="rails test:system"
+    - TASK="TEST_TYPE=mocha rake test:all"
```
jywarren: What if we removed the TEST_TYPE but looked for `TASK == "rake test:system"`?

jywarren: No, because the current output shows that `TEST` is not passed in. But maybe we can change that in the container config...

jywarren:
Weird I got the screenshots by email but not here?

jywarren:
@kaustubh-nair note in relation to #5931 how I'm using an env var to make only one of the 5 tasks actually generate a Danger report!

PR #5967, opened 2019-07-01
Issue templates added 

gauravano:
Hi, inspired by Leaflet.Distortable templates, lets add similar templates here too.


IshaGupta18:
Looks lovely! We also have a similar thing in simple-data-grapher!

jywarren:
Super!!!! 🎉

Issue #5889, opened 2019-06-13
UI Improvements Planning Issue : tag/tagname page 


CleverFool77:
#### Current Design 
![ezgif-4-efcfcf93f14c](https://user-images.githubusercontent.com/26685258/59403385-0e9e3f80-8dc0-11e9-81b2-3189d78f579e.gif)

#### Draft mockup design 
![PublicLab mockups](https://user-images.githubusercontent.com/26685258/59403573-c59abb00-8dc0-11e9-8950-7612725da532.png)

#### New style guide design

![Public Lab Style Guide Draft](https://user-images.githubusercontent.com/26685258/59403269-9a639c00-8dbf-11e9-853a-9c30deeb809c.png)

CleverFool77:
Hi @gautamig54 @jywarren I saw two designs as per earlier draft and new style guide. So may I know which one will be the one that will be implemented ?

gautamig54:
Haha @CleverFool77 We opened the issues at the same time. I have added the checklists according to the new design. Look at #5890

gautamig54:
I think we need to implement the new style guide design.

PR #5979, opened 2019-07-05
Tweak card CSS on tag page

jywarren:
Re #5889 -- some small CSS tweaks! cc @gautamig54 @CleverFool77


jywarren:
Also, I see some odd behaviors for mobile-width screens, but perhaps those will be resolved in an existing open PR? Thanks!

image

Issue #5983, opened 2019-07-05
Login Modal container css distorted


CleverFool77:
### Please describe the problem (or idea)
I guess while I removed the padding and margin for the container for the absolute image in tag/tagname url/. It effected the login modal too. So Now when I'm trying to login , the modal seems to out of screen from above wherease it's working fine in other urls.

### Please show us where to look
It's working fine on other urls like this
![Screenshot from 2019-07-05 22-50-21](https://user-images.githubusercontent.com/26685258/60737150-4ccdef80-9f77-11e9-9d09-31d87c13b8c4.png)

But here tag/tagname, Its like this
![Screenshot from 2019-07-05 22-49-31](https://user-images.githubusercontent.com/26685258/60737169-62dbb000-9f77-11e9-9711-d7885fed2dca.png)

As much I know, If I'm correct then its due to conatiner. Fixing it either at wiki image for tag/tagname at  https://github.com/CleverFool77/plots2/blob/567c035a96745946f5c0ed2085e5718d4fd07eba/app/views/tag/show/_header.html.erb#L68

or tweaking it in modal maybe would do.

CleverFool77:
cc : @jywarren

jywarren:
Ah yes, i think i saw the `.container` style overridden. I think you need
to narrow the scope of the CSS to just your own .container. So, perhaps
using a custom ID or a more specific nested classname?

CleverFool77:
Great idea Reagrding nested class Jeff.

PR #5989, opened 2019-07-06
Small fix: login container

CleverFool77:
Fixes https://github.com/publiclab/plots2/issues/5983

Screenshot from 2019-07-06 16-24-45


jywarren:
Hmm, are you sure this doesn't affect mobile view? Maybe best to narrow the other usage of container instead of removing it from this, what do you think?

CleverFool77:
> Hmm, are you sure this doesn't affect mobile view? Maybe best to narrow the other usage of container instead of removing it from this, what do you think?

I guess it would be better to avoid overridden-css like you said. So I'll try the method you spoke of now.

PR #5997, opened 2019-07-09
Update screenshots_test.rb

jywarren:
cc @gautamig54 @CleverFool77 - adding an automated screenshot test for this page:

image

I think this'll be mostly OK, but just to make you both aware of this alternative sub-view of the tag page! Thanks!

### DIFF
--- a/test/system/screenshots_test.rb
+++ b/test/system/screenshots_test.rb
@@ -33,6 +33,11 @@ class ScreenshotsTest < ApplicationSystemTestCase
     take_screenshot
   end
 
+  def 'tag by author page' do
+    visit '/tag/spectrometer/author/bob'
+    take_screenshot
+  end
+
   test 'stats' do
     visit '/stats'
     take_screenshot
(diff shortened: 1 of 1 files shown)


jywarren:
Aha! @gautamig54 @CleverFool77 note the issue with the title card pushing down here -- once the title text wraps to 2 lines, it starts to overlap the section below. 

image

What should we do when that happens? Once it reaches 3 lines it gets quite a bit worse:

https://stable.publiclab.org/tag/public-comment%20host-an-event

image

And especially worse in mobile display.

gautamig54:
Can we make the height of the grey image added at the back relative to the tags card? Like its height adjusts to the height of the tags card.
Another approach can be adding ellipses like we have done for the tags. Displaying the whole tag on hover else display only as much as can fit in a single line. 
What do you say @CleverFool77 ??

jywarren:
Grey image relative height sounds reasonable, do you want to try it out at
a few different screen sizes to check? And, perhaps at some point then we
use an ellipse, like if the title gets REALLY long? :-)

jywarren:
Maybe... 2 lines? but you may have to try it out a bit to see what looks
good. Thanks!!!

PR #6017, opened 2019-07-14
Changed notification permission prompt to be appea only when user is logged in

### LINKED ISSUE #6016: Ask for notification permission only when user id logged in
Now, we have notification supported to public lab website. It asks for permission whenever the user open the public lab website irrespective of whether the user is logged in or not. Let's change to the case where notification permission prompt only appears and asks for permission when user is logged in.

namangupta01:
Fixes #6016


namangupta01:
@jywarren can we merge this?

jywarren:
Yep!!!

Issue #5490, opened 2019-04-15
Stats downloading returns "Page does not exist" for dates prior to early 2013


skilfullycurled:
I was trying to download stats from the beginning of the site (which was sometime in 2010) until July of 2016 and I received a "Page does not exist" error.  I did some experimentation and while I'm not sure of the exact date before which this occurs, I can say it is sometime between 01-01-2013 and 01-07-2013 (in stats page date format DD-MM-YYYY).  For what it's worth to the diagnosis of the problem, I was "only" downloading up until July of 2016.

Let me know if there's anything else I can tell you (FF 66.0.2, macOS 10.12.6).

Oh, a possible reference where this could be merged: Raw data from stats page, #4654

> **NOTE** for high impact URLs which could slow the main website down, if you want to test them out, please use `stable.publiclab.org` instead of `publiclab.org` and you'll only slow or break the stable test server, which should have very similar code and data. (note by @jywarren)

cesswairimu:
Will take a look at why this is happening thanks @skilfullycurled

cesswairimu:
@skilfullycurled when did you receive the error when submitting the date ranges or the data was returned fine and this occurred when you were trying to "download as"?

skilfullycurled:
No, @cesswairimu, it was the search itself that returned the error.

cesswairimu:
Aha! gotcha this is strange because dates before that are working and after that are working Did a little digging and the culprit is `25-04-2013` any range before that or after that works which is strange. @jywarren any idea if anything happened to the system on this date?

skilfullycurled:
Oh, I forgot about that date.  @cesswairimu, we're never supposed mention that date or even speak about what happened on it.  It's sort of the Voldemort of dates.  

But seriously folks.  

Very interesting.

I'll download around it.

cesswairimu:
Aha gotcha sorry will delete the comment

cesswairimu:
@skilfullycurled maybe we can close this now since its not a code issue?

jywarren:
lol are you folks having too much fun in here? Voldemort dates? 🔮 🤐 🙊 I hope Cess has seen Harry Potter and knows you're joking? Cess, @skilfullycurled has a strange sense of humor please don't hold it against him. 

No but for real, thanks for looking into this. I'm sure it was a tough one to track down esp. given the mysterious date.

cesswairimu:
:laughing: :laughing:

skilfullycurled:
[...]

PR #6050, opened 2019-07-23
Add all time stats option filter

### LINKED ISSUE #5904: Add 'all time' option for stats
I have also encountered this error when attempting to guestimate the start date for our website so that i could view data "For All Time". 

I'm wondering if it's really a bug or perhaps maybe it could be worked around by people not having to guess when we started logging data for our website? 

If it's the latter, here's an idea: 
On the "Choose a Start Date" interface, could we add the exact start date of our website as a "quick pick" OR perhaps make an option to choose "for all time" as the period of inquiry?

_Originally posted by @ebarry in https://github.com/publiclab/plots2/issues/5490#issuecomment-497454355_

cesswairimu:
Fixes #5904

### DIFF
--- a/app/controllers/stats_controller.rb
+++ b/app/controllers/stats_controller.rb
@@ -10,9 +10,11 @@ class StatsController < ApplicationController
 
   def range
     flash.now[:notice] = "Data is cached and recalculated daily"
-    if params[:options].present?
+    if params[:options].present? && params[:options] != 'For all time'
       params[:start] = Time.now - to_keyword(params[:options])
       params[:end] = Time.now
+    elsif params[:options] == 'For all time'
+      all_time_stats
     end
     @start = start
     @end = fin
@@ -35,6 +37,11 @@ class StatsController < ApplicationController
       @contributors = User.contributor_count_for(@start, @end)
       @popular_tags = Tag.nodes_frequency(@start, @end)
     end
+  end
+
+  def all_time_stats
+      params[:start] = Date.new(2012,9,27).to_time
+      params[:end] = Time.now
   end
 
   def index
--- a/app/controllers/stats_controller.rb
+++ b/app/controllers/stats_controller.rb
@@ -40,8 +40,8 @@ class StatsController < ApplicationController
   end
 
   def all_time_stats
-      params[:start] = Date.new(2012,9,27).to_time
-      params[:end] = Time.now
+    params[:start] = Date.new(2014, 01, 01).to_time
+    params[:end] = Time.now
   end
 
   def index
--- a/app/views/stats/index.html.erb
+++ b/app/views/stats/index.html.erb
@@ -14,7 +14,7 @@
         </div>
         <div class=col-lg-2>
             <%= form_tag request.url, method: 'get', enforce_utf8: false do %>
-                <%= select_tag :options, options_for_select(["Past week", "Past month","Past year"], params[:options]), prompt: "Filter period: ",
+                <%= select_tag :options, options_for_select(["Past week", "Past month","Past year", "For all time"], params[:options]), prompt: "Filter period: ",
                     class: " form-control custom-select", onchange: "this.form.submit();" %>
             <% end %>
         </div>


cesswairimu:
@skilfullycurled I think the data error still persists tried a query from Jan2013 - Dec 2013...same error #5490. So I have set it as from 2014 for the `all time option` for now until we can solve the data issue. What do you think?

jywarren:
Hi @cesswairimu just marking this as in-progress until it's ready. thanks!

skilfullycurled:
Sorry for the delay @cesswairimu, I was traveling.  I think what you did (limiting to 2014) is perfect for what I hope will be a short time being.  The good news is that I believe I've narrowed down the day of the problem and I have a theory as to it's cause.

**Short story:** _I believe one of the original offending dates (4/25/2013) is still causing trouble and although the post-fix test which ended in 4/25/2013 passed, this was only because the "start date" is inclusive but the "end date" is exclusive._ 

**Proposed Solution:** Assuming the problem is the same, the solution is the fix @jywarren found in #5490.  There must be remaining NodeTag record(s) which point to a non-existant tid(s) and need to be removed.

My reasoning is below, but you might as well go ahead and try the solution as @jywarren outlined in #5490 and it if doesn't work you can take a look at how I arrived there and see if there is a flaw in my logic.

||
||
||
\\/

**Reasoning:** In #5490, @cesswairimu had narrowed down the problem date to 4/25/13.  @jywarren successfully fixed the issue by removing [NodeTag records pointing at non-existent tids](https://github.com/publiclab/plots2/issues/5490#issuecomment-502931726) afterwhich the test case from [1/1/12 - 4/25/13](https://publiclab.org/stats?start=01-01-2012&end=25-04-2013 ) worked.  Therefore, I figured that the problem must be after that date.  To find the next offending date, I did a month by month query from 4/26/2013 through 12/31/2013 yet everything worked as expected.  

Then I did a series of queries with successively decreasing time windows starting at 4/25/2013 through 4/30/2013 by increasing the start date by one day each query (4/25/2013 - 4/30/2013, 4/26/2013 - 4/30/2013...).  The only query that did not work was the first one, 4/25/2013 - 4/30/2013).

Where it got strange was when I performed a day by day query starting at 4/25/2013 through 4/30/2013 (4/25/2013 - 4/25/2013, 4/26/2013 - 4/26/2013...) and in this case, 4/25/2013 _did_ work.

At this point, I suspected the issue might be one of start/end date inclusion/exclusion, so I repeated the day by day test but this time under the assumption that the start date was inclusive but the end date was exclusive (e.g., 4/25/13 - 4/26/13).  Once again, the date with 4/25/13 returned to causing a failure.  This would also explain why that 4/25/13-4/25/13 works because it doesn't return anything.

cesswairimu:
Great investigation @skilfullycurled :100: ,  I believe your proposed solution will work. I guess we shall wait for Jeff to come back and so that he can remove the fault record(s). I could write a rake task that deletes the nodes with no tids..but I am not sure if its necessary will have to wait for Jeff on that too. Thanks  a lot :+1:

jywarren:
Hi, all, just catching up here, thanks! So, is there a target date I should look for a tid-less NodeTag at? Could I simply look for ALL NodeTags with no tid, or no matching Tag, and delete them? I would hope there aren't many.

skilfullycurled:
> So, is there a target date I should look for a tid-less NodeTag at? 

@jywarren, based upon my investigation (see logic above) I believe the only day that should have one is 4/25/13.  If it was not very time/computationally expensive, you could look from 4/25/13 (inclusive) to the end of the year (inclusive), but I believe the result of my analysis leaves only 4/25/13 as the offending time period.  Since I've downloaded every yearly set of data, I think running the inquiry on the entire database would be overkill but it's the only way to be sure.  I suppose we could split the difference and only look from the beginning until the point at which the tag setup changed in the database but I don't know when that was.  Or, perhaps until Plots2 was deployed.

> Could I simply look for ALL NodeTags with no tid, or no matching Tag, and delete them? 

I think this was your strategy last time.  You can see what you past @jywarren did [here](https://github.com/publiclab/plots2/issues/5490#issuecomment-502931726).

As an aside: I believe the reason why we missed this date was due to confusion over the end date being inclusive when it was exclusive.  Should we be making all dates inclusive?  Or perhaps stating on the page that the start date is inclusive but the end date is exclusive?  Not sure how hard the change is, but my expectation would be that when I ask for a collection of data, it will return that data through the specified end date.

jywarren:
Finally merging this, thank you @cesswairimu !!!!

ebarry:
This is super exciting, thanks @cesswairimu @jywarren @skilfullycurled !!!!

Issue #6085, opened 2019-07-30
Reply By Tweet Bug Discussion


namangupta01:
So I have figured out what the actual problem is after adding the console to print the mid way data in #6064.
The actual problem is short links.
`parent tweet #<Twitter::Tweet:0x00005646ea71df68>
parent tweet full text -> We’ve overhauled the entire https://t.co/PKr9eq2bkF website! Thanks to support from @GoogleOSS and contributors from five continents(!), we’re helping people around the world to better tell their stories through aerial mapping. Read all about it! https://t.co/nBXpX7hCHZ https://t.co/HClNdCS9y4
urls => ["https://t.co/PKr9eq2bkF", "https://t.co/nBXpX7hCHZ", "https://t.co/HClNdCS9y4"]
node => 
node.nil? => true`

![Screen Shot 2019-07-31 at 3 42 55 AM](https://user-images.githubusercontent.com/20709985/62169243-557ecf00-b345-11e9-92c1-0b1894e010bc.png)

So here, https://t.co/nBXpX7hCHZ opens to https://publiclab.org/notes/warren/07-26-2019/community-atlas-an-around-the-world-collaboration-on-mapknitter where we have implemented the algo for the https://publiclab.org/n/20331.

https://github.com/publiclab/plots2/blob/449d047f05209b443b049cdd6d3c2c905cfca51e/app/models/comment.rb#L452

I guess we can try this once with a tweet with this kind of link https://publiclab.org/n/20331
If you want to see the logs, you can see at stable.publiclab.org/cron_log1.log
What do you think?

namangupta01:
cc @jywarren , @gauravano

namangupta01:
Can @icarito @jywarren, please tell me what is the value of ENV["WEBSITE_HOST_PATTERN"] on stable? I need this for further debugging.

namangupta01:
@jywarren I tried on local with the same tweet. I was able to find node on local machine but same was not happening on production. I think it is something to do with the website host pattern env variable. Can anyone of @jywarren or @icarito let me know that what is the ENV["WEBSITE_HOST_PATTERN"] set on production? Is it `//publiclab.org/n/`?

Thanks

jywarren:
Hmm, yes it is! `export WEBSITE_HOST_PATTERN=://publiclab.org/n/`

namangupta01:
I checked the logs and yayy, I found this:
[...]

Issue #4962, opened 2019-03-08
Optimise reply by tweet Feature


namangupta01:

siaw23:
Any errors? Or pointers where I can have a look?

jywarren:
Yes, a few places - optimization is detailed here: https://github.com/publiclab/plots2/pull/3175#discussion_r263211966

but as to functionality I'm not sure, I do believe we saw a Sentry error for it at one point -- here:

* https://sentry.io/share/issue/cd87dd7040b945eca26c906f3f963a5a/
* https://sentry.io/share/issue/e87f95381f20461e835efea1941eabe1/

```
HTTP::ConnectionError
failed to connect: Connection reset by peer - SSL_connect

      parent_tweet = Client.status(in_reply_to_tweet_id, tweet_mode: "extended")
```

That's on this line: https://github.com/publiclab/plots2/blob/master/app/models/comment.rb#L387

jywarren:
https://sentry.io/share/issue/16c09d47269240bdaa4dd6a100e0c258/ also shows a related error!

jywarren:
Added UserTag query optimization here:

https://github.com/publiclab/plots2/commit/a7ca3e9250e8b6b5a8317149f3658685462ac134

namangupta01:
@jywarren all the sentry issues are related to the outbound requests i.e the request that is going out from the server to fetch the data and our system is not able to get the response.

namangupta01:
Can you try for the query ` Client.search(ENV["TWEET_SEARCH"])` in rails console and check if that it is getting some response or not. If it not then there will be something to be done with server configuration.

jywarren:
Yes, it is -
https://gist.github.com/jywarren/ed5ce5c2256ac35487715223c1a46452

Hmm, what log should we look in?

namangupta01:
Strange, It is working in rails console and the same line of code not working in production.

namangupta01:
I guess this should not happen. @jywarren can you please look into the sentry dashboard and tell the total number of time above event occurs and the last occurrence of these events?

jywarren:
Well, why don't I make you a Sentry account? It looks like there are a few inter-related errors, like this one: https://sentry.io/share/issue/0e925d1fc2ab4cc7b5fca0cce5033149/

Try searching for "twitter" in the issues list there!

namangupta01:
[...]

PR #6093, opened 2019-08-01
Reply By Tweet Debugging

namangupta01:


namangupta01:
@gauravano can we merge this?

namangupta01:
Restarted travis.

gauravano:
It's failing for all PRs @namangupta01

namangupta01:
Oh okay. So it's not related to mine. We can merge this then?

app/models/comment.rb
```diff
@@ -439,20 +439,58 @@ def self.check_and_add_tweets(tweets)
   end
 
   def self.get_node_from_urls_present(urls)
+    puts ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
+    puts "all urls ->"
+    puts urls
+    puts ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
     urls.each do |url|
+      puts ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
+      puts "url.include? https:// -> #{url.include? "https://"}"
```
seabl: Maybe you could use the start_with? string method instead of include? !

app/models/comment.rb
```diff
@@ -439,20 +439,58 @@ def self.check_and_add_tweets(tweets)
...
+      puts ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
       next unless url.include? "https://"
 
       if url.last == "."
         url = url[0...url.length - 1]
       end
+
```
seabl: And here (lines 452 and 453) you could use the **end_with?** and **chomp** methods

namangupta01:
Hi @seabl , this is just for debugging. After we are done with the debugging we are going to remove it.
@jywarren can we merge this real quick so that we can further investigate what the actual problem is?
Thanks

app/models/comment.rb
jywarren: Thanks for the review, @seabl!

jywarren:
Hi @namangupta01 i'm afraid I'd prefer to not merge a failing build, as it can mask other errors being merged. Can you try helping to solve #6094 do you think, as this is an urgent situation for all contributors? 

Alternatively, is there really no way to test this happening in your dev environment or in unstable? Perhaps @SidharthBansal may be able to suggest a clever way to test it. Just to be clear, is it possible to get working locally in production mode, or in unstable? Thanks!

namangupta01:
I only have a couple of concern while testing on `unstable` which is `unstable` donot have upto date database to sync with the notes mention in tweets and if there are I don't know which one it is. And another concern is does login via twitter works on unstable.

namangupta01:
Sure I will have look at #6094 and try to solve it along with you guys.
Thanks

jywarren:
In the original issue https://github.com/publiclab/plots2/issues/4962 (i
think) I mentioned some specific tweets you can use. The other option is to
try pointing at a different Twitter account on unstable, like your own, and
to test there using a new user account on unstable, and posting a new test
note.

namangupta01:
Hi, I tried this on local with a tweet from one of my account and replied to that tweet from another account and I was able to find the node from that tweet but in Production, I am not able to. I need a couple of information to get it fix and running it on production. First is what is the environment variable `ENV["WEBSITE_HOST_PATTERN"]` setted in there and second is what is the data we store in UserTag when we do login via twitter -- I need this just to confirm that everything will work fine on production because I am not able to run login via twitter on local and get that value. I guess these are the small things that we have to look into it. Afterwards, it will be good to run. 
Also having discussion at https://github.com/publiclab/plots2/issues/6085
Thanks

jywarren:
OK, i'll pull those data now!

WEBSITE_HOST_PATTERN=://publiclab.org/n/

For my twitter oauth i got: `oauth:twitter:15024966` - you can see it on my
profile page!

namangupta01:
Thanks! Is this also same in stable.publiclab.org i.e WEBSITE_HOST_PATTERN=://publiclab.org/n/
Because I am downloading logs from http://stable.publiclab.org/cron_log1.log

jywarren:
Yes, it is the case there too!

namangupta01:
It is weird. When I set the same on local I get the node id extracted from the link.

namangupta01:
Can we also merge this? Since I have tested on local and it's able to identify the node_id easily which it is not getting on production on checking logs? What do you think?

jywarren:
Yes! No danger that we'll fill up logs too fast or anything else to watch out for? Thanks!

jywarren:
I guess it's fine. But please open up an issue pointing to this PR so that we remember to take out these comments too! Thanks!

jywarren:
OK, i hope to publish today but not sure. Thanks!

jywarren:
I'll leave a note in the chatroom if we manage to!

namangupta01:
Thanks!!! I will create a new issue for removing these comments so that we know we have to remove these comments.

namangupta01:
Hoping this could give us enough information to find out what is wrong exactly. Otherwise, we have to dig for more logs.

PR #5525, opened 2019-04-18
try system test to count navbar typeahead results in dropdown

jywarren:
Re: #4515

PR #6094, opened 2019-08-01
chromedriver version bump after Chrome 74 failures

gauravano:
Addressing https://github.com/publiclab/plots2/issues/6095 in the short term, follow up there for longer-term fix

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN apt-key add /tmp/nodesource.gpg.key && apt-get update -qq \
 RUN apt-get install -y fonts-liberation libappindicator3-1 libasound2 libatk-bridge2.0-0 \
                        libatspi2.0-0 libgtk-3-0 libnspr4 libnss3 libx11-xcb1 libxss1 \
                        libxtst6 xdg-utils phantomjs lsb-release
-RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && \
+RUN wget https://chromedriver.storage.googleapis.com/76.0.3809.68/chromedriver_linux64.zip && \
     dpkg -i google-chrome-stable_current_amd64.deb && \
     apt-get -fy install && \
     wget https://chromedriver.storage.googleapis.com/74.0.3729.6/chromedriver_linux64.zip && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ RUN apt-get install -y fonts-liberation libappindicator3-1 libasound2 libatk-bri
 RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && \
     dpkg -i google-chrome-stable_current_amd64.deb && \
     apt-get -fy install && \
-    wget https://chromedriver.storage.googleapis.com/74.0.3729.6/chromedriver_linux64.zip && \
+    wget https://chromedriver.storage.googleapis.com/76.0.3809.68/chromedriver_linux64.zip && \
     unzip chromedriver_linux64.zip && \
     mv chromedriver /usr/local/bin/chromedriver && \
     chmod +x /usr/local/bin/chromedriver
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ RUN apt-get install -y fonts-liberation libappindicator3-1 libasound2 libatk-bri
 RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && \
     dpkg -i google-chrome-stable_current_amd64.deb && \
     apt-get -fy install && \
-    wget https://chromedriver.storage.googleapis.com/76.0.3809.68/chromedriver_linux64.zip && \
+    wget https://chromedriver.storage.googleapis.com/75.0.3770.140/chromedriver_linux64.zip && \
     unzip chromedriver_linux64.zip && \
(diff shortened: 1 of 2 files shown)


jywarren:
Nice, and noting  # https://sites.google.com/a/chromium.org/chromedriver/ has all versions

jywarren:
Noting specific error: `(unknown error: DevToolsActivePort file doesn't exist)`

jywarren:
https://stackoverflow.com/questions/50642308/webdriverexception-unknown-error-devtoolsactiveport-file-doesnt-exist-while-t ...says that `...implies that the ChromeDriver was unable to initiate/spawn a new WebBrowser i.e. Chrome Browser session.`

It suggests: 

> However as per Add --disable-dev-shm-usage to default launch flags seems adding the argument --disable-dev-shm-usage will temporary solve the issue.

linking to https://github.com/GoogleChrome/puppeteer/issues/1834

test/application_system_test_case.rb
```diff
@@ -1,7 +1,7 @@
 require "test_helper"
 
 class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
-  chromeOptions = %w(--headless  --disable-gpu --no-sandbox --remote-debugging-port=9222)
+  chromeOptions = %w(--headless  --disable-gpu --no-sandbox --disable-dev-shm-usage --remote-debugging-port=9222)
```
jywarren: Seeing this change here...

Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN apt-get install -y fonts-liberation libappindicator3-1 libasound2 libatk-bri
 RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && \
     dpkg -i google-chrome-stable_current_amd64.deb && \
     apt-get -fy install && \
-    wget https://chromedriver.storage.googleapis.com/74.0.3729.6/chromedriver_linux64.zip && \
+    wget https://chromedriver.storage.googleapis.com/76.0.3809.68/chromedriver_linux64.zip && \
```
jywarren: We have to match this to our chrome version; see https://github.com/publiclab/plots2/pull/5525#issuecomment-486043530

jywarren: So that should be ok...

jywarren:
Possibly connected https://github.com/karma-runner/karma-chrome-launcher/issues/198

jywarren:
I think we may be seeing a broader issue, re the error `(The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)`

Some have said using `--disable-dev-shm-usage` fixed the problem, but not for us!

jywarren:
Some people are downgrading to `75.0.3770.140`... i'll try this.

Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN apt-get install -y fonts-liberation libappindicator3-1 libasound2 libatk-bri
 RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && \
```
jywarren: I want to change the version here to fix it to a specific chrome version, but I can't find a version-pegged link to substitute instead of `https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb` -- hmm!

jywarren: Maybe https://chromereleases.googleblog.com/ or ah! Here!

https://www.chromium.org/getting-involved/download-chromium

jywarren: Hm. That says old versions are not provided as they lack security updates. It shows a pathway to finding old binaries but it's complicated and I wasn't able to find one for 75 on my first try. I feel like it's not a best practice to point at old chromes, but we're stuck as chromedriver isn't working.

jywarren: OK, so I'm stuck here, i can't find an old version for chrome 75-- so can't make this change:

```
    wget https://chromedriver.storage.googleapis.com/75.0.3770.140/chromedriver_linux64.zip && \
```

Do we just wait a bit for a solution?

jywarren: OK, found a chrome 74 to point to: https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Linux_x64/638880/

jywarren: Can we use https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Linux_x64%2F638880%2Fchrome-linux.zip?generation=1552011881622744&alt=media ?

gauravano: Ok, trying the above link

gauravano:
Yeah, I tried all chrome-driver versions after 73 and none of them seems to work. So, bumping chrome version maybe tried.

alaxalves:
Ok, I have recently dealt with something like this in https://gitlab.com/noosfero/noosfero/merge_requests/1687/diffs#diff-content-0ec615393fffd605e287c92d72dc85217af8d1dd I'll push on top of this.

gauravano:
Great @alaxalves!

gauravano:
Although, I am all eyes here, but please ping me if your fix works, as we want to merge it asap.

alaxalves:
> Although, I am all eyes here, but please ping me if your fix works, as we want to merge it asap. Thank you!

Fixed this here -> https://github.com/publiclab/plots2/pull/6099

gauravano:
Great. Waiting for tests to end :smiley:

PR #6099, opened 2019-08-03
Locking Chrome version

### LINKED ISSUE #6095: Managing ChromeDriver version (long term)
Follow-up to latest issue solved in https://github.com/publiclab/plots2/pull/6094 -- 

and, i see! so dependabot can't update that line. Ideally we can point at a .zip file that's the latest release... let's look into this.

Here's the last time we updated the version number: 

https://github.com/publiclab/plots2/commit/099a1aba9af851c7dd53504395513ab9a4c1ac12#diff-354f30a63fb0907d4ad57269548329e3R36

Linking to: https://github.com/publiclab/plots2/issues/5725, grew out of long-ago issue https://github.com/publiclab/plots2/issues/5683

Noting chromedriver versions can be found here:  https://sites.google.com/a/chromium.org/chromedriver/

alaxalves:
Fixes #6095 (<=== Add issue number here)

Thanks!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,8 +27,9 @@ RUN apt-key add /tmp/nodesource.gpg.key && apt-get update -qq \
 RUN apt-get install -y fonts-liberation libappindicator3-1 libasound2 libatk-bridge2.0-0 \
                        libatspi2.0-0 libgtk-3-0 libnspr4 libnss3 libx11-xcb1 libxss1 \
                        libxtst6 xdg-utils phantomjs lsb-release
-RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && \
-    dpkg -i google-chrome-stable_current_amd64.deb && \
+RUN wget https://github.com/webnicer/chrome-downloads/raw/master/x64.deb/google-chrome-stable_75.0.3770.142-1_amd64.deb \
+          -O google-chrome.deb && \
+    dpkg -i google-chrome.deb && \
     apt-get -fy install && \
     wget https://chromedriver.storage.googleapis.com/74.0.3729.6/chromedriver_linux64.zip && \
     unzip chromedriver_linux64.zip && \
--- a/Gemfile
+++ b/Gemfile
@@ -111,7 +111,7 @@ group :test, :development do
   gem 'phantomjs'
   gem 'puma', '~> 4.0'
   gem 'rails-perftest'
-  gem 'rake',  '~> 12.3.3'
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -27,8 +27,9 @@ RUN apt-key add /tmp/nodesource.gpg.key && apt-get update -qq \
 RUN apt-get install -y fonts-liberation libappindicator3-1 libasound2 libatk-bridge2.0-0 \
                        libatspi2.0-0 libgtk-3-0 libnspr4 libnss3 libx11-xcb1 libxss1 \
                        libxtst6 xdg-utils phantomjs lsb-release
-RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && \
-    dpkg -i google-chrome-stable_current_amd64.deb && \
+RUN wget https://github.com/webnicer/chrome-downloads/raw/master/x64.deb/google-chrome-stable_75.0.3770.142-1_amd64.deb \
```
alaxalves: Since Google doesn't maintain old chrome versions I had to set this one rather than the official FTP. I tried to keep using the official but upgrading the chromedriver wasn't fixing the issue, so I had to set this instead.

gauravano: Yeah, none of the chromedriver is working with chrome v76. We can't depend on this source for long right. But, for now, we can surely work with this. @jywarren can we merge this?

alaxalves: This source is commited to keep the .deb files there. But we could maintain one of our own. Just in case

gauravano: Yeah, maintaining a copy seems like a good idea. But, we will also look forward to using the official source soon, so let's use this one for now.

gauravano:
@jywarren considering the priority of the issue. I am merging this without waiting for another review. 

As we are just using this source for testing in docker, it doesn't threaten any security as I can think of. But, feel free to revert the PR if you think, we shouldn't use un-official source. 

Thank you!

gauravano:
Thank you so much @alaxalves!!

Dockerfile
jywarren: This is great, and lets us switch to different versions under our own decision making process instead of being tied to Google's release schedule, esp since chromedriver stability per version doesn't seem ironclad. Thanks!

jywarren:
Fantastic. Thanks @alaxalves this seems just right. Let's link the long term issue I created to the repo you identified and perhaps others pointing there will be looking for a sustainable solution as well. Thanks @gauravano too!

Issue #6218, opened 2019-08-28
On Profile pages, have the profile picture and bio info stay on top when screen width decreases


ebarry:
**Please describe the desired behavior.**

I would like to see the profile picture and bio info stay on top when screen width decreases. @CleverFool77 I would like your opinion on this observation and this idea, i feel it is a breakout from https://github.com/publiclab/plots2/issues/6097

**Current state**

Currently, when the screen width narrows, the profile pic is pushed quite far down the page under all the topic boxes. See gif: 

![profile-squeeze](https://user-images.githubusercontent.com/161439/63889284-b7673e80-c9ae-11e9-8898-50efb513c802.gif)

**Future desired state**

when the screen narrows, I would like to see, in order: 

1. the profile picture
2. the bio text
3. barnstars
4. the number of types of content contributed . 
5. the map
6. the topic boxes 

Does this make sense?
Thank you for reading!

gautamig54:
Sure @ebarry! This sounds systematic.

CleverFool77:
Ohkk Cool. This would look better. We did something similar in tags page
when we wanted map to be above in smaller screens as its much more
important. I'll fix the order the columns in small screens and let the
second column be first and above the first one in small screen and this
would be fixed.
Ill open a fix for this by today evening.
Thanks !!
cc: @ebarry

CleverFool77:
Correction : It was people section page where we did something similar.
Thanks !!

ebarry:
awesome! thanks for your brainpower on this!

CleverFool77:
Hi @ebarry Just fixed the issue mentioned above.
Thanks !!

ebarry:
wow awesome!!!!! I'm really looking forward to seeing it!!!!!
thank you so much
--

+1 336-269-1539 / @lizbarry <http://twitter.com/lizbarry> / lizbarry.net

CleverFool77:
This is the fix for above https://github.com/publiclab/plots2/pull/6224
Thanks !!

PR #6224, opened 2019-08-30
On Profile pages, have the profile picture and bio info stay on top when screen width decreases

CleverFool77:
Fixes https://github.com/publiclab/plots2/issues/6218

profile


CleverFool77:
Hi @ebarry I have just updated this. Is this what we needed ? Like having the profile image and bio above in smaller screens.

jywarren:
Awesome, and it'll build on stable again now!

PR #6227, opened 2019-08-31
Fix change on hover and other bug-fix

CleverFool77:
Recently, I guess @ebarry mentioned that we don't have links to some pages. At that time, I found that we do have links for most them except few but to follow to new designs, I've changed the underline and other features. 
Thus unless you know before, You'll get never discover that these are the links not just info.
<hr />

Screenshot from 2019-08-31 22-06-23

<hr />

So I added a small css for this. Is hovered, We'll just show blue info and some text-decoration on hover. By this , we'll at least know that these are links. 

<hr />

Screenshot from 2019-08-31 22-11-01

And there is another small fix added in this PR by removing `|` in 

Screenshot from 2019-08-31 22-21-10

and getting this -

Screenshot from 2019-08-31 22-20-53

### DIFF
--- a/app/views/users/profile.html.erb
+++ b/app/views/users/profile.html.erb
@@ -50,7 +50,7 @@
   <h4>
     <small>
       <% if logged_in_as(['admin']) %><%= @profile_user.email %><% end %>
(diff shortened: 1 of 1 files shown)


SidharthBansal:
Well done @CleverFool77

Issue #6088, opened 2019-07-31
Revise the help text in the "choose a password" box on publiclab.org/signup


ebarry:
### What part of publiclab.org is this a new feature idea for?

This feature idea applies to https://publiclab.org/signup while not signed in. 

### What is the idea? 

This idea is to change the help text in the "choose a password" box that someone sees when creating a new account on PublicLab.org. Using the help text to explain the password requirements would be an improvement on what we have now. 

### Details about this idea: 

The text inside the password field should indicate the password requirements that the system will accept. For instance, the current text simply says "enter your new password": 

<img width="711" alt="Screen Shot 2019-07-31 at 3 00 05 PM" src="https://user-images.githubusercontent.com/161439/62240034-1764c800-b3a4-11e9-8477-d367a19bbe86.png">

However, here's the message a person sees if they fail to type a password that meets system requirements; this message explains the password requirements that were never shown on the previous screen: 

<img width="717" alt="Screen Shot 2019-07-31 at 3 03 01 PM" src="https://user-images.githubusercontent.com/161439/62240180-5f83ea80-b3a4-11e9-8bc7-4d190730d2e7.png">

My feature request is that we use the text inside the password field to state the system requirements for the password, as shown in this example from Typeform: 

<img width="293" alt="Screen Shot 2019-07-31 at 3 04 22 PM" src="https://user-images.githubusercontent.com/161439/62240262-90641f80-b3a4-11e9-9ebc-650aa03476c2.png">

### What would need to be done:

- [ ] Find where we have stored the help text shown in the password box
- [ ] Delete the text "enter your new password"
- [ ] Add the text "password must be at least 8 characters"

newbazz:
Or i think, there can be some dynamic checking done of the password while someone is typing. Unable the submit button until it is atleast 8 characters long.

CleverFool77:
> Or i think, there can be some dynamic checking done of the password while someone is typing. Unable the submit button until it is atleast 8 characters long.

Hi @ebarry Wasn't this feature of password validation added earlier ? I did saw the PR I guess.

ebarry:
cool idea @newbazz! 
@CleverFool77 -- thank you for letting me know! I had this experience last week, so if there's a PR already on the way -- amazing! Once it lands, i'll close this issue! Or -- please advise on the best path to proceed 🙏

CleverFool77:
[...]

PR #6256, opened 2019-09-07
Add validation for username,password,email etc

CleverFool77:
Fixes https://github.com/publiclab/plots2/issues/6088

This PR include validation and messages for various cases :
- [X] Username size
- [X]  Password size
- [X]  Password matching
- [X]  Email format

#### For edit profile page
edit

#### For singup
sign

#### For submit signup
sign

#### For submit edit
edit

### DIFF
--- a/app/views/users/_edit_form.html.erb
+++ b/app/views/users/_edit_form.html.erb
@@ -1,5 +1,5 @@
 <div style="padding:10px;">
-<%= form_for @user, :url => {:controller => 'users', :action => 'update'}, :html => {:class => "row"} do |f| %>
+<%= form_for @user, :url => {:controller => 'users', :action => 'update'}, :html => {:class => "row",:id => 'edit-form' } do |f| %>
   <%= hidden_field_tag "user[ui_update]", "true" %>
   <div class="col-md-10">
     <% if f.error_messages != "" %>
@@ -20,14 +20,14 @@
 
       <div class="form-group col-md-6">
         <label for="username"><%= t('users._form.username') %></label>
-        <%= f.text_field :username, { readonly: true, tabindex: 1, placeholder: "Username", class: 'form-control', id: 'username' , title: 'Username field cannot be changed', disabled: true} %>
+        <%= f.text_field :username, { readonly: true, tabindex: 1, placeholder: "Username", class: 'form-control', id: 'username' , title: 'Username field cannot be changed', disabled: true, name: 'username'} %>
         <input type="hidden" name="id" value="<%= @user.username %>" />
       </div>
 
       <div class="form-group col-md-6">
 
         <label for="email"><%= t('users._form.email') %></label>
-        <%= f.text_field :email, { tabindex: 3, required: true, pattern: '[^@]+@[^@]+\.[a-zA-Z]{2,6}', placeholder: "you@email.com", class: 'form-control', id: 'email' } %>
+        <%= f.text_field :email, { tabindex: 3, required: true, placeholder: "you@email.com", class: 'form-control', id: 'email' ,name: 'email'} %>
       </div>
 
     </div>
@@ -46,7 +46,8 @@
       <%= f.password_field :password, { placeholder: "Enter your new password",
                                         tabindex: 4,
                                         class: 'form-control',
-                                        id: 'password',
+                                        id: 'password1',
+                                        name: 'password1',
                                         onpaste: 'return false;' }
       %>
     </div>
@@ -66,6 +67,7 @@
                                                        tabindex: 5,
                                                        class: 'form-control',
                                                        id: 'password-confirmation',
+                                                       name: 'password2',
                                                        onpaste: 'return false;' }
         %>
       </div>
@@ -105,3 +107,9 @@
 
 <% end %>
 </div>
+
+<style>
+.error {
+    color: #D34836;
+  }
+</style> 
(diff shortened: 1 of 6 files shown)


CleverFool77:
Hi @ebarry As we talked about validation of password that while writing it should show that the text `is less than 8 characters `.
I added this in this PR. 

Hi @jywarren Besides the above feature. I have also added password matching and validation for other fields like username, emails. As I noticed they too don't have validation. 
It would make it more interactive to write. And we need not to submit to checks this fields. . 

Thanks!!

CleverFool77:
The test failure is not related to this PR.
Thanks

app/views/users/_edit_form.html.erb
```diff
@@ -20,14 +20,14 @@
...
-        <%= f.text_field :username, { readonly: true, tabindex: 1, placeholder: "Username", class: 'form-control', id: 'username' , title: 'Username field cannot be changed', disabled: true} %>
+        <%= f.text_field :username, { readonly: true, tabindex: 1, placeholder: "Username", class: 'form-control', id: 'username' , title: 'Username field cannot be changed', disabled: true, name: 'username'} %>
         <input type="hidden" name="id" value="<%= @user.username %>" />
       </div>
 
       <div class="form-group col-md-6">
 
         <label for="email"><%= t('users._form.email') %></label>
-        <%= f.text_field :email, { tabindex: 3, required: true, pattern: '[^@]+@[^@]+\.[a-zA-Z]{2,6}', placeholder: "you@email.com", class: 'form-control', id: 'email' } %>
+        <%= f.text_field :email, { tabindex: 3, required: true, pattern: '[^@]+@[^@]+\.[a-zA-Z]{2,6}', placeholder: "you@email.com", class: 'form-control', id: 'email' ,name: 'email'} %>
```
Anupam-dagar: I guess the `pattern` here is used to check if the email entered is in correct format or not on the server side. Since the pattern is defined for the server, using the same pattern for email validation on the frontend side js will be good too. Although the email identifier from the jquery-validation library will work for sure but there can be a mismatch in the pattern used by the library and the patterned defined here.

The jquery validation plugin uses the following regex for the email match.
`/^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/`
https://github.com/jquery-validation/jquery-validation/blob/master/src/core.js#L1398

@gauravano @jywarren Please provide your review on what needs to be done for the email validation here.
We can have the following possibilities:
1. The current implementation is good to go :tada: .
2. Use server side pattern in jquery validation. :robot: 
3. Change server side pattern according to the pattern used by validation library. :computer:

Other than this, the PR looks good to me and ready to be merged.

CleverFool77: Hi @Anupam-dagar Thanks for mentioning this. I forgot to mention this. In the last PR for the same which has been closed and thus I created a new PR.
* I mentioned this too that this existing pattern can't check emails like `lekhika@lekhika` which is valid email 
https://stackoverflow.com/questions/20573488/why-does-html5-form-validation-allow-emails-without-a-dot
* It doesn't allow the lekhika@hotmail.engineer because the regex limited the length of domain name to max 6 characters which isn't correct..

CleverFool77:
And Hi Gaurav. you mentioned this in last PR that we should not allow the user to click on button if there is an error or validation is not being satisfied. Should I implement this or let it remain like this only as usually user won't click on button until the error is being shown.

app/views/users/_edit_form.html.erb
Anupam-dagar: I guess then it's good to get the regex changed to the library one since it can accept wide range of emails.

CleverFool77:
Hi @jywarren @gauravano @Anupam-dagar Earlier I was missing one condition that when the error is being shown , then the form should not allow submission of it.
Now I've added an submit handler which doesn't let form to get submit unless all error being shown are fixed.
Thanks !!

app/views/users/_edit_form.html.erb
CleverFool77: Hi @Anupam-dagar I've removed the `hard-coded regex` present in code and replaced it with the new one present in library. It would let us check more conditions which hard-coded was not able to do so.

CleverFool77:
Hi @gauravano Referencing from this comment https://github.com/publiclab/plots2/pull/5368#issuecomment-487945054, Now the form cannot be submitted unless all errors are fixed.

Thanks !!

CleverFool77:
unrelated system test failing 
```
ERROR["test_viewing_the_settings_page", #<Minitest::Reporters::Suite:0x00007f8a426381d8 @name="SettingsTest">, 9.19095409099998]
 test_viewing_the_settings_page#SettingsTest (9.19s)
ActionView::Template::Error:         ActionView::Template::Error: undefined method `id' for nil:NilClass
            app/views/users/settings.html.erb:10:in `_app_views_users_settings_html_erb__3360030719601865626_70115903272380'
```

restarting the build.
Thanks !!

jywarren:
Great work! Finally passed system tests :-)

Can you give this a quick try on https://stable.publiclab.org to be sure it works? Thank you!!!

PR #6258, opened 2019-09-07
Newly-created user tag cannot be immediately deleted (#6163)

### LINKED ISSUE #6163: Newly-created user tag cannot be immediately deleted (page reload required)
Found while investigating #5431. Reproducible before the design changes to the user tags section. 
Deletion URL is wrong for newly-created tags.

simonastefan:
Newly-created user tag cannot be immediately deleted

Fixes #6163 

Thanks!

### DIFF
--- a/test/system/profile_test.rb
+++ b/test/system/profile_test.rb
@@ -1,0 +1,25 @@
+require "application_system_test_case"
+
+class ProfileTest < ApplicationSystemTestCase
+  Capybara.default_max_wait_time = 60
+
+  test 'create and delete user tag using AJAX' do
+    visit '/'
+    click_on 'Login'
+    fill_in("username-login", with: "Bob")
+    fill_in("password-signup", with: "secretive")
+    click_on "Log in"
+
+    visit '/profile/bob'
+
+    find_link('tags-section', href: nil).click
+    find_link('tags-open', href: nil).click
+    find('.tag-input').set("test_user_tag\n")
+    assert_selector('span', text: 'test_user_tag')
+
+    accept_alert('Are you sure you want to delete it?') do
+      find_link('x', href: /name=test_user_tag\z/).click
+    end
+    assert_no_selector('span', text: 'test_user_tag')
+  end
+end
--- a/test/system/profile_test.rb
+++ b/test/system/profile_test.rb
@@ -15,11 +15,11 @@ class ProfileTest < ApplicationSystemTestCase
     find_link('tags-section', href: nil).click
     find_link('tags-open', href: nil).click
     find('.tag-input').set("test_user_tag\n")
-    assert_selector('span', text: 'test_user_tag')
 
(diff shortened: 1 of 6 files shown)


simonastefan:
@publiclab/reviewers  There is a 404 error on the profile page ("/photos/medium/missing.png") that I think is related to the user profile photo. I had to make Capybara ignore server errors for this system test because it was blocking it. Let me know if I should log an issue.

test/system/profile_test.rb
```diff
@@ -0,0 +1,28 @@
+require "application_system_test_case"
+
+class ProfileTest < ApplicationSystemTestCase
+  Capybara.default_max_wait_time = 60
+
+  test 'create and immediately delete user tag' do
```
gauravano:
```suggestion
  test 'create and delete user tag using AJAX' do
```

gauravano:
Great work @simonastefan!! Can you commit once after commenting  `Capybara.raise_server_errors` lines. Let's see how they work here? 

Thanks!!

test/system/profile_test.rb
simonastefan: Fixed

simonastefan:
Test fails with that 404 error. It appears if the user has no profile photo set. Unfortunately, this is the case with test users.

gauravano:
Let's ask @jywarren that if for this case, we should ignore the error or try to fix it?

jywarren:
Ah hm. I think it's OK for now since there is a clear fallback image?

jywarren:
It's really odd though. Where is `missing.png` being called from?

jywarren:
I just made a LOT of changes to the tag form. Does this issue persist?

simonastefan:
The 404 error on profile page is gone but unfortunately I [cannot add a user tag](https://user-images.githubusercontent.com/14910846/65717896-0639ee00-e0ab-11e9-8946-6f6f723bce51.gif) any more. Same on local with master up-to-date. I could look into this.

jywarren:
That would be tremendous! I fixed all regular tagging (of nodes) and did my
best not to break profile tagging. We should perhaps separate all the
functions out instead of doing multi-purpose code here, as they keep
breaking one another :-/

Thank you for your help!!!

simonastefan:
I tried not to duplicate code and make the function apply to both. The new changes fix both the add and the original delete issue. I had to change a route to make the deletion path for both user_tag and tag use the same format. Let me know if this is not ok.
The system test should make it easier to detect if it breaks again.

I spoke too soon about the 404 error. It still happens so I opened #6344. Travis will fail here until that is fixed.

simonastefan:
@publiclab/reviewers Ready for review

SidharthBansal:
There are some conflicts. Kindly remove them

SidharthBansal:

PR #6334, opened 2019-09-25
Tag card css tweaks

jywarren:
Follow-up to #6322


jywarren:
Hmm, odd: `Sprockets::FileNotFound: couldn't find file 'debounce/index.js' with type 'application/javascript'`

PR #6422, opened 2019-10-08
Refactor JS Function add_tag() and create add_comment() for Javascript API

### LINKED ISSUE #6411: Refactor JS functions add_tag() and create add_comment() for pure JS API
We have a JavaScript function to add tags, here:

https://github.com/publiclab/plots2/blob/86ea8ba5efafad0fcb0befadb399c5a84e0bf786/app/assets/javascripts/tagging.js#L1-L13

However, it uses the on-page HTML tag form to do this. We should be able to do this directly using AJAX without using an HTML form. 

Secondly, we should implement the same functionality for an `add_comment('Comment body here')` function, similarly AJAX based.

Our functional tests show how these should be possible to run directly, instead of via the HTML page:

https://github.com/publiclab/plots2/blob/86ea8ba5efafad0fcb0befadb399c5a84e0bf786/test/functional/comment_controller_test.rb#L43-L53

https://github.com/publiclab/plots2/blob/86ea8ba5efafad0fcb0befadb399c5a84e0bf786/test/functional/tag_controller_test.rb#L9-L12

We should create a system test to confirm that these javascript functions actually work! You can run javascript in a system test like this:

https://github.com/publiclab/plots2/blob/86ea8ba5efafad0fcb0befadb399c5a84e0bf786/test/system/search_test.rb#L26-L27

nstjean:
Fixes #6411  (<=== Add issue number here)

Creating a Pull Request early to get feedback! I'm not sure if I'm doing this right.

Thanks!

### DIFF
--- a/app/assets/javascripts/ajax.js
+++ b/app/assets/javascripts/ajax.js
@@ -1,0 +1,9 @@
+function sendFormSubmissionAjax(dataObj, selector) {
+  $.ajax({
+    url: $(selector).attr('action'), // grab the URL from the form itself
+    data: dataObj,
+    success: (event, success) => {
+      $(selector).trigger('ajax:success', event);
+    }
+  });
+}
--- a/app/assets/javascripts/ajax.js
+++ b/app/assets/javascripts/ajax.js
@@ -1,9 +1,0 @@
-function sendFormSubmissionAjax(dataObj, selector) {
-  $.ajax({
-    url: $(selector).attr('action'), // grab the URL from the form itself
-    data: dataObj,
-    success: (event, success) => {
-      $(selector).trigger('ajax:success', event);
-    }
-  });
-}
--- a/app/assets/javascripts/comment.js
+++ b/app/assets/javascripts/comment.js
@@ -55,11 +55,6 @@ function insertTitleSuggestionTemplate() {
 // JS API for submitting comments
 function addComment(comment, selector) {
   selector = selector || '.comment-form'; // class instead of id because of the bound function on line 3
-  $.ajax({
-    url: $(selector).attr('action'), // grab the URL from the form itself
-    data: { body: comment },
-    success: (event, success) => {
-      $(selector).trigger('ajax:success', event);
-    }
-  });
+  let data = { body: comment };
+  sendFormSubmissionAjax(data, selector);
 }
--- a/app/assets/javascripts/comment.js
+++ b/app/assets/javascripts/comment.js
@@ -53,8 +53,8 @@ function insertTitleSuggestionTemplate() {
 }
 
 // JS API for submitting comments
-function addComment(comment, selector) {
-  selector = selector || '.comment-form'; // class instead of id because of the bound function on line 3
+function addComment(comment, submitTo) {
+  submitTo = submitTo || '.comment-form'; // class instead of id because of the bound function on line 3
   let data = { body: comment };
-  sendFormSubmissionAjax(data, selector);
+  sendFormSubmissionAjax(data, submitTo);
 }
--- a/app/assets/javascripts/comment.js
+++ b/app/assets/javascripts/comment.js
@@ -53,8 +53,11 @@ function insertTitleSuggestionTemplate() {
 }
 
 // JS API for submitting comments
-function addComment(comment, submitTo) {
+function addComment(comment, submitTo, parentID = 0) {
   submitTo = submitTo || '.comment-form'; // class instead of id because of the bound function on line 3
   let data = { body: comment };
+  if (parentID)  {
+    data.reply_to = parentID;
+  }
   sendFormSubmissionAjax(data, submitTo);
 }
--- a/app/assets/javascripts/comment.js
+++ b/app/assets/javascripts/comment.js
@@ -51,3 +51,15 @@ function insertTitleSuggestionTemplate() {
   var newText = currentText+template;
   element.val(newText);
 }
+
+// JS API for submitting comments
+function addComment(comment, selector) {
+  selector = selector || '.comment-form'; // class instead of id because of the bound function on line 3
+  $.ajax({
+    url: $(selector).attr('action'), // grab the URL from the form itself
+    data: { body: comment },
+    success: (event, success) => {
+      $(selector).trigger('ajax:success', event);
+    }
+  });
+}
--- a/app/assets/javascripts/submit_form_ajax.js
+++ b/app/assets/javascripts/submit_form_ajax.js
@@ -12,7 +12,7 @@ function sendFormSubmissionAjax(dataObj, submitTo) {
     url: url,
     data: dataObj,
     success: (event, success) => {
-      if (url != submitTo) {
+      if (url !== submitTo) {
         $(submitTo).trigger('ajax:success', event);
       }
     }
--- a/app/assets/javascripts/submit_form_ajax.js
+++ b/app/assets/javascripts/submit_form_ajax.js
@@ -1,0 +1,12 @@
+// Takes in a data object that contains the info to be submitted in the form property: dataString
+//   and the selector for a form on the page
+// Allows data to be submitted from anywhere on the page using Javascript without using the form itself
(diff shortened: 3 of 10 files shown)


nstjean:
@cesswairimu  Can you (or any of the mentors!) check my code to make sure I'm doing this correctly? Thank you!!

cesswairimu:
Sure, not sure why travis is failing....taking a look

nstjean:
@jywarren Can you take a look at my code and let me know if this is what you intended? I wasn't sure if you wanted to reference the form at all in the Javascript.

nstjean:
AddComment is done!

I had to rebase in branch before pushing, I don't know if I did it right. Now it's showing me a bunch of codeclimate errors.

jywarren:
Wow, it looks like this may need some rebasing and cleaning up! @nstjean this can be a bit rough, but please save a local copy of your branch with `git branch refactor-js-backup` so nothing goes wrong. 

Would you like to try rebasing again over the latest master branch? I can also help a bit with this!

There's some guidance here, however!

https://publiclab.org/wiki/contributing-to-public-lab-software#Rewinding+the+master+branch

jywarren:
Also, i think this needs parentheses to disambiguate the `||` logic, just a note!

image

The rest of the CodeClimate changes can be ignored for now, as they're mostly stylistic!

jywarren:
Thanks for sticking with this one! It's a really awesome one! Just ping me if you get super stuck!

nstjean:
Yes I'll try rebasing using rewinding the master branch!

nstjean:
@jywarren I reset the HEAD, switched to my branch, rebased. But now when trying to run tests on my local everything is failing, even in Master. Do I have to do something else?

Screenshot Test Errors

nstjean:
When I try to run my test for my javascript function I get this error, which points to a Chrome version error. But I haven't changed anything other than pulling the new version of Master.
```
ERROR["test_adding_a_tag_via_javascript", #<Minitest::Reporters::Suite:0x00005587bc60d350 @name="CommentTest">, 1.5917023420042824]
 test_adding_a_tag_via_javascript#CommentTest (1.59s)
Selenium::WebDriver::Error::SessionNotCreatedError:         Selenium::WebDriver::Error::SessionNotCreatedError: session not created: This version of ChromeDriver only supports Chrome version 78
            test/system/comment_test.rb:9:in `block in <class:CommentTest>'
```

app/assets/javascripts/ajax.js
```diff
@@ -0,0 +1,9 @@
+function sendFormSubmissionAjax(dataObj, selector) {
```
jywarren: This is looking good! Perhaps this file needs a more descriptive name, however; but is it a generic utility? If so, maybe can add a sentence or two on how to use it?

Thanks!

nstjean: Yes, I should do that!

jywarren:
This looks good!

Not sure where this error is coming from though:

```
ERROR["test_find_by_tagname_and_user_id", #<Minitest::Reporters::Suite:0x00005653ad056bb0 @name="NodeTest">, 2.091748925999923]
 test_find_by_tagname_and_user_id#NodeTest (2.09s)
NoMethodError:         NoMethodError: undefined method `title' for nil:NilClass
            test/unit/node_test.rb:529:in `block in <class:NodeTest>'
```

That's for this test:

https://github.com/publiclab/plots2/blob/8c2713fa10bdd023f193e3957f2f047560f20e91/test/unit/node_test.rb#L529

... i guess it seems unrelated? I'll try restarting the unit tests to see what happens..

Thanks for the epic rebase!!!!

test/system/tag_test.rb
```diff
@@ -0,0 +1,27 @@
+require 'test_helper'
```
jywarren: Wow, really awesome tests!!!!

test/system/comment_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    fill_in("password-signup", with: "secretive")
+    click_on "Log in"
+
+    visit "/wiki/wiki-page-path/comments"
+
+    # run the javascript function
+    page.evaluate_script("addComment('fantastic four')")
+
+    # check that the tag showed up on the page
+    assert_selector('#comments-list .comment-body p', text: 'fantastic four')
```
jywarren: 🎉

app/assets/javascripts/submit_form_ajax.js
```diff
@@ -0,0 +1,12 @@
+// Takes in a data object that contains the info to be submitted in the form property: dataString
+//   and the selector for a form on the page
+// Allows data to be submitted from anywhere on the page using Javascript without using the form itself
+function sendFormSubmissionAjax(dataObj, selector) {
+  $.ajax({
+    url: $(selector).attr('action'), // grab the URL from the form
```
jywarren: Could we offer a parameter like `url` which defaults back to this value? That way it could be done without a form!

jywarren:
Could we... Refactor the form so it submits through ajax only? I am not
that worries about the ambiguity but I would be fine consolidating
everything into a single remote call a dropping the direct use of the html
form if it's bothering you! Thanks!

nstjean:
Okay! :)

nstjean:
Ok I allowed either a selector or a url to get passed in and it decides what to do from there. :) I also added a test case for a URL being used in both addComment and addTag. Tests work on local.

app/assets/javascripts/comment.js
```diff
@@ -51,3 +51,10 @@ function insertTitleSuggestionTemplate() {
   var newText = currentText+template;
   element.val(newText);
 }
+
+// JS API for submitting comments
+function addComment(comment, submitTo) {
```
jywarren: Great, and this works appropriately for nested comments (responding to other comments) as well as top-level comments? Just checking!

test/system/tag_test.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    # run the javascript function
+    page.evaluate_script("addTag('bluebell')")
+
+    # check that the tag showed up on the page
+    assert_selector('.tags-list .tag-name', text: 'bluebell')
+
+  end
+
+  test 'adding a tag via javascript with url only' do
```
jywarren: Love this. Just one more question - does adding tags to your profile also still work? I think that might've become broken at some point because we (unfortunately) share code between node tags and user tags, but just curious if this touches that at all? Thanks!

nstjean:
Let me check on those possibilities!

jywarren:
Thanks. What's really nice about your code is that we will be moving away from more brittle implementations like the current one, and backing it up with tests, so the hope is to have fewer "gotchas" like this in the future. Thanks a lot!!!

jywarren:
For what it's worth, i just tried adding a profile tag and got some kind of error i think... it submitted via XHR (ajax) to: https://publiclab.org/profile/warren?remote=true&name=providence also which I think isn't right!

jywarren:
So, what we can do is merge this, pending the comment-responses clarification above, and then in a new PR (if you're willing!) try to fix the profile tagging -- preferably by extending the methods you've written, to have a more rational system for handling profile tagging, too! 

The nice thing is that we are trying to add profile tags at different places around the site, so having a general purpose JS method to do this will be really great.

nstjean:
Ok! I added the ability to add a parent comment ID number to the function call so it will nest it appropriately... and added a test to check.

I can check on the profile tag implementation in another PR!

jywarren:
Awesome ok this is ready to merge then?

nstjean:
It's throwing an error on my test test_adding_a_tag_via_javascript_with_url_only - but it works when I run it on local. I'll try to figure out what's going on with it.

nstjean:
It appears I have fixed it! Should be all set!

jywarren:
Super! Now go take a break! 😁🎉🙌🏽

You'll be able to test it out at stable.publiclab.org soon once it builds!

nstjean:
Thanks! :tada: Enjoy your Thanksgiving break!

PR #6472, opened 2019-10-15
Add questions and comments to profile cards

### LINKED ISSUE #6390: Addition of questions and comments in topic cards in profile page
### Description
We have notes added to the topic cards for that particular tag. We need Questions/comments too. So we need to add this. There are two portions -
- [ ] First we need to add questions and comments in cards
- [ ] Second we need to add the no of questions/comments for that topic in the cards
Similar as notes given below.

Screenshot from 2019-10-04 12-11-18

We have to follow the style draft. Here is the screenshot of the change that is need to be done.
The code of the `question`, `comments` part would be similar to that of `notes` part.
The code of the notes part is already present in the codebase in file [app/views/tag/_profileCard.html.erb](https://github.com/publiclab/plots2/blob/master/app/views/tag/_profileCard.html.erb)

Screenshot from 2019-10-04 12-28-46

Thanks !!

nstjean:
Fixes #6390  (<=== Add issue number here)

- [x] add number for questions posted with that tag/author
- [x] add number for comments posted on nodes with that tag/author
- [x] fix link for notes by tag
- [x] add link for questions by tag
- [x] add link for comments by tag, add route and function in users_controller
- [x] fix style according to style guide
- [x] add link in profile sidebar to all users' comments
- [x] refactor Node.find_by_tag_and_author
- [x] add unit test for Comment.find_by_tag_and_author
- [x] add unit test for Node.find_by_tag_and_author

Thanks!


nstjean:
This is what it looks like currently:
FireShot Capture 010 - 🎈 Public Lab_ user - localhost

nstjean:
This is what it looks like now. :) It all works. I'm not sure if we want to keep the username after each post link, since all of them on this page are by this user.

Please take a look, I'm open to all suggestions and comments!

FireShot Capture 015 - 🎈 Public Lab_ user - localhost

jywarren:
Wow, this is looking really awesome. A few requests if it's OK!

1. could we keep the card background all white? 
2. can we make the words lower case, like `2 questions` ?
3. What about, for compactness, just `1 note | 4 questions | 2 comments by @author` ?

I think that, given that we see these types of cards in other places around the site, it's worth keeping `by @author` in this view, just to make it clear that it's a little different here. 

Also, where do the links for comments lead to? Is there a view for comments by user by topic yet? I know we'd planned this but didn't think it existed yet!

app/views/tag/_profileCard.html.erb
```diff
@@ -1,41 +1,46 @@
 
-<div class="card-view">
+<div class="card-view profile-card">
```
jywarren: Oh wow so this separate template existed but you've made it match the standard topic card! Awesome!

config/routes.rb
```diff
@@ -225,6 +225,7 @@
   get 'dashboard' => 'home#dashboard'
   get 'comments' => 'comment#index'
   get 'profile/comments/:id' => 'users#comments'
+  get 'profile/comments/:id/tag/:tagname' => 'users#comments_by_tagname'
```
jywarren: oh my gosh!!!! amazing.

app/models/comment.rb
```diff
@@ -475,4 +475,13 @@ def render_body
     end
     body
   end
+
+  def self.find_by_tag_and_author(tagname, userid)
```
jywarren: Woohoo!!!! 🎉 

I wonder if we could write a simple unit test for this great method??? Amazing work!

app/models/node.rb
```diff
@@ -877,6 +877,42 @@ def self.questions
         .group('node.nid')
   end
 
+  # finds nodes by tag name, user id, and optional node type
+  def self.find_by_tag_and_author(tagname, user_id, type = 'note')
```
jywarren: Interesting -- very cool! @cesswairimu and I are looking here and I thought we already had a way to do this... maybe? In the controller for https://publiclab.org/tag/water-quality/author/mimiss ? Just wondering where this new code is being used, and if it is possibly redundant with that route? Thanks for helping me understand, this is a big awesome PR!

nstjean: I didn't see anything that matched, but I'll take a look at that page and see what it uses!

app/models/comment.rb
nstjean: Yes, sure!

config/routes.rb
nstjean: > Wow, this is looking really awesome. A few requests if it's OK!
> 
> 1. could we keep the card background all white?
> 2. can we make the words lower case, like `2 questions` ?
> 3. What about, for compactness, just `1 note | 4 questions | 2 comments by @author` ?
> 
> I think that, given that we see these types of cards in other places around the site, it's worth keeping `by @author` in this view, just to make it clear that it's a little different here.
> 
> Also, where do the links for comments lead to? Is there a view for comments by user by topic yet? I know we'd planned this but didn't think it existed yet!

I'll make those changes!

app/models/node.rb
nstjean: So yes, there is similar logic in TagController#show_for_author, but the variables are passed in via the URL. I'll have to figure out how to refactor my code to use this.

nstjean: So the issue with this is that I can pull the logic out into a separate function to use in TagController#show_for_author but the problem is that this function doesn't just generate one list of notes, it creates one for notes, questions, answered questions, wikis and nodes. Several of these lists are referenced in other view templates... basically trying to clean it up breaks a bunch of other things.

nstjean:
I have refactored Node.find_by_tag_and_author to call on Node.questions and be simplified.

I have also updated the look of the links using the translations file:
FireShot Capture 016 - 🎈 Public Lab_ user - localhost

nstjean:
@publiclab/reviewers I'm done with this, unless someone else has suggested changes. :)

jywarren:
This is looking awesome. Fantastic job!

I'm wondering if we should try to get the grey zone to not go outside the boundary of the card, as in this snapshot:

image

Would that be hard to adjust? And perhaps the white stripe below could be eliminated as well? Alternatively we could eliminate the grey tone if these changes are very hard to make. But i like how it highlights this content, so just wondering if it's a hard change to make?

jywarren:
Regarding the | - i think it's OK as it is. It's not perfect but I think the extra division is helpful.

nstjean:
> This is looking awesome. Fantastic job!
> 
> I'm wondering if we should try to get the grey zone to not go outside the boundary of the card, as in this snapshot:
> 
> image
> 
> Would that be hard to adjust? And perhaps the white stripe below could be eliminated as well? Alternatively we could eliminate the grey tone if these changes are very hard to make. But i like how it highlights this content, so just wondering if it's a hard change to make?

Oh wow, my monitor didn't even show that light shade. I adjusted my monitor and removed the grey - it is default in Bootstrap, apparently. It should be a white background now!

jywarren:
Awesome! Great work! Merging this now, fantastic!!!

jywarren:
Yes, congratulations!!!!

Issue #6155, opened 2019-08-19
New front page design!


jywarren:
After a LOT of design work, we've settled on a new design for our front page! It incorporates a lot from [our new Style Guide](https://publiclab.org/notes/warren/05-07-2019/introducing-a-draft-style-guide-for-public-lab), but also with a lot of input from @ebarry @steviepubliclab on our communities team at the Public Lab non-profit. 

Here's the mockup (from our [shared mockups doc](https://docs.google.com/presentation/d/1TCZoTfuhamRVrUak8aDgqJAwSgyhRtZg2Pgacl2_4zc/edit#slide=id.g5a6cac5901_4_0)):

![Screen Shot 2019-08-19 at 2 31 40 PM](https://user-images.githubusercontent.com/24359/63289910-3ad9af00-c28e-11e9-99da-dfce32c0d2fe.png)
![Screen Shot 2019-08-19 at 2 32 06 PM](https://user-images.githubusercontent.com/24359/63289914-3ca37280-c28e-11e9-9312-c3d7bba5a747.png)
![Screen Shot 2019-08-19 at 2 32 29 PM](https://user-images.githubusercontent.com/24359/63289920-4200bd00-c28e-11e9-8fc2-cffeabfeec50.png)

OK, so, it breaks into three sections: 

The top is very much like one of our tag pages (see https://publiclab.org/tag/water-quality), plus a 4-column pane. The 4-column area could be [run from a Feature](https://github.com/publiclab/plots2/blob/master/doc/DATA_MODEL.md#features). We're still working on the final text, so let's use placeholder text for everything in this section for now. 

The middle section is a set of 3 bands, each showing content from a wiki page. The title is the wiki page title, the description is the first paragraph of the wiki page, and people are not quite decided yet -- so we can leave them off in the first implementation. We should fetch these as the most recently 3 pages bearing the tag "front-page" (at least for now).

The bottom is a set of columns showing a wide range of topics (tags). This could be a second Feature (https://github.com/publiclab/plots2/blob/master/doc/DATA_MODEL.md#features). 

So this needs breaking up, and we should do it as a shadow page, I think, before we're ready to publish it -- so at, maybe, `/front-shadow`?

@CleverFool77 @gautamig54 what do you think of this? Would you be interested in trying to implement this new design or are you pretty busy at the end of this month?

Thank you!

CleverFool77:
[...]

PR #6497, opened 2019-10-19
New homepage design

jywarren:
New home page! Solves https://github.com/publiclab/plots2/issues/6155


jywarren:
Oops, `authors` is defined twice:

```
Method Node#authors is defined at both app/models/node.rb:280 and app/models/node.rb:289. NEW
```

jywarren:
Aiming for:

image

jywarren:
```
ERROR["test_basic_node_attributes", #<Minitest::Reporters::Suite:0x000056377b9b1438 @name="NodeTest">, 15.528653303999988]
 test_basic_node_attributes#NodeTest (15.53s)
NameError:         NameError: undefined local variable or method `first' for #<Node:0x000056377b9e9518>
            app/models/node.rb:317:in `summary'
            test/unit/node_test.rb:19:in `block in <class:NodeTest>'
```

jywarren:
missing.png error could be due to the profile pics in the project... We can make sure a missing.png really exists!

jywarren:
Looking good!

jywarren:
Now we have to decide which people to show under each project, and how to identify them (i.e. how to call them up from the database: tags? profile tags? followers of some tag?)

steviepubliclab:
Would it make sense to show profiles in order of who has tagged most to that particular project?

ebarry:
that sounds great @steviepubliclab ! To my understanding, each project has a unique tag, and people will either have posted content to that tag, or have "joined the group" for that tag which results in a profile tag. Should we get more specific in any way?

jywarren:
OK, so for a project page, do we assume that the name of the page is the same as the project tag? Or should we make that customizable as well?

I'm just thinking -- for the project at https://publiclab.org/wiki/the-sand-sentinel-program for example, the tag isn't `the-sand-sentinel-program`, right? If we do want to be able to customize, we could have a power tag like `project:sand-sentinel` you can add to the project page which would override the default `the-sand-sentinel-program` as the "project's tag"?

steviepubliclab:
Currently, the name of the page isn't always the same as the project tag.

jywarren:
OK, so path forward is:

1. default is that a project called `/wiki/ponyhat`, then we look for contributors to the tag `ponyhat`
2. if a project tag and page URL are different, like `/wiki/bagel` and `breakfast`, we use a special powertag on the wiki like `contributors:breakfast` (maybe?) and then it'll display people who've contributed to that tag instead of to `bagel`

SidharthBansal:
I loved the new homepage design

jywarren:
We now just need to make the lead text area into a feature, then do the people listings follow-ups

app/views/home/home.html.erb
```diff
@@ -1,18 +1,123 @@
-<div class="container">
-
 <% cache('feature_home-intro', skip_digest: true) do %>
   <%= feature("home-intro") %>
```
jywarren: This now needs to be updated with a final version of https://gist.github.com/jywarren/ae51e24b4bcc2242ff2d8083838ce41a once published!

jywarren:
Asked on Slack - 

Should I change that to be `project-featured` or something, so that we have more curatorial control? 

Should it show a random sampling from the project pages? 

Should it only show ones that have a lead image?

jywarren:
🎉

PR #6204, opened 2019-08-24
Short-code generator for PublicLab.org 

sagarpreet-chadha:
Fixes https://github.com/publiclab/leaflet-environmental-layers/issues/167

* [ ] PR is descriptively titled 📑 and links the original issue above 🔗
* [ ] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [ ] code is in uniquely-named feature branch and has no merge conflicts 📁
* [ ] screenshots/GIFs are attached 📎 in case of UI updation
* [ ] ask `@publiclab/reviewers` for help, in a comment below

PR #6591, opened 2019-11-01
Flow: Button to post marker + Popup working for multiple maps

sagarpreet-chadha:
Fixes: https://github.com/publiclab/plots2/pull/6204#issuecomment-535252025 and https://github.com/publiclab/leaflet-environmental-layers/issues/274

PR fixes: 
1. Popup working for multiple maps.
2. Adds "add a post that will be added to map" button to each inline map.
3. Same markers are not loaded to map now in inline maps.

To-Do :

 

- [ ] Make 2 new layers - people and content in LEL
- [ ]  add iFrame button
- [ ]  add fullScreen button
- [ ]  use new simplified form mockups
- [ ]  Change marker popup content
- [ ]  Write tests

### DIFF
--- a/app/views/map/_inlineLeaflet.html.erb
+++ b/app/views/map/_inlineLeaflet.html.erb
@@ -6,8 +6,12 @@
   <style>
     #map<%= unique_id %> { width:100%; height:300px; margin: 0; position: relative;}
   </style>
-  <div class="leaflet-map" id="map<%= unique_id %>"></div>
 
+  <div class="leaflet-map" id="map<%= unique_id %>"></div>
+  
+   <p><i><small>
+    Add a <button id="postOnMap<%= unique_id %>" class="btn btn-primary">post</button> that'll appear in this map.
+  </small></i></p>
 
 <script>
     var bounds = new L.LatLngBounds(new L.LatLng(84.67351257 , -172.96875) , new L.LatLng(-54.36775852 , 178.59375)) ;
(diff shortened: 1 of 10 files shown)


sagarpreet-chadha:
@jywarren ...used button as recommended by you :

Screenshot 2019-11-02 at 12 49 11 AM

jywarren:
Let's call it "Add a story +"?

and, is that a standard Bootstrap button? could we use one do you think? Thank you so much Sagarpreet!

I had kind of mocked up a manually created button here for an example of styling: https://publiclab.org/wiki/rimap

jywarren:
Next step is to ensure that the location in the /post form recognizes the tags, right?

sagarpreet-chadha:
We just need to add tags in the form? Do we need to load the map also on that page (3rd step of form)?

app/views/map/_inlineLeaflet.html.erb
```diff
@@ -5,24 +5,40 @@
...
         maxBounds: bounds ,
         maxBoundsViscosity: 0.75
     }).setView([<%= lat %>,<%= lon %>], <%= lat.to_s.length.to_i %> + 6);
 
+    var button<%= unique_id %> = document.getElementById('postOnMap<%= unique_id %>');
+
+    button<%= unique_id %>.addEventListener('click', function(event) {
+      var center_lat = map<%= unique_id %>.getCenter().lat;
+      var center_lng = map<%= unique_id %>.getCenter().lng;
+      window.location.href='/post?tags=lat:' + center_lat + ',lon:' + center_lng + '&title=This post will apppear on maps';
```
sagarpreet-chadha: Hey @jywarren ,

Changed the CSS : 

Screenshot 2019-11-02 at 1 12 53 AM

This line opens the /post form with tags lat:XX and lon:XX 😄 . I have added the next steps in the PR description --- do check them please.

The tags appear as (calculated from map's center) : 

Screenshot 2019-11-02 at 1 29 34 AM

jywarren:
Perfect, is this ready then? Thanks! And yes, we need the map to recognize
these tags and center the form map based on them. But in another pr!
Perhaps it needs to be able to override the given tags too if you move the
form map?

sagarpreet-chadha:
Yes this is ready to merge then! We need to change PL.editor to recognize the lat:XX and lon:XX tags or some other param in URL to show map .

PR #6794, opened 2019-11-18
Revert "turn off begin/rescue loop in mailman_server script"

icarito:
Reverts publiclab/plots2#5840
This is what we are actually running.
It was in response to reports of failed emails.


govindgoel:
@icarito  Please merge there #6872 as @jywarren  said.

Issue #5678, opened 2019-05-07
/inline-maps page showing some errors


jywarren:
https://publiclab.org/wiki/inline-maps seems to be stalling and erroring a bit, and I know @sagarpreet-chadha has worked on this before, so I feel bad bringing it up again, but there are some 

I see a few things going on which I think is what makes this so complicated.

1. This is inconsistent, but I'm seeing a lot of the maps on that page greyed out, as if they're stalling while loading
2. Oddly, when I changed the width of the page, some more markers showed up
3. The ones further down are really stuck (see screenshots below)

Also, this error is shown on the page:

```
Uncaught TypeError: Cannot read property 'featureLayer' of undefined
    at wisconsinLayer (eval at <anonymous> (application-98ba6c8dece3e3db273f0b0692b8692bfdaa79300cf55d6636e3ed1f2f5aa3b1.js:350), 
```

I think this really may be multiple overlapping issues here! Maybe once we get it totally working, we should write a system test to protect this, since there's so much JS/Ruby integration going on here: #5316 

![image](https://user-images.githubusercontent.com/24359/57333182-4e4f7880-70ea-11e9-81b9-e88f00cb4abd.png)

![image](https://user-images.githubusercontent.com/24359/57333174-48f22e00-70ea-11e9-9f51-e93c485f3e7d.png)

sagarpreet-chadha:
More details here : https://github.com/publiclab/plots2/issues/4722

I made a copy of this page as a `note` of the main `wiki` page here : https://publiclab.org/notes/sagarpreet/01-30-2019/inline-map-testing  and this note is working fine . 

I am investigating on why the same content is not working on the `wiki page` : https://publiclab.org/wiki/inline-maps . Most probably it is due to external JS code exclusively coded in /wiki . Thanks :)

jywarren:
ok, awesome! Thanks so much, Sagarpreet!

sagarpreet-chadha:
@jywarren ...i think i found the solution to this. 
Do look at the PR https://github.com/publiclab/plots2/pull/6699 . It is because of some function in `short-code-forms` lib.

jywarren:
Let's double check in production once this publishes!

sagarpreet-chadha:
Yes, you are depolying now?

jywarren:
I'm trying to fix this first:
https://github.com/publiclab/plots2/pull/6581/ but
either tonight or tomorrow i hope!

jywarren:
[...]

PR #6802, opened 2019-11-18
map API using common global variable.

sagarpreet-chadha:
Fixes https://github.com/publiclab/plots2/issues/5678#issuecomment-555145346 .


jywarren:
Fantastic!

PR #6862, opened 2019-11-26
Remove limit on typeahead tags display length

jywarren:


jywarren:
Hmm this didn't work although it was in the typeahead docs...

Issue #6854, opened 2019-11-26
Error occurs on tag/TAGNAME/stats/ if tagname does not have any posts


nstjean:
**Please note we are preparing to participate in Google Code-in, and have reserved this issue for GCI participants - but we'd love to have your help with another one! Please check out https://code.publiclab.org to see more.**

When someone tries to view the stats page for a tag that does not exist we get an error. It would be really nice if we could display a page that says, for example, "Wiggles not found."

Example: https://publiclab.org/tag/wiggles/stats

In the tag_controller there is the function `stats` that control what content is shown on that page:
https://github.com/publiclab/plots2/blob/833e0b69a7124f61e0f02ea42d549f7871ddd600/app/controllers/tag_controller.rb#L533-L551

In https://github.com/publiclab/plots2/blob/master/app/views/tag/stats.html.erb we will also have to use some IF statements to display the message if there's no content.

nstjean:
@SidharthBansal Here's another one!

SidharthBansal:
We still need 70 more issues. Can you please help me in labelling some more issues?
It will be great help!

nstjean:
Yes, I will add more this weekend!

SidharthBansal:
Published on GCI website

PR #6892, opened 2019-12-03
Custom error page when tagname has no posts

chen-robert:
Resolves https://github.com/publiclab/plots2/issues/6854

I've added a custom error page when the tag cannot be found. 
image

The error was caused by `@tags` being empty, resulting in `@tags.first` being nil. I added an additional if statement for `@tags.size.empty?` in both the ruby code and html, skipping the corresponding calculations and rendering if `@tags` was empty. 

I also refactored `@tags.first.name` into `@tag_name`. 

@nstjean if you could take a look, that'd be great!

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -534,13 +534,18 @@ class TagController < ApplicationController
     @start = params[:start] ? Time.parse(params[:start].to_s) : Time.now - 1.year
     @end = params[:end] ? Time.parse(params[:end].to_s) : Time.now
     tagname = params[:id]
-
+    
+    @tag_name = params[:id]
     @tags = Tag.where(name: params[:id])
-    @tag_notes = @tags.first.contribution_graph_making('note', @start, @end)
-    @tag_wikis = @tags.first.contribution_graph_making('page', @start, @end)
-    @tag_questions = @tags.first.quiz_graph(@start, @end)
-    @tag_comments = @tags.first.comment_graph(@start, @end)
-    @subscriptions = @tags.first.subscription_graph(@start, @end)
+
+    if @tags.size != 0 
+      @tag_notes = @tags.first.contribution_graph_making('note', @start, @end)
+      @tag_wikis = @tags.first.contribution_graph_making('page', @start, @end)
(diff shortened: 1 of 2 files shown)


app/controllers/tag_controller.rb
```diff
@@ -534,13 +534,18 @@ def stats
...
+    
+    @tag_name = params[:id]
     @tags = Tag.where(name: params[:id])
-    @tag_notes = @tags.first.contribution_graph_making('note', @start, @end)
-    @tag_wikis = @tags.first.contribution_graph_making('page', @start, @end)
-    @tag_questions = @tags.first.quiz_graph(@start, @end)
-    @tag_comments = @tags.first.comment_graph(@start, @end)
-    @subscriptions = @tags.first.subscription_graph(@start, @end)
+
+    unless @tags.empty?
```
cesswairimu: Hey @chen-robert, thanks for opening this :+1: almost there..just one little improvement that we could add to stop code climate from complaining...lets change this to `return if @tags.empty`

app/controllers/tag_controller.rb
```diff
@@ -534,13 +534,18 @@ def stats
...
-    @tag_comments = @tags.first.comment_graph(@start, @end)
-    @subscriptions = @tags.first.subscription_graph(@start, @end)
+
+    unless @tags.empty? 
+      @tag_notes = @tags.first.contribution_graph_making('note', @start, @end)
+      @tag_wikis = @tags.first.contribution_graph_making('page', @start, @end)
+      @tag_questions = @tags.first.quiz_graph(@start, @end)
+      @tag_comments = @tags.first.comment_graph(@start, @end)
+      @subscriptions = @tags.first.subscription_graph(@start, @end)
+    end
```
cesswairimu: Then we won't need this. Thanks

chen-robert:
Thanks for catching that @nstjean! I accidentally dropped a `</div>` tag, the new commit https://github.com/publiclab/plots2/pull/6892/commits/2dab2fbb47c8803bf2e32df8ac941d938790a012which should fix this issue. 

Could you please take a look again?

chen-robert:
@cesswairimu I have also implemented your suggestions, it now passes codeclimate 😄 Could you please review again? Thanks!

PR #6930, opened 2019-12-06
Datauri posting!

### LINKED ISSUE #6594: Allow data-url image upload using file input (from Infragram)
This is a project I'm working on to allow Infragram images to be posted to PublicLab.org easily. They are, for better or worse, data-urls. So i want to make a JS call or a URL that opens https://publiclab.org/post with some parameters (GET or POST) that then make the editor open with the image already set as the lead image. 

https://stackoverflow.com/questions/4998908/convert-data-uri-to-file-then-append-to-formdata

I think we can use this to intercept the data-url (in a POST because GET can't accommodate long data-uris) and then set the main image input to that. It's a stretch but i think it could work. 

This failed: https://stable.publiclab.org/post?main_image=data%3Aimage%2Fpng%3Bbase64%2CiVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8%2F9hAAAAAklEQVR4AewaftIAAAApSURBVKXBAQEAAAiAIPP%2F5mqEMAtHIJFEEkkkkUQSSSSRRBJJJJFEEj1UXgIdx5M8pQAAAABJRU5ErkJggg%3D%3D

(as a first attempt)

jywarren:
Fixes #6594

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,6 @@
 include ActionView::Helpers::DateHelper # required for time_ago_in_words()
 class ApplicationController < ActionController::Base
-  protect_from_forgery
+  protect_from_forgery unless: -> { hostname_excepted }
   layout 'application'
 
   helper_method :current_user_session, :current_user, :prompt_login, :sidebar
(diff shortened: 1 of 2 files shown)


jywarren:
Uh oh, whoops! Need to switch it to the origin header, not the request host... coming up in a sec!

jywarren:
This might be impossible as currently imagined! See https://stackoverflow.com/questions/22397072/are-there-any-browsers-that-set-the-origin-header-to-null-for-privacy-sensitiv#22625354 -- the Referer header is not always there (esp cross http/https) and the Origin header is changed to `null` in some circumstances, which seems to be happening to us.

PR #6942, opened 2019-12-09
PublicLab scroll-to-top issue#6435

Asiedu13:
Fixes #6435

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [ ] PR is descriptively titled 📑 and links the original issue above 🔗
* [ ] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [ ] code is in uniquely-named feature branch and has no merge conflicts 📁
* [ ] screenshots/GIFs are attached 📎 in case of UI updation
* [ ] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software 

Thanks!

PR #6944, opened 2019-12-09
Scroll top fix @publiclab

### LINKED ISSUE #6435: Remove "scroll to top" arrow key
Please Note:  **We are preparing to participate in Google Code-in, and have reserved this issue for participants in GCI - but we'd love to have your help with another one! Please check out https://code.publiclab.org to see more**

### Problem

We can find that the scroll-to-top arrow key also appears when we are at the top of the page.
[here](https://publiclab.org/) 

Screenshot (21)_LI

hence, we should add some javascript scroll function so that it becomes visible only when the user starts to scroll the page

Asiedu13:
Fixes #6435  (<=== Add issue number here)
Please review or comment@publiclab
Thanks!

### DIFF
--- a/scroll-func.js
+++ b/scroll-func.js
@@ -1,0 +1,7 @@
+let scrollBtn = document.querySelector(".back-to-top");
+window.addEventListener("scroll", function(){
+      let position = window.scrollY;
+    if(position < 200){
+      scrollBtn.style.opacity = "0";
+    }else{scrollBtn.style.opacity = "1";}
+})
--- a/scroll-func.js
+++ b/scroll-func.js
@@ -1,7 +1,10 @@
 let scrollBtn = document.querySelector(".back-to-top");
 window.addEventListener("scroll", function(){
       let position = window.scrollY;
-    if(position < 200){
+      if(position < 200){
       scrollBtn.style.opacity = "0";
-    }else{scrollBtn.style.opacity = "1";}
+    }
+      else{
+            scrollBtn.style.opacity = "1";
+      }
 })


scroll-func.js
```diff
@@ -0,0 +1,7 @@
+let scrollBtn = document.querySelector(".back-to-top");
+window.addEventListener("scroll", function(){
+      let position = window.scrollY;
```
seabl: @Asiedu13 Could you fix the indentations please, and make them uniform.

Asiedu13: Sure

IshaGupta18:
This looks good! @jywarren @SidharthBansal please merge it if you find it okay!

cesswairimu:
Hi all, is the scrollbar ever visible locally? I pulled the changes and could not see it...Thanks all

anthony-zhou:
@Asiedu13 Please see my comment on PR #6942 for some tips on integrating this script so that it can be used on the site and seen locally. Thanks!

SidharthBansal:
Can we merge this @cesswairimu

cesswairimu:
Yes we can

cesswairimu:
Thanks everyone

SidharthBansal:
Thanks all

scroll-func.js
```diff
@@ -0,0 +1,10 @@
...
+      if(position < 200){
+      scrollBtn.style.opacity = "0";
```
frappelatte28, line 5: @cesswairimu should not we use "visibility=hidden" instead of "opacity=0" as in this case we still are able to interact with the "scrollBtn"

cesswairimu: @frappelatte28 I thought the two are the same...but I am not the best at styling :laughing: ...would you like to give it a shot and open a PR with the visibility option, see how it goes?

Issue #6950, opened 2019-12-11
Identical failing tests on different pull requests. These tests may not work at all


Uzay-G:
I noticed a problem with the Travis CI testing but probably also the testing in general on the repo since yesterday. My [pull request](https://github.com/publiclab/plots2/pull/6948) and [@chen-robert's](https://github.com/publiclab/plots2/pull/6949) have both been getting the same errors on our travis report that seem totally unrelated to the changes of our pull requests. You can see the details [here](https://travis-ci.org/publiclab/plots2/jobs/623312422?utm_medium=notification&utm_source=github_status) and [here](https://travis-ci.org/publiclab/plots2/jobs/623755453?utm_medium=notification&utm_source=github_status).
 
I think this is pretty important if it actually a problem with the testing because then any new pull requests would then be blocked.

What do you think @SidharthBansal and @jywarren?

cesswairimu:
Hey @Uzay-G , yeah, we have been having this problem with travis for a while.. could be it is running some dependent tests before other and hence the failures... most times restarting travis usually does the trick but I agree could be great to investigate and possibly fix it. I will restart both of your builds for the time being. Thanks for opening this.

cesswairimu:
Okay, so these errors are more frequent than usual and none seems too pass after restart...investigating more on these errors and checking recent merges to see if I can point out the root cause

nstjean:
The tests that are failing:

```
 FAIL["test_that_NodeShared_can_be_used_to_convert_short_codes_like_[map:content:lat:lon]_into_maps_which_display_notes", #<Minitest::Reporters::Suite:0x000055c4d437d5b8 @name="NodeSharedTest">, 7.643299305000028]
 test_that_NodeShared_can_be_used_to_convert_short_codes_like_[map:content:lat:lon]_into_maps_which_display_notes#NodeSharedTest (7.64s)
        Expected: 1
          Actual: 0
        test/unit/node_shared_test.rb:153:in `block in <class:NodeSharedTest>'
 FAIL["test_that_NodeShared_can_be_used_to_convert_short_codes_like_[map:tag:blog:lat:lon]_into_maps_which_display_notes,_but_only_those_tagged_with_\"blog\"", #<Minitest::Reporters::Suite:0x000055c4d41fecc8 @name="NodeSharedTest">, 7.680251494999993]
[...]

Issue #6910, opened 2019-12-04
Map markers for nodes only appear after map interaction, not on initial load 


jywarren:
Pages like https://publiclab.org/wiki/sandbox-3 show off the great new features of inline maps, but I noticed that the map markers for content (pages on PL) don't appear on initial load; only after interacting with the map a bit:

Initial load (no black map markers):

![image](https://user-images.githubusercontent.com/24359/70154003-c7fa1580-167d-11ea-9bbb-2ce0ff93b30b.png)

After dragging map:

![image](https://user-images.githubusercontent.com/24359/70153969-b9136300-167d-11ea-8700-55c675018f85.png)

I want to guess that the trigger for loading the remote API call for that layer is not triggered on initial page load, only on map move. Or if it is, it's getting missed perhaps due to load order? 

@sagarpreet-chadha would you mind taking a look here to see if we can adjust this? Thank you!

sagarpreet-chadha:
Yes i think i figured why this is happening! I will work on this issue on
weekend. Thanks.

jywarren:
awesome, thank you Sagarpreet!!!!

jywarren:
@sagarpreet-chadha @nstjean if you happen to put an eye on this that would be great! I bet it's a one-liner to trigger the markers and we just forgot to do an initial page-load trigger.

nstjean:
Sure! I bet that's exactly what the issue is. @sagarpreet-chadha you said you had figured it out?

jywarren:
Just noting here - possible additional issue which we believe is solved, but linking:

- markers don't appear at all now; could be linked to the fix -- https://github.com/publiclab/plots2/pull/7299 - this is the LEL bump that was causing testing errors, we talked about default loading
- https://publiclab.org/notes/warren/11-16-2019/mashapaug-pond - example of this in sidebar map

jywarren:
I think i found the event handlers:

https://github.com/publiclab/plots2/blob/23e83cdc51a8496f23db932054cbe25ee329e821/app/assets/javascripts/leaflet_helper.js#L109-L140

nstjean:
I can fix this, I had to do something similar on `/map/`. I'll make a PR

PR #6951, opened 2019-12-11
Fix new system tests for adding tags to node and user

### LINKED ISSUE #6904: new system test fails due to probably timing issue
https://github.com/publiclab/plots2/blob/3c0273364cb9b2a22169bea32af6f11aad3ade91/test/system/tag_test.rb#L27-L46

These lines seem to fail - esp `assert_selector('.tags-list .card-body h5', text: 'roses')`, in what I think is just a timing failure - that we should wait longer for it to complete, as it's an AJAX request. 

I've seen a failure for this on https://github.com/publiclab/plots2/pull/6758 (which I restarted) and on https://github.com/publiclab/plots2/pull/6858#issuecomment-561337577 (which resolved on restart, but the error output is there):

```
 FAIL["test_adding_a_tag_via_javascript_with_url_only", #<Minitest::Reporters::Suite:0x00007fc289832e00 @name="TagTest">, 116.16007206999996]
 test_adding_a_tag_via_javascript_with_url_only#TagTest (116.16s)
        expected to find css ".tags-list .card-body h5" but there were no matches
        test/system/tag_test.rb:44:in `block in <class:TagTest>'
[Screenshot]: tmp/screenshots/test_viewing_the_dropdown_menu.png
[Screenshot]: tmp/screenshots/test_viewing_the_dropdown_menu.png
[Screenshot]: tmp/screenshots/test_viewing_the_settings_page.png
  43/43: [=================================] 100% Time: 00:03:05, Time: 00:03:05
```

We could lengthen `  Capybara.default_max_wait_time = 60` but there are other subtle ways to fix timings. Oddly, one is to add an `assert_select` as it is supposed to naturally keep looking for up to 60 seconds for stuff to appear on the page when looking for specific CSS/HTML with that method. Here, that doesn't seem to be working.

Referencing some previous timing-related issues which don't seem so helpful, but maybe:

https://github.com/publiclab/plots2/issues/5683 and #6179 (maybe?)

This is an odd one! I wonder what's going on, @nstjean !? Stumped!

nstjean:
Fixes #6904  (<=== Add issue number here)

I have added some logic to the updating of the page after adding tags so now tags added via URL (instead of the form) will also be updated immediately on the page, just like it did when tags were added by the form. I also added a check so that it would only show on the page if you were adding a tag to the same node/user as the one that's being displayed.

I have updated the system tests to check for this instead of refreshing the page. Hopefully this will stop the timing issue we were seeing!

@jywarren

### DIFF
--- a/app/assets/javascripts/submit_form_ajax.js
+++ b/app/assets/javascripts/submit_form_ajax.js
@@ -1,7 +1,7 @@
 // Takes in a data object that contains the info to be submitted in the form property: dataString
 //   and the url to submit to the controller
 // Allows data to be submitted from anywhere on the page using Javascript without using the form itself
-function sendFormSubmissionAjax(dataObj, submitTo) {
+function sendFormSubmissionAjax(dataObj, submitTo, responseEl = "") {
   let url = '';
   if(submitTo.slice(0,1) === "/") {
     url = submitTo;
@@ -12,8 +12,8 @@ function sendFormSubmissionAjax(dataObj, submitTo) {
     url: url,
     data: dataObj,
     success: (event, success) => {
-      if (url !== submitTo) {
-        $(submitTo).trigger('ajax:success', event);
+      if (responseEl !== "") {
+        $(responseEl).trigger('ajax:success', event);
       }
     }
   });
--- a/app/assets/javascripts/tagging.js
+++ b/app/assets/javascripts/tagging.js
@@ -1,8 +1,15 @@
-
 // Taking the prompt+value retrieved in promptTag() or the links in the drop-down menu and populating the form field before submitting it
 // Instead we want to take the tag value and directly submit it with AJAX
-function addTag(tagname, submitTo) {
+// responseEl is the page element that we want the messages being appended to
+function addTag(tagname, submitTo, responseEl = "") {
   submitTo = submitTo || '#tagform';
+  if (responseEl == "") {
+    if(submitTo.slice(0,1) === "/") {
+      responseEl = '#tagform';
+    } else {
+      responseEl = submitTo;
+    }
+  }
   if (tagname.slice(0,5).toLowerCase() === "place") {
     place = tagname.split(":")[1];
     place.replace("-", " ");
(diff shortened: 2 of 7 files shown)


nstjean:
I did have to modify the controllers by sending back the node or user id in the response on success... otherwise there was no way to check if it should be displayed or not! I have updated the tests for those controllers.

jywarren:
Huh, i wonder what happened? These seem unrelated; i'll try restarting them?

``` FAIL["test_that_NodeShared_can_be_used_to_convert_short_codes_like_[map:content:lat:lon]_into_maps_which_display_notes", #<Minitest::Reporters::Suite:0x0000555dec4a74d8 @name="NodeSharedTest">, 14.961872337999978]
 test_that_NodeShared_can_be_used_to_convert_short_codes_like_[map:content:lat:lon]_into_maps_which_display_notes#NodeSharedTest (14.96s)
        Expected: 1
          Actual: 0
        test/unit/node_shared_test.rb:153:in `block in <class:NodeSharedTest>'
 FAIL["test_that_NodeShared_can_be_used_to_convert_short_codes_like_[map:tag:blog:lat:lon]_into_maps_which_display_notes,_but_only_those_tagged_with_\"blog\"", #<Minitest::Reporters::Suite:0x0000555dec165090 @name="NodeSharedTest">, 14.988351848999969]
 test_that_NodeShared_can_be_used_to_convert_short_codes_like_[map:tag:blog:lat:lon]_into_maps_which_display_notes,_but_only_those_tagged_with_"blog"#NodeSharedTest (14.99s)
        Expected: 1
          Actual: 0
        test/unit/node_shared_test.rb:160:in `block in <class:NodeSharedTest>'
```

jywarren:
@sagarpreet-chadha any ideas here? This is layers stuff i guess, but what's going on? Could it be related to #6910 ?

nstjean:
These are the same tests that are failing for others https://github.com/publiclab/plots2/issues/6950

jywarren:
Awesome work! Thanks!

PR #6957, opened 2019-12-12
Add default location to /post

### LINKED ISSUE #6946: Geographic: Set a default location of user's profile location on /post/
My goal here is to:
- check if the currently logged in user has a location saved
- if they do, pre-fill the location form with that info
- present a way for them to remove that info

nstjean:
Fixes #6946 (<=== Add issue number here)

This adds in the user's profile location as a default value for location.

- Updated button to change to "Remove location" when it has a location. It will toggle back and forth. Location values are only saved when map is visible.
- Added a function `getPrecisionFromNum` - I recommend this be placed in leaflet-blurred-location. The precision number is then used to call `setZoomByPrecision` which already exists.
- Should we update the code here https://github.com/publiclab/plots2/blob/7a6d9316193e1cb5b9318aa49dca665d52c54f27/app/views/map/_leaflet.html.erb#L24 with the new calculator, since it appears to be doing a different calculation than what is used in LBL
- PL.Editor does has `lat:` and `lon:` in the constructor, but not `zoom:`. I called `editor.mapModule.blurredLocation.setZoomByPrecision` after the map was contructed to set the zoom level. Perhaps zoom could be added to the constructor?

FireShot Capture 108 - 🎈 Public Lab_ Post - localhost
FireShot Capture 109 - 🎈 Public Lab_ Post - localhost


nstjean:
Still having that NodeShared test issue. :disappointed:

app/views/editor/rich.html.erb
```diff
@@ -404,6 +406,29 @@
       }
     });
 
+    editor.mapModule.blurredLocation.setZoomByPrecision(getPrecisionFromNum("<%= @lat %>"));
```
sagarpreet-chadha: Awesome @nstjean ! 
But this `getPrecisionFromNum` function needs to be removed. I have added more details here: https://github.com/publiclab/plots2/issues/6946

Briefly - the zoom level calculation needs to be inside a function in LBL.
The PL.editor will just take that zoom level as input and just set map to that zoom in map_module of PL.editor.
If zoom is present in the form of zoom:xx in plots2, let's pass that otherwise call the function of LBL to get zoom and then pass it. Thanks 😄 !!!

nstjean: Ahhh, ok! That makes sense!

nstjean:
- I removed `getPrecisionFromNum` function, now using `blurredLocation.getZoomFromCoordinates`

 **https://github.com/publiclab/leaflet-blurred-location/issues/222 must be merged first so it works!** 

- This checks the user for a zoom level, if it exists it uses it on the map. If it doesn't exist it gets a zoom from blurredLocation and then saves it as a powertag `zoom:xx`

app/views/editor/rich.html.erb
```diff
@@ -414,11 +423,43 @@
       }
     });
 
+    <% if @lat and @lon %>
+      <% if @zoom %>
+        editor.mapModule.blurredLocation.setZoom(<%= @zoom %>);
+      <% else %>
+        let zoom = editor.mapModule.blurredLocation.getZoomFromCoordinates(<%= @lat %>, <%= @lon %>);
+        addTag('zoom:'+zoom, '/profile/tags/create/<%= current_user.uid %>');
```
sagarpreet-chadha: Why we are explicitly adding tag here? 
Pl.editor will automatically add zoom tag, right?
Moreover if user changes zoom level, we will have 2 zoom tags? One from here and other from PL.editor.

nstjean: This only calculates zoom in the instances that the user has a lat and lon in their profile but no zoom level saved. So it calculates a zoom and saves it to their profile using the addTag javascript function (I worked on that a few weeks ago). It does not add the tag into the map, just to their profile. :)

sagarpreet-chadha: Oh i see, the name is misleading. Very cool place to put this though :)

nstjean:
@jywarren I think this all set but codeclimate is complaining!

jywarren:
Sorry for slow catchup here, i just resolved that. All good to merge then? Thanks!

jywarren:
Ah, we now have a couple merge conflicts to resolve. Would you mind taking a look? Thanks!

jywarren:
Hmm, looks like this is the error, odd! 

```
ERROR["test_posting_from_the_editor", #<Minitest::Reporters::Suite:0x00007f8ff6c707f8 @name="PostTest">, 77.13617552799997]
 test_posting_from_the_editor#PostTest (77.14s)
Capybara::ElementNotFound:         Capybara::ElementNotFound: Unable to find css ".wk-wysiwyg"
            test/system/post_test.rb:19:in `block in <class:PostTest>'
```

https://travis-ci.org/publiclab/plots2/jobs/631972389#L3767

nstjean:
That's weird. That class is in PublicLab.Editor - I'll try `yarn upgrade` and push again to see what happens.

nstjean:
@jywarren Fixed it! Should be all set to merge!

sagarpreet-chadha:
Awesome!

PR #6961, opened 2019-12-12
Adjust featured sidebar logic

jywarren:
This is because https://publiclab.org/donate is stuck somehow not showing an Edit button - i think because it has a unique URL and the slug/path logic isn't working.


jywarren:
If we can merge this we can test it at https://stable.publiclab.org/donate

Issue #6303, opened 2019-09-16
Correct Sign-up Form Validation message 


SidharthBansal:
**We are preparing to participate in Google Code-in, and have reserved this issue for participants in GCI - but we'd love to have your help with another one! Please check out https://code.publiclab.org to see more.**

# Problem

When a user tries to submit an empty sign up form, and also does not check off the recaptcha box, This output is shown: 

![image](https://user-images.githubusercontent.com/41092741/64667642-7b10f500-d428-11e9-8273-fe91767a290e.png)

# Solution

Instead, we want all the relevant errors ("field cannot be empty" or so) to be listed along with the failed recaptcha verification. 

Each box in this problem is worth some points. We will accept answers that complete one part and not the other, but encourage you to go for all 3. Completion of all three will result in bonus points :)

- [ ] fix the `create` action so that the correct errors show
- [ ] fix the error messages themselves as well. If you see below, there are three different errors for email, but only one of them is correct. 
- [ ] write a test reflecting this fix. Guidelines: the test must fail without your changes, and your changes are what makes it pass.

![image](https://user-images.githubusercontent.com/41092741/64667496-faea8f80-d427-11e9-85b9-725f1c9cb19d.png)

SidharthBansal:
Codelinks 
1. Sign Up form - https://github.com/publiclab/plots2/blob/master/app/views/users/_create_form.html.erb
2. Dynamic form - https://github.com/publiclab/plots2/blob/5c0d96274bd0be15feba69d8f30093082402ace5/lib/dynamic_form/action_view/helpers/dynamic_form.rb
3. Local files - https://github.com/publiclab/plots2/blob/3a96771279cf4fbf31ff6279f99f068a6d7a2a47/config/locales/en.yml

ananyaarun:
This has been added to GCI dashboard, Thanks !

SidharthBansal:
From #3340

chen-robert:
Can I have this task assigned to me?

SidharthBansal:
Are you notdeghost?

chen-robert:
Yes

SidharthBansal:
Go ahead

chen-robert:
Great, thanks!

chen-robert:
@SidharthBansal I've fixed the first two parts but I'm not too sure how to write the tests. Should it be under system, unit, integration, or something else? I would greatly appreciate it if you could point me in the right direction, thanks!

[...]

PR #6967, opened 2019-12-13
Fixed Recaptcha verification error

chen-robert:
Fixes part of https://github.com/publiclab/plots2/issues/6303

I noticed that user.save populated many of the relevant errors. This was not being called when the recaptcha failed because the if statement short circuited the method call.
```ruby
if ((@spamaway&.valid?) || recaptcha) && @user.save
```
To make this code more declarative, I refactored the result of @user.save into a new variable saved. This way, the errors would be populated every time.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -14,7 +14,9 @@ class UsersController < ApplicationController
     using_recaptcha = !params[:spamaway] && Rails.env == "production"
     recaptcha = verify_recaptcha(model: @user) if using_recaptcha
     @spamaway = Spamaway.new(spamaway_params) unless using_recaptcha
-    if ((@spamaway&.valid?) || recaptcha) && @user.save
+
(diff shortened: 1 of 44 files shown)


app/controllers/users_controller.rb
```diff
@@ -14,7 +14,9 @@ def create
     using_recaptcha = !params[:spamaway] && Rails.env == "production"
     recaptcha = verify_recaptcha(model: @user) if using_recaptcha
     @spamaway = Spamaway.new(spamaway_params) unless using_recaptcha
-    if ((@spamaway&.valid?) || recaptcha) && @user.save
+
+    saved = @user.save
```
SidharthBansal: changed saved->saved_user
Write descriptive variable names while working in prod.
Else LGTM

SidharthBansal:
Awesome
Pls check other prs if there is any conflict

PR #6969, opened 2019-12-13
Clean up styling of location form

### LINKED ISSUE #6968: Location form on /post/ has inconsistent styling
This is the mockup from my proposal for publiclab.org/post/

Post_2_with_text_2_BORDERED

It evens up the spacing of the elements. The code needs cleaning up too.

I'll be doing this in a PR very shortly. :)

nstjean:
Fixes #6968 (<=== Add issue number here)

- "learn more" now opens in a new window
- Using `.form-group` for spacing items out
- Using `<label>` and `<p>` in a standard way for each element
- Moved styling to `location.css`

To make the "location privacy here" link open in a new window that will need to be done in `PL.Editor`

The "Remove location" button has been added in another PR #6957

### DIFF
--- a/app/assets/stylesheets/location.css
+++ b/app/assets/stylesheets/location.css
@@ -1,0 +1,16 @@
+.leaflet-marker-icon {
+  background: url("https://raw.githubusercontent.com/pointhi/leaflet-color-markers/master/img/marker-icon-black.png") ; 
+}
+
+#map_content p {
+  margin: 0;
+}
+#map_content label {
+  margin-bottom: .2rem;
+}
+#map_content .form-group {
+  margin-bottom: 1.4rem;
+}
+#map_content .form-group .btn {
+  margin-top: .4rem;
+}
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -35,4 +35,5 @@
  *= require jquery-confirm/css/jquery-confirm.css
  *= require jquery.atwho
  *= require profile
+ *= require location
 */
--- a/app/views/editor/rich.html.erb
+++ b/app/views/editor/rich.html.erb
@@ -21,9 +21,7 @@
 <%= javascript_include_tag('/lib/leaflet-spin/example/leaflet.spin.min.js') %>
 
 <style type="text/css">
-  .leaflet-marker-icon {
-    background: url("https://raw.githubusercontent.com/pointhi/leaflet-color-markers/master/img/marker-icon-black.png") ; 
-  }
+  /* Fix a background-color error on this page. */
   body {
     background-color: white;
   }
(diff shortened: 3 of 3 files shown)


jywarren:
Nice! Is this just code cleanup or UI changes as well? 

unnamed

nstjean:
Ahh, sorry! This it the form on /post/. Both code cleanup and some very minor UI adjustments.

FireShot Capture 114 - 🎈 Public Lab_ Post - localhost

jywarren:
This looks good. One idea that could be added - could we make "by entering coordinates" be a link that shows those two inputs, and hide them by default, to make this area visually simpler? 

And, would you be able to share a screenshot at a mobile device width? You can click the "tablet" icon in the Dev tools console and select a device here:

image

Thank you!

nstjean:
Sure, I like the idea of hiding the coordinates box by default. :) I'll do
that and then add the screenshots.

nstjean:
FireShot Capture 119 - 🎈 Public Lab_ Post - localhost

FireShot Capture 118 - 🎈 Public Lab_ Post - localhost

FireShot Capture 117 - 🎈 Public Lab_ Post - localhost

jywarren:
This is almost perfect - I think if it says "Or _by entering coordinates_"
where only the _text_ is a link, it makes it a little clearer that it can
be clicked, versus it appearing all blue. Alternatively we could add an
underline to make sure people are aware it's a link. I can just imagine
some people not being aware it's clickable... sorry this is such a small
thing but I hope it makes sense! Thank you!

nstjean:
Yes that makes sense, I was wondering if it was going to be obvious enough for people to know what to do! This ok?

FireShot Capture 120 - 🎈 Public Lab_ Post - localhost

jywarren:
Awesome!!! Merging!!

app/assets/stylesheets/location.css
```diff
@@ -0,0 +1,16 @@
+.leaflet-marker-icon {
+  background: url("https://raw.githubusercontent.com/pointhi/leaflet-color-markers/master/img/marker-icon-black.png") ;
```
crisner: @nstjean, is there a reason behind adding this code? This is currently turning all marker icons' background to black.

nstjean: @crisner Looks like I just moved that code from rich.html.rb into a stylesheet. So feel free to remove it if it's causing issues!

sagarpreet-chadha: Aha, yes i think this will overwrite LEL css also. 
@nstjean , ideally we should make this a functionality in PL.editor to change marker color by taking input from user. 
But for now let's move this css back as it may cause issues in future 😄

nstjean: Interesting. I don't actually see it changing the marker color in /post/ - I'm wondering if we can just remove it altogether. @crisner where are you seeing the color change?

crisner: Here:https://publiclab.org/map
publiclab org_map

And here: https://publiclab.org/wiki/unearthing-pvd
publiclab org_wiki_unearthing-pvd

I guess wherever we have maps.

nstjean: Ahhh, now I see where it is!
Fix in PR #7222

crisner: Awesome!

PR #7046, opened 2019-12-26
Use default location lat lon for /map/

### LINKED ISSUE #7045: Fix /map/ when no layers are specified
- Fix route error so /map/ displays the same as /map/:layernames
- If lat and lon isn't specified in the params, display the user's profile location. If the user has no profile location, show a default lat and lon.

This is part of https://github.com/publiclab/plots2/issues/6993

nstjean:
Fixes #7045  (<=== Add issue number here)

- [x] Fix route error so /map/ displays the same as /map/:layernames
- [x] If lat and lon isn't specified in the params, display the user's profile location. If the user has no profile location, show a default lat and lon.

### DIFF
--- a/app/views/map/map.html.erb
+++ b/app/views/map/map.html.erb
@@ -32,14 +32,17 @@
       var user_lon = null;
       var profile_zoom = null;
       <% if @user_lat && @user_lon %>
-        user_lat = <%= @user_lat %>;
-        user_lon = <%= @user_lon %>;
+        user_lat = "<%= @user_lat %>";
+        user_lon = "<%= @user_lon %>";
         profile_zoom = 7;
       <% end %>
 
-      var lat = parseFloat(params.lat) || user_lat || 23;
-      var lon = parseFloat(params.lon) || user_lon || 77;
+      var lat = params.lat || user_lat || 23;
+      var lon = params.lon || user_lon || 77;
       var zoom = parseFloat(params.zoom) || profile_zoom || 3;
+      // These must be parsed after evaluating above or else 0 will not be displayed
+      lat = parseFloat(lat);
+      lon = parseFloat(lon);
 
       var map = L.map("map", {
         maxBounds: bounds,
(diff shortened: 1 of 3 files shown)


nstjean:
When logged in with a profile location saved:
🎈 Public Lab: Map - Google Chrome_006

When not logged in / no profile location:
🎈 Public Lab: Map - Google Chrome_005

nstjean:
I discovered that if a lat or lon of 0 was passed in via the params or by a user's profile that it was evaluating as false and using the default value instead. I have corrected this.

app/views/map/map.html.erb
```diff
@@ -28,9 +28,21 @@
         new L.LatLng(-54.36775852, 178.59375)
       );
 
-      var lat = parseFloat(params.lat)|| 23;
-      var lon = parseFloat(params.lon) || 77;
-      var zoom = parseFloat(params.zoom) || 4;
+      var user_lat = null;
```
sagarpreet-chadha: Hey, i think we should do all these calculations in the controller. From the controller we should pass only lat, lon, zoom, and layer_name. Thanks!

nstjean: Ahhh, that's a good point. Can the controller access the urlHash though? It's in Javascript. I'll take a look at it all and see what I can move to the controller!

jywarren: I think we can't... Without reimplementing urlhash in Ruby w request.fullpath maybe? Do you think it's worth it? What are the pros and cons?

nstjean: My thought is that the Controller is only called once on page load, and the url hash is accessed many times in the map. It makes sense to keep the url hash calculations all on one page.

I do see some other things that can be cleaned up and moved to the controller though!

nstjean:
I made the clean up changes in #7061

jywarren:
Should this be merged now? Before or after #7061? Thanks!!!!

nstjean:
@jywarren  Before, thank you!!!

jywarren:
:-) all good i guess! Thanks!

Issue #6993, opened 2019-12-17
Follow-up tasks for new /map... route


jywarren:
@sagarpreet-chadha updated us that:

> I had implemented a new route /map (https://github.com/publiclab/plots2/pull/6874) with a full screen map with URL Hash implemented. Check this link:

https://publiclab.org/map/mapknitter#lat=60.23981116999893&lon=5.976562500000001&zoom=4

It looks great!

![image](https://user-images.githubusercontent.com/24359/71008012-e9afbf80-20b5-11ea-91ca-b0d840b95800.png)

Follow up tasks:

* [x] use CSS to hide the footer OR (maybe better) let's make the map 90% of window height, but fixed position, so you can scroll down to content lower on the page, since we may add some stuff there later. Not absolute-positioning it may naturally push the footer down. https://github.com/publiclab/plots2/issues/7047
* [x] There seems to be an error when we just go to https://publiclab.org/map with no layers specified - https://github.com/publiclab/plots2/issues/7045

Copied from https://github.com/publiclab/plots2/issues/6801

* [x] Make `/map/_wiki_slug_` routes that source lat/lon/zoom from matching wiki page #7061
* [x] get permalinks on `/map/_wiki_slug_#zoom/lat/lon` pages as you drag #7061
* [x] get notes, wikis, and people all showing on `/map/SLUG` pages
* [x] make “#” link in grey appear on wiki page popups that show `/map/SLUG` permalinks
* [x] add “View map” button to geotagged page sidebars, opening `/map/SLUG` for wikis, and `/map#zoom/lat/lon` permalink for notes #7061
* [x] add a “Add to map +” button on `/map/…` pages to add a note, question or wiki at that location (go to a selection page between these options, at `/post/choose?tags=lat:10,lon:10` -- preserving params!)

![image](https://user-images.githubusercontent.com/24359/71007382-e831c780-20b4-11ea-948c-55d8020df0dd.png)

jywarren:
A couple notes that may be helpful to @nstjean @sagarpreet-chadha @crisner --

https://github.com/publiclab/plots2/blob/570a3bb96ef0c467f9ddd582e0e6b177c2e6703f/app/assets/stylesheets/style.css#L545-L561

These styles can bust out of the Bootstrap margins to become full-width, when applied to divs in the main content column. I used them in https://publiclab.org/wiki/inline-maps and they work on mobile too!

For the `View on a map` buttons, we can use the same style buttons as the "Add a location" buttons:

[...]

PR #7048, opened 2019-12-26
Push footer below map in /map/

### LINKED ISSUE #7047: Push footer below map in /map/
Currently the footer text is all overlayed on top of the map. We want to push it below the map and set the map to approximately 90% of the screen height.

nstjean:
Fixes #7047 (<=== Add issue number here)

I tried using the class `inline-fullwidth-content` as suggested in https://github.com/publiclab/plots2/issues/6993 but for some reason the map won't go full-width no matter what I do. I can get it to go full-width using other methods, but it would be awful nice to be able to use one standard class to achieve it!

This currently has all kinds of footer info in layouts/_footer.html.erb just for testing purposes. I will remove it when I am done!


nstjean:
This is what is happening when I use `inline-fullwidth-content`

FireShot Capture 130 - 🎈 Public Lab_ Map - localhost

nstjean:
How it looks with my overrides:

FireShot Capture 148 - 🎈 Public Lab_ Map - localhost

FireShot Capture 149 - 🎈 Public Lab_ Map - localhost

sagarpreet-chadha:
Yes it looks great! You can remove the testing data from this PR and this is ready to got then!

nstjean:
@sagarpreet-chadha Removed the test data, this is all set to merge!

jywarren:
This is AWESOMEEEEEEEE

PR #7058, opened 2019-12-30
Set map bounds in /post/ and location modal

### LINKED ISSUE #7052: Map allows scrolling out-of-bounds and saves incorrect longitude
When entering a profile location with the popup modal I've noticed that the lon just keep going even when you scroll past the map bounds of -180/180 longitude. If we allow users to keep scrolling their map then we need to correct the numbers to stay within the bounds when we save the info in plots2.

Saving a number out of bounds is causing an issue on /map/ when loading a user location.

Example:

FireShot Capture 139 - 🎈 Public Lab_ admin - localhost

The incorrect number is saved, this should be corrected to [0, 115], in which -245 + 360 = 115

FireShot Capture 140 - 🎈 Public Lab_ admin - localhost

Instead of showing Indonesia in /map/ this is showing somewhere in the middle of the Pacific ocean:

FireShot Capture 141 - 🎈 Public Lab_ Map - localhost

On the modal it also allows scrolling past the edge of the north and south poles but it does not allow the numbers to go beyond -90/90 latitude, so this appears to be a longitude only problem.

FireShot Capture 144 - 🎈 Public Lab_ admin - localhost

I will be looking into this and figuring out if this is plots2, LBL, or LEL?

nstjean:
Fixes #7052 (<=== Add issue number here)

So right now this is the "quick fix" version by setting the map bounds after the map is created. If we want to allow it being set on initialization then we'll have to change add options to both LBL and PL.Editor.

I am a little concerned with how this looks when you get to the edge of the map. There is no visual boundary, so I'm worried that people trying to select Hawaii will have problems doing so and not understand why it's bouncing away from Hawaii when zoomed out. I know this is just how Leaflet works, but it doesn't seem intuitive to a user.

Peek 2019-12-30 10-44

Zoomed in you can select Hawaii.

Peek 2019-12-30 10-45

The same happens when trying to select Fiji, Tonga.

### DIFF
--- a/app/views/editor/rich.html.erb
+++ b/app/views/editor/rich.html.erb
@@ -314,6 +314,7 @@
 
 </div>
 <script>
+
   var editor;
 
   (function() {
@@ -414,6 +415,12 @@
       }
     });
 
+    var bounds = new L.LatLngBounds(
+      new L.LatLng(90, -180),
+      new L.LatLng(-90, 180)
+    );
+    editor.mapModule.blurredLocation.map.setMaxBounds(bounds);
+
     $('#coord_button').click((event) => {
       event.preventDefault();
       $('#coord_input').toggle();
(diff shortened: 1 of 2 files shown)


jywarren:
Fantastic!!!

PR #7078, opened 2020-01-01
Add zoom override to maps in sidebar and top of page

### LINKED ISSUE #7077: Add zoom override to maps in sidebar and top of page
The powertag `zoom:xx` has been added to maps that are being saved, but currently maps displayed in the sidebar and top of page don't check if that zoom tag exists. We need to change it so that it uses zoom if it is specified!

I'll add this. :)

nstjean:
Fixes #7077 (<=== Add issue number here)

- This adds in the zoom tag being passed to the `map/leaflet` partial.
- If the contents of that tag is empty then it uses the default `lat.to_s.length.to_i + 6` as it was using before.

Wiki page that has lat and lon but no zoom tag:

FireShot Capture 152 - 🎈 Public Lab_ Software Outreach - localhost

Wiki page that has lat, lon and zoom tags:

FireShot Capture 153 - 🎈 Public Lab_ Media - localhost

Wiki page that has lat, lon, zoom and place tags:

FireShot Capture 151 - 🎈 Public Lab_ Events - localhost

Note that has lat, lon and zoom tags:

FireShot Capture 154 - 🎈 Public Lab_ Notes on England - localhost

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -90,12 +90,12 @@ module ApplicationHelper
   end
 
   # we should move this to the Node model:
-  def render_map(lat, lon)
-    render partial: 'map/leaflet', locals: { lat: lat, lon: lon, top_map: false }
+  def render_map(lat, lon, zoom = '')
+    render partial: 'map/leaflet', locals: { lat: lat, lon: lon, zoom: zoom, top_map: false }
   end
 
-  def render_top_map(lat, lon)
-    render partial: 'map/leaflet', locals: { lat: lat, lon: lon, top_map: true }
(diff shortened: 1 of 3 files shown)


nstjean:
I noticed that after you pan the map a location marker appears. Should the marker show when the map is first loaded? Or it working as intended?

FireShot Capture 156 - 🎈 Public Lab_ Notes on England - localhost

jywarren:
I think this is a variation of a bug --
https://github.com/publiclab/plots2/issues/6910 -- shall we try to address
that here and there? Thanks!

nstjean:
Oh ok! Yes this is ready for merge!

jywarren:
Hooray!

PR #5954, opened 2019-06-27
Remove coveralls, add codecov

kaustubh-nair:
Partially fixes https://github.com/publiclab/plots2/issues/5931

PR #7089, opened 2020-01-02
Fix map on profile and add check for zoom tag

### LINKED ISSUE #7088: Fix map on user profile
The map at the top of the user profile should show up when the user has a location saved. However it's not showing anything at the moment!

FireShot Capture 164 - 🎈 Public Lab_ admin - localhost

I'll take care of this!

nstjean:
Fixes #7088  (<=== Add issue number here)

I have the map working now, and it's checking for the user's zoom value.

FireShot Capture 163 - 🎈 Public Lab_ admin - localhost

I notice this is very very similar to the maps in the page sidebar. Is there a reason we are using a separate template for this map? Should we consider consolidating into one template?

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -210,9 +210,11 @@ class UsersController < ApplicationController
         @count_activities_attempted = Tag.tagged_nodes_by_author("replication:*", @profile_user).count
         @map_lat = nil
         @map_lon = nil
+        @map_zoom = nil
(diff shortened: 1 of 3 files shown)


jywarren:
I think we can consolidate, yes! I'm going to merge this now to test a theory about CodeCov we're seeing in #5954 -- cc @Uzay-G 

Thank you!

nstjean:
Ok! I'll consolidate in a different PR. :)

PR #7105, opened 2020-01-03
Consolidate userLeaflet.html.erb into leaflet.html.erb

nstjean:
This is continuing from https://github.com/publiclab/plots2/pull/7089

Profile map when user's location is blurred:true

FireShot Capture 166 - 🎈 Public Lab_ admin - localhost

Profile map when user's location is not blurred:

FireShot Capture 167 - 🎈 Public Lab_ admin - localhost


app/views/map/_leaflet.html.erb
```diff
@@ -2,6 +2,8 @@
 <% top_map = top_map || false %>
 <% zoom = zoom || lat.to_s.length.to_i + 6 %>
 <% unique_id = rand(1000) %>
+<% user = user || @user # allow overriding w/ local variable %>
+<% blurred = blurred || false %>
```
Uzay-G: Hey! This is a great change that cleans up a lot of code :+1:. You should use the Ruby [`||=`](https://www.rubyguides.com/2018/07/ruby-operators/#Assignment_Operators) when assigning the `user` and `blurred` variables. It does exactly what you wrote: if the variable being assigned is nil, it gets reassigned, otherwise it stays the same.
Good job!

nstjean: Ahhh, excellent!! Thanks for the tip! I've made the change. :)

jywarren:
Yay, awesome!

PR #7106, opened 2020-01-03
Add threshold to codecov.yml and set certain files to ignore

Uzay-G:
This will add a threshold so that even if  a pull request reduces coverage by less than 1% it will still succeed in codecov. I only added this for the project status for now because I wanted to talk about the patch status. The patch only looks at the changes in the pull request and their coverage without comparing to the general project. You can learn more about this at https://docs.codecov.io/docs/commit-status#section-patch-status. 

I think we need to edit the config for the patch status because for example with css changes it's almost impossible to test and we can remove the patch status for those files who will always be marked with a coverage of 0% relative to the pull request.

What do you think @jywarren?


jywarren:
I think that's right. Even this which has a good test only reached 70% for the patch, and is failing even though it increased coverage for the whole project: https://github.com/publiclab/plots2/pull/7061

(cc @nstjean we're trying to help from here!)

jywarren:
Hmm, oddly this doesn't seem to have worked for project status threshold? -- `codecov/project — 80.68% (-0.13%) compared to d251b5f` still failed. Not sure why!

Uzay-G:
I'm not sure if it takes into account changes until you merge them. I might be wrong though

jywarren:
I think it should... is there a way to see the decision it made in the logs?

Uzay-G:
Where can I view the logs for the test run?

jywarren:
Oh, you can press the red X in the Travis results and it opens the logs!

jywarren:
Or in this case the green check mark, or the word Details above:

Screenshot_20200103-222208

Uzay-G:
Oh yeah I know I just thought there was something even more detailed like with travis

Uzay-G:
I renamed it to codecov.yml after a long search for the bug when I found this stack overflow answer: https://stackoverflow.com/questions/55730544/codecov-doesnt-recognize-flags-from-codecov-yml-config

Uzay-G:
I think we should ignore some of the project directories because they are very difficult provide coverage for like for example stylesheets. Is there anything else you think we should remove?

jywarren:
Let's do but let's merge this first since it's holding up others!

jywarren:
So would next step be to add a very wide threshold (could be like 80%?) for the patch? I haven't seen enough of these to know what's a usual amount of coverage for a PR. We should encourage but not require coverage esp. for very small PRs. I wonder: maybe we turn off the check entirely for patches? I'm thinking of first-timers especially.

Uzay-G:
Yeah I think we should remove the patch coverage because I don't find it that useful especially in cases like CSS changes and ftos. Although the patch status shows a more obvious representation of the coverage diff for each pull request, at the end we can measure the impact almost as well with the loss or gain of coverage for the project. 

Should I remove it?

jywarren:
Take a look at the last few PRs I merged. Maybe it's not necessary?

Uzay-G:
Yeah alright no problem! Are there any other issues or features you think we should enable for the coverage?

jywarren:
Not that I can think of now! Maybe later we can make an issue where we encourage people to find undercovered areas of code to write tests for!

Issue #4268, opened 2018-12-11
Validation error messages for the signup modal


SidharthBansal:
We expect the errors will be shown on the modal instead of redirecting to sign up page.
Please ensure the following:
- [ ] Sign Up form validations are working fine for incorrect input fields(testing)
- [ ] Sign Up modal validations are working fine for incorrect input fields(testing)

SidharthBansal:
It needs to be done after #4209

SidharthBansal:
@kevinzluo may be you will be willing to work on this?
As it is related to your pr #3984

vishvanatarajan:
Hey @SidharthBansal can i work on this issue?

SidharthBansal:
Yes

SidharthBansal:
@vishvanatarajan in case of any help needed, please free to ping me.

vishvanatarajan:
@SidharthBansal okay no problem. I will create a PR and link it as soon as possible. Thank you for the support :+1:

SidharthBansal:
How is it going @vishaldeepak ?

divyabaid16:
@SidharthBansal  Can I work on this issue?

SidharthBansal:
Sure

divyabaid16:
So do I need to use the alert to show the error message or just show it below the field?

SidharthBansal:
You can do it anyways.

SidharthBansal:
Published on GCI dashboard

VladimirMikulic:
@SidharthBansal isn't this issue same as #3439? In both of these issues, we need to validate sign up form/modal. Right?

VladimirMikulic:
I am working on this by the way.

SidharthBansal:
Point 1: Please check whether it is solved or not. If they are solved then ping me for adequate action
Point 2: They are different
Point 3: Create modular PR for different functionality.
Thanks

VladimirMikulic:
> Point 1: Please check whether it is solved or not. If they are solved then ping me for adequate action

It's not solved that's why I am working on it.

> Point 2: They are different

Could you please explain the difference between this issue and issue #3439?

> Point 3: Create modular PR for different functionality.

For this issue, it won't be necessary.

SidharthBansal:
Form is different from modal
Also password confirmation is through js
While validation errors are through rails
So they are different but related I guess

VladimirMikulic:
Whoops, I didn't know that. I picked the wrong task then 😅
[...]

PR #7086, opened 2020-01-02
ADD: Client-side Sign Up form validation

VladimirMikulic:
Changes:
 - Removed realtime_usernmae_validation.js
 - CSS classes for valid and invalid input elements
 - Minor changes in _create_form.html.erb

The old sign up form would only validate a username.
This change provides a nice UI to the user while filling the form.

[Demo](https://streamable.com/lqluk)

Resolves #3439 & #4268

PR #7110, opened 2020-01-04
FIX: Username validation

VladimirMikulic:
When the form is filled(valid) and the user deletes username, submit button is
enabled but should be disabled(missing username).

[Demo](https://streamable.com/0509a)

Minor changes:
 - Add space in the password error message
 - Update system test to match the new password error message

This commit resolves the issue.

Part of #7086


VladimirMikulic:
@jywarren I've provided the tests for this change.

VladimirMikulic:
@jywarren you forgot to close #4268, my PR closes this also.

PR #7125, opened 2020-01-06
[WIP] Redesign of location modal form

### LINKED ISSUE #7031: New design for location popup modal
I'd really like to redesign the popup modal so that it looks and feels like the in-line form at /post/ . Here are two options I came up with. Which do you prefer? Any other suggestions? Open to all ideas!

Also I had a question of whether that Scale slider is needed or not? We already will have zoom buttons and mouse zoom in the map itself.

Option 1: 

Option 1

Option 2:

Option 2

And here's how it would look like on an iphone:

iPhone 6-7-8 – 1

nstjean:
Fixes #7031  (<=== Add issue number here)

This is not complete yet, but feel free to give feedback on this so far!

FireShot Capture 174 - 🎈 Public Lab_ Donate - localhost

I feel like the "recently used locations" make it look clutered, but I understand it's probably quite useful to use them! Should we make them grey to make them stand out less? Move them to the top above the map? Or just leave it as is?

### DIFF
--- a/app/assets/stylesheets/location.css
+++ b/app/assets/stylesheets/location.css
@@ -10,7 +10,49 @@
 #map_content .form-group .btn {
   margin-top: .4rem;
 }
+#map_content .form-group.recent-locations {
+  /* margin-bottom: 1.2rem; */
+}
+#map_content .recent-locations label {
+  margin-bottom: 0;
+}
+#map_content .slide-container label {
+  margin: 0;
+}
+#map_content .slide-description {
+  display: inline;
+  font-size: 0.9rem;
+  color: #808080;
+}
 
+.modal-header {
+  border-bottom: 0;
+  padding-bottom: 0;
+}
+.modal-header .close {
+  padding-bottom: 0.5rem;
+}
+#placenameDisplay {
+  padding: 0;
+  border: 0;
+  color: #212529;
+  font-weight: bold;
+}
+
+.modal-content .badge-group {
+  margin-left: -6px;
+}
+.modal-content .badge.badge-info {
+  background-color: #337ab7;
+  margin-left: 7px;
+  margin-bottom: 2px;
+  padding: 3px;
+}
+.modal-content .badge {
+  background-color: white;
+  color: #337ab7;
+  margin: 0;
+}
 a:not([href]):not([tabindex]).btn-location,
 .btn-location {
   position: inline;
--- a/app/assets/stylesheets/location.css
+++ b/app/assets/stylesheets/location.css
@@ -118,3 +118,7 @@ a:not([href]):not([tabindex]).btn-outline-secondary.btn-location:hover i,
 .btn-outline-secondary.btn-location:hover i {
   color: white;
 }
+
+#coord_button {
+  text-decoration: underline;
+}
--- a/app/assets/stylesheets/location.css
+++ b/app/assets/stylesheets/location.css
@@ -19,7 +19,8 @@
 #map_content .slide-container label {
   margin: 0;
 }
-#map_content .small-description {
+
+#blurred-location-modal .small-description {
   font-size: 0.9rem;
   color: #808080;
 }
@@ -49,9 +50,57 @@
 }
 #blurred-location-modal .modal-footer {
   justify-content: space-between;
+  flex-wrap: wrap;
+}
+#blurred-location-modal #place_name {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+}
+#blurred-location-modal .modal-footer .location-box {
+  flex: 0 1 100%;
+}
+#blurred-location-modal #placenameDisplay {
+  flex: 0 1 100%;
 }
 #blurred-location-modal .modal-footer label {
-  margin-bottom: .1rem;
+  flex: 0 1 100%;
+  margin: 0;
+}
+#blurred-location-modal .modal-footer .button-box {
+  flex: 0 1 100%;
+  display: flex;
+  justify-content: center;
+  margin: 0.5rem 0 0 0;
+}
+
+@media (min-width: 576px) {
+  #blurred-location-modal .modal-footer {
+    justify-content: space-between;
+    flex-wrap: nowrap;
+  }
+  #blurred-location-modal .modal-footer .button-box {
+    flex: 0 0 10%;
+    display: inline;
+    margin: 0 0 0 3rem;
+  }
+}
+
+@media (min-width: 992px) {
+  #blurred-location-modal #place_name {
+    flex-wrap: nowrap;
+  }
+  #blurred-location-modal .modal-footer .location-box {
+    flex: 1 1 50%;
+  }
+  #blurred-location-modal #placenameDisplay {
+    flex: 1 1 60%;
+  }
+  #blurred-location-modal .modal-footer label {
+    flex: 1 1 40%;
+    margin: 0;
+    padding-left: 0.5rem;
+  }
 }
 
 a:not([href]):not([tabindex]).btn-location,
--- a/app/assets/stylesheets/location.css
+++ b/app/assets/stylesheets/location.css
@@ -19,40 +19,41 @@
 #map_content .slide-container label {
   margin: 0;
 }
-#map_content .slide-description {
-  display: inline;
+#map_content .small-description {
   font-size: 0.9rem;
   color: #808080;
 }
 
-.modal-header {
+#blurred-location-modal .modal-header {
   border-bottom: 0;
   padding-bottom: 0;
 }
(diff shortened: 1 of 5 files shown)


nstjean:
On a small screen size:

FireShot Capture 175 - 🎈 Public Lab_ Donate - localhost

jywarren:
Looking nice! For recently used, what if we used a dropdown so they're there but less visible?

Uzay-G:
Yeah it looks really nice! I agree that the recent locations are cluttery. I'm not sure they are useful if they only show latitude and longitude and not the place's name. Maybe we only display the recent locations that have the `name` attribute? I'll make a suggestion on the pull request if you think it's a good idea!

nstjean:
I like the idea of a dropdown, that would be cleaner. I'll try that!

> Yeah it looks really nice! I agree that the recent locations are cluttered. I'm not sure they are useful if they only show latitude and longitude and not the place's name. Maybe we only display the recent locations that have the `name` attribute? I'll make a suggestion on the pull request if you think it's a good idea!

Yes it does look kind of odd not having place names, but I'm not sure not displaying them is the ideal option. Right now nothing is saving with a name, I'll have to figure that out!

sagarpreet-chadha:
Yes dropdown sounds more tidy! 😄

nstjean:
I definitely like this better. :) In my local db my recent locations don't have any names so I can't test how that would look right now.

FireShot Capture 185 - 🎈 Public Lab_ Stats - localhost

jywarren:
It is really nice!!!

Maybe "recent" belongs below, like where "by coordinates" is, and the
biggest prompt should be the search by place name input? But i'm not sure
about that. What do you think?

Uzay-G:
Yeah it looks great! Maybe just one thing is that there are two columns in your design, the map and the settings and I feel that the save map part kind of conflicts with that. Maybe we could move the save button to the right part?

nstjean:
> It is really nice!!! Maybe "recent" belongs below, like where "by coordinates" is, and the biggest prompt should be the search by place name input? But i'm not sure about that. What do you think?

I think search by place name is probably the one that will be used most often so it makes sense to have it at the top.

> Yeah it looks great! Maybe just one thing is that there are two columns in your design, the map and the settings and I feel that the "save map" part kind of conflicts with that. Maybe we could move the save button to the right part?

Hmm, let me try two different things.

nstjean:
@jywarren I'm looking at what the "recent locations" and "save as place name" actually does in the system. It's saving the place name as a `place:xx` tag. This means a few things:
1. it has to be an editable field for the user to type something in
2. there should be an explanation of what that's going to do for the wiki/note to add that place name, maybe a tooltip on hover?
3. It should be removed if this is a location for a user profile because place is not used, correct?

jywarren:
Ah, yes. great. I remember this system - it's a little bit obscure, huh?
Yes, let's do what you're proposing; and one more thing is that I worry
when "unknown location" or whatever the error message is appears in that
field.

As to user profile, I suppose not, although we could add it to that system
too -- but let's think of that as a possible follow-up and not include it
in this.

How does that sound? Thanks for catching this!

nstjean:
Yes I agree, I don't like "unknown location", I initially assumed it was an error message that meant something wasn't working! I'll make the changes. :+1:

jywarren:
I think it's possible we actually screen for that string in plots2... i
have some half-baked memory of this. just FYI!

nstjean:
I did see it on that page, checking the value to see if it was different from `unknown location`. I'll check to make sure it's not somewhere else too!

nstjean:
It's in LBL... this field is supposed to display the google reverse-lookup location.
https://github.com/publiclab/leaflet-blurred-location/blob/550b8ac78144b32e0e01a0dba242a09009f155c8/src/core/Geocoding.js#L5
Which does not appear to be working. Then it would save that value as the `place:xx` tag.

Is this useful behavior? Do we want every map location looked up and saved as a place power tag?

Uzay-G:
I it would be useful if the place power tag was saved but we should make sure it is not saved for users with `location:blurred`. I think we should save the `place` attribute otherwise.

nstjean:
Why shouldn't we save place with `location:blurred`?

Uzay-G:
Because, users with `location:blurred` don't want their location to be publicly displayed and only want approximate coordinates to be shown so they can maintain their privacy.

nstjean:
Ahh. The maps take that into account. When a blurred location is saved the numbers are truncated, so that instead of a specific location like 17.957832 instead it is 17.9 - so their specific location is never saved to our database, just the "blurred" number which gives an approximate location. Also for maps where there is a tag `blurred:true` it does not display a location marker on a specific spot.

nstjean:
Save button on the right in a separate section? I'm not fully sold on the idea, but it's an option.

FireShot Capture 195 - 🎈 Public Lab_ admin - localhost

Or maybe centered but separated - would have to find a spot for the "save as place name"

FireShot Capture 197 - 🎈 Public Lab_ admin - localhost

Uzay-G:
I like the Save button on the right in a separate section design. What do you think @jywarren?

jywarren:
oh, i LOVE the modal footer with the input on the left and the save button
on the right! Great solution!!! Maybe we could put the label text on the
right side of the input, and change it to "How your location will be shown
in text" or something like that?

nstjean:
> It's in LBL... this field is supposed to display the google reverse-lookup location.
https://github.com/publiclab/leaflet-blurred-location/blob/550b8ac78144b32e0e01a0dba242a09009f155c8/src/core/Geocoding.js#L5
Which does not appear to be working. Then it would save that value as the place:xx tag.

So I tried to overwrite the "Location unavailable" but right now it looks like LBL is overwriting anything I put in there. I would prefer for it to just be blank if the function isn't working!

@jywarren  Assuming it was working and returning a location name properly, do we want every location to have a place:xx tag? Is that the desired behavior? Or do we want the user to be able to choose if it gets saved as a place or not?

Uzay-G:
I think it would start becoming a bit too complex if we also ask users if they want the place to be saved on top of the `location:blurred` value. What do you think @jywarren?

jywarren:
I believe the original design was that it generates the `place:____` automatically, but if you modify it, it respects your modification. Could we do that (it may involve an upstream change to LBL) and then if you actually delete it so it's blank, it doesn't save `place:` at all? 

That way the basic interaction is pretty simple, but if you want to customize, you're still able to. 

How does that sound? Thanks all, for this careful thinking!

nstjean:
Yes we can do that! I'm working in LBL right now at https://github.com/publiclab/leaflet-blurred-location/pull/232

jywarren:
awesome.

sagarpreet-chadha:
Great work so far @nstjean !!!

nstjean:
Pics of the modal at the different breakpoints:

FireShot Capture 213 - 🎈 Public Lab_ four - localhost

FireShot Capture 214 - 🎈 Public Lab_ four - localhost

FireShot Capture 215 - 🎈 Public Lab_ four - localhost

jywarren:
This looks really great. I think it's basically ready if you agree. If you'd like to make one small tweak, i think a `»` character (in HTML it's `&raquo;`) after "entering coordinates" could make it seem more actionable. The color change isn't a strong indicator, esp. without an underline. Underlining could also work, i just want people to know they can click it!

Thanks! Ping me when it's ready to merge, this is awesome!!! Fantastic job!

nstjean:
Oh yes! In other pages the `»` is used to indicate opening in a new window, so I'll go with underline in order to match the form on /post/ .

Or maybe we could change the modal and /post/ forms to have an arrow down?
FireShot Capture 221 - Angle Down Icon - Font Awesome - fontawesome com

nstjean:
Here's with an underline.

FireShot Capture 224 - 🎈 Public Lab_ admin - localhost


I currently don't have "save as placename" on location for a user. Do you want that turned on or no?

Uzay-G:
> 
> 
> Oh yes! In other pages the `»` is used to indicate opening in a new window, so I'll go with underline in order to match the form on /post/ .
> 
> Or maybe we could change the modal and /post/ forms to have an arrow down?
> FireShot Capture 221 - Angle Down Icon - Font Awesome - fontawesome com


We could also use `&darr;` to have a downwards arrow like this: ↓

jywarren:
> I currently don't have "save as placename" on location for a user. Do you
want that turned on or no?

Yes, let's do, please! Thank you!

nstjean:
Ok I've enabled it for profiles as well!

jywarren:
Great, let's do this!!! 🙌

sagarpreet-chadha:
🎉

Issue #6297, opened 2019-09-15
Issues with cross-wired comment preview buttons - break this up


sashadev-sky:
# Problem

Heres the page used for gif and screenshot: https://publiclab.org/questions/nickyshen0306/09-12-2019/how-to-design-a-camera-for-calculating-endvi

Easier to just show... the preview button on an editor should not be coupled to multiple editors on the page:

![exx](https://user-images.githubusercontent.com/41092741/64926446-8e76e400-d7cb-11e9-9069-75a94e3f645c.gif)

Besides being unusual behavior, it results in a lot of other problems like losing your text, confusing output (Has previously been documented in #3534). 

<br>

The issue continues:  this is a screenshot of my page with 3 editors on it at a time, of two different varieties. The top editor is to comment on the last comment thread on the page, the middle editor is to comment on my own comment, and the bottom editor is to create a new comment: 

<img width="513" alt="Screen Shot 2019-09-15 at 3 05 30 PM" src="https://user-images.githubusercontent.com/41092741/64926458-c5e59080-d7cb-11e9-8441-cd8d1b711782.png">

# Solution 

We may want to take start by removing the functionality to comment on comments and just leave creating new comments with one editor (just remove it front-end, not back-end data). When we're ready, we can start breaking apart tasks to re-implement a more complex interface. It seems like re-implementing will be much easier than debugging through it.

jywarren:
Hi @sashadev-sky - thanks for the great documentation. I wanted to CC @ViditChitkara who worked on this recently as did @gauravano -- the relevant code is here:

https://github.com/publiclab/plots2/blob/f29f90d6690d5a3c6d578221af79c0cd2b7bafee/app/views/comments/_form.html.erb#L92-L100

I believe the `Publish` button works properly - independently, that is. But my initial thought on the `Preview` button is that it is not properly addressing the correct container in toggling the preview.

I actually think this is solvable directly, BUT I think a priority should be to write a system test which protects it from future breakages. The reply-to-comment system has been some long hard work to get working correctly and I'd love to ensure we preserve it -- tests seem like the best way to do that!

sashadev-sky:
@jywarren ok hopefully one of them @gauravano @ViditChitkara can either solve this or open it up as  a GCI task! Then I can close this. I would put this as a "medium" task given that you provided the solution

Issue #6833, opened 2019-11-21
Give comment upload script unique IDs to properly place images


jywarren:
When uploading an image to a comment, esp if it's not the first comment, it looks like we're displaying the progress bar and inserting the image into the wrong textarea, probably due to non-unique html id attributes! See:

![Screenshot_20191121-153429](https://user-images.githubusercontent.com/24359/69374767-cb9c9e00-0c74-11ea-8aac-37f2fdac66cf.png)

Instead of showing the file here, it appeared in the "reply to" textarea of the top comment:

![Screenshot_20191121-153420](https://user-images.githubusercontent.com/24359/69374794-deaf6e00-0c74-11ea-9e18-addbdb4e0002.png)

Let's fix this!

jywarren:
Code for the comment form is here:

https://github.com/publiclab/plots2/blob/master/app/views/comments/_form.html.erb

And it's includes into each note page here:

https://github.com/publiclab/plots2/blob/master/app/views/notes/_comments.html.erb

And the JavaScript for uploading is here:

https://github.com/publiclab/plots2/blob/master/app/assets/javascripts/dragdrop.js

SidharthBansal:
Published on GCI dashboard

PR #5531, opened 2019-04-19
comment id unique

CleverFool77:
Fixes https://github.com/publiclab/plots2/issues/4771 (<=== Add issue number here)

* [x] PR is descriptively titled 📑 and links the original issue above 🔗
* [x] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [x] code is in uniquely-named feature branch and has no merge conflicts 📁
* [x] screenshots/GIFs are attached 📎 in case of UI updation
* [x] ask `@publiclab/reviewers` for help, in a comment below

Thanks!

PR #7140, opened 2020-01-07
Fix preview button text to go back editing

### LINKED ISSUE #7122: Wiki "Preview" state does not show a way to go back to editing
I noticed today that when you edit a Wiki page it has the options of "Save" "Preview" and "Cancel." When you click on "Preview" the page displays as a preview just as it should, but there is no button that says "Edit" or shows how you can return back to editing.

FireShot Capture 171 - 🎈 Public Lab_ Editing 'Power Tags' - publiclab org

I learned from trial and error that pushing the "Preview" button again brings me back to the Edit mode. The text on the button should change to reflect this change in function. When you press the "Preview" button to show the preview the text should change to "Edit" so you know to push that button to go back.

DouglasLutz:
### Bug:

The preview button text was not being altered when clicked. With this you would need to click 'Preview' again to go back to editing.

I found this happening both in comments (create and edit) and in the legacy editor.

* Old behaviour:
Not working new comment

### How it was fixed:

There were two bugs in `app/assets/javascripts/editor.js` file, the `toggle_preview` was getting the button by it's class but there were cases that multiple preview buttons were in the same page causing it to fail. Also the jquery `.button('text')` function was not altering the button text.

It has been created a parameter to receive the button object to ensure the right button is being accessed.

* New behaviour:
Working new comment

### DIFF
--- a/test/system/comment_test.rb
+++ b/test/system/comment_test.rb
@@ -122,5 +122,23 @@ class CommentTest < ApplicationSystemTestCase
     # Make sure that image has been uploaded
     page.assert_selector('#preview img', count: 1)
   end
+ 
+  test 'the text change on preview button click' do
+    visit '/'
 
+    click_on 'Login'
+
+    fill_in("username-login", with: "jeff")
+    fill_in("password-signup", with: "secretive")
+    click_on "Log in"
+
+    visit "/wiki/wiki-page-path/comments"
+
+    click_on 'Preview'
+    assert_selector('.preview-btn', text: "Edit")
+
+
+    click_on 'Edit'
+    assert_selector('.preview-btn', text: "Preview")
+  end
 end
--- a/app/views/comments/_edit.html.erb
+++ b/app/views/comments/_edit.html.erb
@@ -127,12 +127,12 @@
 
     <div class="control-group">
       <button type="submit" class="btn btn-primary"><%= translation('comments._edit.publish') %></button>
-        <a class="btn btn-default preview-btn"  data-previewing-text="Hide Preview"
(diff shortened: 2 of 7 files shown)


DouglasLutz:
@nstjean Sure! Now that you said I looked up and also found the preview button on `app/views/features/_form.html.erb` and `app/views/editor/question.html.erb`, I will apply the changes on those too

jywarren:
This is fantastic. I think this has been broken for a little while. Would you be able to write a short system test to protect this fix for the future? 

It could be based on this; and it could be pretty minimal but it'd be great to just check each of these use cases: https://github.com/publiclab/plots2/blob/master/test/system/comment_test.rb

That way this excellent fix gets protected. I think that we're re-using code a lot which is efficient, but it means people may not realize when they break a different use case for the same code!

BTW, Does this fix #6297? 🎉 

Also noting possible links to #5531, #6833

Awesome!!!!

DouglasLutz:
@jywarren Sure, I'll work on the tests.

DouglasLutz:
I noticed that #6297 is still occurring with the changes

jywarren:

test/system/comment_test.rb
```diff
@@ -61,4 +61,22 @@ class CommentTest < ApplicationSystemTestCase
     assert_selector("#{parentid} .comment .comment-body p", text: 'batman')
   end
  
+  test 'the text change on preview button click' do
```
jywarren: Hm, so did the original fix not resolve it, or is the test not passing but it should be?

SidharthBansal:
There are some conflicts. Kindly resolve

SidharthBansal:

PR #7160, opened 2020-01-09
Add tooltip explaining number of wiki authors

jywarren:
I believe `first-timers-bot` should pick this up and we can close this PR. But just in case it's not working, i'm leaving this open!


jywarren:
Fixed, appeared in https://github.com/publiclab/plots2/issues/7170 !

Issue #7062, opened 2019-12-30
"Internal Server Error" when posting research note


wmacfarl:
**NOTE: immediate issue solved, and several more changes have been made to prevent this in the future, but one of the several causes is still open pending, see [this comment below](https://github.com/publiclab/plots2/issues/7062#issuecomment-572811285)**

### Please describe the problem (or idea)

> What happened just before the problem occurred? Or what problem could this idea solve?

I attempted to save a research note as a draft.  I also followed up by attempting to post a new "test" research note and got the same result/error.

> What did you expect to see that you didn't?

When I clicked "Save" the "Save" button was replaced with the spinning progress indication icon as usual but nothing happened.  When I looked in the developer console I saw the following errors:

![image](https://user-images.githubusercontent.com/757598/71604871-daf3ee80-2b32-11ea-9405-f42e17949d6c.png)

### Please show us where to look

I have attached the markdown text of the note I was initially trying to post, but I had the same problem using a different browser with a research note that simply said "test" so I don't think it's a problem with the body of the research note.

### What's your PublicLab.org username?
wmacfarl

### Browser, version, and operating system
Windows 10, using Firefox and Google Chrome

> Many bugs are related to these -- please help us track it down and reproduce what you're seeing!

****

## Thank you!

Your help makes Public Lab better! We *deeply* appreciate your helping refine and improve this site. 

To learn how to write really great issues, which increases the chances they'll be resolved, see:

https://publiclab.org/wiki/developers#Contributing+for+non-coders

[sensor_platform_build_guide_draft.md.txt](https://github.com/publiclab/plots2/files/4011450/sensor_platform_build_guide_draft.md.txt)

jywarren:
Thanks for reporting -- I'm checking this on Sentry now.

jywarren:
Yes, looks like first incident on Dec 20th:

![image](https://user-images.githubusercontent.com/24359/71674737-10265980-2d4a-11ea-86d1-510473fde679.png)


```
ActiveRecord::RecordNotUnique
[...]

PR #7173, opened 2020-01-09
test bad revisions

jywarren:
re #7062

### DIFF
--- a/test/unit/revision_test.rb
+++ b/test/unit/revision_test.rb
@@ -23,7 +23,7 @@ class RevisionsTest < ActiveSupport::TestCase
     assert node.save!
     # in testing, uid and id should be matched, although this is not yet true in production db
     node_revision = Revision.new(title: 'My new node',
-                                 body: '',
+                                 body: 'I hope this test works -- by which I mean fails! 😅',
                                  uid: users(:bob).id,
                                  nid: nodes(:one).nid)
     assert node_revision.save!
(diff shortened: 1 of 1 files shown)


test/unit/revision_test.rb
```diff
@@ -15,6 +15,21 @@ class RevisionsTest < ActiveSupport::TestCase
...
   end
+  
+  test 'create a bad revision' do
+    node = Node.new(uid: users(:bob).id,
+                    type: 'page')
+    node.title = 'My new node for revision testing'
+    assert node.save!
+    # in testing, uid and id should be matched, although this is not yet true in production db
+    node_revision = Revision.new(title: 'My new node',
+                                 body: 'I hope this test works -- by which I mean fails! 😅',
```
jywarren: If this causes a test failure, let's try to fix it with https://github.com/publiclab/plots2/pull/7172

```ruby
        # prevent vid non-unique bug in https://github.com/publiclab/plots2/issues/7062
        raise ActiveRecord::Rollback if !node.valid? || node.vid == 0
```

jywarren:
Nope! This did not cause a test failure here, probably for the same mysterious reason that it did not in #3007 ... difference between Travis and our live site containers...

SidharthBansal:
@VladimirMikulic I think you will be willing to take this one too

VladimirMikulic:
@SidharthBansal as always, just tag me :)

SidharthBansal:
Lol

PR #7174, opened 2020-01-10
ADD: System test for return_to query parameter

### LINKED ISSUE #5585: System test for Preserve `return_to` param when switching from /login to /signup

VladimirMikulic:
Tests that return_to query parameter is preserved while switching from
/login to /signup.

Resolves #5585

@SidharthBansal @jywarren could you review this? Thanks.

### DIFF
--- a/test/system/screenshots_test.rb
+++ b/test/system/screenshots_test.rb
@@ -71,6 +71,14 @@ class ScreenshotsTest < ApplicationSystemTestCase
     take_screenshot
   end
 
+  test 'return_to query parameter preserved while switching from /login to /signup' do
+    visit '/login?return_to=page'
+    click_on 'sign up'
+
+    path = URI.parse(current_url).request_uri
+    assert_equal path, '/signup?return_to=page'
+  end
+
   test 'tags' do
     visit '/tags'
     take_screenshot


SidharthBansal:
Please ask uzay for a review

Uzay-G:
Sorry, I think we should change which file we put the test in. It shouldn't be in `screenshots_test` but the `sessions_test` would be more appropriate.

VladimirMikulic:
@Uzay-G the issue title says that it should be a system test. I am not an expert RoR tester but we are not dealing with sessions here. We just check for the URL, no logging in/out, filling forms...

Uzay-G:
No I know it's a system test but it should be `system/sessions_test` instead of `system/screenshots_test`. The screenshots file is just for the test screenshots. Sessions handle the login functionality.

VladimirMikulic:
Oh, then it's my mistake. @Uzay-G thank you for the clarification I'll submit the new PR.

Uzay-G:
No problem! :+1:

SidharthBansal:
Please do pr to rectify it
Thanks both

VladimirMikulic:
@Uzay-G @SidharthBansal here it is #7185

PR #7189, opened 2020-01-10
push map below floating nav bar

### LINKED ISSUE #7188: Wider nav bar covers up top of map
Currently the navbar is fixed and the map is within the `container` div in the content. `container has a fixed `padding-top`, which works when the nav bar is standard size but when there is an extra notice at the top it does not work!

FireShot Capture 202 - 🎈 Public Lab_ Map - publiclab org

I have a fix for this!

nstjean:
Fixes #7188  (<=== Add issue number here)

Because the #header is has a fixed position it does not take up space on the page flow. Most of the content relies on a large margin at the top so it won't get covered, but the map is supposed to be positioned directly below it.

I used javascript to check the height of the navbar and set the gap to that size. Now no matter what size the header is the map should be directly below it.

### DIFF
--- a/app/views/map/map.html.erb
+++ b/app/views/map/map.html.erb
@@ -8,7 +8,7 @@
     max-width: 100%;
   }
   body>.container {
-    padding-top: 28px;
+    padding-top: 0px;
   }
   .footer {
     margin: 0;
@@ -29,6 +29,9 @@
 
    <script>
 
+      // set the map just under the navbar
+      $('.body-container').css('padding-top', $('#header').css('height'));
+
       var urlHash = urlMapHash();
 
         // if there are location params in URL let them override, otherwise save lat/lon/zoom from controller
@@ -41,7 +44,7 @@
             <% end %>
           <% end %>
         }
-
+        
       const params = urlHash.getUrlHashParameters();
      
       var lat = parseFloat(params.lat);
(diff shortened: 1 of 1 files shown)


nstjean:
Ahh yes, sorry.

Peek 2020-01-13 08-59

FireShot Capture 204 - 🎈 Public Lab_ Map - localhost

nstjean:
@jywarren It is now failing the codecov after rebase?

jywarren:
I'll try retriggering them... :-( 

sorry for the trouble!

jywarren:
Hm. I'm not too stressed about this but it does seem like an error in CodeCov, don't you think? cc @keshav234156  who I believe was able to set this up -- is that right? And this page of documentation: 

https://docs.codecov.io/docs/unexpected-coverage-changes

nstjean:
I don't know if it's an error, or how the way the rebase changed how codecov calculates all of its data. Because it's mentioning files that have nothing to do with this patch, such as `tag_controller.rb`. I'm rebasing again to see what happens.

nstjean:
That worked! @jywarren This is set to merge.

keshav234156:
@jywarren i didn't set up codecov in this project.i think you wanted to mention someone else.

jywarren:
Sorry @keshav234156 i reviewed a lot of PRs this week and my brain is mush now. My mistake! Thanks @nstjean too!

PR #7214, opened 2020-01-12
ADD: Better sign up form validation messages

### LINKED ISSUE #7212: Invalid email error on signup page
@VladimirMikulic please fix this. You missed this.
image
I already gave 5 points for signup form and modal validations.

VladimirMikulic:
Resolves #7212

### DIFF
--- a/app/assets/javascripts/validation.js
+++ b/app/assets/javascripts/validation.js
@@ -34,7 +34,7 @@ $(document).ready(function() {
   });
 });
 
-$(document).ready(function () {
+$(document).ready(function() {
   // The two forms have same ID
   var forms = document.querySelectorAll("#create-form");
 
@@ -128,7 +128,7 @@ SignUpFormValidator.prototype.isFormValid = function() {
 };
 
 function validateUsername(obj) {
-  var username = this.value;
+  var username = this.value.trim();
   var self = this;
 
   if (username.length < 3) {
@@ -151,10 +151,10 @@ function validateUsername(obj) {
 }
 
 function validateEmail(obj) {
-  var email = this.value;
+  var email = this.value.trim();
 
-  if (email === "") {
-    obj.updateUI(this, false, "This field can't be empty");
+  if (email.length === 0) {
+    obj.updateUI(this, false, "The email cannot be empty.");
     return;
   }
 
@@ -165,10 +165,10 @@ function validateEmail(obj) {
 }
 
 function validatePassword(confirmPasswordElement, obj) {
-  var password = this.value;
+  var password = this.value.trim();
 
-  if (password === "") {
-    obj.updateUI(this, false, "This field can't be empty");
+  if (password.length === 0) {
+    obj.updateUI(this, false, "The password cannot be empty.");
     return;
   }
 
@@ -189,16 +189,20 @@ function validatePassword(confirmPasswordElement, obj) {
 }
 
 function validateConfirmPassword(passwordElement, obj) {
-  var confirmPassword = this.value;
+  var confirmPassword = this.value.trim();
   var password = passwordElement.value;
 
-  if (confirmPassword === "") {
-    obj.updateUI(this, false, "This field can't be empty");
+  if (confirmPassword.length === 0) {
+    obj.updateUI(this, false, "The password confirmation cannot be empty");
     return;
   }
 
   if (confirmPassword !== password || !isPasswordValid(password)) {
-    obj.updateUI(this, false, "Passwords must be equal");
+    obj.updateUI(
+      this,
+      false,
+      "Password and Password Confirmation should be the same."
+    );
     return;
   }
 
--- a/app/assets/javascripts/validation.js
+++ b/app/assets/javascripts/validation.js
@@ -132,9 +132,7 @@ function validateUsername(obj) {
   var self = this;
 
   if (username.length < 3) {
-    restoreOriginalStyle(this);
-    obj.disableSubmitBtn();
-    removeErrorMsg(self);
+    obj.updateUI(this, false, "Username has to be at least 3 characters long");
   } else {
     $.get("/api/srch/profiles?query=" + username, function(data) {
       if (data.items) {
@@ -154,6 +152,12 @@ function validateUsername(obj) {
 
 function validateEmail(obj) {
   var email = this.value;
+
+  if (email === "") {
+    obj.updateUI(this, false, "This field can't be empty");
+    return;
+  }
+
   var emailRegExp = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
   var isValidEmail = emailRegExp.test(email);
 
@@ -162,6 +166,11 @@ function validateEmail(obj) {
 
 function validatePassword(confirmPasswordElement, obj) {
   var password = this.value;
+
+  if (password === "") {
+    obj.updateUI(this, false, "This field can't be empty");
+    return;
+  }
 
   if (!isPasswordValid(password)) {
     obj.updateUI(
@@ -183,7 +192,12 @@ function validateConfirmPassword(passwordElement, obj) {
   var confirmPassword = this.value;
   var password = passwordElement.value;
 
-  if (confirmPassword !== password) {
+  if (confirmPassword === "") {
+    obj.updateUI(this, false, "This field can't be empty");
+    return;
+  }
+
+  if (confirmPassword !== password || !isPasswordValid(password)) {
     obj.updateUI(this, false, "Passwords must be equal");
     return;
   }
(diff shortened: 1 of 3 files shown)


SidharthBansal:
Travis broken
This is quick
@Uzay-G kindly review

VladimirMikulic:
Thanks, @SidharthBansal. Functional test for banning users fails, no idea why.
My change doesn't have connections with the test.

On localhost, this test passes. Strange.

Uzay-G:
I will look at the travis errors once the build is done! I hope we can figure this out @VladimirMikulic :+1:

VladimirMikulic:
Yes, I fixed the Travis :tada: :tada: :tada:

VladimirMikulic:
@Uzay-G now it's your turn :smile:

Uzay-G:
Awesome! I will check it out now :+1:

app/assets/javascripts/validation.js
```diff
@@ -154,6 +152,12 @@ function validateUsername(obj) {
 
 function validateEmail(obj) {
   var email = this.value;
+
+  if (email === "") {
```
Uzay-G: Here instead of checking if the `string === ""`, we can use regex to test if the string has any non-whitespace characters. That way the message will be displayed even if the email field contains  `"   "`. You can write it like this:
`if (/\S/.test(email)) {`

VladimirMikulic: Thanks, @Uzay-G. I love regexes also, but for someone new, it is not readable.
I can trim email and check for length. Much more readable and it has the same effect.

app/assets/javascripts/validation.js
```diff
@@ -163,6 +167,11 @@ function validateEmail(obj) {
 function validatePassword(confirmPasswordElement, obj) {
   var password = this.value;
 
+  if (password === "") {
```
Uzay-G: We can also do the same here

app/assets/javascripts/validation.js
```diff
@@ -183,7 +192,12 @@ function validateConfirmPassword(passwordElement, obj) {
   var confirmPassword = this.value;
   var password = passwordElement.value;
 
-  if (confirmPassword !== password) {
+  if (confirmPassword === "") {
```
Uzay-G: And here

VladimirMikulic: Yeah, I've noticed that I've missed it after I pushed.

VladimirMikulic:
It should be good now.

test/functional/users_controller_test.rb
```diff
@@ -85,7 +85,7 @@ def setup
     # checks flash
     assert_equal flash[:error], I18n.t('users_controller.user_has_been_banned')
     # checks duplicated flash is not present
-    assert_equal flash[:notice], nil
+    assert_nil flash[:notice]
```
Uzay-G: Nice, this is really useful since that type of assertion will be deprecated soon in rails :+1:

SidharthBansal:
Can we merge this now?

app/assets/javascripts/validation.js
```diff
@@ -180,10 +189,15 @@ function validatePassword(confirmPasswordElement, obj) {
...
   var password = passwordElement.value;
 
-  if (confirmPassword !== password) {
+  if (confirmPassword.length === 0) {
+    obj.updateUI(this, false, "This field can't be empty");
+    return;
+  }
+
+  if (confirmPassword !== password || !isPasswordValid(password)) {
     obj.updateUI(this, false, "Passwords must be equal");
```
Uzay-G: I think this should say "Passwords should be the same.". I totally get why you did that though. You see the `confirmPassword !== password` and you start thinking like the computer :smile:

test/functional/users_controller_test.rb
VladimirMikulic: I've noticed that also, so I update it :rocket: Thanks @Uzay-G.

app/assets/javascripts/validation.js
VladimirMikulic: @Uzay-G I think that should is not meant to be placed here.
If the user want to complete the form successfully, passwords must(have to) be equal.
Should is not that enforcing. 

@SidharthBansal your opinion on this? Thank you both.

Uzay-G: Oh no that's not I was focusing on. It should be: `Passwords should be the same.`

Uzay-G: I just meant it should "the same" instead of "equal". I totally agree that we should use `must`

VladimirMikulic: In that case, I apologise. Let me quickly update this.

SidharthBansal:
I have to review it too before merging :rocket:

app/assets/javascripts/validation.js
```diff
@@ -153,15 +151,26 @@ function validateUsername(obj) {
 }
 
 function validateEmail(obj) {
-  var email = this.value;
+  var email = this.value.trim();
+
+  if (email.length === 0) {
+    obj.updateUI(this, false, "This field can't be empty");
```
SidharthBansal: Don't use pronouns in documentations.
Remove `this`

app/assets/javascripts/validation.js
```diff
@@ -153,15 +151,26 @@ function validateUsername(obj) {
...
 
   obj.updateUI(this, isValidEmail, "Invalid email");
 }
 
 function validatePassword(confirmPasswordElement, obj) {
-  var password = this.value;
+  var password = this.value.trim();
+
+  if (password.length === 0) {
+    obj.updateUI(this, false, "This field can't be empty");
```
SidharthBansal: this

app/assets/javascripts/validation.js
```diff
@@ -180,11 +189,16 @@ function validatePassword(confirmPasswordElement, obj) {
...
 
 function validateConfirmPassword(passwordElement, obj) {
-  var confirmPassword = this.value;
+  var confirmPassword = this.value.trim();
   var password = passwordElement.value;
 
-  if (confirmPassword !== password) {
-    obj.updateUI(this, false, "Passwords must be equal");
+  if (confirmPassword.length === 0) {
+    obj.updateUI(this, false, "This field can't be empty");
```
SidharthBansal: this

VladimirMikulic:
We can merge this :tada:

test/system/screenshots_test.rb
```diff
@@ -32,7 +32,7 @@ class ScreenshotsTest < ApplicationSystemTestCase
     assert_equal( username_error_msg, "Username already exists" )
     assert_equal( email_error_msg, "Invalid email" )
     assert_equal( password_error_msg, "Please make sure password is at least 8 characters long with minimum one numeric value" )
-    assert_equal( confirm_password_error_msg, "Passwords must be equal" )
+    assert_equal( confirm_password_error_msg, "Passwords must be the same" )
```
SidharthBansal: Password and Password Confirmation should be the same.

app/assets/javascripts/validation.js
```diff
@@ -180,11 +189,16 @@ function validatePassword(confirmPasswordElement, obj) {
...
 
-  if (confirmPassword !== password) {
-    obj.updateUI(this, false, "Passwords must be equal");
+  if (confirmPassword.length === 0) {
+    obj.updateUI(this, false, "The field can't be empty");
+    return;
+  }
+
+  if (confirmPassword !== password || !isPasswordValid(password)) {
+    obj.updateUI(this, false, "Passwords must be the same");
```
SidharthBansal: Password and Password Confirmation should be the same.

app/assets/javascripts/validation.js
```diff
@@ -153,15 +151,26 @@ function validateUsername(obj) {
...
 
   obj.updateUI(this, isValidEmail, "Invalid email");
 }
 
 function validatePassword(confirmPasswordElement, obj) {
-  var password = this.value;
+  var password = this.value.trim();
+
+  if (password.length === 0) {
+    obj.updateUI(this, false, "The field can't be empty");
```
SidharthBansal: Password cannot be empty.

app/assets/javascripts/validation.js
```diff
@@ -153,15 +151,26 @@ function validateUsername(obj) {
 }
 
 function validateEmail(obj) {
-  var email = this.value;
+  var email = this.value.trim();
+
+  if (email.length === 0) {
+    obj.updateUI(this, false, "The field can't be empty");
```
SidharthBansal: The email cannot be empty.

app/assets/javascripts/validation.js
VladimirMikulic: :+1:

app/assets/javascripts/validation.js
VladimirMikulic: :+1:

app/assets/javascripts/validation.js
VladimirMikulic: :+1:

test/system/screenshots_test.rb
VladimirMikulic: :+1:

VladimirMikulic:
@SidharthBansal I've updated it as you requested.

SidharthBansal:
It may be possible that I am wrong somewhere. You can discuss things if you think that is not a worthwhile effort. It is important to learn quantitatively and qualitatively both in 21st century.

SidharthBansal:
The marks were already given here. This was a follow up issue.

PR #7222, opened 2020-01-13
Remove styling for leaflet-marker-icon

nstjean:
Addresses problem mentioned in #6969

### DIFF
--- a/app/assets/stylesheets/location.css
+++ b/app/assets/stylesheets/location.css
@@ -1,7 +1,3 @@
-.leaflet-marker-icon {
-  background: url("https://raw.githubusercontent.com/pointhi/leaflet-color-markers/master/img/marker-icon-black.png") ; 
-}
-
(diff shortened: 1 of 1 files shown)


nstjean:
This styling was originally in rich.html.erb for the /post/ location form. The style was causing an issue with some map layers, but wasn't doing anything to the color of the marker in /post/ .

nstjean:
I am not understanding the codecov report. This patch affects 0% of the coverage, but it's showing project is -4.89%?

jywarren:
Hmm. It's possible there is a problem with it... let's try restarting it. I'll take a screenshot so we can compare once it finishes again. 

image

nstjean:
@jywarren It worked this time! All set to merge.

jywarren:
Great! I assume @crisner sees this but just cc'ing in case. Thanks!

PR #7216, opened 2020-01-13
ADD: Login forms validation

VladimirMikulic:
Added login modal form validation and pl.o/login form validation.

[Demo](https://streamable.com/we4ue)

Resolves #4250

@SidharthBansal @Uzay-G could you review this? Thanks.

PR #7223, opened 2020-01-13
ADD: System test for login modal form validation

VladimirMikulic:
This test is for #7216.

@Uzay-G can you confirm that this is good? Thanks.

### DIFF
--- a/test/system/screenshots_test.rb
+++ b/test/system/screenshots_test.rb
@@ -61,6 +61,23 @@ class ScreenshotsTest < ApplicationSystemTestCase
     take_screenshot
   end
 
+  test 'login modal form validation' do
+    visit '/'
+    click_on 'Login'
+
+    fill_in 'user_session[username]', with: 'Bob'
+    # The length of a password should be minimum 8 characters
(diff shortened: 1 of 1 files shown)


test/system/screenshots_test.rb
```diff
@@ -61,6 +61,22 @@ class ScreenshotsTest < ApplicationSystemTestCase
...
+
+    fill_in 'user_session[username]', with: 'Bob'
+    fill_in 'user_session[password]', with: 'invalid'
+
+    click_on 'Log in'
+
+    # Get the error message (remove '×' that closes the modal and remaining whitespaces)
+    error_msg = find('.error-msg-container').text.gsub('×', '').strip()
+
+    assert_equal( error_msg, 'Invalid username or password' )
```
Uzay-G: Yeah this is good! Could you explain with a comment just above that the password is invalid because there are less than 8 characters? Thanks

VladimirMikulic: :+1:

jywarren:
Very nice!!

PR #7225, opened 2020-01-13
FIX: redirect to previous page when subscribing to multiple tags

### LINKED ISSUE #5600: Redirect user to the same page after multi tag subscription
I think we should redirect the user to the same url after multi tag subscription, instead of /subscriptions so that the user will remain on the current research note or wiki page. Implemented the same in the pr.

Uzay-G:
Fixes #5600 

I included a test to verify that the redirection was working correctly

Here is a gif showing the functional redirection:
redirection-for-subscribe

### DIFF
--- a/app/controllers/subscription_controller.rb
+++ b/app/controllers/subscription_controller.rb
@@ -1,6 +1,6 @@
 class SubscriptionController < ApplicationController
   respond_to :html, :xml, :json
-  before_action :require_user, only: %i(create delete index digest multiple_add)
+  before_action :require_user, only: %i(create delete index digest)
 
   def index
     @title = "Subscriptions"
@@ -165,6 +165,9 @@ class SubscriptionController < ApplicationController
             end
           end
         end
+      else
+        flash[:warning] = "You must be logged in to subscribe for email updates!"
+        redirect_to "/login?return_to=" + request.fullpath
       end
     end
   end
--- a/app/controllers/subscription_controller.rb
+++ b/app/controllers/subscription_controller.rb
@@ -123,10 +123,10 @@ class SubscriptionController < ApplicationController
       redirect_to return_to
     else
       tag_list = if params[:tagnames].is_a? String
-                    params[:tagnames].split(',')
-                  else
-                    params[:tagnames]
-                  end
+                   params[:tagnames].split(',')
+                 else
+                   params[:tagnames]
+                 end
       if current_user
         if params[:type] == "tag"
           tag_list.each do |t|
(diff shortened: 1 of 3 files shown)


Uzay-G:
What do you think @SidharthBansal?

SidharthBansal:
check your screenshot. there is dashboard always

Uzay-G:
It worked. If you try on publiclab.org, you will be redirected to /subscriptions. In this case it redirected to the page the user was on

SidharthBansal:
Please don't try at PL.org
Your previous prs merged are not deployed yet
Does this makes sense?

Uzay-G:
Hey @SidharthBansal, I'm not at home right now and can't record my screen but when I subscribe to several tags on the stable version of PL I am redirected to /subscriptions:
Capture

On the other hand, as you can see in the gif, with my changes, I am redirected to the previous page I was on.

Uzay-G:
Hey! I made another gif where you can see it working:
redirection-for-subscribe gif

There is a problem with the flash though...

SidharthBansal:
@jywarren kindly review this. It looks good to me.
But will like to review once more after you review. 
@Uzay-G thanks for this. We can move to MK and SWB tests now I guess

app/controllers/subscription_controller.rb
```diff
@@ -117,57 +117,58 @@ def digest
   end
 
   def multiple_add
-    if !params[:tagnames] || params[:tagnames] == ''
+    return_to = params[:return_to] || "/subscriptions?_=" + Time.now.to_i.to_s
+    if params[:tagnames].blank?
       flash[:notice] = "Please enter tags for subscription in the url."
-      redirect_to "/subscriptions" + "?_=" + Time.now.to_i.to_s
```
SidharthBansal: Can you please see why git is showing many lines of change eventhough it is two line change?

test/functional/subscription_controller_test.rb
```diff
@@ -55,4 +55,12 @@ def setup
     get :add, params: { type: 'tag', name: 'blog' }, xhr: true
     assert users(:bob).following(:blog)
   end
+  
+  test "should redirect properly when subscribing to multiple tags" do
```
SidharthBansal: Great

app/views/tag/_tags_popover.html.erb
```diff
@@ -1,6 +1,7 @@
 <div class="pop_content subscribe-btn">
   <p> Follow related tags: </p>
   <%= form_with remote: true, url: '/subscribe/multiple/tag' do |f| %>
+    <input type="hidden" name="return_to" class="subscription-popover-redirect" value="<%= request.fullpath %>" />
```
SidharthBansal: Cool

app/controllers/subscription_controller.rb
SidharthBansal: sorry 4 lines of change

SidharthBansal:
Oh the code is inside the else part so it showing indented code snippets. 
Great well done!

Uzay-G:
Thank you so much! I couldn't really work today but will be available tomorrow! :+1:

PR #7233, opened 2020-01-14
Fix conditional in Node model

seabl:
No issue opened

Thanks!

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -900,7 +900,7 @@ class Node < ActiveRecord::Base
   # finds nodes by tag name, user id, and optional node type
   def self.find_by_tag_and_author(tagname, user_id, type = 'notes')
 
-    node_type = 'note' if type == 'notes' || 'questions'
+    node_type = 'note' if type == 'notes' || type == 'questions'
     node_type = 'page' if type == 'wiki'
     # node_type = 'map' if type == 'maps'  # Tag.tagged_nodes_by_author does not seem to work with maps, more testing required
 


VladimirMikulic:
```ruby
 # These are the same!
 node_type = 'note' if type == 'notes' || 'questions'
 node_type = 'note' if type == 'notes' || type == 'questions'
```

Uzay-G:
Codeclimate suggests we refactor this to:
`node_type = 'note' if ["notes", "questions"].include? type`

Which is the same as:
`node_type = 'note' if w%(notes questions).include? type`

seabl:
@VladimirMikulic Thank you for the review. 
I think these 2 are not the same as `'notes' || 'questions'` will always evaluate to `'notes'` and therefore `type == 'questions'` will never be `true`

seabl:
@Uzay-G Do you think though that Codeclimate's suggestion is a good one in this case? 
Let's see what @jywarren thinks. Thank you.

Uzay-G:
> @Uzay-G Do you think though that Codeclimate's suggestion is a good one in this case?
> Let's see what @jywarren thinks. Thank you.

I mean it's just syntax but codeclimate follows ruby style standards so I think we should follow the suggestion. It is a bit cleaner I find

jywarren:
Is this true? Can we test this in a ruby console? 

I think @seabl is right, actually! I must've written this 😅

jywarren:
Tested:

```ruby
irb(main):001:0> type = 'marshmallows'
=> "marshmallows"
irb(main):002:0> type == 'notes' || 'questions'
=> "questions"
irb(main):003:0> (type == 'notes' || 'questions') != false
=> true
```

jywarren:
Thank you!!!!

seabl:
So... I was not exactly right there... `==` has higher precedence than `||`, which means that the line as it is in our codebase will always evaluate to `true`.

jywarren:
Cool. If you'd like to refactor that's fine too, but this at least addressed the logical error I'd made. Good catch!

seabl:
@jywarren Thanks. 
Actually, I found this purely by chance. _For once_, I gave the test warnings a closer look.

PR #7250, opened 2020-01-15
Update test blog page with location modal

nstjean:
Fixes the test error in https://github.com/publiclab/plots2/pull/7248

The test was failing by trying to fill in text in an input that was disabled due to the API key not being valid on the testing server. I removed the `fill_in` for now.


nstjean:
@jywarren This is ready to merge so that we can bump LBL in https://github.com/publiclab/plots2/pull/7248

PR #7251, opened 2020-01-15
ADD: System test for comments

Uzay-G:
For part of #5316.

I have added a system test for commenting on a note. The test verifies that the UI reacts properly to the comment and that the preview button works well.

Are there any other comment tests you think would be useful? What do you think @VladimirMikulic @publiclab/reviewers?

### DIFF
--- a/test/system/comment_test.rb
+++ b/test/system/comment_test.rb
@@ -4,16 +4,14 @@ require "application_system_test_case"
 
 class CommentTest < ApplicationSystemTestCase
   Capybara.default_max_wait_time = 60
-
-  test 'adding a comment via javascript' do
+  def setup
     visit '/'
-
     click_on 'Login'
-
     fill_in("username-login", with: "jeff")
     fill_in("password-signup", with: "secretive")
     click_on "Log in"
-
+  end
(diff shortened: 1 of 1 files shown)


VladimirMikulic:
Nice one @Uzay-G. The test fails on my server, we'll see what Travis says.

Uzay-G:
Yeah you are right. There is a problem I hadn't noticed. I will fix it :+1:

Uzay-G:
Alright I fixed it. I also refactored things to a `setup` method that will login before each test instead of putting the commands in the tests each time.

Uzay-G:
Does this work for you @VladimirMikulic?

SidharthBansal:
You both have to complete tests for 6 features. Then I will approve tasks
for both of you. And you both will get 15 points each. After that we can
repeat this activity as many times as you want.
I hope this will give you a lot of points and learning too

VladimirMikulic:
@SidharthBansal it will for sure :)

jywarren:
Thank you!!!!

SidharthBansal:
5th test here 
We just need to approve at the GCI dashboard now

PR #7253, opened 2020-01-15
ADD: Tags adding system test

VladimirMikulic:
Part of #5316

@Uzay-G your opinion? Thanks.

### DIFF
--- a/test/system/post_test.rb
+++ b/test/system/post_test.rb
@@ -31,6 +31,25 @@ class PostTest < ApplicationSystemTestCase
 
   end
 
+  test 'adding tags to the post' do
+    visit '/wiki/wiki-page-path/comments'
+
+    find('a[data-target="#loginModal"]', text: 'Login').click()
+
+    fill_in 'user_session[username]', with: 'jeff'
+    fill_in 'user_session[password]', with: 'secretive'
+
+    click_on 'Log in'
+
+    find('a#tags-open').click()
+
+    find('.tag-input').set('nature').native.send_keys(:return)
+    find('.tag-input').set('mountains').native.send_keys(:return)
+
+    # Make sure that the 2 tags are added
+    page.assert_selector('.tags-list p.badge', :count => 2)
+  end
+
   # Utility methods:
   
   def assert_page_reloads(message = "page should reload")
--- a/test/system/post_test.rb
+++ b/test/system/post_test.rb
@@ -50,6 +50,28 @@ class PostTest < ApplicationSystemTestCase
     page.assert_selector('.tags-list p.badge', :count => 2)
   end
 
+  test 'removing tags from the post' do
+    visit '/wiki/wiki-page-path/comments'
+
+    find('a[data-target="#loginModal"]', text: 'Login').click()
+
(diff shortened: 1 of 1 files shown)


test/system/post_test.rb
```diff
@@ -31,6 +31,25 @@ class PostTest < ApplicationSystemTestCase
...
+    find('a[data-target="#loginModal"]', text: 'Login').click()
+
+    fill_in 'user_session[username]', with: 'jeff'
+    fill_in 'user_session[password]', with: 'secretive'
+
+    click_on 'Log in'
+
+    find('a#tags-open').click()
+
+    find('.tag-input').set('nature').native.send_keys(:return)
```
Uzay-G: This looks good! Could you explain what the `.native.send_keys(:return)` does? I am just curious!

VladimirMikulic: As you can probably guess it emits the "Enter" keyboard key. In order to add a tag, the user has to press the "Enter" key since this is a test, I have to emit it somehow.

Uzay-G: Just to understand, why wouldn't the `fill_in` method work here?

VladimirMikulic: As the name suggests `fill_in` will just fill the input with some text, but that's it!
It won't trigger any submit/enter events. For a tag to be added, user has to press the "Enter" key.
That's why we need `.native.send_keys(:return)`.

I hope that it's clear now. Thank you for the interest though.

Uzay-G: Oh alright I get it. We can only use that for forms with submit buttons. I'm not experienced with system testing. Thanks for your help :+1:

VladimirMikulic: Anything else? :)

VladimirMikulic:
@SidharthBansal unless you have some suggestions, this is ready to be merged :)

SidharthBansal:
Let's take review from Uzay.
You have to consider all kinds of test. please ensure completeness. Also
please ensure no redundant tests

test/system/post_test.rb
```diff
@@ -31,6 +31,25 @@ class PostTest < ApplicationSystemTestCase
...
+    find('.tag-input').set('mountains').native.send_keys(:return)
+
+    # Make sure that the 2 tags are added
+    page.assert_selector('.tags-list p.badge', :count => 2)
```
Uzay-G: We could also make sure that the tag text is correctly displayed by searching for "mountains" and "nature"

SidharthBansal:
We will require tests for multiple tag subscriptions also.

SidharthBansal:
Please ensure that this pr deals with all types of TAGS SYSTEM TESTS.
It may be possible that we will change the scoring system for System Tests to be fair to all the candidates. I hope you all will be okay with it.

test/system/post_test.rb
VladimirMikulic: I think it would be redundant to do that.

jywarren:
This is super! Thanks so much!!!

SidharthBansal:
Yeah we need more tests here too

PR #7255, opened 2020-01-16
ADD: Comment image upload system test

VladimirMikulic:
Part of #5316 

@jywarren could you review this? Thanks.

### DIFF
--- a/test/system/comment_test.rb
+++ b/test/system/comment_test.rb
@@ -97,6 +97,73 @@ class CommentTest < ApplicationSystemTestCase
     page.assert_selector('#preview img', count: 1)
   end
 
+  test 'comment image drag and drop upload' do
+    Capybara.ignore_hidden_elements = false
+    visit "/wiki/wiki-page-path/comments"
+
+    find("p", text: "Reply to this comment...").click()
+
+    reply_preview_button = page.all('#post_comment')[0]
+
+    # Upload the image
+    drop_in_dropzone("#{Rails.root.to_s}/public/images/pl.png")
+
+    # Wait for image upload to finish
+    wait_for_ajax
+    Capybara.ignore_hidden_elements = true
+
+    # Toggle preview
+    reply_preview_button.click()
+
+    # Make sure that image has been uploaded
+    page.assert_selector('#preview img', count: 1)
+  end
+
+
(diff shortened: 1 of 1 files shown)


test/system/comment_test.rb
```diff
@@ -74,4 +74,40 @@ def setup
     assert_equal( comment_preview_button.text, "Preview" )
   end
 
+  test 'comment image upload' do
```
SidharthBansal: There should be a test for exceeding image size too

SidharthBansal:
@Uzay-G @VladimirMikulic Please create a pr for Comments.
In the comments pr you can add all the system tests regarding comments.
For example, image upload tests, Cntrl+Enter should result in posting the comment and so on.
It will result in pr floods if you will add one pr per system test. I hope this makes sense.
However if you think that you require functional changes with the system tests, then you can create a discrete pr.

SidharthBansal:
The prs requiring the functional changes will be graded separately. We will be fair to you all in terms of marks.

test/system/comment_test.rb
VladimirMikulic: The question is, do you have a limit set? I've uploaded 60mb image without a problem.

VladimirMikulic: If you don't display some kind of error message or something, then I can't test it.

SidharthBansal: Kindly check. We need to set the limit if not set yet. pinging @jywarren for limit

Uzay-G:
> @Uzay-G @VladimirMikulic Please create a pr for Comments.
> In the comments pr you can add all the system tests regarding comments.
> For example, image upload tests, Cntrl+Enter should result in posting the comment and so on.
> It will result in pr floods if you will add one pr per system test. I hope this makes sense.
> However if you think that you require functional changes with the system tests, then you can create a discrete pr.


My comment test has already been merged but we will do this if there are more comments tests :+1:

VladimirMikulic:
@jywarren I think you'll be glad to see this. Both image upload tests have been repaired and are working (even drag & drop)! The system tests are awesome :smile: Love them!

jywarren:
This is EPIC! So great! I'm so glad you like system testing. It's so powerful. Thank you a ton for this! I'll close my original attempt!

SidharthBansal:
Yeah we should do more tests.

PR #7267, opened 2020-01-16
Add better error validation message for username

### LINKED ISSUE #5829: Username with a space in it needs better validation error message
This is small but important - someone wrote in saying they got a "username invalid" message when trying to sign up; I tried, and realized that if you enter a space in a username, it says "invalid" but doesn't say why. 

Screen Shot 2019-06-03 at 6 42 29 PM

We should offer a validation error message specifying the formatting rules for usernames!

I think this should go here: 

https://github.com/publiclab/plots2/blob/faa5a12e86bf7944dca43134f649947f03ca96a6/app/models/user.rb#L60-L61

Tlazypanda:
Fixes #5829 

# Screenshot  
Screenshot from 2020-01-17 02-03-19

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -35,6 +35,7 @@ class User < ActiveRecord::Base
 
   acts_as_authentic do |c|
     c.crypto_provider = Authlogic::CryptoProviders::Sha512
+    c.validates_format_of_login_field_options = { with: Authlogic::Regex::LOGIN, message: I18n.t('error_messages.login_invalid', default: "can only consist of alphabets, numbers, underscore '_', and hyphen '-'.") }
   end
 
   has_attached_file :photo, styles: { thumb: '200x200#', medium: '500x500#', large: '800x800#' },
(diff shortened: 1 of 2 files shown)


SidharthBansal:
@VladimirMikulic kindly review

Tlazypanda:
@jywarren @VladimirMikulic updated the pr and added ss of the change.Removed authlogic login for this case since the message was fixed in that case...also noticed that tests were failing because there was no check for length of username so added that check.Kindly review :v: Thanks :)

app/models/user.rb
```diff
@@ -6,9 +6,6 @@ def validate(record)
   end
 end
 
-# Overwrites authlogic username regex to allow one character usernames
-Authlogic::Regex::LOGIN = /\A[A-Za-z\d_\-]*\z/
```
VladimirMikulic: @jywarren is this line safe to remove?

jywarren: I would say probably not - i think we need that? Can we look at the pull request that generated that line, using the "blame" view of that file on GitHub?

Thank you!

Tlazypanda:
@jywarren @VladimirMikulic updated the pr :) Kindly review :v:

app/models/user.rb
Tlazypanda: @jywarren @VladimirMikulic Have updated the pr keeping in mind the changes.Thanks :)

SidharthBansal:
:tada:

jywarren:
Awesome!!

PR #7268, opened 2020-01-16
add system tests for moderation

Uzay-G:
For #5316.

Using system tests to verify moderation has been really useful! I have made a few tests that I think have value and think some more could be added. 
What do you think @jywarren @VladimirMikulic @SidharthBansal?

### DIFF
--- a/test/system/moderation_tests.rb
+++ b/test/system/moderation_tests.rb
@@ -1,0 +1,75 @@
+require "application_system_test_case"
+
+class ModerationTest < ApplicationSystemTestCase
+
+  def setup
+    visit "/"
+    click_on "Login"
+    fill_in 'user_session[username]', with: 'palpatine'
+    fill_in 'user_session[password]', with: 'secretive'
+    click_on "Log in"
+  end
+
+  test "banning and unbanning a user" do
+    visit "/profile/#{users(:bob).username}"
+
+    find("a#info-ellipsis").click
+    accept_confirm "Are you sure? The user will no longer be able to log in or publish, and their content will be hidden." do
+      find("a[href='/ban/#{users(:bob).id}'").click()
+    end
+    assert find("div.alert", text: "That user has been banned.")
+  
+    find("a#info-ellipsis").click
+    find("a[href='/unban/#{users(:bob).id}'").click()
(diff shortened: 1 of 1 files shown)


test/system/moderation_tests.rb
```diff
@@ -0,0 +1,75 @@
...
+    fill_in 'user_session[password]', with: 'secretive'
+    click_on "Log in"
+  end
+
+  test "banning and unbanning a user" do
+    visit "/profile/#{users(:bob).username}"
+
+    find("a#info-ellipsis").click
+    accept_confirm "Are you sure? The user will no longer be able to log in or publish, and their content will be hidden." do
+      find("a[href='/ban/#{users(:bob).id}'").click()
```
VladimirMikulic: Is there a particular reason for inconsistency?
On line 16, you call `.click` and on line 18 you call `.click()`, similar on other lines also.

Uzay-G: Haha, I will this minor problem :+1:. What do you think is better?

VladimirMikulic: In my home opinion, since most of the programmers come from C based languages, the `.click()` is probably more readable and indicates that it is a function call. What do you think?

Uzay-G: Yeah that's what I did :+1:

test/system/moderation_tests.rb
```diff
@@ -27,13 +27,13 @@ def setup
   test "moderating and unmoderating a user" do
     visit "/profile/#{users(:bob).username}"
 
-    find("a#info-ellipsis").click
+    find("a#info-ellipsis").click()
```
jywarren: Are `()` required in ruby? just curious!

Uzay-G: No they aren't but @VladimirMikulic thinks it's a bit more readable.

jywarren: ok!!

jywarren:
This is great! Merging it!!!

SidharthBansal:
I think if you have 5 tests merged then we can get approved system tests 1.

SidharthBansal:
3 tests already merged here

PR #7292, opened 2020-01-18
FIX: messages when logging in without password for oauth

Uzay-G:
Fixes https://github.com/publiclab/plots2/pull/7291#issuecomment-575913424.

This problem is now fixed:
> If a user signs up with Oauth but does not set their password, and then tries to login with their email as username, they will get a `password_invalid` error instead of a message prompting them to set their password.
> 
> On the other hand, the correct message is displayed when trying to login with username.

### DIFF
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -117,7 +117,7 @@ class UserSessionsController < ApplicationController
 
   def handle_site_login_flow
     username = params[:user_session][:username] if params[:user_session]
-    u = User.find_by(username: username)
+    u = User.find_by(username: username) || User.find_by(email: username)
     if u && u.password_checker != 0
       n = u.password_checker
       hash = { 1 => "Facebook", 2 => "Github", 3 => "Google", 4 => "Twitter"  }
--- a/test/functional/user_sessions_controller_test.rb
+++ b/test/functional/user_sessions_controller_test.rb
@@ -18,7 +18,7 @@ class UserSessionsControllerTest < ActionController::TestCase
     assert_redirected_to '/dashboard'
   end
 
-  test 'should login and redirect to corresct url' do
+  test 'should login and redirect to correct url' do
     session[:return_to] = '/post?tags=question:question&template=question'
     post :create, params: { user_session: { username: users(:jeff).username, password: 'secretive' } }
     assert_redirected_to '/post?tags=question:question&template=question'
@@ -303,7 +303,7 @@ class UserSessionsControllerTest < ActionController::TestCase
     assert_redirected_to "/notes/liked?_=#{Time.now.to_i.to_s}"
   end
   
-  test "logging in through omniauth and then through normal login should display error and redirect" do
+  test "logging in through omniauth and then logging in with username should display correct error and redirect" do
     request.env['omniauth.auth'] =  OmniAuth.config.mock_auth[:github1]
     # login through omniauth 
     post :create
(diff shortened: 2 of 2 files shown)


SidharthBansal:
Please open a corresponding issue so that you can get one point for suggestion too.
Task: https://codein.withgoogle.com/dashboard/tasks/5916202187096064/

app/controllers/user_sessions_controller.rb
```diff
@@ -117,7 +117,7 @@ def handle_social_login_flow(auth)
 
   def handle_site_login_flow
     username = params[:user_session][:username] if params[:user_session]
-    u = User.find_by(username: username)
+    u = User.find_by(username: username) || User.find_by(email: username)
```
SidharthBansal: Kindly explain

Uzay-G: Oauth users are supposed be able to login with their email as username. So we need to check if the username they used to login is identical to an email in the db. That way, the code can check that `u` has a password. Otherwise, if a user logs in with their email,  `u` is nil and the code does not check if `u` has a password. If that happens, the correct error message is not displayed

SidharthBansal: >  login with their email as username. 

Wrong. I have made a recursive hash for this. We don't use emails as usernames. Consider
bansal.sidharth@rediffmail.com
and bansal.sidharth@gmail.com
so I stored usernames bansal_sidharth and bansal_sidharth213 where 213 is some random hash created. Kindly check.

Uzay-G: No, what i mean is that oauth users can put their email in the username field when they login. For example, here:
https://github.com/publiclab/plots2/blob/d8c644a5304211ad461374d2bba4c90b25f86100/app/controllers/user_sessions_controller.rb#L131-L134

It checks if the username entered by the user is actually their email, and if this is the case and if their credentials are correct, they are logged in.

SidharthBansal:
So here is an inconsistency on which we need to ponder too. :-(

SidharthBansal:
I will think about it and let you know. You can resume your other prs and
system test 2 task in the mean time.
Do as many as you can. We will review it whenever we are free. Thanks good
night.

app/controllers/user_sessions_controller.rb
SidharthBansal: Try to think regex of username and email are different. So, these queries are worthless. Does this makes sense?
cc\ @jywarren

Uzay-G: This isn't validation though. Currently, the login system allows oauth users to enter their email in the username field. It makes sense that if a user tries to log in with their email as username, they should get the same error messages as if they used their actual username.

SidharthBansal: I am sorry, but I disagree here.
We can't treat emails as usernames in the db. 
> the login system allows oauth users to enter their email in the username field.

Oauth takes the email prefix as username. username has different regex. Username can't contain dots etc.  
Things are different.

Uzay-G: We don't treat the emails as usernames. We just check if the user entered their email instead of their username.

If this is not the expected behavior, other parts of the codebase that already work like this need to be changed

SidharthBansal:
This is critical issue then. Let's ping @jywarren on this.
This is really important thing. You can work on other tasks in the meantime.
I will approve this task. Please submit for review. Thanks.

Uzay-G:
I figured it out @SidharthBansal! I am so happy :smile: :tada:!
Pushing the changes now :+1: 

It was really really difficult but it's worth it

Uzay-G:
Sorry wrong pr

app/controllers/user_sessions_controller.rb
jywarren: > We can't treat emails as usernames in the db.

I think the difference is we're recognizing in the UI that people sometimes put their email instead of username, and the system adapts to accept that. It doesn't actually store the email as a username in the db, it just relabels the username input as an email during the auth process. Does that help clarify? 

It's a convenience but it does significantly reduce the incidence of people either forgetting their usernames, or just getting mixed up and using an email instead (which people do by mistake sometimes).

SidharthBansal: >  It doesn't actually store the email as a username in the db

Ok. So, these changes 
> || User.find_by(email: username)

are not needed. @Uzay-G can you remove this change and see if test passes. I think test should pass.

Uzay-G: No it doesn't work if I do that. Instead the code continues and a `password invalid` error is displayed.

Uzay-G: I tried the test only passes with the changes

SidharthBansal: Can you please explain what is happening at regex level?
If test is failing, we need to merge it. Hence, merging it.

SidharthBansal: This is not going to break anything as changes are OR GATE. So, it is safe to merge. Hence merged it.

SidharthBansal: This is not going to break anything as changes are OR GATE. So, it is safe to merge. Hence merged it.

Issue #7303, opened 2020-01-19
Travis sometimes fails due to redirection fixes


Uzay-G:
In #7113, oauth redirection was fixed and now the user is redirected to the page he was on before logging in. But the user is for some reason redirected to the previous page plus a parameter with the current time:
https://github.com/publiclab/plots2/blob/b80f193b60827f90b70bef020648d8b2813468fa/app/controllers/user_sessions_controller.rb#L21-L22

So, at the time I did not question this and left it as is. But, I also made some tests for redirection:
```ruby
  test "redirects to previous page when logging in through oauth" do
    request.env['omniauth.origin'] = "/notes/liked"
    request.env['omniauth.auth'] =  OmniAuth.config.mock_auth[:github1]
    # sign up
    post :create
    assert_redirected_to "/notes/liked?_=#{Time.now.to_i.to_s}"
  end
```

Now I am noticing that the `_` parameter is sometimes causing travis to fail at the `    assert_redirected_to "/notes/liked?_=#{Time.now.to_i.to_s}"` line. This happens because there is a slight variation between the `Time.now` parameter where the test redirects and the `Time.now` that the test compares the url to. Here is an example of the Travis error:
```
FAIL["test_user_that_links_provider_to_existing_account_should_not_be_redirected_to_dashboard_on_oauth_signup_for_Google_provider", #<Minitest::Reporters::Suite:0x00005573a6e48120 @name="UserSessionsControllerTest">, 21.083390820999966]

 test_user_that_links_provider_to_existing_account_should_not_be_redirected_to_dashboard_on_oauth_signup_for_Google_provider#UserSessionsControllerTest (21.08s)

        Expected response to be a redirect to <http://test.host/notes/liked?_=1579461863> but was a redirect to <http://test.host/notes/liked?_=1579461862>.

        test/functional/user_sessions_controller_test.rb:344:in `block in <class:UserSessionsControllerTest>'
```

As you can see, the user is redirected to `http://test.host/notes/liked?_=1579461862` but there is a very small difference between that time (1579461862) and the time when the test is run (1579461862). This is failing certain prs like https://github.com/publiclab/plots2/pull/5666/.

Should I open a pr to fix this or is the `Time.now` parameter useful @publiclab/community-reps?

SidharthBansal:
@jywarren

SidharthBansal:
@cesswairimu what are your views

cesswairimu:
[...]

PR #7307, opened 2020-01-20
ADD: Post question system tests

VladimirMikulic:
I've refactored post_question_test.rb file to log in the user before each test because the user can't submit a question if he is not logged in.

Tests:
 - posting a question (https://github.com/publiclab/PublicLab.Editor/issues/113)
 - form validation messages

### DIFF
--- a/test/system/post_test.rb
+++ b/test/system/post_test.rb
@@ -119,32 +119,35 @@ class PostTest < ApplicationSystemTestCase
     # Make sure that image has been uploaded
     page.assert_selector('#preview img', count: 1)
   end
-  
-  test "changing and reverting versions works correctly for wiki" do
-    wiki = nodes(:wiki_page)
 
-    visit wiki.path
+  test "changing and reverting versions works correctly for wiki" do
+    visit '/wiki/wiki-page-path/'
+
     # save text of wiki before edit
-    old_wiki_content = find("#content").text
+    old_wiki_content = find("#content p").text
 
     find("a#edit-btn").click()
     find("#text-input").set("wiki text")
     find("a#publish").click()
 
     # view wiki
-    current_wiki_content = find("#content").text
+    current_wiki_content = find("#content p").text
     # make sure edits worked and text is different
     assert current_wiki_content != old_wiki_content
 
     find("a[data-original-title='View all revisions for this page.']").click()
+
     accept_confirm "Are you sure?" do
-      # revert to previous version of wiki
-      all("a", text: "Revert")[1].click()
+      # revert to the previous version of wiki
+      find('#row0 a[data-confirm="Are you sure?"]', text: "Revert").click()
     end
-    visit wiki.path
+
+    visit '/wiki/wiki-page-path/'
+
+    wiki_content = find("#content p").text
 
     # check old wiki content is the same as current content after revert
-    assert old_wiki_content == find("#content").text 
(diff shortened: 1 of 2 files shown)


VladimirMikulic:
I've fixed @Uzay-G's test that would mess up other system tests due to improper elements selecting.

test/system/post_test.rb
```diff
@@ -124,27 +124,30 @@ def assert_page_does_not_reload(message = "page should not reload")
...
-    old_wiki_content = find("#content").text
+    old_wiki_content = find("#content p").text
 
     find("a#edit-btn").click()
     find("#text-input").set("wiki text")
     find("a#publish").click()
 
     # view wiki
-    current_wiki_content = find("#content").text
+    current_wiki_content = find("#content p").text
```
Uzay-G: Sorry this might pass in testing but it doesn't accomplish the required purpose. The `#content` div is composed of different elements like headings, paragraphs and many others. Not just p tags

VladimirMikulic: Not really, there is only one p tag, I can add :first-of-type if you insist though.

VladimirMikulic: Still better than `#content` alone.

Uzay-G: For example, if you go to https://publiclab.org/about/. You will see the #content div has many different elements. Is there a way to get the text of all those elements and not just the p tags?

VladimirMikulic: This is testing page.

Uzay-G: Alright; but say later someone edits the test and adds a heading. The test will fail because the p tags will not have changed so the content will be marked equal

VladimirMikulic: No, it won't fail, the `p` tag is first and it would fetch it.

VladimirMikulic: It failed again :(

VladimirMikulic: I think it's because of `visit`, let's try again by setting fixed path.

test/system/post_test.rb
```diff
@@ -119,32 +119,33 @@ def assert_page_does_not_reload(message = "page should not reload")
     # Make sure that image has been uploaded
     page.assert_selector('#preview img', count: 1)
   end
-  
+
   test "changing and reverting versions works correctly for wiki" do
-    wiki = nodes(:wiki_page)
+    visit '/wiki/wiki-page-path/'
```
Uzay-G: Why did you change this? Also could you please explain how my test is affecting others?

VladimirMikulic: I am trying to figure that out. Every time I submit a PR with system tests, your test fails.
Let's wait for Travis, but it might be possible that it's fetching the wrong row, so I'll have to update it again.

VladimirMikulic: On localhost, it works fine though.

VladimirMikulic:
Finally, I've fixed your bug @Uzay-G.
You were asserting the text with `find` method dynamically fetching the element and of course it fails.
Also, please do fix #7303, my tests (and I believe tests from others) too were failing because of this issue.

Update your repo and be careful next time you are writing tests.

VladimirMikulic:
#7303 needs to be patched ASAP. Travis failed again due to it. @SidharthBansal ?

VladimirMikulic:
Ready to be merged :tada: :rocket:

PR #7322, opened 2020-01-22
ADD: Missing system tests

VladimirMikulic:
I've added the important system tests that were missing as a part of my last GCI task :rocket: 

Tests added:
 - "adding a location to the wiki"
 - "notification settings are working correctly"
 - "subscribing to a tag"
 - "unsubscribing from a tag"

Part of #5316

@jywarren could you review this? Thanks.


yarn.lock
```diff
@@ -18,6 +18,11 @@
     esutils "^2.0.2"
     js-tokens "^4.0.0"
 
+"@fortawesome/fontawesome-free@^5.12.0":
```
SidharthBansal: do we need this file?
I think you added it by mistake

SidharthBansal:
Maybe you are willing to work on more issues. Just I will say that we will not allow you to work till 30th January 2020 !!!
One week we expect you to take a break and enjoy. :-) 🚀 
cc\ @Uzay-G @HarshKhandeparkar

Uzay-G:
From now until the 30th?

yarn.lock
VladimirMikulic: No, I did not add this by mistake. We need it.

SidharthBansal:
Gci students will be on off from 24 January 18 utc to 30 January so that
they can relax

SidharthBansal:
@Uzay-G kindly review.

PR #7328, opened 2020-01-22
FIX: Removing tags Travis error

VladimirMikulic:
Travis was failing because find command would find 2 same elements.
We have to take a break from 24th to 30th of January as per GCI rules and I don't want to leave my code failing.

I've noticed that the last build failed because of my "removing" tags test.
This is my patch for it. I apologise for the inconvenience.

### DIFF
--- a/test/system/post_test.rb
+++ b/test/system/post_test.rb
@@ -48,11 +48,20 @@ class PostTest < ApplicationSystemTestCase
 
     find('a#tags-open').click()
 
-    find('.tag-input').set('nature').native.send_keys(:return)
-    find('.tag-input').set('mountains').native.send_keys(:return)
+    tag_input_box = find('.tag-input')
 
-    find('.tags-list p.badge .tag-delete').click()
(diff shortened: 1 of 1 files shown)


VladimirMikulic:
@SidharthBansal the patch is ready.

SidharthBansal:
Great

SidharthBansal:
Break is for enjoyment and relaxation 
There is no policy that you have to strictly take a break. We are just recommending it

VladimirMikulic:
You're right @SidharthBansal. I misquoted the rules. It's the review period.
The main build is still failing for some reason, but the tags controller is the problem this time.

Could you manually restart the build?
This test is a week old and it has never failed before. On the localhost, it also works.

SidharthBansal:
I don't have permissions. Let's ping @cesswairimu

VladimirMikulic:
The build is stable, the patch did work :rocket: 
It was probably some minor accident with the controller.

cesswairimu:
Awesome :tada:

PR #7329, opened 2020-01-22
ADD: system tests for moderation, posts and comments

Uzay-G:
For #5316

I have included tests for the following functionality in this pr:

- spamming and unspamming a page using the `/spam` admin dashboard
- spamming a comment
- reacting to a comment in the same way we can react on github
- giving barnstar works correctly

What do you think @SidharthBansal?


Uzay-G:
Thanks a bunch @SidharthBansal! :+1:

PR #7338, opened 2020-01-23
Fix tag suggestions to show all relevant ones

### LINKED ISSUE #7289: Very few suggestions show up for subscription
On this page [link](https://publiclab.org/subscriptions) when I try to add new tags to subscribe to then very few suggestions show up in the drop down like balloon mapping, spectrometer and thermal flashlight .
Screenshot from 2020-01-18 16-27-00
In this field 
bug

What I was expecting was to find all tags orders in some manner to appear when I type a letter like if I type 'a' then all the tags with 'a' in it should appear which doesn't seem to happen .

imajit:
Fixes #7289 
The [subscriptions](https://publiclab.org/subscriptions) page does not give tag suggestions based on tags saved in database .Instead it has hard-coded values .
<br>
The current results
Screenshot from 2020-01-24 02-53-48
<br>
Changes 
Screenshot from 2020-01-24 02-54-20
Instead of fixed values it now shows values by fetching them from database .


app/models/tag.rb
```diff
@@ -407,6 +407,16 @@ def self.graph_data(limit = 250)
     end
   end
 
+  def self.all_tags_by_popularity
+    not_power_tags = []
+    Tag.all.order('count DESC').each do |tag|
```
Uzay-G: You can refactor this `each` block to a [select iterator](https://www.rubyguides.com/2019/04/ruby-select-method/). They conserve all the elements of the array that match your condition. It would be something like this:
`not_power_tags = Tag.all.order('count DESC').select { |tag| tag.name.include? ":" }`

What do you think?

This is good work :+1:

imajit: Sure will do this in next commit .

app/models/tag.rb
```diff
@@ -407,6 +407,10 @@ def self.graph_data(limit = 250)
     end
   end
 
+  def self.all_tags_by_popularity
+    not_power_tags = Tag.all.order('count DESC').select { |tag| !(tag.name.include? ":") }.uniq(&:name).map(&:name)
```
Uzay-G: Could you remove the assignment to `not_power_tags` and just keep the expression? Thanks!

imajit:
@Uzay-G can you please review this pr

Uzay-G:
Please be more descriptive in the titles for your pr's. Thanks!

PR #7343, opened 2020-01-24
Refresh markers on initial page load

### LINKED ISSUE #6910: Map markers for nodes only appear after map interaction, not on initial load

nstjean:
Fixes #6910 (<=== Add issue number here)

I don't know why the LEL layers aren't showing up at all on my localhost, but this shows on a page refresh that the content markers are behaving as they should be.

Peek 2020-01-24 11-49

### DIFF
--- a/app/assets/javascripts/leaflet_helper.js
+++ b/app/assets/javascripts/leaflet_helper.js
@@ -108,6 +108,7 @@
 
        if(typeof mainLayer !== "undefined" && mainLayer !== ""){
            if(mainLayer === "people"){
+               peopleLayerParser(map, markers_hash);
                
                map.on('zoomend' , function () {
                   peopleLayerParser(map, markers_hash);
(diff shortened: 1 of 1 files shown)


nstjean:
@jywarren This is ready to go, other than the travis test error that seems unrelated?

jywarren:
Awesome! Yeah... annoying system test! Hm. Maybe need to ask for input from vladimir again, but he's hard at work on https://github.com/publiclab/Leaflet.DistortableImage/pull/512

jywarren:
Great!

PR #7358, opened 2020-01-27
Set map default zoom level

### LINKED ISSUE #7357: View on a map from notes pages results in "NaN" in URL
Hi @nstjean - i think maybe this is an unintended effect, but I see "view on a map" on pages like this:

https://publiclab.org/notes/warren/11-16-2019/mashapaug-pond

But it leads to this URL, probably because we aren't getting a proper slug from it, since it's not a wiki page? 

https://publiclab.org/map#NaN/41.79/-71.43

But, I'm not sure about that. I guess we should be using `node.id` on notes, since they require more than the slug as a unique id. What do you think?

This seems like it'll be a relatively simple fix, would you mind taking a look? 

All the buttons across the site now look AMAZING!!! 🎉

nstjean:
Fixes #7357 (<=== Add issue number here)

This sets a default zoom, lat and lon in the map page so that even if a parameter is missing the map still loads.

### DIFF
--- a/app/views/map/map.html.erb
+++ b/app/views/map/map.html.erb
@@ -51,9 +51,9 @@
         
       const params = urlHash.getUrlHashParameters();
      
-      var lat = parseFloat(params.lat);
-      var lon = parseFloat(params.lon);
-      var zoom = parseFloat(params.zoom);
+      var lat = parseFloat(params.lat) || 10;
+      var lon = parseFloat(params.lon) || 1;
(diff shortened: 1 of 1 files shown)


nstjean:
Can we re-start the test, @jywarren ? This shouldn't have anything to do with post tags.

jywarren:
Yep, doing so now, thank you!

jywarren:
Awesome! I'll merge and publish. Thanks for the quick response!!!

PR #7371, opened 2020-01-28
Move map dependencies to header to fix error with map popups not working

### LINKED ISSUE #7281: Map popups not working
@sagarpreet-chadha reported this, we need to collect more information on it.

***

([original post](https://github.com/publiclab/plots2/issues/6887#issuecomment-575665530))

> Regarding popups not showing on markers:

> If we change this line `popup = new Popup(options).setContent(popup);` to  `popup = new Popup(options).setContent(popup._content ? popup._content : popup);` in leaflet repo line:

> https://github.com/Leaflet/Leaflet/blob/d1a1e97b8290f642eb677284af53c2db64199a76/src/layer/Popup.js#L320

> Everything works fine then!

> One way to change this line is to use https://www.npmjs.com/package/patch-package , this changes that line in node modules folder after we do yarn install. The advantage is that we can still update leaflet version if there is any (we will kind of become static user of leaflet if we fork leaflet and do this change).

***

My response:

> oh goodness, i totally misunderstood, please forgive me @sagarpreet-chadha ! It's a Leaflet bug? I guess I'd prefer to fork Leaflet and fix it and point to our branch, but is it a known issue, that might be fixed by them soon? Or, is there any way for us to override this line in our code, after including Leaflet? 

> @crisner @nstjean what do you think? I guess we should not hold back the publication of PublicLab.org if we don't have a quick path to fix this yet. I'm opening an issue where we can discuss this a bit more, would love to hear from you. Thanks!

Possibilities:

* Fork and fix Leaflet
* Find issue at Leaflet repo and see how fast their solution is coming
* File solution PR with Leaflet if it's not there already
* Locally override the function in our code until it's fixed upstream

What do folks think? Let's dig a bit on the Leaflet project to see what's up?

nstjean:
Fixes #7281  (<=== Add issue number here)

Map dependencies cannot be included multiple times on a map without causing an error with the map popups. I have moved the include to the header file and removed it from the other pages. Now the popups work and all the other maps are still working as well.

FireShot Capture 241 - 🎈 Public Lab_ Inline Maps - localhost

FireShot Capture 243 - 🎈 Public Lab_ Notes on England - localhost

FireShot Capture 244 - 🎈 Public Lab_ Map - localhost

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -197,3 +197,5 @@
       opacity:0.5;
   }
 </style>
+
+<%= render :partial => "map/mapDependencies" %>
--- a/app/views/map/_inlineLeaflet.html.erb
+++ b/app/views/map/_inlineLeaflet.html.erb
@@ -1,5 +1,3 @@
-<%= render :partial => "map/mapDependencies" %>
-
 <% unique_id = rand(1000) %>
 
   <style>
--- a/app/views/map/_leaflet.html.erb
+++ b/app/views/map/_leaflet.html.erb
@@ -1,4 +1,3 @@
-<%= render :partial => "map/mapDependencies" %>
 <% top_map ||= false %>
 <% zoom ||= lat.to_s.length.to_i + 6 %>
 <% unique_id = rand(1000) %>
--- a/app/views/map/_peopleLeaflet.html.erb
+++ b/app/views/map/_peopleLeaflet.html.erb
@@ -1,4 +1,3 @@
-<%= render :partial => "map/mapDependencies" %>
 <% unique_id = rand(1000) %>
 
   <style>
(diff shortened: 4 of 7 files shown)


jywarren:
Rad!!! 🎉

jywarren:
This is live! Can you confirm it's working? thanks so much!!!

nstjean:
It looks great!!

FireShot Capture 247 - 🎈 Public Lab_ Inline maps - publiclab org

jywarren:
Hi, sorry, i'm just noticing some oddness with the unearthing layer here: 

https://publiclab.org/wiki/onecranston#location

image

It seems that the zooming and layer loading are a little cross-wired. Like, i zoom but the unearthing layer doesn't. Could this have something to do with this PR, or something else we recently changed? cc @crisner as well -- thank you!

nstjean:
For me it looks like it does zoom properly after I am done with the zoom motion, but it doesn't react in between. Interesting. I don't think moving the dependency would have affected that but we are seeing some other weirdness too. I'll take a look.

PR #7378, opened 2020-01-28
DOCUMENT: System Tests

VladimirMikulic:
@jywarren as I promised. Here is the comprehensive guide on system tests :)

### DIFF
--- a/SYSTEM_TESTS.md
+++ b/SYSTEM_TESTS.md
@@ -1,0 +1,349 @@
+ 
+
+# Guide on System Tests in Ruby
+
+System tests in Ruby are special types of tests.
+
+They allow us to interact with our web app, test the UX and JavaScript of our application.
+
+This repository uses [Capybara](https://github.com/teamcapybara/capybara) library to perform system tests. It simulates user's interaction with our web app (visiting pages, clicking buttons, filling forms...)
+
+System tests are the slowest type of tests and we need to be careful when writing them.
+
+
+
+Visit the official [Capybara Docs](https://www.rubydoc.info/github/jnicklas/capybara/Capybara) to learn more.
+
+
+
+## Test's file structure
+
+The example code below shows how the normal test file looks like and introduces you to the basics of system tests.
+
+```ruby
+# Import system test configuration (driver, options...)
+require "application_system_test_case"
+
+# Class name should be descriptive and align with the filename
+class EditorTest < ApplicationSystemTestCase
+    # The time Capybara should wait for a function to complete (visit, find, all...)
+    Capybara.default_max_wait_time = 60
+    # Other includes and settings
+    
+    def setup
+       # This function doesn't exist in every test file, it's optional.
+       # If we define it, it'll run before each test.
+       # For example, we want to login the user before each test.
+    end
+    
+    # The standard test structure
+    test 'creating 3x3 table in Markdown' do
+       # Code ...
+    end
+    
+    # Other tests
+end
+```
+
+
+
+## Basic functions
+
+#### **visit '/path'**
+
+`visit` navigate  to the specific URL. If the page doesn't load in certain period of time(timeout) the test fails.
+
+
+
+#### assert_equal 'expected_value', 'actual_value'
+
+`assert_equal` does the assertion. If it evaluates to false, the test fails.
+
+
+
+#### click_on 'text'
+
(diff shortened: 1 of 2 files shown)


Uzay-G:
This is nice @VladimirMikulic! I will review this :+1:

@jywarren I think it would also be interesting to document how to test omniauth integration in our `login_systems.md` file. I had to search quite a bit to find how to write those tests so I think it will be useful and I will create an issue

SYSTEM_TESTS.md
```diff
@@ -0,0 +1,349 @@
...
+
+# Guide on System Tests in Ruby
+
+System tests in Ruby are special types of tests.
+
+They allow us to interact with our web app, test the UX and JavaScript of our application.
+
+This repository uses [Capybara](https://github.com/teamcapybara/capybara) library to perform system tests. It simulates user's interaction with our web app (visiting pages, clicking buttons, filling forms...)
+
+System tests are the slowest type of tests and we need to be careful when writing them.
```
Uzay-G: Could you add detail here to explain that they need to be minimal and that because of their slow speed, every test need to be crucial and efficient? Thanks!

SYSTEM_TESTS.md
```diff
@@ -0,0 +1,349 @@
...
+
+## Basic functions
+
+#### **visit '/path'**
+
+`visit` navigate  to the specific URL. If the page doesn't load in certain period of time(timeout) the test fails.
+
+
+
+#### assert_equal 'expected_value', 'actual_value'
```
Uzay-G: This part is common to all tests in the default rails testing framework. Maybe we can remove it?

VladimirMikulic: That's why I put it in the 'Basic functions' sections.

SYSTEM_TESTS.md
VladimirMikulic: The sentence on line 11 says exactly that.

SYSTEM_TESTS.md
```diff
@@ -0,0 +1,349 @@
...
+   # Select the element that has the **exact same** text
+   find("#btn-submit", exact_text: "Submit")
+end
+```
+
+> IMPORTANT NOTE: If `find` selects multiple elements, the test fails. If you are expecting multiple elements you should either use `all` or provide 'match' argument to select a specific element.
+
+
+
+I hope you see now why `find` is one of my favourite functions when I need to select an element. Its possibilities are endless.  If this doesn't cover your case, you can always check out the [official docs](https://rubydoc.info/github/jnicklas/capybara/master/Capybara%2FNode%2FFinders:find).
```
Uzay-G: Maybe we can make this part less personal with something like this: "`find` is an incredibly useful function to select elements and it is a crucial component for system testing."

SYSTEM_TESTS.md
Uzay-G: No I know but in just says `we need to be careful` and I think it would be nice to detail that that means the tests need to be efficient and minimal

SYSTEM_TESTS.md
VladimirMikulic: :+1:

SidharthBansal:
Ping me when you both arrive to climax(that I can review and merge the pr) 🎉 !

jywarren:
Wow this looks amazing. Shall we add links to it from the welcomebot messages (maybe) and the README or existing testing documentation? Thank you!!!

SidharthBansal:
Yeah I agree with @jywarren and @Uzay-G

VladimirMikulic:
Thanks, @jywarren & @Uzay-G. It's nice to hear that you like it.
I've added the link to the SYSTEM_TESTS.md in README.
I don't think it's necessary to add it to the welcome bot-message though :)

cesswairimu:
This is awesome :tada: :tada:

SYSTEM_TESTS.md
VladimirMikulic: Well, I've covered some of the edge cases of system tests. It really depends on what someones mean by minimal. In the context of tests in general, I don't think that 'minimal' is the right word to use. A person writing tests should know what code needs to be tested.

In the end, @jywarren decides if a test should be added or not. In other words, is the test crucial to certain functionality?

cesswairimu:
this one is ready for merge @VladimirMikulic ?

VladimirMikulic:
Yes, it is @cesswairimu :)

PR #7393, opened 2020-01-30
Move password confirmation field that was in wrong place

### LINKED ISSUE #7391: Password confirmation box in wrong place
On the Edit Profile page `/profile/edit` I see username, email, Change your password, Current password, and Confirmation.

Confirmation says "Retype your password to confirm". But this is next to "Current password". So what are we confirming, the new password or the old one? We should be confirming the new one, to make sure the user knows what they typed in is correct. The old password is being checked against what is in the database, it does not need to be confirmed.

FireShot Capture 254 - 🎈 Public Lab_ Edit - localhost

"Confirmation" should be next to "Change your password".

DouglasLutz:
Fixes #7391

Altered field order in the form and the tab indexes.

Page after alterations:
change-password

### DIFF
--- a/app/views/users/_create_form.html.erb
+++ b/app/views/users/_create_form.html.erb
@@ -59,7 +59,7 @@
       </div>
 
       <div class="form-group col-md-6">
-        <label for="password_confirmation"><%= t('users._form.confirmation') %></label>
+        <label for="password_confirmation"><%= t('users._form.password_confirmation') %></label>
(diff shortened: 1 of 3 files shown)


SidharthBansal:
Thanks for the PR.
Kindly move `current password` above `Change your password` and `Password Confirmation`
Also change `Confirmation` to  `Password Confirmation`
Superb work so far :tada:

DouglasLutz:
@SidharthBansal 
Also had to change `Confirmation` to `Password Confirmation` in new user form since it was using the same translation.
after-changes

PR #7319, opened 2020-01-21
Header line removed and Menu icon made more visible for mobile view

NitinBhasneria:
Fixes #7312 (<=== Add issue number here)

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [x] PR is descriptively titled 📑 and links the original issue above 🔗
* [x] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [x] code is in uniquely-named feature branch and has no merge conflicts 📁
* [x] screenshots/GIFs are attached 📎 in case of UI updation
* [x] ask `@publiclab/reviewers` for help, in a comment below

Screenshot from 2020-01-21 17-30-24
Screenshot from 2020-01-21 17-30-22
> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software 

Thanks!

PR #7396, opened 2020-01-30
FIX: Wiki revisions test

VladimirMikulic:
This test has already been fixed before but recently it has failed again
in #7319. The problem might be related to our selection of "Revert"
button.

Inside of the button, we have an empty space before the text
and a font-awesome icon(\<i>).
We were selecting the button with #row0 and the text of "Revert" which
probably caused the issue.

I've added the `btn-revert` class to avoid this strange behaviour.


VladimirMikulic:
Thanks @Uzay-G. @SidharthBansal it's ready.

PR #7405, opened 2020-01-31
Makes user card size equal

### LINKED ISSUE #7402: Ensure the user card height is equal for all
If the user does not have badges and profile picture the user card will shrink.
Screenshot from 2020-01-31 18-42-00

urvashigupta7:
Fixes #7402 
### Results
Screenshot from 2020-01-31 19-49-12


urvashigupta7:
@cesswairimu I think, It is all good now.

PR #7511, opened 2020-02-16
Updating post time to published time 

### LINKED ISSUE #7498: Change post date on draft to publish date
When someone starts a post on Public Lab it captures the date of the draft was started. Then when it's finally published, it holds the original date. This means it doesn't show up on the top of the feed and new posts end up buried in the history. can we change this so that the data it is published is what is used rather than the date the draft started. here's an example of a post I'm talking about. Was started in August, but posted yesterday. 

https://publiclab.org/notes/WendyBrawer/08-20-2019/how-to-get-street-trees-for-your-community?_=1581707285

@ebarry to follow

imajit:
Fixes #7498 
After the changes 
ezgif com-video-to-gif(3)


cesswairimu:
@imajit we need to find a better way to capture this...This will be updated any time there is a update. say, change of title or smth...any ideas?

SidharthBansal:

jywarren:
I think we would need to check if it's a draft becoming a non-draft, and set the timestamp at that moment.

PR #7520, opened 2020-02-17
Reload fucntion added upon adding location tags

NitinBhasneria:
Fixes #7359 (<=== Add issue number here)

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [x] PR is descriptively titled 📑 and links the original issue above 🔗
* [x] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [x] code is in uniquely-named feature branch and has no merge conflicts 📁
* [x] screenshots/GIFs are attached 📎 in case of UI updation
* [x] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software 

Thanks!

PR #4888, opened 2019-02-27
try creating a system test and running in Travis

jywarren:
Closes #3683

https://guides.rubyonrails.org/testing.html#implementing-a-system-test

https://chriskottom.com/blog/2017/04/full-stack-testing-with-rails-system-tests/

PR #7562, opened 2020-02-25
WIP: Formatting and improving map popups

nstjean:
On /map we now have people displaying from the PLpeople layer in LEL, displayed via LeafletBlurredLocationDisplay.

We also have all the site content being displayed leaflet_helper.js. This PR updates the styling of those popups.

FireShot Capture 309 - 🎈 Public Lab_ Map - localhost

### DIFF
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -358,7 +358,7 @@ module Srch
               longitude: model.lon,
               blurred: model.blurred?,
               created_at: model.created_at,
-              doc_image_url: !model.profile_image? ? model.profile_image : "",
+              doc_image_url: model.profile_image ? model.profile_image : "",
             )
           end
           DocList.new(docs, search_request)
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -356,7 +356,9 @@ module Srch
               doc_title: model.username,
               latitude: model.lat,
               longitude: model.lon,
-              blurred: model.blurred?
+              blurred: model.blurred?,
+              created_at: model.created_at,
+              doc_image_url: !model.profile_image? ? model.profile_image : "",
             )
           end
           DocList.new(docs, search_request)
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -311,9 +311,11 @@ module Srch
 
         if results.present?
           docs = results.map do |model|
+            doctype = model.has_power_tag('question') ? 'QUESTION' : 'NOTE'
+            doctype = 'WIKI' if model.type == 'page'
             DocResult.new(
               doc_id: model.nid,
-              doc_type: 'PLACES',
+              doc_type: doctype,
               doc_url: model.path(:items),
               doc_title: model.title,
               doc_author: model.user.username,
@@ -321,7 +323,11 @@ module Srch
               score: model.answers.length,
               latitude: model.lat,
               longitude: model.lon,
-              blurred: model.blurred?
+              blurred: model.blurred?,
+              place_name: model.has_power_tag('place') ? model.power_tag('place') : '',
+              created_at: model.created_at,
+              # time_since: distance_of_time_in_words(model.created_at, Time.current, { include_seconds: false, scope: 'datetime.time_ago_in_words' }),  # works, but really slows down the search results
+              # comment_count: model.comments_viewable_by(current_user).length  # causes an error because of current_user?
             )
           end
           DocList.new(docs, search_request)
--- a/app/api/srch/search.rb
+++ b/app/api/srch/search.rb
@@ -358,7 +358,7 @@ module Srch
               longitude: model.lon,
               blurred: model.blurred?,
               created_at: model.created_at,
-              doc_image_url: model.profile_image ? model.profile_image : "",
+              doc_image_url: model.profile_image ? model.profile_image : ""
             )
           end
           DocList.new(docs, search_request)
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -59,3 +59,4 @@
 //= require validation.js
 //= require submit_form_ajax.js
 //= require urlMapHash.js
+//= require timeago.js
--- a/app/assets/javascripts/leaflet_helper.js
+++ b/app/assets/javascripts/leaflet_helper.js
@@ -86,11 +86,12 @@
 
                if(markers_hash.has(mid) === false){
                   var popup_content = "";
-
-                  console.log(image_url);
-                  if (image_url) popup_content += "<img src='" + image_url + "' class='popup-thumb' /><br>";
+                  if (image_url) popup_content += "<img src='" + image_url + "' class='popup-thumb' />";
                   popup_content += "<h5><a href='" + url + "'>" + limit_words(title, 10)  + "</a></h5>";
-                  popup_content += "<span>" + nodetype + " by <a href='https://publiclab.org/profile/" + author + "'>@" + author + "</a> " + time_since + "</span><br>";
+                  popup_content += "<div class='popup-two-column'>";
+                     popup_content += "<div class='popup-stretch-column'>" + nodetype + " by <a href='https://publiclab.org/profile/" + author + "'>@" + author + "</a> " + time_since + "</div><br>";
+                     if (nodetype.toLowerCase() === "wiki") popup_content += "<div class='map-slug popup-shrink-column'><a href='/map/" + url.split('/').pop() + "'>#</a></div>";
+                  popup_content += "</div>";
                   // if (place_name) popup_content += "<span><b>Place: </b>" + place_name + "</span><br>";
 
                   var popup = L.popup({
--- a/app/assets/javascripts/leaflet_helper.js
+++ b/app/assets/javascripts/leaflet_helper.js
@@ -87,10 +87,11 @@
                if(markers_hash.has(mid) === false){
                   var popup_content = "";
 
+                  console.log(image_url);
                   if (image_url) popup_content += "<img src='" + image_url + "' class='popup-thumb' /><br>";
                   popup_content += "<h5><a href='" + url + "'>" + limit_words(title, 10)  + "</a></h5>";
                   popup_content += "<span>" + nodetype + " by <a href='https://publiclab.org/profile/" + author + "'>@" + author + "</a> " + time_since + "</span><br>";
-                  if (place_name) popup_content += "<span><b>Place: </b>" + place_name + "</span><br>";
+                  // if (place_name) popup_content += "<span><b>Place: </b>" + place_name + "</span><br>";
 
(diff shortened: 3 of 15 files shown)


nstjean:
This is how they look on inline maps - currently they are styled the same but everything can easily be overridden for inline maps if we want to adjust the size of them.

FireShot Capture 313 - 🎈 Public Lab_ Inline Maps - localhost

FireShot Capture 314 - 🎈 Public Lab_ Inline Maps - localhost

nstjean:
Question: On the popups do we want to display the location lat and lon? I can also display other information about the post if we want to.

I tried to display the number of comments but that caused an error because it depends on if the person viewing the page is a moderator/admin or not. The API doesn't like doing calculations based on current user, it causes an error.

nstjean:
Added the # link for wikis to link to their /map/SLUG

FireShot Capture 315 - 🎈 Public Lab_ Map - localhost

nstjean:
I don't know what's up with the tests... it's having failures and timeouts on things I didn't touch.

cesswairimu:
Restarting travis

alaxalves:
@nstjean Whats the output when you locally run the system tests? I had to suppress the logs for it in Travis with `> /dev/null` because it could not support their size.

nstjean:
I get some failures when I run the system tests locally, but it does run properly. It looks like travis isn't even starting the tests.

cesswairimu:
:thinking: Taking a look

jywarren:
> Question: On the popups do we want to display the location lat and lon? I can also display other information about the post if we want to.

I think not, it seems extraneous!

This is looking awesome though!!!

alaxalves:
> I get some failures when I run the system tests locally, but it does run properly. It looks like travis isn't even starting the tests.

It seems it doesnt run because of the suppressed logs, but the system tests run properly on Travis. The problem was that the Travis machine could support to output the amount of logs we produce (mostly byte-represented screenshot images).

jywarren:
Thanks for the suggestion, @alaxalves !!!

jywarren:
Taking a closer look at this now...

jywarren:
OK, 11 days ago this change was made, and that seems to have added the `> /dev/null` to the system tests line: https://github.com/publiclab/plots2/pull/7568

Looking deeper...

jywarren:
So @alaxalves you're saying that we can't see the error because the output of `script: bundle exec rails test:system > /dev/null` is being dumped to `/dev/null`? Why was it not an issue before? And, could we store it but then filter out the images before outputting it to the Travis log?

jywarren:
Would one possible step be to remove the `> /dev/null` temporarily in this PR to try to get a look at the error output?

alaxalves:
> So @alaxalves you're saying that we can't see the error because the output of `script: bundle exec rails test:system > /dev/null` is being dumped to `/dev/null`? Why was it not an issue before? And, could we store it but then filter out the images before outputting it to the Travis log?

Sorry for late reply Jeff,
Yes, the error log is being dumped to /dev/null since Travis could handle the log size and crashed every build. I think that it has not been an issue before because we had been using Docker to internally run our tests, and Docker might have been suppressing the error messages.

nstjean:
So is there anything I can do here? @jywarren @alaxalves

jywarren:
Hmm. Let me think -- wouldn't some of our system tests have failed if that were happening? I'm going to look at the commits/test runs leading up to the change to see if I can learn more...

jywarren:
A number of test failures predated this but presumably they all passed at some point but not after the merge (when they get re-run). I remember we were having some inconsistent test results and rebooting them had seemed to work. I'm now going back to the most recent one that did work:

image

https://github.com/publiclab/plots2/pull/7520

jywarren:
We were getting this error: https://travis-ci.org/github/publiclab/plots2/jobs/653371760#L3915

```
ERROR["test_removing_tags_from_the_post", #<Minitest::Reporters::Suite:0x00007f62dc8433e8 @name="PostTest">, 93.41379365400002]
 test_removing_tags_from_the_post#PostTest (93.41s)
Minitest::UnexpectedError:         Mysql2::Error: 
            app/controllers/tag_controller.rb:396:in `delete'
```
which could be a disk space error...

jywarren:
Other errors include lots of:

```
Minitest::UnexpectedError:         Selenium::WebDriver::Error::TimeoutError: timeout: Timed out receiving message from renderer: 10.000
          (Session info: headless chrome=75.0.3770.142)
          (Driver info: chromedriver=74.0.3729.6 (255758eccf3d244491b8a1317aa76e1ce10d57e9-refs/branch-heads/3729@{#29}),platform=Linux 4.15.0-1028-gcp x86_64)
            test/system/search_test.rb:24:in `block in <class:SearchTest>'
```

which could also be disk space related. I'm going to try removing the /dev/null portion to see if we can get it to intermittently pass. 

Then we should consider ways to not log the screenshots, rather than dumping the whole log.

jywarren:
So, the screenshots don't seem to be getting logged: 

```
[Screenshot]: tmp/screenshots/test_front_page_with_navbar_search_autocomplete.png
[Screenshot]: tmp/screenshots/test_searching_an_item_from_the_homepage.png
[Screenshot]: tmp/screenshots/test_viewing_the_dashboard.png
[Screenshot]: tmp/screenshots/test_viewing_the_dashboard.png
[Screenshot]: tmp/screenshots/test_viewing_the_settings_page.png
[Screenshot]: tmp/screenshots/test_viewing_question_post.png
[Screenshot]: tmp/screenshots/test_simple-data-grapher_powertag.png
[Screenshot]: tmp/screenshots/test_blog.png
```

however they are getting saved. If disk space is the issue, could we delete them each after they're uploaded? I would have thought `/tmp/` files would get auto deleted... but not sure about that.

jywarren:
@alaxalves - what exactly did you see about the log size being too big? Is it storage space, or something else? https://travis-ci.org/github/publiclab/plots2/jobs/651968744 doesn't seem too wildly long - system tests logs are about 3k lines. Is that breaking something?

jywarren:
Aha - ok, https://travis-ci.org/github/publiclab/plots2/jobs/663629236#L1865 does show the data-urls inline. its giant, and got:

> The job exceeded the maximum log length, and has been terminated.

But, based on comparison to previous logs, i don't think we used to put these into logs. Looking at where that happens and if we can suppress it just for the screenshots...

jywarren:
Hmm. I can't see where system test output is configured. When originally installed in https://github.com/publiclab/plots2/pull/4888/files and https://github.com/publiclab/plots2/pull/5868/files we don't seem to be customizing output. Let's see if capybara has a config to customize the log output...

jywarren:
I'm also not convinced this is a data-url:

```
]1337;File=name=dGVzdF9ibG9nLnBuZw==;height=400px;inline=1:iVBORw0KGgoAAAANSUhEUgAABXgAAAV4CAYAAAAaC7U3AAAAAXNSR0IArs4c6QAAIABJREFUeJzs3XdcVHe+//E3MPTeBAGxUCwINlTsGmM0ppje2+5mk2ySm7ubbffu3m3Z397d7O69d7MlvfdimibGqDH2gmLDCoKA0ofemYH5/YFMREBBgeGY1/PxyCPTzjmf+TJOeX+/5/t1Sp453yYAAAAAAAAAgOE4O7oAAAAAAAAAAMCFIeAFAAAAAAAAAIMi4AUAAAAAAAAAgyLgBQAAAAAAAACDIuAFAAAAAAAAAIMi4AUAAAAAAAAAgyLgBQAAAAAAAACDIuAFAAAAAAAAAIMyOboAAAAAYDBxdXXTkPBwBYeEyMvL29HlXJT6+jqVmc0qKSqSxdLcL8e4lNrrQg1EOwMAAHTHKXnmfJujiwAAAAAGA1dXN8WPGaOqykqVmc1qaKh3dEkXxdPTS8EhIfIPCFDG0aN9Hj5eau11ofq7nQEAAM6FgBeXPH9/f82ZM0fJycmKiIhwdDnAoBQVFiI/Px9Hl9FBRWWlNmzcrFffeEtlZeWOLgfAt0TksGg5STp1Ms/RpfSp/npel2p7XSjaAwAAOAJTNOCS5u/vr0cffVRBAX6Kjxmp8LBQR5cEDEpNzRZHl9BJYECArl92jebPna37vv8DQl4AAyI4JESZR486uow+V242K27MmD4PHi/V9rpQ/dXOAAAA58Iia7ikzZkzRyFBAZo7cxrhLmBQgYGB+u69dzu6DADfEl5e3pfkNAMNDfX9Mj/updpeF6q/2hkAAOBcCHhxSUtOTtbY+FhHlwHgIs2eNcPRJQAAAAAAMCgR8OKSFhERoeCgAEeXAeAiBQcFOboEAAAAAAAGJQJe
```

jywarren:
`dGVzdF9ibG9nLnBuZw==` is `test_blog.png` in base-64, so for some reason we are dumping a file, it seems, to log. When did this begin?

jywarren:
Hmm, yeah - even the last PR before the #7568 changes does not try to output the files into the log: https://travis-ci.org/github/publiclab/plots2/jobs/654281714#L3717 although it does fail for some reason that could be related or not...?

jywarren:
So, in summary, before #7568 we were not outputting files into the logs, and the logs were reasonable length. Some other issue may have been causing intermittent test failures but it's not clear to me what the connection is unless it's disk space. 

If we can figure out why it's trying to output files into logs, we can probably solve this?

jywarren:
Hmm, i'm pretty stuck here, i think. @alaxalves maybe after i documented some of this research and digging, you have any new insights? Thank you!!!

keshav234156:
@jywarren I got a similar error here https://travis-ci.org/github/publiclab/plots2/jobs/653371760#L3915 .Restarting the Travis fixed it.

cesswairimu:
@nstjean could you try rebasing see if travis passes

alaxalves:
Hey @jywarren @cesswairimu Could you guys sum up the discussion here?

jywarren:
Hi @alaxalves -- thanks, I think my [last comment](https://github.com/publiclab/plots2/pull/7562#issuecomment-600250810) is unfortunately still the latest we know:

For some reason the new setup outputs screenshot files to the logs, but I confirmed that this wasn't the case before #7568, and so the goal is to figure out /why/ this is happening in the new tests setup and stop them from being logged, so that we can see the system tests log output properly and get this PR and others merged. Thanks!

nstjean:
> @nstjean could you try rebasing see if travis passes

Ok, pushing it up to see!

alaxalves:
@nstjean @jywarren I have fixed this PR here https://github.com/publiclab/plots2/pull/7715. I had to bypass some test asserts, but Natalie could work on those later. The problem was that we have a method named `take_screenshot` and it would output the screenshot binaries, now I have configured it to only ouput the screenshot path.

PR #7609, opened 2020-03-08
Fix incorrect tag count in /tags route

### LINKED ISSUE #7590: Incorrect count of tags shown

In the /tags route the cards for individual tags have wrong count of tags marked.
image
image

Here there are only 2 posts with this tag but 20 is shown as the count.

https://publiclab.org/tags

Tlazypanda:
Fixes #7590 
New method counter added in tag.rb file to compute count of tag correctly in /tags route.

## Screenshots   
Screenshot from 2020-03-08 17-54-42  
image

Thanks!


Tlazypanda:
@VladimirMikulic I have taken all wiki pages,questions and notes in the count ...let me know if only notes is to be taken :sweat_smile: can you kindly review? Thanks :v:

VladimirMikulic:
Looks good to me. Before approving, @Uzay-G could you take a look at this as well? Thanks.

Tlazypanda:
@Uzay-G can you kindly review? Thanks :v:

app/views/tag/_topicCard.html.erb
```diff
@@ -32,7 +32,7 @@
       </div>
 
       <div class="card-footer" style="background-color: inherit; border:none;">
-        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= tag.count %> <%= t('tag.index.more_posts') %> &raquo;</a>
+        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= Tag.counter(tag.name)-shown_nids.count - Tag.find_nodes_by_type(tag.name, type = 'note', limit = 3).where.not(nid: shown_nids).count %> <%= t('tag.index.more_posts') %> &raquo;</a>
```
Uzay-G: this looks good! I am just wondering why u r subtracting from the `counter` method. Shouldn't all the tags be counted in that method without requiring further operations?

Maybe I am missing something though. What does that subtraction do?

Tlazypanda: Actually this shows "x more posts" on the website so I am subtracting the number of posts that is shown already as you can see in this screenshot 
image

Tlazypanda:
@cesswairimu @SidharthBansal @jywarren can you kindly review? Thanks :v:

app/views/tag/_topicCard.html.erb
Uzay-G: Oh alright thanks. This looks good!

app/models/tag.rb
```diff
@@ -115,6 +115,10 @@ def self.find_nodes_by_type(tagnames, type = 'note', limit = 10)
         .order(order)
   end
 
+  def self.counter(tagname)
+    find_nodes_by_type(tagname, 'note', false).count + find_nodes_by_type(tagname, 'page', false).count + find_nodes_by_type("question:#{tagname}", 'note', false).count
```
jywarren: This looks really good. I do think we need to watch out a bit for performance issues on this, and will just ask for folks to reflect on performance once before we merge it -- see how the `find_nodes_by_type()` method is relatively heavy: 

https://github.com/publiclab/plots2/blob/80a649ab55b1ac574ac29ca3335ac98f80ccca60/app/models/tag.rb#L94-L116

And we're calling it ~4 times on this page then? Not a blocking concern but just one to watch out for. Thanks -- any thoughts on how to optimize or do folks think we can go ahead and just keep an eye on this page's load time using Skylight? (https://oss.skylight.io/app/applications/GZDPChmcfm1Q/recent/6h/endpoints)

Tlazypanda: @jywarren should I create another method for this entirely not calling any existing method?

Tlazypanda:
@jywarren @cesswairimu  I was hoping to record this contribution for Outreachy since it is a good issue can you let me know if there's anything else required to complete it or would it better if I link the issue as my contribution? Thanks :v:

sashadev-sky:
@Tlazypanda gonna chime in here and say that if you think you can make a more optimal query - that would be awesome and definitely give it a go! If not then we can merge and watch skylight.

@VladimirMikulic or @Uzay-G do you have any ideas for optimizing ?

VladimirMikulic:
Maybe combining all queries into a single function call?

Tlazypanda:
@sashadev-sky @jywarren @VladimirMikulic After some optimization here's the code that I have written can you have a look? If it works I will push it .. It is giving the same results as the previous code on testing locally. 

`  def self.counter(tagname)  
    Node.where(type: ['note','page'])  
        .where('term_data.name = ?', tagname)  
        .includes(:node_tag, :tag)  
        .references(:term_data)  
        .count()  
  end`

app/models/tag.rb
Uzay-G: I'm thinking it would be more efficient to query all nodes with that tagname with a single query instead of applying a type for each search. That way the query would ignore the type and we would only have to use one instead of 4. I'm not very familiar with how the tag organization works in the db. Do you know who could help with this @jywarren?

jywarren: yes indeed it would be fine to query once for all types. Good idea!

jywarren:
I think this looks right, @Tlazypanda! Thank you!

Tlazypanda:
@jywarren Pushed the code :v: Thank you :smile: Can you kindly review?

app/models/tag.rb
```diff
@@ -115,6 +115,14 @@ def self.find_nodes_by_type(tagnames, type = 'note', limit = 10)
         .order(order)
   end
 
+  def self.counter(tagname)
```
Uzay-G, line 118: Nice refactor! You tested this locally and it worked?

jywarren:
This looks great. Merging! Thank you!!!

PR #7772, opened 2020-03-31
upgrading LBLD to v1.3.0

jywarren:
okkkkk will need to upgrade Leaflet.BlurredLocation first...


jywarren:
OK, so we actually need to do this first: https://github.com/publiclab/leaflet-environmental-layers/pull/422 so I'm going to close this!

Issue #5095, opened 2019-03-15
Mysterious error message "User owmloading not found" shown


HarshKhandeparkar:
I got this error when I opened my profile
![Screenshot_20190316_012912](https://user-images.githubusercontent.com/34770591/54458837-f56b0000-478a-11e9-9144-738ea5616b24.jpg)

I have used the website for the first time so please correct me if I have asked something wrong. Also my images were not posting on comments and even on profile image. Is this some internet problem or something else?

madeofhuman:
@gauravano @jywarren I want to look into this next.

HarshKhandeparkar:
Can someone explain to me what this bug is?

HarshKhandeparkar:
What does the error mean?

gauravano:
That would be great @madeofhuman!

gauravano:
@HarshKhandeparkar Actually, this alert is shown in some instances, mostly in login flow. The reason for this is still not known.

HarshKhandeparkar:
Uh ok. I was also not able to upload imgs. What could be the reason?

gauravano:
Locally or in prod?

HarshKhandeparkar:
prod!

HarshKhandeparkar:
I have good internet. 8MB/s upload

gauravano:
There's an issue open for upload of the image in the comments. You faced issue there or somewhere else?

HarshKhandeparkar:
Yes, in comments.

CleverFool77:
Hi
I faced similar issue while I was editing my note.

HarshKhandeparkar:
@guaravano I also couldn't upload a profile img.

skilfullycurled:
If it helps diagnose the problem, I've had this recently as well.  For me, it reveals itself after clicking on a link to see a filtered list of my content from my profile.  

For example: `https://publiclab.org/tag/activity:*/author/bsugar`

Doesn't effect any functionality.

Firefox 66.0.3
macOS 10.12.6

jywarren:
I think i got this. `public/lib/leaflet-environmental-layers/dist/LeafletEnvironmentalLayers.js:             imageLoadingUrl: 'owmloading.gif', // URL of loading image rela`

this is an artifact from LEL, for the "loading" gif. it may try to load this at the wrong location here:

[...]

PR #4633, opened 2019-01-16
LBL Slider added and spinner working (during auto-locate me !) .

sagarpreet-chadha:
### **Slider :**

screenshot 2019-01-16 at 11 08 55 pm

🔹 🔹 🔹 🔹 🔹 🔹 🔹 🔹 🔹 🔹 🔹 🔹 🔹 🔹 🔹 

### **Spinner :** 

screenshot 2019-01-16 at 11 11 14 pm

PR #7939, opened 2020-05-21
Minify bootstrap slider script and add defer tag

### LINKED ISSUE #7918: Replace bootstrap-slider.js with minified js and add defer tag
Currently, the `bootstrap-slider.js` file is called from cdn and can be replace with minified js for better performance. Since we don't want to wait for fetching and executing the script , we can use the defer tag for it to execute after dom construction (since it depends on jquery we shouldn't use async tag).

Tlazypanda:
Fixes #7918 
Bootstrap slider script minified and defer tag added to allow execution after html parsing (not used async because this script requires jquery).

Thanks!


Tlazypanda:
@jywarren @cesswairimu @emilyashley can you kindly review ? Thanks :v:

Tlazypanda:
@VladimirMikulic Thanks a lot for the review :tada:

app/views/layouts/application.html.erb
```diff
@@ -91,11 +91,11 @@
      background: #BABABA;
     }
 
-  </style>  
-  
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-slider/10.2.0/bootstrap-slider.js"></script>
+  </style>
+
+  <script defer src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-slider/10.2.0/bootstrap-slider.min.js"></script>
```
jywarren, line 96: Hmm, i'm just trying to think of where this is used so that it's clear it'll still work! 

See how you can use the "blame" view to find out: https://github.com/publiclab/plots2/blame/master/app/views/layouts/application.html.erb#L95

Given it links here: https://github.com/publiclab/plots2/pull/4633

Could you test that function to be sure, and just note that in this PR? That'll give us extra confidence! But this does look good to go!

Tlazypanda:
Hey @jywarren @cesswairimu ! So while going through the function and testing it out here are some observations- 
These observations are from publiclab.org and not local 

-  This is invoked by adding location from dashboard but on clicking share location from profile page nothing happens.
- On adding location modal, the slider works by zooming in but when we save it and check it out in the profile page it is saved without any zoom.
- Also after adding your location it still shows add location on dashboard
- Also after adding location for first time, it doesn't seem to update for a second time
- Some weird alert comes sometimes couldn't catch it but it was something like owning not found ...

So while this pr won't pose a problem but first maybe we should work these things out to get a better understanding of the adding location workflow.Can someone else try to reproduce and help me out? Thanks :v:

cesswairimu:
Great findings here @Tlazypanda :+1: I will try and reproduce and let you know what I find. Thanks

Tlazypanda:
@cesswairimu @jywarren Adding more to the weird alert (last point from above comment) - this time took a screenshot from publiclab.org and not local
err

jywarren:
Thank you, i agree you've found some issues with the workflow and would love help in correcting them! Perhaps we can run this by @ebarry once you establish what you think the workflow /ought/ to be, just for a +1? 

and we can for sure do that all in another issue!

As for `owmloading` -- i've seen that before! I believe it was an assets issue related to an upstream Leaflet library last time. This time, take a look and see if we could have caused it another way? https://github.com/publiclab/plots2/issues/5095#issuecomment-513397183

Tlazypanda:
> Hey @jywarren @cesswairimu ! So while going through the function and testing it out here are some observations-
> These observations are from publiclab.org and not local
> 
> * This is invoked by adding location from dashboard but on clicking share location from profile page nothing happens.
> * On adding location modal, the slider works by zooming in but when we save it and check it out in the profile page it is saved without any zoom.
> * Also after adding your location it still shows add location on dashboard
> * Also after adding location for first time, it doesn't seem to update for a second time
> * Some weird alert comes sometimes couldn't catch it but it was something like owning not found ...
> 
> So while this pr won't pose a problem(since it is a performance improvement) but first maybe we should work these things out to get a better understanding of the adding location workflow.Can someone else try to reproduce and help me out? Thanks

Hey @jywarren @cesswairimu  @ebarry can you try reproducing these? Once I am confirmed that this is actually what's happening we can start establishing the workflow in a new issue :sweat_smile: And fix issues if any with the slider and all  

As for the mysterious owmloading thing I have left a comment at #5095 :sweat_smile: Hopefully will be able to set the ghost free :ghost:

jywarren:
Sorry! Ok;

* use from dashboard - clicking `Add your location` gets me the below, so is it on actually saving that it doesn't do anything for you? I confirm the `Save` button doesn't do anything (at least visible to the user!)

image

Just checked, and I can confirm your observations on all of this. I think we should try to break this list out into some bugfix issues, using `break-me-up`?

> * On adding location modal, the slider works by zooming in but when we save it and check it out in the profile page it is saved without any zoom.
> * Also after adding your location it still shows add location on dashboard
> * This is invoked by adding location from dashboard but on clicking share location from profile page nothing happens.
> * Also after adding location for first time, it doesn't seem to update for a second time

jywarren:
We also ought to add a system test for adding profile tags - like, one of these modified for the dashboard!

https://github.com/publiclab/plots2/blob/86776c2c72a505b0db563f52d1a8165c36c6eb37/test/system/post_test.rb#L188-L211

We could add one of these assertions afterwards that checks that the profile tags are properly created: 

https://github.com/publiclab/plots2/blob/86776c2c72a505b0db563f52d1a8165c36c6eb37/test/system/tag_test.rb#L40-L51

The ideal workflow is:

1. See "add a location" on dashboard, click it
2. See modal and check content of one or more of the inputs
3. enter lat/lon and zoom, or use the slider
4. press `Save` and see an alert (maybe?) that says "profile location was saved. Learn more about location privacy" - linking to https://publiclab.org/location-privacy
5. confirm dashboard button now says "View nearby content
6. confirm tags were added and appear on profile page

This is really too much for this PR, so let's create a new issue around it! 

The more basic usage would be when adding locations using the "Add a location" button in the tags input on pages like https://stable.publiclab.org/notes/anngneal/12-08-2017/7-ways-to-measure-monitor-and-evaluate-water-quality -- we can test this out there to finish this PR:

image

I'l move the parts about the dashboard button into its own issue!

Thanks @Tlazypanda !!

jywarren:
Great, moved dashboard stuff to https://github.com/publiclab/plots2/issues/8566! Thanks!

jywarren (approved):

Issue #2537, opened 2018-03-24
Layer icon missing on profile map .


sagarpreet-chadha:
### Please describe the problem (or idea)

<img width="312" alt="screen shot 2018-03-25 at 12 54 19 am" src="https://user-images.githubusercontent.com/14952645/37868077-13ed760c-2fc7-11e8-922c-83fce7221480.png">

The top- right corner has missing icon .
Should be like this : 

<img width="318" alt="screen shot 2018-03-25 at 12 55 12 am" src="https://user-images.githubusercontent.com/14952645/37868084-31f0dfcc-2fc7-11e8-9423-b90293b4d7ac.png">

### Please show us where to look

https://publiclab.org/profile/sagarpreet

### Solutions : 

1.) Overwrite the CSS .
2.) Add leaflet css cdn links 

https://cdnjs.cloudflare.com/ajax/libs/leaflet/1.3.1/images/layers.png
https://cdnjs.cloudflare.com/ajax/libs/leaflet/1.3.1/images/layers-2x.png

****

jywarren:
I'm interested in maybe doing this in an inline `<script>` tag so that it's easier to break out into its own library. OR -- if we plan to use Leaflet all over the site, which i guess we do already, finding a good place in a top-level stylesheet for this. Maybe `/app/assets/stylesheets/leaflet.css` and then including it into `application.css` ?

sagarpreet-chadha:
Okay 😄 , will submit the PR in a day or two . 
Thanks !

sagarpreet-chadha:
@jywarren ...I think this icon is not shown because we have installed **leaflet library** via **bower.json** and **required** in `application.js`  , hence the icon will look for url `/assets/leaflet/dist/images/marker-icon-2x.png`  which does not exists , see this :

<img width="448" alt="screen shot 2018-04-07 at 5 48 27 pm" src="https://user-images.githubusercontent.com/14952645/38454817-f25a978e-3a8b-11e8-9f2b-d588301b6606.png">

And we are able to see the Full-Screen icon because we are using **cdn link**  and do not have a local-copy as we have for **leaflet** .

So if we pre-compile the images by adding this to `environment/production.rb` : 
`config.assets.precompile += %w[*.png *.jpg *.jpeg *.gif] ` .

What do you think ?

sagarpreet-chadha:
This is solved 🎉  !
Closing this issue .

PR #7945, opened 2020-05-22
 replace leaflet files from assets.rb in questionRich.html and _mapDependencies.html

### LINKED ISSUE #7922: Replace leaflet files in app/views/editor/questionRich.html.erb
Currently the leaflet files in `app/views/editor/questionRich.html.erb` are referenced by `script` and `link `tag but since some of them are already present in the asset pipeline they can be replace with `javascript_include_tag` and ` stylesheet_include_tag` inorder to speed up time of execution.

**Additional Context**
Some of the leaflet files were flagged by lighthouse to reduce javascript execution time and minification of scripts. By serving them from the asset pipeline both these things will be taken care of.

Tlazypanda:
Fixes #7922
Replace leaflet files which are already present in asset pipeline by using javascript_include_tag and stylesheet_link_tag. Also removed files which are already served from application.js and application.css. Added some files to assets.rb for precompilation as well.

Edited _mapDependencies as well since some files there also needed to be replaced and in testing inorder to reduce js execution time.

## Screenshots 

Before   

7945_before

After   
 
7945_after

Thanks!


Tlazypanda:
@jywarren @cesswairimu Should we add the remaining files of /lib mentioned in this file for precompilation by augmenting the assets.rb array? Thanks :v:

Tlazypanda:
Hey @jywarren @cesswairimu @emilyashley Looking forward to your input on this :smile:

Tlazypanda:
@VladimirMikulic can you help me with the travis tests failure? Thanks :v:

VladimirMikulic:
Hi @Tlazypanda. The error in Travis tells us that the route `/assets/images/layers.png` doesn't exist.
`Leaflet` lib is using those layers. You've probably removed the wrong asset. `leaflet.js` maybe?

Tlazypanda:
Hmm..weird I removed those assets which would otherwise be available from the asset pipeline @VladimirMikulic `leaflet.js` is in `application.js` then shouldn't it be present in the asset pipeline already bundled in application.js :sweat_smile:

jywarren:
I believe layer.png often causes issues in the asset pipeline, actually! Maybe this helps? https://github.com/publiclab/plots2/issues/2537

But i recall this kind of thing with Leaflet image assets, esp. those referred to in CSS, because CSS doesn't respect asset pipeline routing (i think that's right!). So it could be worth a broader search for `leaflet layers.png asset pipeline` -- maybe on the Leaflet repo or StackOverflow?

Good luck, this looks great and thanks to both of you!

Tlazypanda:
@jywarren Found the issue! So basically whenever we are trying to move this into the asset pipeline it is searching in the route `assets/images` for the css images but in the link form it is searching in the `node_modules` folder where the css file is ...so a fix would be to move the layers.png file to `assets/images` but I am not sure if I should proceed with it ...what are your thoughts on this? Thanks :v:

Tlazypanda:
Hey @cesswairimu @jywarren any update on what should be my course of action  :sweat_smile: Thanks :v:

jywarren:
OK this could be linked to https://github.com/publiclab/plots2/issues/5095#issuecomment-650806234 -- is that right?

For the broader question, originally we included assets in the editor because we thought it optimal to only include them here, instead of on any page (the editor has an especially large number of assets, not needed elsewhere on the site). But I totally trust your judgement and analysis of whether re-organizing which assets are loaded where and how they're minified, so as long as nothing breaks 😅  please go for it! Thank you!!!

Tlazypanda:
Hey @jywarren @cesswairimu I am also fairly confused as to what can be done here :sweat_smile: The reason being if I am supposed to reference the images without adding them to assets/images route then I will have to add the path of the specific images dir to the `config.assets.path` but the issue with this is that I will have to reference the image css with `image_url(image.png)` but since this is an external library I can't change its code. Maybe we can remove the css file from the asset pipeline to prevent adding these static images for precompilation?

https://guides.rubyonrails.org/asset_pipeline.html#coding-links-to-assets
https://makandracards.com/makandra/29567-managing-vendor-libraries-with-the-rails-asset-pipeline

Tlazypanda:
@cesswairimu @jywarren can you kindly review? Thanks :v:

cesswairimu:
@Tlazypanda quick question...are the contents of the files in /assets identical to the ones in /lib?

Tlazypanda:
Hey @cesswairimu I am not sure I completely follow :sweat_smile: on precompilation all are put together in assets directory but since here we are referencing image from css we need to add relative path so since this is an external lib and we can't make changes to the code (since it is yarn package) so I removed the file that required the css leaflet.css (as you can see in _mapDependencies now it is referenced from /lib suggesting that is not part of asset pipeline)

cesswairimu:
aha! gotcha thanks

app/views/editor/questionRich.html.erb
```diff
@@ -5,21 +5,15 @@
 <link href="/lib/publiclab-editor/dist/PublicLab.Editor.css" rel="stylesheet">
 
 <!-- required for MapModule -->
-<link href="/lib/leaflet/dist/leaflet.css" rel="stylesheet">
-<script src="/lib/leaflet/dist/leaflet.js"></script>
```
sagarpreet-chadha, line 9: I think we should keep this and remove from `application` level, because as you are also working on performance improvement, having a light entry page should take less time, right?
When these libraries are required they should be loaded only then (lazy loading), what do you think?
😄 🎉

jywarren: Hi @Tlazypanda any comment on these last questions from @sagarpreet-chadha ? My apologies for replying late to these. Thank you!!!

Tlazypanda: Hey @jywarren @sagarpreet-chadha I totally missed this :sweat_smile: my apologies ...ummm can we do this in a follow up pr because I think the leaflet.js and css files are actually required on many routes since the map is loaded on dashboard/user profile/all editor pages/map routes etc so might have to investigate it more to make sure that removing it won't break anything. What do you think? Thanks :v:

jywarren: That's totally fine. Always game to split up PRs! Does that make this PR otherwise good to merge? Thank you!

jywarren:
Retriggering once more!

jywarren:
Great work, @Tlazypanda !!!

PR #7954, opened 2020-05-26
fix hyphenation in tags search

jywarren:
re #2335 fixes!


ebarry:
This is awesome!!!!
Thank You!!!

Issue #7737, opened 2020-03-28
Change function t to translation in app/views/user_sessions/_form.html.erb


nicoleiocana:
Hi, this is a [first-timers-only issue](https://code.publiclab.org/#r=all). This means we've worked to make it more legible to folks who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**. 

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

We know that the process of creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

If you have contributed before, **consider leaving this one for someone new**, and looking through our general [help wanted](https://github.com/publiclab/plots2/labels/help-wanted) issues. Thanks!

### 🤔 What you will need to know.

Nothing. This issue is meant to welcome you to Open Source :) We are happy to walk you through the process.

### 📋 Step by Step

- [ ] 🙋 **Claim this issue**: Comment below. If someone else has claimed it, ask if they've opened a pull request already and if they're stuck -- maybe you can help them solve a problem or move it along!

- [ ] 📝 **Update** the file [app/views/user_sessions/_form.html.erb](https://github.com/publiclab/plots2/blob/master/app/views/user_sessions/_form.html.erb) in the `plots2` repository (press the little pen Icon) and edit the line as shown below. 

[See this page](https://code.publiclab.org/#r=all) for some help in taking your first steps!

Below is a "diff" showing in red (and a `-`) which lines to remove, and in green (and a `+`) which lines to add:

```diff
- 6: <h2><%= raw t('user_sessions.new.log_in') %></h2>
+ 6: <h2><%= raw translation('user_sessions.new.log_in') %></h2>

- 9: <%= raw t('user_sessions.new.sign_up') %></a> <%= raw t('user_sessions.new.to_join') %>
+ 9: <%= raw translation('user_sessions.new.sign_up') %></a> <%= raw translation('user_sessions.new.to_join') %>

- 28: <label for="username"><%= t('user_sessions.new.username') %></label>
+ 28: <label for="username"><%= translation('user_sessions.new.username') %></label>

- 32: <label class="control-label" for="password"><%= t('user_sessions.new.password') %></label>
+ 32: <label class="control-label" for="password"><%= translation('user_sessions.new.password') %></label>

- 47: tabindex="3"><%= t('user_sessions.new.log_in') %></button>
[...]

PR #7892, opened 2020-05-08
Issue code #7737 Change function t to translation in app/views/user_sessions/_form.html.erb

Rohit-786:
Fixes #0000 (<=== Add issue number here)

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [x] PR is descriptively titled 📑 and links the original issue above 🔗
* [ ] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [ ] code is in uniquely-named feature branch and has no merge conflicts 📁
* [ ] screenshots/GIFs are attached 📎 in case of UI updation
* [ ] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software 

Thanks!

PR #7979, opened 2020-06-03
Update _form.html.erb

ioampatzidis:
Fix Issues of #7737


emilyashley:
Thanks for this! I want to note that it looks like there may already be a Pull Request open that covers this fix? https://github.com/publiclab/plots2/pull/7892 Let's make sure they aren't duplicate fixes before merging :)

PR #8021, opened 2020-06-11
contributors tables fixes

### LINKED ISSUE #7817: Table display issue on Tag Contributor pages

There are a couple small display issues on Tag Contributor pages* that if fixed would entirely upgrade the user experience of someone searching on Public Lab for people researching topics that are also of interest to you. 

*Tag Contributor pages have the URL structure: https://publiclab.org/contributors/foo  , where `foo` represents a tag. 

> What happened just before the problem occurred? 

I was browsing the `evidence` tag via https://publiclab.org/tag/evidence and wanted to see who contributed this type of content. I clicked on "23 contributors" -- see screenshot below of where this link is on the Tag Page: 
Screen Shot 2020-04-14 at 2 02 47 PM

The link brought me to https://publiclab.org/contributors/evidence , which is great. However, the layout on this page has a couple issues: 

1. Both columns in the two column table are too narrow for the content they contain and for their header labels. The column headers are labelled "People who've posted" and "Notes". "People who've posted" breaks onto two lines. The content in Notes is forced to break onto two lines even though it contains only 7 characters. 
2. There is a third column which only appears once the user scrolls down to the bottom row of the first two columns. Then this third column appears at right, labelled "People who are following". This column lacks content. 

See gif below: 
contributors-display

Tag Contributor pages such as https://publiclab.org/contributors/evidence

There are some closed issues related to this page which can be seen at #993 (create this type of page), #2444 and #6675 (counting the number of contributors correctly)

liz

jywarren:
Fixes #7817

I've made a few changes that get it to render decently at 3 sizes, shown below. I think this is good enough for now but we do have a stalled conversation about the layout of the tags page overall that may address where the "related tags" zone goes. But this looks good for the time being!

image

image

image

### DIFF
--- a/app/views/tag/_contributors.html.erb
+++ b/app/views/tag/_contributors.html.erb
@@ -34,33 +34,35 @@
 <% elsif @note_count.nil? || @note_count == 0 %>
   <p><%= raw translation('tag.contributors.no_contributors', :tag => params[:id]) %>:</p>
 <% else %>
-  <div class="col-md-6" style="display: inline-block;">
-    <table class="table">
-      <tr>
-        <th style="width:75%;"><%= translation('tag.contributors.people_who\'ve_posted') %></th>
-        <th><%= translation('tag.contributors.notes') %></th>
-      </tr>
-      <% if @users %>
-        <% @users.each do |user| %>
-          <tr class="users-row">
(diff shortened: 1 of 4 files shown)


ebarry:
This looks great! Let's do it.

PR #8034, opened 2020-06-14
spam2 stable issues, UI update and modals added

keshavsethi:
There were few bugs in the UI of the spam management dashboard. 
- Wrong ids in /comments and /revisions 
- created modal for comment and revisions 
- default order nodes wrt time 
- **UI is updated** as there were few issues related to banner in stable. Apart from that this new UI requires very less external css and it is based on bootstrap mostly. it has everything which was in old UI. This is fully responsive and better than the previous UI. I have attached some screenshots and gifs.
Screenshot from 2020-06-20 18-14-00
Screenshot from 2020-06-20 18-10-04
Screenshot from 2020-06-20 18-09-47

There are a few issues that I am facing currently: 
1. /spam2 and /spam2/revision are giving error in stable while wiki and comment are fine, I think it might be due to invalid ID or something. I have changed it but not sure if it is the only issue. 
I have added (&.) instead of (.) it will discard all values where node.author is nill. Please review this.
` <a href="/profile/<%= node.author&.name %>"><%= node.author&.name %> <%= node.author&.new_contributor%></a> `
Please review @jywarren @SidharthBansal @cesswairimu @pydevsg @ananyaarun @VladimirMikulic @Uzay-G @emilyashley 
Thanks!!

PR #8037, opened 2020-06-14
stable revision issue

keshavsethi:
/spam has some problems in stable for example https://stable.publiclab.org/spam/revisions?page=3 this will give error while page 4 is fine.
https://stable.publiclab.org/spam?page=232 this will give error while any other page even more than 232 will show empty table
I checked it on sentry it was due to
` ActionView::Template::Error undefined method name for nil: NilClass`

`<a href="/profile/<%= revision.author.name %>"><%= revision.author.name %></a> `

I have edited it a bit in /spam2 and /spam, please review @jywarren @pydevsg @VladimirMikulic @Uzay-G @cesswairimu @ananyaarun 
Thanks!!

### DIFF
--- a/app/views/admin/_revisions.html.erb
+++ b/app/views/admin/_revisions.html.erb
@@ -16,9 +16,7 @@
         <%= revision.body.truncate(40) %>
       </td>
       <td style="width:200px;">
-      <% unless revision.author.nil? %>
-        <a href="/profile/<%= revision.author.name %>"><%= revision.author.name %></a> 
-      <% end %>
+        <a href="/profile/<%= revision.author&.name %>"><%= revision.author&.name %></a> 
       </td>
       <td><div class="btn-toolbar" style="margin:0;">
         <div class="btn-group">
--- a/app/views/admin/_revisions.html.erb
+++ b/app/views/admin/_revisions.html.erb
@@ -16,7 +16,9 @@
         <%= revision.body.truncate(40) %>
       </td>
       <td style="width:200px;">
+      <% unless revision.author.nil? %>
         <a href="/profile/<%= revision.author.name %>"><%= revision.author.name %></a> 
+      <% end %>
       </td>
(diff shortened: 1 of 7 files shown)


jywarren:
Hi! This looks great. Should this merge before https://github.com/publiclab/plots2/pull/8034 ? Let's link these together in your planning issue and perhaps use the Spam milestone, so we can see everything in order, thank you!!!

app/views/admin/_revisions.html.erb
```diff
@@ -16,7 +16,9 @@
         <%= revision.body.truncate(40) %>
       </td>
       <td style="width:200px;">
+      <% unless revision.author.nil? %>
         <a href="/profile/<%= revision.author.name %>"><%= revision.author.name %></a>
```
jywarren, line 20: Here, I propose:
```suggestion
        <a href="/profile/<%= revision.author&.name %>"><%= revision.author&.name %></a> 
```
instead of using the `unless` -- because then it'll be simpler than the conditional logic, and will fail gracefully! What do you think for this and other instances here?

keshavsethi: Sure, It will work here.  So, should I change in all the places of /spam and /spam2 where conditional logic is used??

keshavsethi:
> Hi! This looks great. Should this merge before #8034? Let's link these together in your planning issue and perhaps use the Spam milestone, so we can see everything in order, thank you!!!

#8034 should be merged before this.

app/views/admin/_revisions.html.erb
keshavsethi: @jywarren I have made changes as suggested. Please review!
Thanks!!

keshavsethi:
I have made all  changes made here in https://github.com/publiclab/plots2/pull/8034
Please review https://github.com/publiclab/plots2/pull/8034
Thanks!!

Issue #8019, opened 2020-06-11
How do you "vet" external libraries when adding them to a project? Best practices.


emilyashley:
I've seen a few questions, issues, PR for review on vetting which external libraries to bring in? Not sure if this is the best place, but wanted to share what I generally look for when adding new dependencies while it is on my mind. I can check these out for you, esp if you post a list of all the ones you may be considering :) 

There's no need to write all of our software from scratch, but we also want to balance convenience with the control we'll have over the maintainability of our software when we add external components

Things I look for:

- **The Docs** does it have readable documentation? feel free to link to the docs of new dependencies in the pull request adding them, others can scope it out and see if they have questions and if they would easily know how to use this
- **Their Release Process** does the repository have clear release and patch communication? Does it seem brittle or upredictable? ie will the dependency change underneath us without warning? Do you see a lot of "rollbacks?" If so, make sure to use appropriate syntax to pin it to a specific version dependency management file to minimize that risk. 
- **Contributor Community Health** does the repo have a good # of contributors? are open issues getting comments and resolutions? There's no right number, just check it out for red flags (i.e. 2 contributors that haven't responded in over a year might not be a reliable choice)
- **The License** make sure there's clear permission to use the software in the way we intend. 

Bonus:
- **Do they seem performant?** are they bundled and loaded (or can they be bundled and loaded) in a way that doesn't significantly slow our system? If there's only one or two functions or features we need, see if they've provided a way to custom load just those. (this differs a lot across languages and libraries and may not be possible) 

A thing I've done on teams before is created a pull request of the bare minimum code to add the dependency to a project and briefly listed the features I found useful from the library and how I might be using them. That way our asynchronous team had a place for discussion and group "vetting" of a library - each with different concerns (devops, security, performance, complexity etc). Dunno if that works here, but sharing what I know :)

Tlazypanda:
@emilyashley Thanks a lot for the detailed explanation on what all to look out for! :heart:

emilyashley:
Was just mentioning with @jywarren that we also bet @alaxalves has thoughts on this. Please chime in and share if you can :)

emilyashley:
LOVE TO SEE THIS. 
This is a really great example from @Tlazypanda and @cesswairimu vetting a library! SHOUT OUT! 🎉🙌

[...]

PR #8043, opened 2020-06-16
lazy load images using lazyload-rails gem

### LINKED ISSUE #7919: Lazy load images using lazyload-rails gem
Currently the pages are loaded only after completely loading all the images. We can lazyload images to improve first paint time. For this we can use lazyload-rails [gem](https://github.com/jassa/lazyload-rails).

**Additional context (optional)**
According to Pagespeed insight, lazyloading images is a good suggestion.
image

Tlazypanda:
Fixes #7919 

Lazy loaded images by using lazyload-rails gem. Converted the img tags into image_tag helpers for the gem to function properly.

## GIFs
lazy1
lazy2
lazy3
lazy4
lazy5

Thanks!

### DIFF
--- a/app/views/dashboard/_node_default.html.erb
+++ b/app/views/dashboard/_node_default.html.erb
@@ -4,9 +4,9 @@
     <%= render partial: 'dashboard/node_moderate', locals: { node: node } %>
 
     <% if node.main_image %>
-        <a class="img" href="<%= node.path %>"><img src="<%= node.main_image.path(:default) %>" style="width:100%;" /></a>
+        <a class="img" href="<%= node.path %>"><%= image_tag(node.main_image.path(:default), style:'width:100%;') %></a>
     <% elsif node.scraped_image %>
-        <a class="img" href="<%= node.path %>"><img src="<%= node.scraped_image %>" style="width:100%;" /></a>
+        <a class="img" href="<%= node.path %>"><%= image_tag(node.scraped_image, style:'width:100%;') %></a>
     <% end %>
 
     <h4><a href="<%= node.path %>"><%= node.title %></a> <% if node.status == 3 %><span style="font-size: small;" class="badge badge-success">Draft</span><% end %></h4>
@@ -17,3 +17,9 @@
 
   </div>
 </div>
+
+<script>
+$(function(){
+  $("img").lazyload();
+});
+</script>
--- a/app/views/notes/_card.html.erb
+++ b/app/views/notes/_card.html.erb
@@ -1,8 +1,8 @@
 <div class="card">
   <% if node.main_image %>
-    <a class="card-img-top img" style="overflow: hidden; height:10em;"<% if @widget %>target="_blank"<% end %> href="<%= node.path %>"><img src="<%= node.main_image.path(:default) %>" style="width:100%;"/></a>
+    <a class="card-img-top img" style="overflow: hidden; height:10em;"<% if @widget %>target="_blank"<% end %> href="<%= node.path %>"><%= image_tag(node.main_image.path(:default), style:'width:100%;') %></a>
   <% elsif node.scraped_image %>
-    <a class="card-img-top img" style="overflow: hidden; height:10em;" href="<%= node.path %>"><img src="<%= node.scraped_image %>" style="width:100%;" /></a>
+    <a class="card-img-top img" style="overflow: hidden; height:10em;" href="<%= node.path %>"><%= image_tag(node.scraped_image, style:'width:100%;') %></a>
   <%  else %>
     <a class="imgg" style="height:10em; margin-bottom: 10px;padding: 1rem;">
     <i class="fa fa-picture-o note-not aria-hidden="true" style="color: #ccc; font-size: 6em;"></i>
@@ -66,3 +66,9 @@
     </div>
   </small>
 </div>
+
+<script>
+$(function(){
+  $("img").lazyload();
+});
+</script>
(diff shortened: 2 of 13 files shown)


vendor/assets/javascripts/jquery.lazyload.js
```diff
@@ -0,0 +1,254 @@
+/*!
```
jywarren, line 1: Ah! Is this something we could include via `yarn` instead, pegged to a specific version? Thank you!!!

Tlazypanda: Yeah done! I still have to add more routes and refine this so no review required right now, I will ping when done :sweat_smile:

Tlazypanda:
@jywarren just checking in to ask if for the placeholder should we have like an animated spinner my only concern is that the size of it should be small. Or we can leave it at the gray placeholder as well :sweat_smile: Thought? :v:

cesswairimu:
@Tlazypanda great job finding the gem. I have a few concerns with it though...its seems that its not frequently updated(last updated 14 months ago), it is crucial we look at this as it could hinder our future updates..also @emilyashley guidelines on vetting libraries https://github.com/publiclab/plots2/issues/8019 ...I am not discarding it just giving you smth to think about/ consider . Thanks

Tlazypanda:
Hey @cesswairimu I totally understand your concerns over this :sweat_smile: and I actually did check it out against the vetting points by @emilyashley so the thing is that although this gem is not updated as frequently, the current implementation is something that won't need updates in the sense it is sufficient in itself. It is also the only implementation for lazy-loading in rails applications. And after doing a lot of searches, I could only find this tool being mentioned in blogs to speed up rails apps. The documenation is sufficient, there aren't any rollbacks in the commits as such and the license is MIT.

cesswairimu:
cool :+1:

Tlazypanda:
Hey @cesswairimu @jywarren I have added some more routes now can you please review this? :sweat_smile:

jywarren:
Hey, i just want to say the discussion and priorities of maintainability and the vetting and everything that happened in this PR is AWESOME 🙌🙌🙌🙌 ❤️❤️❤️

emilyashley:
Reading y'all vet this just made me so happpppy! Good work.

app/views/dashboard/_node_default.html.erb
```diff
@@ -4,9 +4,9 @@
     <%= render partial: 'dashboard/node_moderate', locals: { node: node } %>
 
     <% if node.main_image %>
-        <a class="img" href="<%= node.path %>"><img src="<%= node.main_image.path(:default) %>" style="width:100%;" /></a>
+        <a class="img" href="<%= node.path %>"><%= image_tag(node.main_image.path(:default), style:'width:100%;') %></a>
```
jywarren, line 7: Hi @Tlazypanda - this seems to have disrupted the loading of some images which are "blob" type - see for example the admittedly odd ones in this view:

https://publiclab.org/embed/grid/grid:infragram-upload

See how these actually use a dataurl as the image? I know its weird but I see the `image_tag` method is replacing the filename with `blob` - can we include a workaround, perhaps, if the image path is a data url? 

This is a pretty weird workflow... i feel like maybe I'm missing something!

Also, I'm trying to fix the somewhat related assets on that template (for use in embedded content), here:
https://github.com/publiclab/plots2/commit/36c54a21a3fc1b94c8e8c0981b09e76cb1e9a1af

Tlazypanda: Hey @jywarren I was thinking of writing a conditional rendering for this but am having some difficulty in defining the condition I was thinking of something like:- 
`<% elsif current_user.photo.url.start_with?("data:image")  %>` but doesn't seem to work... My understanding of the models may not be at point here can you help me out :sweat_smile:

jywarren: Yeah, hmm. Are we mixing up blobs with data_urls? This is a bit mysterious! Uh, is it possible that the lazy load gem is converting a data-url into a blob? 

AH! I remember, i wrote some really obscure code here. Let's see...

OMG YES -- https://github.com/publiclab/plots2/pull/6931 and https://github.com/publiclab/plots2/pull/6930 -- deep mysterious code! But yes, it has the data-uri-to-blob method there. That should help us!

jywarren: Should we consider just turning off lazy loading for this one instance in order to recover the functionality? @TildaDares @Manasa2850 @17sushmita are any of you interested in this particular bug? Thank you all!

17sushmita: I was looking into this and was playing around with some modifications on the unstable but I'm not sure how to reproduce this on my local. Would love to get any help regarding this.

jywarren: Unfortunately I'm not sure how to reproduce... I'm not sure how the blob images are saved! Can we dig into the raw html shown on unstable to try to learn more?

jywarren: My best guess here is that we should just turn off lazy loading for this one type of image.

jywarren: OK this file is for the old dashboard! So ignore this particular file, we'll revert the one specifically for the new card-style display of nodes, below, i'll leave a comment.

app/views/notes/_card.html.erb
```diff
@@ -1,8 +1,8 @@
 <div class="card">
   <% if node.main_image %>
-    <a class="card-img-top img" style="overflow: hidden; height:10em;"<% if @widget %>target="_blank"<% end %> href="<%= node.path %>"><img src="<%= node.main_image.path(:default) %>" style="width:100%;"/></a>
```
jywarren, line 3: So let's revert this entire file, only. That should solve it!

PR #8048, opened 2020-06-18
Correct tag count

### LINKED ISSUE #7965: Remove "-" before count of posts on Topic Cards when sorting by "# of People"
### Please describe the problem

There appears to be a tiny "-" before the count of posts on Topic Cards when the "# of People" sorting option is clicked, making it appear like there is a negative count of posts. See screenshot: 

Screen Shot 2020-05-20 at 10 36 42 AM (1)

i expected to see positive integers. 

https://publiclab.org/tags?sort=followers&order=asc

liz

urvashigupta7:
Fixes #7965 
### Screenshot

Screenshot from 2020-06-18 11-03-32
It can be seen there are 4 research notes, 1question,5 wiki pages with tag **one** which sums to 10

Screenshot from 2020-06-18 11-04-59
The topic card for tag **one** is now showing the correct count.

### DIFF
--- a/app/views/tag/_topicCard.html.erb
+++ b/app/views/tag/_topicCard.html.erb
@@ -34,7 +34,7 @@
       </div>
 
       <div class="card-footer" style="background-color: inherit; border:none;">
-        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= Tag.counter(tag.name) - count %> <%= translation('tag.index.more_posts') %> &raquo;</a>
+        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= [Tag.counter(tag.name) - shown_nids.count - Tag.find_nodes_by_type(tag.name, type = 'note', limit = 3).where.not(nid: shown_nids).count, 0].max %> <%= translation('tag.index.more_posts') %> &raquo;</a>
         <div id="follow-unfollow-column-<%=tag.tid%>" style="float:right;margin:10px 0 10px 10px;">
           <% if current_user %>
             <% if !current_user.following(tag.name) %>
--- a/app/views/tag/_topicCard.html.erb
+++ b/app/views/tag/_topicCard.html.erb
@@ -12,8 +12,10 @@
       </div>
       <div class="card-body" style="padding:0.8em;">
         <div class="node-list">
+          <% count = 0 %> 
           <% Tag.find_nodes_by_type(tag.name, type = 'note', limit = 3).where.not(nid: shown_nids).each do |node|  %>
             <% shown_nids << node.id %>
+            <% count = count + 1 %>
             <div class="node-body">
               <% if node.main_image %>
                 <img class="rounded-circle pull-left" id="profile-photo" style="width:20px; height:20px; margin-right:8px; display: inline-block;" src="<%= node.main_image.path(:default) %>" />
@@ -32,7 +34,7 @@
       </div>
 
       <div class="card-footer" style="background-color: inherit; border:none;">
-        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= Tag.counter(tag.name)-shown_nids.count - Tag.find_nodes_by_type(tag.name, type = 'note', limit = 3).where.not(nid: shown_nids).count %> <%= translation('tag.index.more_posts') %> &raquo;</a>
+        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= Tag.counter(tag.name) - count %> <%= translation('tag.index.more_posts') %> &raquo;</a>
         <div id="follow-unfollow-column-<%=tag.tid%>" style="float:right;margin:10px 0 10px 10px;">
           <% if current_user %>
             <% if !current_user.following(tag.name) %>
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -116,10 +116,10 @@ class Tag < ApplicationRecord
   end
 
   def self.counter(tagname)
-    Node.where(type: %w(note page))
-        .where('term_data.name = ?', tagname)
-        .includes(:node_tag, :tag)
-        .references(:term_data)
+    Node.where(status: 1, type: %w(note page))
+        .includes(:revision, :tag)
+        .references(:term_data, :node_revisions)
+        .where('term_data.name = ? OR term_data.parent = ?', tagname, tagname)
         .count
   end
 


urvashigupta7:
@jywarren  @cesswairimu Please review

IshaGupta18:
This looks good to me @urvashigupta7

app/views/tag/_topicCard.html.erb
```diff
@@ -32,7 +34,7 @@
       </div>
 
       <div class="card-footer" style="background-color: inherit; border:none;">
-        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= Tag.counter(tag.name)-shown_nids.count - Tag.find_nodes_by_type(tag.name, type = 'note', limit = 3).where.not(nid: shown_nids).count %> <%= translation('tag.index.more_posts') %> &raquo;</a>
```
jywarren, line 35: Aha - so, i think here we actually do need to keep `-shown_nids.count` because that takes the total, but deducts the currently displayed posts because the link says `X more posts` -- does that make sense? I think your fix should still work because it resolves the spam filtering on line 119 of the controller. 

Can you re-add the offset for the already-displayed posts? Thank you!!!

ebarry:
Thank you for taking this up @urvashigupta7 ! 🌟🔢

app/views/tag/_topicCard.html.erb
urvashigupta7: Just a doubt @jywarren 😅, for example, we have 2 posts with tag **one** and 1 post with tag **one** and **two** both,  
and all the 3 posts gets displayed in the topic card for tag **one**. So, topic card of **two** should show **0 more posts since the post is displayed already in tag-one**. I guess this is the thing we want here, right ?

urvashigupta7:
@ebarry 😄

app/views/tag/_topicCard.html.erb
jywarren: Aha - yes. So we could maybe use this syntax -- 

```ruby
[Tag.counter(tag.name)-shown_nids.count - Tag.find_nodes_by_type(tag.name, type = 'note', limit = 3).where.not(nid: shown_nids).count, 0].max
```

?

app/views/tag/_topicCard.html.erb
```diff
@@ -32,7 +34,7 @@
...
+        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= Tag.counter(tag.name) - count %> <%= translation('tag.index.more_posts') %> &raquo;</a>
```
jywarren, line 37: OK, making suggestion here:

```suggestion
        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= [Tag.counter(tag.name) - shown_nids.count - Tag.find_nodes_by_type(tag.name, type = 'note', limit = 3).where.not(nid: shown_nids).count, 0].max %> <%= translation('tag.index.more_posts') %> &raquo;</a>
```

Is that right?

urvashigupta7: Screenshot from 2020-07-01 08-50-07

Screenshot from 2020-07-01 08-50-48

As you can see, there are 7 research notes, 1 question and 3 wiki pages which is equal to 11 and the topic card shows 2 more posts. Also, none of the posts with tag two is shown in topic card one. So, It should have shown 8 more posts.
P.S. This is after the changes you suggested.

urvashigupta7: What I feel is we should subtract **Tag.find_nodes_by_type(tag.name, type = 'note').where(nid: shown_nids).count** from the total count.  What it will do is count all the posts which are already shown in the topic cards above it.

ebarry:
Thanks @jywarren ! Thank you @urvashigupta7 , what do you think of Jeff's last comment ?

urvashigupta7:
@ebarry I tried the changes suggested by him. But, I think it is not producing the right results 🤔 Correct me if I am wrong 😅

app/views/tag/_topicCard.html.erb
jywarren: Apologies for the slow reply!

OK, let's see:

total should be 11, agreed!

so the current logic will do: `[Tag.counter(tag.name) - shown_nids.count - Tag.find_nodes_by_type(tag.name, type = 'note', limit = 3).where.not(nid: shown_nids).count, 0].max` which doesn't really make sense... i wonder. 

It used to be: `Tag.counter(tag.name)-shown_nids.count - Tag.find_nodes_by_type(tag.name, type = 'note', limit = 3).where.not(nid: shown_nids).count`

Let's try to write it in English:

(all nodes with that tag) - (count of nodes we've shown on the page anywhere) - (all nodes with that tag except for the shown nodes)

Yeah, I can't figure out why it's like that! 

You're right, it should just be:

```ruby
Tag.counter(tag.name) - Tag.find_nodes_by_type(tag.name, type = 'note').where(nid: shown_nids).count
```

Let's try that! I'm also going to rebase it so we get it to build in GitPod and we can test it out there too.

app/views/tag/_topicCard.html.erb
```diff
@@ -32,7 +34,7 @@
       </div>
 
       <div class="card-footer" style="background-color: inherit; border:none;">
-        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= Tag.counter(tag.name)-shown_nids.count - Tag.find_nodes_by_type(tag.name, type = 'note', limit = 3).where.not(nid: shown_nids).count %> <%= translation('tag.index.more_posts') %> &raquo;</a>
+        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= [Tag.counter(tag.name) - shown_nids.count - Tag.find_nodes_by_type(tag.name, type = 'note', limit = 3).where.not(nid: shown_nids).count, 0].max %> <%= translation('tag.index.more_posts') %> &raquo;</a>
```
jywarren, line 37:
```suggestion
        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= [Tag.counter(tag.name) - Tag.find_nodes_by_type(tag.name, type = 'note').where(nid: shown_nids).count, 0].max %> <%= translation('tag.index.more_posts') %> &raquo;</a>
```

jywarren:
I'm really sorry - due to errors in our GitPod setup (https://github.com/publiclab/plots2/issues/8177#issuecomment-662121295) we can't test this out in GitPod. If you're able to pull this down to see if it works, we could complete it that way! Or I can try myself in the next few days. Thank you for your patience!!!

urvashigupta7:
https://github.com/publiclab/plots2/blob/main/app/models/tag.rb#L95

This method suplies default value for limit, that means it can return maximum 10,but we don't want that, we want count of all the tags shown above it . Should I make another method which doesn't limit the tag count? Or Is there any other approach for doing this?

urvashigupta7:
We can make a new method which takes three arguments - tag name, type, shown_nids and it will directly return the count of tags above it.

jywarren:
I believe that due to the way `.limit()` works in Rails, we can pass `false` and it will not limit!

https://github.com/publiclab/plots2/blob/cd9bd3a2988569632f09c4dd0fad89aa30abbe20/app/models/tag.rb#L114

> We can make a new method which takes three arguments - tag name, type, shown_nids and it will directly return the count of tags above it.

i can see the logic here but i worry it's not a widely enough used function to make it worthwhile! However, you're right that we could then write a simple unit test for it and so better protect it. I'm going to try pulling down this code to manually test, and see how we do.

jywarren:
Also closing and reopening to try to retrigger the tests, which look to have gotten stuck!

jywarren:
Hi, @urvashigupta7 I've just had an insight that makes me feel kind of silly -- why do we really need to say `X more posts` anyways? Isn't it simpler to just show the count itself and just say `X posts`? That makes the math... nonexistent. 

I also finally found that we had both a `tag.count` attribute maintained by the `tag.run_count` method, AND a `Tag.counter(tag.name)` method, so I searched and found `Tag.counter(tag.name)` is only used here. I'm going to switch it to the simpler and more efficient `tag.count`, and make some other small adjustments. in https://github.com/publiclab/plots2/pull/8249/. 

Thank you so much for going through this journey together! It's been pretty convoluted and I want to be sure you feel good about it, and know how much we appreciate it. Is there another issue you'd like to try to take up that might end up being a little easier? This one was really not very typical, i have to say, but you've been great to work with. Thank you!!!

urvashigupta7:
Thank you soo much @jywarren. I will look into other issues for sure.

PR #8062, opened 2020-06-23
fix profile location selection

### LINKED ISSUE #8061: addLocation is not defined on profile page location adding modal
On profile pages, removing location tags and re-adding a location using the modal fails, with:

`addLocation() is not defined`

OK, so this used to exist:

https://github.com/publiclab/plots2/pull/1965/files#diff-0e96eb89432977c7610f8598db819f24R17-R22

```html
    <script>
      function addLocation() {
        $('.blurred-location-input').click() ;
      }
    </script>
```

Where'd it go? 

Looks like this commit consolidated some Leaflet code but lost this:

https://github.com/publiclab/plots2/commit/0545b0fa306b4326e5242ec7ae6ac737538b5e9a

let's just not use a separate function but apply the click directly?

jywarren:
fixes #8061

This also involved moving the modal to the top of the DOM so it works better, following guidance on the layout at https://getbootstrap.com/docs/4.4/components/modal/#modal-components

### DIFF
--- a/app/views/tag/_location.html.erb
+++ b/app/views/tag/_location.html.erb
@@ -2,28 +2,29 @@
 <div class="blurred-location-container" style="position:absolute;top:0;left:0;"></div>
 
 <script>
+  function createBlurredLocationInput() {
+    $(this).attr("disabled", "disabled");
+    let modalUrl = '/locations/modal';
+    let firstVar = true;
+    <% if (params[:action] == "list") || (params[:action] == "profile") %>
+      modalUrl += firstVar ? '?' : '&';
+      modalUrl += 'reload=true';
+      firstVar = false;
+    <% end %>
+    if( $(this).attr("submit_to") ) {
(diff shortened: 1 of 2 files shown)


jywarren:
Hmm. System tests failed with:

```
SimpleCov failed with exit 1
The command "bundle exec rails test:system" exited with 1.
```

jywarren:
Aha - 
```
Minitest::UnexpectedError:         Capybara::ElementNotFound: Unable to find css "#map_content #coord_button"
            test/system/post_test.rb:195:in `block in <class:PostTest>'
```

jywarren:
Hmm. not sure why:

https://github.com/publiclab/plots2/blob/a02ba052d305ab7ef71b61f7f5430dfbc056d50a/test/system/post_test.rb#L188-L195

jywarren:
OK, widened click event handler setup to get the wiki page "add a location" buttons too! Good system test!

PR #8099, opened 2020-06-30
display methods in grid upon filtering

jywarren:


jywarren:
Hmm. the older `continuous-integration/travis-ci/pr` one seems to have finished faster!

PR #8106, opened 2020-07-02
wip: Add alternate text for lead images

### LINKED ISSUE #7991: Alternate text for Uploaded images
Currently we dont have any alternate text for images across the app. While for the static ones we can add some alt text but for the ones uploaded by the user, its not possible for us to apply the same. So what should be the plausible solution for it?

- One possible way is to add a field for the user to add alternate text for the image whenever they upload but this won't always ensure that the user does add the alternate text or if they do whether it will be meaningful text relevant to the picture. We can have a special message asking them to add since it will help accessibility of the app and help more sections of society to access it with ease
- The other solution is to use some image recognition service like aws/microsoft services to determine the image text and accordingly add it as alt text. This might slow down the app and might also require us to tag all our previous data or add our data to their servers.

Tlazypanda:
Fixes #7991 

work in progress:add alternate text for image

Thanks!

### DIFF
--- a/app/views/editor/rich.html.erb
+++ b/app/views/editor/rich.html.erb
@@ -112,6 +112,9 @@
               >Select an optional main image for your post.</span
             >
           </p>
+          <p>
+            <input class="form-control input-lg"  type="text" id="alt-text" placeholder="Add alternate text for image" value="<% if @node && @node.main_image.photo_text %><%= @node.main_image.photo_text %><% else %><%= params[:photo_text] %><% end %>"/>
+          </p>
         </div>
       </div>
     </div>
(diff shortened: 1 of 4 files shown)


app/views/editor/rich.html.erb
```diff
@@ -111,6 +111,9 @@
           <p class="ple-help">
             <span class="ple-help-minor">Select an optional main image for your post.</span>
           </p>
+          <p>
+            <input class="form-control input-lg"  type="text" id="alt-text" placeholder="Add alternate text for image" value="<% if @node && @node.main_image.photo_text %><%= @node.main_image.photo_text %><% else %><%= params[:photo_text] %><% end %>"/>
```
cesswairimu, line 115: @Tlazypanda found that the issue is because some notes do not have images hence the nil  error..we can use a safe navigator to cater for the nodes without 'main_images' like so

`input class="form-control input-lg"  type="text" id="alt-text" placeholder="Add alternate text for image" value="<% if @node && @node&.main_image&.photo_text %><%= @node.main_image.photo_text %><% else %><%= params[:photo_text] %><% end %>"/>`

cesswairimu: Hi @Tlazypanda I will prioritize this coming week so we can collaborate and have it merged.

jywarren: wow this is v cool! can we see a screenshot at some point? Thanks!!

PR #8122, opened 2020-07-08
Try setting up GitPod.io

jywarren:
Based on https://github.com/gitpod-io/rails_sample_app/blob/master/.gitpod.yml

https://www.gitpod.io/ -- this can create an interactive version of the site for each PR! It can let us interact with the code changes proposed in each. 

quoting myself from the chatroom 😄 

> It's a cool feature for trying out new code, troubleshooting, and also could be great for new contributors. I mention you two because it seems it kind of mirrors the Docker config to get the database set up and I'm struggling a bit to set it up

Watch for the button below. It may take some tweaking to get it to boot properly, but this should make reviewing UI changes MUCH easier and also potentially make it easier to skip the setup steps if you're a newcomer!

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -3,6 +3,7 @@ ports:
     onOpen: open-preview
 tasks:
   - init: >
+      rvm use 2.4.4 --default
       bundle install --without production &&
       rails db:migrate
     command: >
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -3,7 +3,7 @@ ports:
     onOpen: open-preview
 tasks:
   - init: >
-      rvm get stable --auto-dotfiles && rvm use 2.4.4 --default
+      rvm install 2.4.4
       bundle install --without production &&
       rails db:migrate
     command: >
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,12 @@
+ports:
+  - port: 3000
+    onOpen: open-preview
+tasks:
+  - init: >
+      bundle install --without production &&
+      rails db:migrate
+    command: >
+      rails server
+github:
+    prebuilds:
+        pullRequestsFromForks: true
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -3,7 +3,7 @@ ports:
     onOpen: open-preview
 tasks:
   - init: >
-      rvm use 2.4.4 --default
+      rvm get stable --auto-dotfiles && rvm use 2.4.4 --default
       bundle install --without production &&
       rails db:migrate
     command: >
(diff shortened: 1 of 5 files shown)


.gitpod.yml
```diff
@@ -0,0 +1,13 @@
+ports:
+  - port: 3000
+    onOpen: open-preview
+tasks:
+  - init: >
+      rvm get stable --auto-dotfiles && rvm use 2.4.4 --default
```
jywarren, line 6: This isn't working, somehow.

jywarren: Worst case maybe this: https://community.gitpod.io/t/error-on-installing-older-ruby-version-with-rvm/1090/14

jywarren:
```
> }; exit
Warning! PATH is not properly set up, /workspace/.rvm/bin is not at first place.
         Usually this is caused by shell initialization files. Search for PATH=... entries.
         You can also re-add RVM to your profile by running: rvm get stable --auto-dotfiles
         To fix it temporarily in this shell session run: rvm use .rvm
         To ignore this error add rvm_silence_path_mismatch_check_flag=1 to your ~/.rvmrc file.
Unrecognized command line argument: bundle
Run `rvm help` to see usage information
```

hmm. running `rvm install 2.4.4` gets us:

```
Updating system...
Error running 'requirements_debian_update_system ruby-2.4.4',
please read /home/gitpod/.rvm/log/1594174105_ruby-2.4.4/update_system.log
Requirements installation failed with status: 1.
```

Could we upgrade to one of the preinstalled ones?

```
   ruby-2.5.8 [ x86_64 ]
 * ruby-2.6.6 [ x86_64 ]
```

jywarren:
aha!

```
Could not load database configuration. No such file - ["config/database.yml"]
```

.gitpod.yml
```diff
@@ -0,0 +1,14 @@
+ports:
+  - port: 3000
+    onOpen: open-preview
+tasks:
+  - init: >
+      bundle install --without production &&
+      cp config/database.yml.example config/database.yml &&
+      cp db/schema.rb.example db/schema.rb &&
```
jywarren, line 8: May also need:

https://github.com/publiclab/plots2/blob/f34e477df8fc260016ba8dc6df67dafe177a5bd8/.travis.yml#L40-L50

jywarren:
OK...

```
/workspace/.rvm/gems/mysql2-0.5.3/lib/mysql2/client.rb:90:in `connect': Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2) (Mysql2::Error::ConnectionError)
```

jywarren:
OK, so we need to set up the database:

https://www.gitpod.io/docs/self-hosted/latest/install/database/

jywarren:
Moving to https://github.com/publiclab/plots2/pull/8145

PR #8149, opened 2020-07-15
[WIP]: add print preview

### LINKED ISSUE #8128: Print specific template for public lab content
Currently the publiclab website uses a set of css hiding strategies to only print important content from a note or wiki. We need to replace this with print specific template for public lab content. 

This print specific template can make use of js manipulation to only pull important part of content from the note and add css styling accordingly.

Tlazypanda:
Fixes #8128 
DO NOT MERGE WIP
Add print preview for wikis and notes 

Thanks!


jywarren:
Hi @Tlazypanda I am reading through this and i guess I'm just wondering if it really might be more maintainable to work around a simpler .html.erb template, using CSS specific to that page: https://github.com/publiclab/plots2/blob/d855ffb1ae36341b2acc07c450ba3babe16f1968/app/views/notes/show.html.erb#L63-L65

I'm just worried about how complex the JS functions are here and how easy they are to read, maintain, etc. -- and if they really do show a major advantage over an HTML and ERB-based solution. What do you think?

Tlazypanda:
Hey @jywarren Sure, I will look into this and figure out how we can do this :tada: Might take me sometime to revert :sweat_smile: 

I guess we are through with most of the styling elements for this in discussion with @ebarry so this should take fairly less time than before :sweat_smile:

Tlazypanda:
Hey @jywarren @cesswairimu  I was looking into finding how to have print specific layouts for rails and I couldn't find any solutions where they were creating specific html templates without the use of external gems/libraries to add those methods into the controllers (prince, wkhtmltopdf etc). Most solutions here focused on print specific stylesheets.

 For changing the print preview most solutions relied on js dom manipulations (https://neroscript.io/pretty-printing-site-javascript/). Also with this one js file since we are selecting components from the page we can have 2 separate functions for wiki/notes printing without creating 2 html files if we depended on that. 

Can you please share any resources or any idea on how we can do this here on html templates? My question here primarily lies that if we are not redirecting to the html template page then without that when we choose the window.print method then how can we get it to fetch the print html template and print it instead in real-time without loading because in our example we are fetching already from the loaded page the particular elements we want :sweat_smile: 

I am sorry if this is unclear but I have tried looking for alternatives but js manipulation/ css styling I what I am able to find. I am sure there are better ways to do this but I just haven't been able to find them. Please guide me :sweat_smile:

cesswairimu:
Hey @Tlazypanda I found this https://stackoverflow.com/questions/468881/print-div-id-printarea-div-only/7532581#7532581 for using css alone and this with js https://stackoverflow.com/questions/2255291/print-the-contents-of-a-div. You could check and see if any of these work for our case.

Tlazypanda:
Hey @cesswairimu! Thanks so much for researching and sending these links :tada: I have also setup this pr using only js manipulations and adding css for those manipulations in a js script. (unless I misunderstood what you meant and am deeply sorry :sweat_smile:) I guess the problem faced right now is that the js script doesn't seem that maintainable but I am not sure what other solutions we have other than js ..maybe we can write in a more modular fashion and break it down into as many functions as possible but that's the best I am getting at 

As suggested by @jywarren I tried looking into separate templates for html for the print preview but so far haven't been able to find any that work without using any gems :cry:

cesswairimu:
Huh :joy: I see, I totally misunderstood what you were asking..no worries...pinging @sagarpreet-chadha  who is a js guru and might have a few pointers on how to refactor what you already have. Thanks

jywarren:
Hi @Tlazypanda my apologies for replying slowly... i think it's OK for the print button to open a new page with the same content on a more minimal .html.erb template. Previously our print function and CSS spent a lot of time hiding elements like navbar, footer, sidebar, which were not necessary for printing. Having a stripped-down separate .html.erb template without those elements should make styling the page much simpler. So perhaps it would be in `notes_controller.rb` as `def print`, or in the `wiki_controller.rb` similarly, and we would keep a simplified print template which just shows title, lead image, author, same as the normal display, but without sidebar, navbar, comments, or footer. As you said, we can simply use a lot of the styling you've developed here on that page. How does that sound? So the print button could open `/notes/print/<nid>` or `/wiki/print/<nid>`, i would imagine?

jywarren:
This also reduces the chances that changes to the node or wiki pages would "break" the print page, or vice versa, something that's definitely happened in the past.

Tlazypanda:
@jywarren Ohh sure definitely if it's on anew page then that shouldn't be a problem :smile: I wasn't able to figure out how we can do it from the same print preview screen without it getting time to load :sweat_smile: This is much clearer!! Thank you :heart:

PR #8161, opened 2020-07-17
externalise jquery using jquery-rails-cdn and jquery-rails fallback

### LINKED ISSUE #7920: Externalize jquery with jquery-rails gem and cdn option
Currently the jquery is fetched from public/lib. We can add it as a jquery-rails gem and its cdn gem.
Cdn option will allow us to load jquery faster since it might be already present in the cache. If the cdn goes down then the jquery-rails gem will load jquery for us as a fallback. The link of the gems can be found [here](https://github.com/styx/jquery-ui-rails-cdn),[here](https://github.com/kenn/jquery-rails-cdn) and [here](https://github.com/rails/jquery-rails).

Most importantly it will allow us to run the application.js file with defer/asynvc attribute. Earlier it would not have been possible since jquery was present in application.js so if we tried to run it with async tab then inline scripts in html files requiring jquery would have through an error.

Tlazypanda:
Fixes #7920 

Externalise jquery using jquery-rails-cdn and jquery-rails fallback to improve performance in terms of - 
Automatically fallback to jquery-rails' bundled jQuery when:

On top of that, if you're using asset pipeline, you may have noticed that the major chunks of the code in combined application.js is jQuery. Implications of externalizing jQuery from application.js are:

- Updating your JS code won't evict the entire cache in browsers.
- Cached jQuery in the client browsers will survive deployments.
- rake assets:precompile will run faster and use less memory.

Thanks!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -70,7 +70,8 @@ gem 'lemmatizer', '~> 0.2.2'
 gem 'mailman', require: false
 # To implement fontawesome v4.7.0
 gem "font-awesome-rails"
- gem "lazyload-rails"
+gem "lazyload-rails"
+gem 'jquery-rails-cdn'
 
 # To convert html to markdown
 gem 'reverse_markdown'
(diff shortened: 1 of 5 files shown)


Tlazypanda:
@jywarren @cesswairimu can you kindly review? Thanks :v: Should I first push on unstable? :sweat_smile:

PR #8152, opened 2020-07-16
fixing GitPod mysql db, and try opening rails console in a gitpod terminal

jywarren:
Follow-up to #8145 

Still seeing:

> Wow, that leads us all the way back to: https://github.com/publiclab/plots2/issues/3120 - it relates to myqsl version number, and this apparently worked on version 5.7. https://github.com/publiclab/plots2/pull/4266#issuecomment-446462483

> The version GitPod has is: `mysql  Ver 8.0.20-0ubuntu0.20.04.1 for Linux on x86_64 ((Ubuntu))`

> Alternatively we could insist on `mariadb` which is what we use in production. 

> Could we specify this via dockerfile? https://www.gitpod.io/blog/docker-in-gitpod/

> i wonder -- could we use our existing Dockerfile? https://github.com/publiclab/plots2/blob/main/Dockerfile#L20 -- i guess?

PR #8162, opened 2020-07-17
Try reducing Gitpod errors by adding redis

keshavsethi:
I have changed the **development configuration** in _config/database.yml.gitpod_ Now it is working fine.
I have also **added the Redis server** for better UI testing.
Screenshot from 2020-07-18 04-25-34
Screenshot from 2020-07-18 04-26-31

@jywarren @cesswairimu @icarito Please review. Thanks!!


.gitpod.dockerfile
```diff
@@ -1 +1,7 @@
-FROM gitpod/workspace-mysql
\ No newline at end of file
+FROM gitpod/workspace-full
+
+# Install Redis.
+RUN sudo apt-get update \
```
alaxalves, line 4: We usually don't need `sudo` for the Dockerfile cmds :thinking:

.gitpod.yml
```diff
@@ -7,15 +7,14 @@ tasks:
...
-        mysql -e "CREATE DATABASE plots;" &&
         cp db/schema.rb.example db/schema.rb &&
         rake db:setup &&
         yarn install
 
       command: >
         passenger start
-
+        
+    - command: redis-server
```
alaxalves, line 17: Maybe instead of creating a new `command` task you can just include the `redis-server` below the `passenger-start` since the `>` supports multiline bash.

config/database.yml.gitpod
```diff
@@ -1,10 +1,6 @@
 development:
-  adapter: mysql2
-  username: root
-  password: 
-  encoding: utf8
-  database: plots
-  strict: false
+  adapter: sqlite3
```
alaxalves, line 2: Changing database might be dangerous :thinking:  I'll  wait for the other guys review

cesswairimu: @alaxalves since this is only specific to gitpod and we just started using it...what issues do you anticipate we will encounter? I know we have folks who use sqlite3 for development and the only drawback is that some tests fail locally. But you are more equipped  in the devops stuff and I would love to hear your thoughts thanks :smiley:

.gitpod.dockerfile
keshavsethi: Actually I don't know this, I found this in gitpod docs.

keshavsethi:
@jywarren can you please review it. Thanks!!

config/database.yml.gitpod
jywarren: Basically here, we started with an sqlite3 db, but our app is complex enough that i got locked database errors pretty often upon initial startup. So I thought perhaps we ought to try to reproduce our entire production environment in the dev environment. This may also be important because we use some MySQL features such as full-text search!

jywarren:
OMG love the Redis addition!!! Thanks everyone this is exciting!

jywarren:
I think we're making progress in #8152, so if you could remove the sqlite3 line here, we can just use this to get Redis running! Thank you @keshavsethi !!

keshavsethi:
> can

@jywarren Done, Thanks!!

SidharthBansal:
Redis is a NOSQL database and sqlite3 is a Relationa DB. So, just be sure we want this transition? Obviously redis is faster and has space for heterogeneity but please rethink @jywarren

jywarren:
Sorry @keshavsethi for late reply, would you mind resolving the conflicts? Thank you!!!

jywarren:
Thank you @keshavsethi! This will help a ton in #8297 for the Editor team!

PR #8196, opened 2020-07-28
Spam2: Moderator's Queue  feature

keshavsethi:
Here, moderators and admin can moderate only those Nodes which has a tag that moderator/admin is following. As in the following screenshot, we can see that "one" tag is being followed by admin. so admin can moderate only those nodes which has "one as  "tag". I have mentioned more details in my [proposal](https://publiclab.org/notes/keshav_sethi0004/03-12-2020/gsoc-2020-proposal-spam-management-dashboard#4.+The+Queue:). 
Screenshot from 2020-08-05 05-38-54
Please try this out on gitpod
@jywarren @cesswairimu @pydevsg @VladimirMikulic @Uzay-G @ananyaarun @ebarry @emilyashley Please review!!
Thanks!


app/views/spam2/_spam.html.erb
```diff
@@ -36,31 +38,37 @@
...
+        <a class="dropdown-item <% if params[:type] == "queue" %> active <% end %>" id="queue_drop" href="/spam2/filter/queue/30" ><i class="drop fa fa-clock text-primary"></i> Queue</a>
         <a class="dropdown-item <% if params[:action] == "_spam_revisions" %> active <% end %>" href="/spam2/revisions"><i class="drop fa fa-list text-primary"></i> Revision</a>
         <a class="dropdown-item <% if params[:action] == "_spam_comments" %> active <% end %>" href="/spam2/comments/filter/all/30" ><i class="drop fa fa-comment text-primary"></i> Comments</a>
         <a class="dropdown-item <% if params[:action] == "_spam_flags" %> active <% end %>"  href="/spam2/flags/filter/all/30" ><i class="drop fa fa-flag text-primary"></i> Flag </a>
         <a class="dropdown-item  <% if params[:action] == "_spam_users" %> active <% end %>" href="/spam2/users/filter/all/30"><i class="drop fa fa-user text-primary"></i> User moderation</a>
       </div>
     </div>
 </div>
-
+<script>
```
VladimirMikulic, line 51: Could you please indent properly this whole script? Thanks,

ebarry:
This is pretty amazing @keshavsethi , thank you for creating this! Very excited to see this functionality come to life. Is there a view where an admin like myself can see which tags moderators are subscribed to? The use case i'm imagining is "Oh, air-quality only has one moderator, while water-quality has 15 -- does anyone want to help moderate over in air-quality?"

app/models/tag.rb
```diff
@@ -334,7 +334,7 @@ def followers_who_dont_follow_tags(tags)
 
   # https://github.com/publiclab/plots2/pull/4266
   def self.trending(limit = 5, start_date = DateTime.now - 1.month, end_date = DateTime.now)
-    Tag.select('term_data.name, plots.term_data.count') # ONLY_FULL_GROUP_BY, issue #8152 & #3120
```
jywarren, line 337: good catch! Thanks!

app/models/user.rb
```diff
@@ -277,6 +277,20 @@ def unmoderated_in_period(start_time, end_time)
         .distinct
   end
 
+  def moderator_queue
```
jywarren, line 280: Does this make more sense in a controller, maybe? Or, could you explain your reasoning in a comment in the code? Thanks!

jywarren:
Would this affect everyone? Perhaps we should think about how this might be the case for moderators but not admins? @ebarry what do you think about this as a policy? 

Also, if we merge this, we should add an explanatory text just after `Queue` perhaps, something like `Spam for <a href="/subscriptions">topics you follow</a>` -- you know? So people don't get confused about the scope. 

Thanks, @keshavsethi this is very exciting!

jywarren:
> Is there a view where an admin like myself can see which tags moderators are subscribed to?

This is a good idea; for now we have the opposite - you can see who is subscribed to a given topic, but not which topics a given person is subscribed to. 

I want to suggest we show some kind of "moderator" icon (dunno what! but with an explanatory tooltip, maybe it could be a "certificate" icon? https://fontawesome.com/v4.7.0/icon/certificate) next to users in the subscribers listing, such as on https://publiclab.org/contributors/water-quality

image
 
This would maybe be easier to fulfill @ebarry's use case:

> The use case i'm imagining is "Oh, air-quality only has one moderator, while water-quality has 15 -- does anyone want to help moderate over in air-quality?"

because the page already exists. Could we open an issue for this to discuss and finalize, @keshavsethi ? It could even be a first-timers-only issue as the badge would be pretty easy to implement!

app/views/spam2/_spam.html.erb
keshavsethi: Done, Thanks!!

app/models/user.rb
keshavsethi: @jywarren, I have added this in the controller. Thanks!!

keshavsethi:
> > Is there a view where an admin like myself can see which tags moderators are subscribed to?
> 
> This is a good idea; for now we have the opposite - you can see who is subscribed to a given topic, but not which topics a given person is subscribed to.
> 
> I want to suggest we show some kind of "moderator" icon (dunno what! but with an explanatory tooltip, maybe it could be a "certificate" icon? https://fontawesome.com/v4.7.0/icon/certificate) next to users in the subscribers listing, such as on https://publiclab.org/contributors/water-quality
> 
> image
> 
> This would maybe be easier to fulfill @ebarry's use case:
> 
> > The use case i'm imagining is "Oh, air-quality only has one moderator, while water-quality has 15 -- does anyone want to help moderate over in air-quality?"
> 
> because the page already exists. Could we open an issue for this to discuss and finalize, @keshavsethi ? It could even be a first-timers-only issue as the badge would be pretty easy to implement!

@jywarren @ebarry, I have added count of moderators and admins who are following that tag as shown in the screenshot. Here  "one" is being followed by 1 moderator. I will add a section where a moderator can follow or unfollow a tag with a count of moderators already following them.
Screenshot from 2020-08-04 09-29-53

keshavsethi:
Here I have created a separate controller(batch controller) for  Bulk moderation as code climate is giving error for controller more than 250 lines. Please review it from gitpod as well.
Thanks!!

keshavsethi:
@pydevsg @ananyaarun @ebarry @jywarren @cesswairimu @VladimirMikulic @Uzay-G Please review!!
Thanks!!

jywarren:
Hi @keshavsethi were you able to add explanatory text about the narrower tag-filtered display after the word Queue? 

Also, 

> I have added count of moderators and admins who are following that tag as shown in the screenshot. Here "one" is being followed by 1 moderator.

This is great. Is there a tooltip perhaps that could explain what the number means, as it could seem mysterious to some folks? Thank you! Love the controller split too, great work. 

> I want to suggest we show some kind of "moderator" icon

Let's create a new issue for this idea of a moderator icon and split out that task. Sound good, @keshavsethi ?

Finally, one more question - i'm thinking of how people could circumvent this filtering and just moderate everything. I can think of 3 ways:

1. moderators are filtered by tag, but not admins, who see everything
2. we have a separate route like `/spam2/all` (or something?) and have explanatory text that says `Here are moderation tools for topics you follow. Click here to see content for all topics.` which links there, so you can still access a full listing
3. we explain how you can subscribe to everything (see below) so that if people /want/ to, they can simply moderate all content. This would require us to respect the "everything" subscription when filtering for this page, in the controller (make sense?)

See the "all research" subscription below, which some admins use:

image

keshavsethi:
@jywarren Thanks for the review, I have made some changes here. now moderator can filter out nodes as per tag. Suppose the moderator wants to filter out nodes only based on tag "XYZ" then all the spammed, unmoderated, and flagged nodes will get rendered. This Will remove the problem of everything tag as admin/moderator can just filter nodes based on the tag.
Screenshot from 2020-08-05 04-22-40
Screenshot from 2020-08-05 04-21-59

@jywarren @ebarry Please review
Thanks!!

jywarren:
Oh this is great! Can we name them "all tags" and "selected tags" for clarity? And what is the default when you first open it?

keshavsethi:
@jywarren 
> Is there a tooltip perhaps that could explain what the number means, as it could seem mysterious to some folks?

Yes, I have added few tooltips with tags and everything filter which will help moderators.

>  Let's create a new issue for this idea of a moderator icon and split out that task

I am bit confused here, should I add moderator icon in the tag. If so, then how a user will be able to tag node with icon??
The basic idea of the Queue is that the moderator can divide some work like some moderator can moderate only nodes that are related to "random tech". In that case, moderator can just filter that tag and moderate those nodes. 

> moderators are filtered by tag, but not admins, who see everything

This will be solved by tag filter

Thanks!!

keshavsethi:
> Oh this is great! Can we name them "all tags" and "selected tags" for clarity? And what is the default when you first open it?

Sure, I will do that.
Thanks!!

keshavsethi:
@jywarren @cesswairimu can we merge this if it is fine?
Thanks!!

cesswairimu:
Yeah this  good to merge.. Tremendous  work @keshavsethi :rocket: 
Thanks @VladimirMikulic @pydevsg and @jywarren for the reviews :tada:

Issue #7334, opened 2020-01-23
Wrong wiki count in Dropdown by type


keshavsethi:
In https://publiclab.org/contributors/newsletter dropdown by type Wiki pages is showing 3 but on clicking it shows only 1
Image is shown below
![Screenshot from 2020-01-23 17-47-30](https://user-images.githubusercontent.com/36025262/72984149-0653bb80-3e09-11ea-91e5-5c905567f5a1.png)

keshavsethi:
If we click on by Views button then it shows all 3 but again if try to sort all 3 of them by title it again shows 1 newsletter

gabrielbaldao:
Can I try fix this?

keshavsethi:
yes, you can try this but before doing this get it reviewed by @SidharthBansal @nstjean @jywarren

SidharthBansal:
Sure. Please provide a fux

NitinBhasneria:
@gabrielbaldao are you doing this or can I proceed?

SidharthBansal:
Go ahead

jywarren:
Note we should make this match the code and tests here: https://github.com/publiclab/plots2/pull/7476

and the idea for a "single source of truth" here: https://github.com/publiclab/plots2/issues/6855#issuecomment-578331475

Thank you!

jywarren:
And noting a more detailed investigation here -- I think we need to be sure all counts are from a unified body of code maintained in the models, and unit tested!

https://github.com/publiclab/plots2/issues/7965#issuecomment-644980326

jywarren:
So we should confirm that all counts are coming from the respective `Tag.count` style methods. Key areas are here:

https://github.com/ckchu8/plots2/blob/0fbcc07bd61fd238d88779e49a3fbd2e300ef041/app/controllers/tag_controller.rb#L542-L561

And on tag cards, for example on https://publiclab.org/tags

jywarren:
Just noting that the original issue at top may be a special case. I see in the DB that we're fetching these three pages:

* https://publiclab.org/n/3928 https://publiclab.org/wiki/newsletter (`Public Lab Community Newsletter`)
* https://publiclab.org/n/4965 https://publiclab.org/newsletter (`Newsletter`)
* https://publiclab.org/wiki/subscribe

Only the middle one is the correct one. The first may be some kind of strange clone... not sure. From the early days of the site? And the last, doesn't look like a page anymore?

Here are the full records:

```
[...]

PR #8243, opened 2020-08-04
paginate tag page listings

jywarren:
Re: #7334, though not a complete solution

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -118,6 +118,7 @@ class TagController < ApplicationController
     end
 
     nodes = nodes.paginate(page: params[:page], per_page: 24).order(order_by)
+    @paginated = true
 
     if @start && @end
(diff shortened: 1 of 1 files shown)


jywarren:
Let's test at https://stable.publiclab.org/wiki/tag/method once it builds!

jywarren:
Huh! No luck!

image

app/controllers/tag_controller.rb
```diff
@@ -118,6 +118,7 @@ def show
     end
 
     nodes = nodes.paginate(page: params[:page], per_page: 24).order(order_by)
+    @paginated = true
```
jywarren, line 121: This did not work!

jywarren:
Whaaaaaat? This page /does/ have pagination:

image

https://stable.publiclab.org/wiki/tag/method?page=2

app/controllers/tag_controller.rb
jywarren: Wait, see below:

PR #8245, opened 2020-08-04
Fixes to tag counting logic

### LINKED ISSUE #8244: Total Tag count on tag stats pages seems incorrect for tag "beach"
Reporting in here from @ebarry and Jeanette, the graph at https://publiclab.org/tag/beach/stats seems correct, but below it the "Total Tag count" of 8 seems wrong, by their count.

Screen Shot 2020-07-27 at 10 56 45 AM

That code is driven from here:

https://github.com/publiclab/plots2/blob/c60aaabb56c018962c76abf3f5f2d114f8492d15/app/views/tag/_graph.html.erb#L4-L8

Strange -- `@tags.first.count` would seem to be running `Tag.where(name: params[:id]).first.count`, which doesn't make sense to me. 

Looking deeper, cc @cesswairimu as well...

jywarren:
Fixes #8244 

If this fails properly, we should then add this to line 40 of tag.rb:

```ruby
    self.count = NodeTag.joins(:node).where(tid: tid).where('node.status = 1').count
```

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -37,7 +37,7 @@ class Tag < ApplicationRecord
   end
 
   def run_count
-    self.count = NodeTag.where(tid: tid).count
+    self.count = NodeTag.joins(:node).where(tid: tid).where('node.status = 1').count
     save
   end
 
--- a/test/unit/tag_test.rb
+++ b/test/unit/tag_test.rb
@@ -24,7 +24,7 @@ class TagTest < ActiveSupport::TestCase
     assert_nil tag.count
     assert_not_nil tag.run_count
     assert_not_nil tag.count
-    assert_equal 1, tag.count
+    assert_equal 3, tag.count
 
     tag = tags(:awesome)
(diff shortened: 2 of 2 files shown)


jywarren:
Great -- 

```
 FAIL["test_tag_counting", #<Minitest::Reporters::Suite:0x00000000093e2cc8 @name="TagTest">, 3.483404433000004]
 test_tag_counting#TagTest (3.48s)
        Expected 3 to be nil.
        test/unit/tag_test.rb:30:in `block in <class:TagTest>'
```

jywarren:
OK, good! 

```
 FAIL["test_tag_counting", #<Minitest::Reporters::Suite:0x000000000af08fe8 @name="TagTest">, 13.415067058000005]
 test_tag_counting#TagTest (13.42s)
        Expected: 0
          Actual: 1
        test/unit/tag_test.rb:33:in `block in <class:TagTest>'
```

Now the fix...

PR #8275, opened 2020-08-08
bcc on moderator spam digest email

jywarren:
Hi @keshavsethi - noted that the spam digest email showed all moderator emails, which is not that big a deal as they are all approved and managed by PL staff, but I felt it was probably better to put them in bcc, unless @ebarry feels differently!

Screen Shot 2020-08-08 at 3 36 13 PM

But, the email is definitely working! Great job!

### DIFF
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -129,6 +129,10 @@ class AdminMailer < ActionMailer::Base
     end
     moderators = User.where(role: %w(moderator admin)).collect(&:email)
     @nodes = nodes
-    mail(to: moderators, subject: @subject)
+    mail(
+      to: "moderators@#{ActionMailer::Base.default_url_options[:host]}",
+      bcc: moderators,
+      subject: @subject
+    )
   end
 end


app/mailers/admin_mailer.rb
```diff
@@ -129,6 +129,10 @@ def send_digest_spam(nodes, frequency_digest)
     end
     moderators = User.where(role: %w(moderator admin)).collect(&:email)
     @nodes = nodes
-    mail(to: moderators, subject: @subject)
+    mail(
+      to: "moderators@#{ActionMailer::Base.default_url_options[:host]}",
+      bcc: moderators,
```
jywarren, line 134: small change!

keshavsethi: @ebarry @jywarren Sorry for the late reply. This is really awesome. Is there any other issue with the digest?? and  I will surely work on its UI in few days 
Thanks @jywarren @ebarry :tada: :heart:

jywarren: Awesome. It seems to be working otherwise!

PR #8450, opened 2020-10-02
Translation helper follow-up to #7881

jywarren:
Replacing/rebasing https://github.com/publiclab/plots2/pull/7881


jywarren:
Oops, too many changes here... trying again with a simple find/replace in https://github.com/publiclab/plots2/pull/8451

PR #8451, opened 2020-10-02
Translation helper follow-up to #7881 #8450

jywarren:
Replacing #7881 #8450


jywarren:
Let's test this in stable, though...

PR #8474, opened 2020-10-07
Fixes odd looking css in image module

Sagarpreet:
Fixes part of https://github.com/publiclab/PublicLab.Editor/issues/596

BEFORE: 
Screenshot 2020-10-07 at 1 49 54 PM

AFTER:
Screenshot 2020-10-07 at 1 52 50 PM


app/views/editor/rich.html.erb
```diff
@@ -79,7 +79,7 @@
 
     <!-- main_image module -->
     <div class="ple-module-main_image ple-module container">
-      <div class="ple-module-guide col-md-3"><h2>2</h2></div>
+      <div class="ple-module-guide" style="flex: 0 0 25%; max-width: 25%;"><h2>2</h2></div>
```
jywarren, line 82: Ah, is there something we can do about this in the CSS file rather? Or, using a native Bootstrap class? I'm just wary of inline CSS in the example files. Thank you!!!!

jywarren (changes requested):

jywarren (approved):

jywarren:
Nice!!!

Issue #8545, opened 2020-10-12
Refactor error messages in Ruby on Rails forms to get rid of "dynamic_form" gem


jywarren:
We use the `dynamic_form` gem in many places for reasons that we can't even remember -- so if we try to remove that gem, as we did in https://github.com/publiclab/plots2/pull/4575, tests fail.

Let's go through the uses of this gem everywhere we can find it and replace it with a more modern and standard form syntax! They look like this, using the `error_messages` property of the `form_for` block:

```ruby
<%= form_for :revision, :as => :revision, :url => url, :html => {:class => "form"} do \|f\| %>
  <% if f.error_messages != "" %>
    <div class="alert alert-danger"><%= f.error_messages :header_message => "Your note couldn't be saved." %></div>
  <% end %>
<% end %>
```

But they should look like this:

```ruby
  <% if @model.errors.any? %>
    <div id="error_explanation">
      <h2><%= pluralize(@model.errors.count, "error") %> prohibited this model from being saved:</h2>
      <ul>
      <% @model.errors.full_messages.each do |msg| %>
        <li><%= msg %></li>
      <% end %>
      </ul>
    </div>
  <% end %>
```
We can search to find where changes should be made, and the simpler of these may potentially be good first-timers-only issues, although if updating these cause tests to need updating too, that could be a potential barrier for a newcomer, so let's be sure to support people well in these!

Here are many instances that could be broken out into their own issues: 

https://github.com/publiclab/plots2/search?q=error_messages&unscoped_q=error_messages

* [ ] https://github.com/publiclab/plots2/blob/c60aaabb56c018962c76abf3f5f2d114f8492d15/app/views/features/new.html.erb#L10
* [ ] 
* [ ] 
* [ ] 
* [ ]

jywarren:
I've made a first-timers-only issue from one of these! See how it looks to perform one of these updates by seeing this commit:

https://github.com/publiclab/plots2/commit/eada0c0970ebf1bcce1e6dcf384dc43813e9384a

stacytonui:
Hey can I try working on this?

jywarren:
Hi @stacytonui -- sure, can you choose one from the checklist above, and open a pull request with that one file change, similar to https://github.com/publiclab/plots2/commit/eada0c0970ebf1bcce1e6dcf384dc43813e9384a ? And link to it from here? Thank you so much!

stacytonui:
Okay let me work on the first one

stacytonui:
[...]

PR #8552, opened 2020-10-12
Refactor error message in _edit_form.html.erb, (partial fix for #8545)

noi5e:
Partial fix for #8545 (in plots2/app/views/users/_edit_form.html.erb)


noi5e:
This is getting a little messy, I'm going to close and resubmit this PR from a feature branch on the repo, instead of merging from noi5e/main.

PR #8553, opened 2020-10-12
Refactor error message in _edit_form.html.erb

### LINKED ISSUE #8545: Refactor error messages in Ruby on Rails forms to get rid of "dynamic_form" gem

noi5e:
Partially fixes #8545 in one location: plots2/app/views/users/_edit_form.html.erb

### DIFF
--- a/app/views/users/_edit_form.html.erb
+++ b/app/views/users/_edit_form.html.erb
@@ -2,8 +2,15 @@
 <%= form_for @user, :url => {:controller => 'users', :action => 'update'}, :html => {:class => "row",:id => 'edit-form' } do |f| %>
   <%= hidden_field_tag "user[ui_update]", "true" %>
   <div class="col-md-10">
-    <% if f.error_messages != "" %>
-      <div class="alert alert-danger"><%= f.error_messages %></div>
+    <% if @model.errors.any? %>
+      <div id="error_explanation">
+        <h2><%= pluralize(@model.errors.count, "error") %> prohibited this form from being saved:</h2>
+        <ul>
+          <% @model.errors.full_messages.each do |msg| %>
+            <li><%= msg %></li>
+          <% end %>
+        </ul>
+      </div>
     <% end %>
 
       <div class='col-md-12' style="display:flex; justify-content: center;">


noi5e:
@jywarren I'm guessing I should be concered about Travis CI not passing, any tips?

I have to admit, I have some familiarity with Ruby, but more with JS. I wonder if there's something wrong with my syntax that's causing any errors?

Also, maybe I should actually test this out locally. 🤦‍♂️ I'm still getting used to the codebase-- do you know where exactly this code is used, and how to see it in action? I'm guessing this is for something like updating userinfo in the profile page. Thanks for the help

jywarren:
Hi, no problem! The test logs should show us what went wrong. Indeed, this should be when editing your profile, at `/profile`

Let's see -- 

image

We could try these out by opening the code in GitPod using the button above. But, let's see if the errors are readable.

An integration test failed:

```

ERROR["test_should_choose_i18n_for_user/_photo_+_user/edit", #<Minitest::Reporters::Suite:0x000000000b29edd8 @name="I18nTest">, 110.85859752900004]
 test_should_choose_i18n_for_user/_photo_+_user/edit#I18nTest (110.86s)
Minitest::UnexpectedError:         ActionView::Template::Error: undefined method `errors' for nil:NilClass
            app/views/users/_edit_form.html.erb:5:in `block in _app_views_users__edit_form_html_erb___2356090617493035158_99485120'
            app/views/users/_edit_form.html.erb:2:in `_app_views_users__edit_form_html_erb___2356090617493035158_99485120'
            app/views/users/edit.html.erb:7:in `_app_views_users_edit_html_erb___747168809122195961_97933260'
            app/controllers/users_controller.rb:89:in `edit'
            test/integration/I18n_test.rb:218:in `block (2 levels) in <class:I18nTest>'
            test/integration/I18n_test.rb:204:in `each'
            test/integration/I18n_test.rb:204:in `block in <class:I18nTest>'
```

And a functional test failed:

```
ERROR["test_rejecting_malformated_email_while_updating_profile", #<Minitest::Reporters::Suite:0x000000001326bea8 @name="UsersControllerTest">, 63.357368742000006]
 test_rejecting_malformated_email_while_updating_profile#UsersControllerTest (63.36s)
Minitest::UnexpectedError:         ActionView::Template::Error: undefined method `errors' for nil:NilClass
            app/views/users/_edit_form.html.erb:5:in `block in _app_views_users__edit_form_html_erb___4035843046761948351_160617800'
            app/views/users/_edit_form.html.erb:2:in `_app_views_users__edit_form_html_erb___4035843046761948351_160617800'
            app/views/users/edit.html.erb:7:in `_app_views_users_edit_html_erb___3256878848318758383_160603060'
            app/controllers/users_controller.rb:72:in `update'
            test/functional/users_controller_test.rb:298:in `block in <class:UsersControllerTest>'
```

Both show `ActionView::Template::Error: undefined method `errors' for nil:NilClass` followed by a template line reference:

`            app/views/users/_edit_form.html.erb:5:in `block in _app_views_users__edit_form_html_erb___4035843046761948351_160617800'`

 -- could we be using the wrong instance variable for the model? Let's take a look...

app/views/users/_edit_form.html.erb
```diff
@@ -2,8 +2,15 @@
 <%= form_for @user, :url => {:controller => 'users', :action => 'update'}, :html => {:class => "row",:id => 'edit-form' } do |f| %>
   <%= hidden_field_tag "user[ui_update]", "true" %>
   <div class="col-md-10">
-    <% if f.error_messages != "" %>
-      <div class="alert alert-danger"><%= f.error_messages %></div>
+    <% if @model.errors.any? %>
```
jywarren, line 5: Aha! Yes, so here, the `@model` is a placeholder - we can replace it with `@user` and this should start to work! For all lines below as well. Sorry that was a bit ambiguous! This way it'll get the errors for the `@user` object. 

Let's try that! Great work so far!

jywarren (changes requested):

jywarren (approved):
Great! It worked!!! Great job!!

cesswairimu:
:tada:

PR #8586, opened 2020-10-15
default sort order fix of tags

### LINKED ISSUE #7758: Sort by name in /tag route returns in descending order
Usually when we sort by name we expect it to be sorted in the ascending order but in /tag route it returns in the descending order. This behaviour can be modified to return in ascending order by default instead of clicking it again to toggle the asc/desc.
image

https://publiclab.org/tags

Code in tag_controller, index method

akp2603:
Fixes #7758 (<=== Add issue number here)

Thanks!

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -8,7 +8,7 @@ class TagController < ApplicationController
 
     @title = I18n.t('tag_controller.tags')
     @paginated = true
-    @order_type = (params[:order].blank? || params[:order] == "desc") ? "asc" : "desc"
+    @order_type = params[:order].blank? || params[:order] == "desc" ? "asc" : "desc"
     powertag_clause = params[:powertags] == 'true' ? '' : ['name NOT LIKE ?', '%:%']
 
     if params[:search]
@@ -533,9 +533,9 @@ class TagController < ApplicationController
 
   def order_string
     if params[:search] || @toggle == "uses"
-      (params[:order].blank? || params[:order] == "asc") ? "count ASC" : "count DESC"
+      params[:order].blank? || params[:order] == "asc" ? "count ASC" : "count DESC"
     else
-      (params[:order].blank? || params[:order] == "asc") ? "name ASC" : "name DESC"
+      params[:order].blank? || params[:order] == "asc" ? "name ASC" : "name DESC"
     end
   end
 
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -8,7 +8,7 @@ class TagController < ApplicationController
 
     @title = I18n.t('tag_controller.tags')
     @paginated = true
-    @order_type = params[:order] == "desc" ? "asc" : "desc"
+    @order_type = (params[:order].blank? || params[:order] == "desc") ? "asc" : "desc"
     powertag_clause = params[:powertags] == 'true' ? '' : ['name NOT LIKE ?', '%:%']
 
     if params[:search]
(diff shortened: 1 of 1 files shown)


app/controllers/tag_controller.rb
```diff
@@ -8,7 +8,7 @@ def index
 
     @title = I18n.t('tag_controller.tags')
     @paginated = true
-    @order_type = params[:order] == "desc" ? "asc" : "desc"
+    @order_type = params[:order].blank? || params[:order] == "desc" ? "asc" : "desc"
```
jywarren, line 11: Ooh, wow the original code is quite hard to interpet, huh?

`@order_type = params[:order] == "desc" ? "asc" : "desc"`

I'm actually not sure the order of the operations... it almost looks like it would assign the value `true` or `false` (the output of the `==` evaluation) to `@order_type`? But i must be confused. 

What if we broke this up onto different lines, or added some parentheses to make it more readable? 

What i did to try to understand was to run the following in my `irb` console:

```ruby
Somerville:~ warren$ irb
irb(main):001:0> "desc" == "desc" ? "asc" : "desc"
=> "asc"
irb(main):002:0> "asc" == "desc" ? "asc" : "desc"
=> "desc"
irb(main):003:0> "desc".blank? || "desc" == "desc" ? "asc" : "desc" 
NoMethodError: undefined method `blank?' for "desc":String
	from (irb):3
	from /usr/bin/irb:11:in `<main>'
irb(main):004:0> "desc".empty? || "desc" == "desc" ? "asc" : "desc"
=> "asc"
irb(main):008:0> "asc".empty? || "asc" == "desc" ? "asc" : "desc"
=> "desc"
irb(main):007:0> "".empty? || "" == "desc" ? "asc" : "desc"
=> "asc"
```

So one thing is that I think it would have to be `.empty?` instead of `.blank?`? (i believe, although maybe the Ruby version affects this?)

But, is the purpose here to have it default to "desc" if it's empty? If so maybe something simpler like this could work? 

```ruby
param = "desc" if param.empty?
```

akp2603: Ohh hi @jywarren , thanks for taking out the time to review this. 
> Ooh, wow the original code is quite hard to interpet, huh?
Haha, true. After all the setting up and debugging, it all came down to this line, in the end. If only one could see my face, then.

> I'm actually not sure the order of the operations... it almost looks like it would assign the value true or false (the output of the == evaluation) to @order_type? But i must be confused.

From what I observed in the code is order_type is to toggle between the current default ordering of the tags. Hence it's "asc" when the current value(that is passed in `params[:order]`) has a value "desc" and vice versa.

> What if we broke this up onto different lines, or added some parentheses to make it more readable?

rubocop (the linter for ruby used here), does not recommend using parenthesis for ternary operators, but can surely change it if you want. let me know in response.

> "desc".blank? || "desc" == "desc" ? "asc" : "desc" 
> So one thing is that I think it would have to be .empty? instead of .blank?? (i believe, although maybe the Ruby version affects this?)
 This is because `.blank?` is a rails method and not a `ruby` method. I prefer it to be `blank?`, mainly because `"".blank?` is true and `"   ".blank?` is also true. Hence every string with only whitespaces is considered to be `blank`. Just a convention, I personally follow and prefer. But again, `empty?` also works, just that `"   ".empty?` is false, hence should not be preferred, IMO.

Thoughts?

akp2603: @jywarren ?

cesswairimu (approved):
Thanks a lot for working on this :tada: 

The fix works great :rocket:, I will keep the issue open incase @jywarren may want us to refactor it further. All the best on hackoberfest @akp2603 :100:

Issue #8560, opened 2020-10-13
Add a new /preview route to open editor preview in new tab


Sagarpreet:
>  Create a POST route at /editor/preview which simply renders a simulated "note" or "wiki page" using the Markdown in the "body" and "title" POST variables. Nothing is saved in the db, and we repoint all editor Preview buttons at this. (not sure about comment previews though, probably not those). We also make a Preview button for the rich editor pointed at this. All open in a new tab, and should use the [->] "open in new tab" icon from FontAwesome

> Add "Preview" button next to the Publish button [ON WHICH EDITORS?] that opens a preview of your content in a new tab

Sagarpreet:
More details here: https://github.com/publiclab/PublicLab.Editor/issues/490#issuecomment-669930675

jywarren:
For this, i'm thinking we could make a notes controller method for `preview` and then in the templates, we could do a `@ node.title || params[:title]` substitution for both `title` and `body` -- 




https://github.com/publiclab/plots2/blob/3664a280d660f6f059fbadfcd38f059129f7645f/app/views/notes/show.html.erb#L23

https://github.com/publiclab/plots2/blob/3664a280d660f6f059fbadfcd38f059129f7645f/app/views/notes/show.html.erb#L64


...but there are lots of references to `@node` on the template, so another way we could try would be to manually assemble a `@node` with a revision and author (current author) but not save any of it...? so it's just an instance variable and never entered in the database? That would mean no template changes needed, except perhaps for an alert message that it is just a preview?

we might also need to pass in the `main_image` URL as a parameter; it will have been uploaded and saved, so that should be OK...

sagarpreet-chadha:
Okay @jywarren that makes sense 👍 
We have to add another button in /post editor page that will open a new tab and call this route, right?

RuthNjeri:
Hi @sagarpreet-chadha I'd like to try this out...please assign it to me.

sagarpreet-chadha:
Hey @RuthNjeri , sure go ahead! I was just going to start on this myself :)
You may start with a different route /preview and pass the form details as body params and render the preview.
Let me know if you need any help 😄 
Meanwhile I will ask @ebarry or @jywarren on where exactly we have to show this preview button, we can take that separately. Thanks 💯

sagarpreet-chadha:
This will be pretty interesting feature, we will also make sure that we write proper tests for this 😄 
[...]

PR #666, opened 2016-08-16
Add documentation for Question page

ananyo2012:
Make sure these boxes are checked before your pull request is ready to be reviewed and merged. Thanks!

* [x] All tests pass -- `rake test:all`
* [x] code is in uniquely-named feature branch, and has been rebased on top of latest master (especially if you've been asked to make additional changes)
* [x] pull request are descriptively named
* [x] if possible, multiple commits squashed if they're smaller changes
* [ ] reviewed/confirmed/tested by another contributor or maintainer
* [x] `schema.rb.example` has been updated if any database migrations were added

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/wiki/contributing-to-public-lab-software

We have a loose schedule of reviewing and pulling in changes every Tuesday and Friday, and publishing changes on Fridays. Please alert developers on plots-dev@googlegroups.com when your request is ready or if you need assistance.

Thanks!

Add modal view for Q & A help documentation

PR #8620, opened 2020-10-19
Notes Preview

RuthNjeri:
Fixes #https://github.com/publiclab/plots2/issues/8560 (<=== Add issue number here)

Thanks!


RuthNjeri:
Hi @sagarpreet-chadha I have made a couple of changes, I would appreciate any feedback of pointers you may have on my current changes before I begin writing tests.., also tagging @jywarren for this :)

app/models/node.rb
```diff
@@ -715,6 +722,23 @@ def self.new_note(params)
     [saved, node, revision]
   end
 
+  def self.new_preview_note(params)
+    author = User.find(params[:uid])
+    node = Node.new(uid:     author.uid,
+                    title:   params[:title],
+                    comment: 2,
+                    type:    'note')
+    revision = node.new_revision(uid:   author.uid,
```
Sagarpreet, line 731: Is this working?
We have not saved `node ` so there would be no `id` for this node, so in `new_revision` method 'id' would be null, right?

https://github.com/publiclab/plots2/blob/fc11ad2b1b03aab6e3f313e8eab1d701cae49714/app/models/node.rb#L667

Sagarpreet: Can you verify this?

app/models/node.rb
```diff
@@ -715,6 +722,23 @@ def self.new_note(params)
...
+                                title: params[:title],
+                                body:  params[:body])
+    node.status = 4 if author.first_time_poster
+    node.draft if params[:draft] == "true"
```
Sagarpreet, line 735: Niceee 🎉

app/models/node.rb
```diff
@@ -390,8 +390,10 @@ def response_count(key = 'response')
   end
 
   # power tags have "key:value" format, and should be searched with a "key:*" wildcard
-  def has_power_tag(key)
-    !power_tag(key).blank?
+  def has_power_tag(key, preview_tags = [])
+    return !power_tag(key).blank? if preview_tags.blank?
+    preview_tag = preview_tags.join(" ").match(key)
+    !preview_tag.blank?
```
Sagarpreet, line 396: We are using these 2 lines again in line 482,483...can we make a small util function and call it at both places?

app/controllers/notes_controller.rb
```diff
@@ -434,4 +444,22 @@ def show_banned_flash
     flash.keep[:error] = I18n.t('notes_controller.you_have_been_banned').html_safe
     redirect_to '/logout'
   end
+
+  def add_preview_tags
```
Sagarpreet, line 448: Awesome 💯

app/views/notes/show.html.erb
```diff
@@ -1,38 +1,39 @@
...
     <%= feature("alert-#{@node.power_tag('alert')}") %>
   <% end %>
-  <% if @node.has_power_tag('with') && current_user && @node.coauthors.collect(&:username).include?(current_user.username) %> <div class="alert alert-success"> You are a coauthor of this post. This attributes your contribution, and gives you edit access to the note. <a href="https://publiclab.org/wiki/power-tags#Coauthorship"> Learn more </a></div><% end %>
-  <% if @node.has_power_tag('upgrade') %><div class="alert alert-success"> This is an upgrade for <a href="/<%= @node.power_tag('upgrade') %>"><%= @node.power_tag('upgrade') %></a>. Try building it and <a href="/n/<%= @node.power_tag('upgrade') %>"> report back </a>how it goes.</div><% end %>
+  <% if @node.has_power_tag('with', @preview_tags) && current_user && @node.coauthors.collect(&:username).include?(current_user.username) && !@preview %> <div class="alert alert-success"> You are a coauthor of this post. This attributes your contribution, and gives you edit access to the note. <a href="https://publiclab.org/wiki/power-tags#Coauthorship"> Learn more </a></div><% end %>
```
Sagarpreet, line 13: Correct! We should not show this 👍

app/views/notes/show.html.erb
```diff
@@ -1,38 +1,39 @@
...
+  <% if @node.has_power_tag('upgrade', @preview_tags) %><div class="alert alert-success"> This is an upgrade for <a href="/<% if !@preview %><%= @node.power_tag('upgrade') %>"><%= @node.power_tag('upgrade') %><% end %></a>. Try building it and <a href="/n/<% if !@preview %><%= @node.power_tag('upgrade') %><% end %>"> report back </a>how it goes.</div><% end %>
```
Sagarpreet, line 14: Correct, we cannot call `power_tag ` as it makes DB calls 💯

app/views/notes/show.html.erb
```diff
@@ -1,38 +1,39 @@
...
+  <% if @node.main_image || @img %>
+    <a class="main-image" style="max-width:100%;width:800px;" href="<%= @node.main_image.path(:original) || @img.path(:original) %>">
+      <%= image_tag(@node.main_image.path(:large) || @img.path(:large), class:"rounded d-none d-lg-inline", style:'max-height:600px;max-width:100%;') %>
+      <%= image_tag(@node.main_image.path(:large) || @img.path(:large), class:"rounded d-lg-none", style:'width:100%;') %>
     </a>
     <!--<div class="expand"><a onClick="$('.main-image').toggleClass('compressed');"><i class="fa fa-angle-down"></i></a></div>-->
   <% end %>
 
-  <h1 style="margin-top: 40px;"><%= @node.title %></h1>
+  <h1 style="margin-top: 40px;"><%= @node.title ||  %></h1>
```
Sagarpreet, line 24: We can remove redundant `||` here

config/routes.rb
```diff
@@ -142,6 +142,7 @@
   get 'notes/liked' => 'notes#liked'
   get 'notes/image/:id' => 'notes#image'
   get 'notes/delete/:id' => 'notes#delete'
+  post 'notes/preview' => 'notes#preview'  # check this!
```
Sagarpreet, line 145: We can remove this comment now

app/views/notes/show.html.erb
```diff
@@ -61,9 +62,10 @@
   <%= render :partial => "notes/gallery" %>
 
   <div id="content" class="pl-content note">
-  <%= raw auto_link(insert_extras(@node.latest.render_body), :sanitize => false) %>
+  <%= raw auto_link(insert_extras(@node.latest.render_body || @body.render_body), :sanitize => false) %>
   </div>
 
+  <%# TODO8: Is this depreciated code? Ask @jywarren %>
```
Sagarpreet, line 68: @jywarren

Sagarpreet (changes requested):
Hey @RuthNjeri , great progress so far!!!
Couple of changes and verification needed.
Also the existing tests are failing, I think it is due to ruby error here (redundant `||`)

I want to test this in gitpod after you do the changes, can you also add a curl request?
Or you can share screenshot (with and without image)

We can then start writing tests for this 🎉 

Thanks 😄

app/models/node.rb
RuthNjeri: I've verified it, it creates the revision with both ids being nil
Screenshot 2020-10-21 at 17 22 15

app/views/notes/show.html.erb
```diff
@@ -82,18 +84,18 @@
   <div class="d-print-none">
     <hr />
 
-    <% if @node.has_power_tag('prompt') && feature("prompt-#{@node.power_tag('prompt')}") != "" %>
+    <% if @node.has_power_tag('prompt', @preview_tags) && feature("prompt-#{@node.power_tag('prompt')}") != "" %>
       <% if current_user && (current_user.id == @node.uid || current_user.admin? || current_user.moderator? || current_user.is_coauthor?(@node)) %>
         <%= feature("prompt-#{@node.power_tag('prompt')}") %>
       <% end %>
     <% end %>
```
RuthNjeri, line 92: @Sagarpreet should these lines be displayed? I am not really sure what `prompt` means..

RuthNjeri:
Thanks for the feedback @Sagarpreet! I have implemented most of them, I'm currently just about to test the post request and add a screenshot...

RuthNjeri:
Hi @Sagarpreet I tried creating a CURL request but got stuck with adding a user and csrf token.. 
I took an alternative route of drafting a simple test that hits the preview route.. which work.. but.. I am currently stuck because I noticed that the preview route's node `@node` contains a `nid` which shouldn't be the case since it should not be saved.. I am not sure what is causing this.. I tried looking into the model callbacks but didn't find anything, I also tried transferring the `new_preview_note` method to the controller, but @node still ends up being saved..could this be a Rails thing?

app/controllers/notes_controller.rb
```diff
@@ -157,6 +158,17 @@ def create
     end
   end
 
+  def preview
+    return show_banned_flash unless current_user.status == User::Status::NORMAL
+    @node, @img, @body = new_preview_note
+    binding.pry
```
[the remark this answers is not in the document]
Sagarpreet: Remove this after debugging

app/views/notes/show.html.erb
Sagarpreet: We have to mock `power_tag` model function also, but for now we can remove display of lines where  `power_tag` is used.
https://github.com/publiclab/plots2/blob/fc11ad2b1b03aab6e3f313e8eab1d701cae49714/app/models/node.rb#L398

test/functional/notes_controller_test.rb
```diff
@@ -1061,4 +1061,21 @@ def test_get_rss_feed
      assert_equal I18n.t('notes_controller.saved_as_draft'), flash[:notice]
      assert_redirected_to '/notes/' + users(:jeff).username + '/' + Time.now.strftime('%m-%d-%Y') + '/' + title.parameterize
    end
+
+  # Preview test
```
Sagarpreet, line 1065: Is the test not working? Seems correct 👍

Sagarpreet (commented):
Everything looks great 🎉

Sagarpreet:
Can you re-upload the Screenshot, it seems to be broken

test/functional/notes_controller_test.rb
RuthNjeri: I was using it to test the route, but it's not passing yet because I discovered that the node is actually being created..
I have updated the screenshot 

https://github.com/publiclab/plots2/pull/8620#issuecomment-713725564

app/controllers/notes_controller.rb
```diff
@@ -434,4 +453,17 @@ def show_banned_flash
     flash.keep[:error] = I18n.t('notes_controller.you_have_been_banned').html_safe
     redirect_to '/logout'
   end
+
+  def add_preview_tags
```
Sagarpreet, line 457:
```suggestion
  def add_preview_tags
     tags = []
```

Sagarpreet (changes requested):
Also this change:

Screenshot 2020-10-22 at 4 07 58 PM

test/functional/notes_controller_test.rb
Sagarpreet: It is not getting saved for me:

Screenshot 2020-10-22 at 4 12 42 PM

Sagarpreet:
@RuthNjeri 
1. You need to do 2 changes mentioned above

2. You can use this curl for testing:

```
curl --location --request POST 'http://localhost:3000/notes/preview' \
--header 'Connection: keep-alive' \
--header 'Accept: */*' \
--header 'X-CSRF-Token: <CSRF_TOKEN>' \
--header 'X-Requested-With: XMLHttpRequest' \
--header 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.80 Safari/537.36' \
--header 'Content-Type: application/x-www-form-urlencoded; charset=UTF-8' \
--header 'Origin: http://localhost:3000' \
--header 'Sec-Fetch-Site: same-origin' \
--header 'Sec-Fetch-Mode: cors' \
--header 'Sec-Fetch-Dest: empty' \
--header 'Referer: http://localhost:3000/post' \
--header 'Accept-Language: en-GB,en-US;q=0.9,en;q=0.8' \
--header 'Cookie: _ga=GA1.1.291287496.1600765372; plots2_locale=en; _session_id=<YOUR_SESSION>; user_token=<YOUR_TOKEN>
--data-raw 'title=testingNotes&body=testingNotestestingNotestestingNotestestingNotes&authenticity_token=<<YOUR_TOKEN>>draft=false&tags=&has_main_image=&main_image=&node_images=&image%5Bphoto%5D='
```

test/functional/notes_controller_test.rb
RuthNjeri: Interesting.. I'll retry it

RuthNjeri:
Thank you @Sagarpreet, I'll keep working on it and update you

Sagarpreet:
🎉 🎉 🎉

RuthNjeri:
Hi @Sagarpreet I have tried running the test to ensure that a record is not being saved but it seems that a new record is being created in the test..I'm not sure what's happening..
Screenshot 2020-10-22 at 18 05 48

jywarren:
Huh... Are we sure .new is working as we expect? And .count? What happens if you try this manually in the rails console with "rails c"? Like try running Node.count then Node.new, then Node.count again...?

How are you sure .new is working as expected in the running app? Thanks, this is a tough one! If you're still stuck in a few hours I'll try to run it myself in GitPod!

RuthNjeri:
Hi @jywarren, @Sagarpreet was able to confirm via gitpod that it's not being saved with the id being nil [here](https://github.com/publiclab/plots2/pull/8620#discussion_r510061594)

Screenshot 2020-10-23 at 19 46 31

I also tested it here.. the test shows it as saved versus when I make a request via the route the controller shows it as not saved... I think this issue is specific to the test environment..

> Hi @Sagarpreet I have tried running the test to ensure that a record is not being saved but it seems that a new record is being created in the test..I'm not sure what's happening..
> 
> The failing build test..
> 
> Screenshot 2020-10-22 at 18 05 48
> 
> This seems to happen only during testing.. on the controller..the nid is nil
> Screenshot 2020-10-22 at 19 45 20

sagarpreet-chadha:
Hey @RuthNjeri , 
When you make the post request, do you see any SQL queries like CREATE new node being executed in rails server terminal?

Is it possible that the node is actually getting saved in some model function?

RuthNjeri:
Hi @Sagarpreet , I'm using Postman to send a request, and same thing happens, the node is not saved..
Screenshot 2020-10-23 at 21 53 16

Sagarpreet:
@RuthNjeri tests passed now 🎉 Kindly check now!

Sagarpreet:
Also, we can use this: https://stackoverflow.com/a/3951843 to easily open a new `window` and pass the params using a hidden form in `/post` page 😄

RuthNjeri:
Thanks @Sagarpreet 🎉 ... I was actually about to share that as well.. I added multiple breaking points checking the node value and discovered that its being created after the draft method 😄 

This was a tough one 😅 , thanks again @Sagarpreet.. I'll keep adding tests and check out the link you've sent!

Screenshot 2020-10-25 at 12 56 03

RuthNjeri:
> @RuthNjeri tests passed now 🎉 Kindly check now!
> 
> I am not sure how the `draft` flow works but anyways we do not have to pass `draft`param = **true** for preview as it somehows save the node.

@Sagarpreet the draft method was the culprit.. it saves the object.. 🤦🏾 
Screenshot 2020-10-25 at 12 58 34

sagarpreet-chadha:
🎉

jywarren:
OMG!!! Amazing work here. Are you still working on this (i see you mentioned adding tests)?

> @Sagarpreet the draft method was the culprit.. it saves the object.. 🤦🏾

Shall we rename `draft` method to `save_draft` maybe, for clarity? This could be an FTO perhaps although we should watch out for a unit test breaking if we rename. 

Hey, really amazing deep work on this -- very impressive. @RuthNjeri you should be very proud of this, it was a very obscure issue!

RuthNjeri:
Thanks, @jywarren 😄 , I'm really learning more about the codebase through this task which is great!
Yes, I've written one test so far, I'm currently looking into making the POST request open a new window then write a few more tests while receiving feedback on the next changes I'm making once pushed.

I can check whether renaming the draft method breaks any tests and create an FTO issue for it 👌🏾

cesswairimu:
@RuthNjeri is this ready for review? if so or when ready please remove the `wip` thanks

KarenEfereyan:
Waow this is a lot. Well done @RuthNjeri

RuthNjeri:
Hi @cesswairimu, it's not completely ready.. I'm working on the frontend bit in addition to the tests.. I think it'll be ready by next Monday/Tuesday since I'm currently making revisions on the Outreachy Proposal.. Once It's ready I'll remove the WIP tag and add a comment..

cesswairimu:
No worries and no pressure,  just thought I got a review request. 
You are really doing a great job here :tada: ,

sagarpreet-chadha:
Hey @RuthNjeri , let me know if you are stuck somewhere? Thanks 😄

RuthNjeri:
Hi @Sagarpreet thanks for reaching out.. I had a knee injury yesterday and couldn't do anything productive...I'm now back and I'll be pushing in some changes..I'll let you know how it goes 😄

RuthNjeri:
Hi @Sagarpreet .. I have made some progess.. though there are a couple of things I am trying to understand..

1. The preview window is being opened in another tab, I implemented an Ajax request when the preview button is clicked. 
Screenshot 2020-11-03 at 20 26 51


So far, I can pass the Title and body, which is great 🎉 
Every time a user presses the preview button, it rewrites the previous data, these changes are[ here](https://github.com/publiclab/plots2/pull/8620/files#diff-2978988608dfe16e254a7a5390e6a7164f0d287a961d0c90a4de88a19dd0d58bR525-R561)

What's currently pending:
1. Images: I've tried testing image uploads locally and no upload is happening, should I have done a specific setup locally for images?
2. Location: Figuring out how I can fetch the location and display it in the preview section..
3.  Testing whether related tags show up on the side bar of a preview

What's confusing:
1. The footer I have for the application locally is not similar to the hosted version(production). In the preview section, the footer is not fixed at the bottom and does not have the footer details. I'm still trying to figure this out..


That's my update so far 😄

app/views/editor/rich.html.erb
```diff
@@ -518,4 +522,44 @@
...
+
+  function preview() {
+    let url = 'notes/preview'
+    $.ajax({
+      type: "POST",
+      url: url,
+      data:
+        {
+          'title': editor.titleModule.value(),
+          // 'image': editor.mainImageModule,
```
sagarpreet-chadha, line 535: So when we upload image, they are saved in backend. I think we can simply pass the url

app/views/editor/rich.html.erb
```diff
@@ -518,4 +522,44 @@
...
+          // 'location': editor.titleModule.value(),
```
sagarpreet-chadha, line 536: Similary location tags are added when we click `publish`, however we can read the location from map module

sagarpreet-chadha:
Hey @RuthNjeri , sorry to hear about the knee injury. Kindly take care 💯 
Great progress so far, I think the location and image params can be exactly same which are sent on publishing the post so that is a minor change. Once that is done, we can merge this PR as it is pending from long time now, what do you think?

app/views/editor/rich.html.erb
sagarpreet-chadha: Okay when we upload an image, then a post call `/image` goes to backend.
The response of this api, we can url and `id`. We are interested in this `id` only, we can pass this `id` in the same manner as it get passed when we publish post (inside param: `main_image `)

app/views/editor/rich.html.erb
sagarpreet-chadha: tags can be passed as `tags: lat:1.324592,  lon:13.687162,  zoom:5`, and on preview the map will be rendered automatically.

RuthNjeri:
Hi @Sagarpreet thanks! I agree we can merge the PR once the image and location tags are complete and all changes approved.

I've been trying to understand how the Publish button is triggered to send the necessary params to the create controller... If I can figure out how the editor form details are consolidated into the create params, I think this task will be complete... 

I'm currently tracking how the tags and the image are added like you've suggested..

app/views/editor/rich.html.erb
RuthNjeri: This is what I've come up with so far..

```suggestion
    let image_url = editor.mainImageModule.image.src
    let image_id = image_url.split("/").pop()
```

RuthNjeri:
I've discovered that the editor.data supplies the params for publishing.. though not all details are displayed when I click on the preview button.. still looking into this.. hopefully, I complete this by EOD🤞🏾

app/views/editor/rich.html.erb
RuthNjeri: I have managed to figure out how to access the location...

this is the current POST data that I'm sending
```suggestion
               data:
        {
          'title': editor.titleModule.value(),
          'body': editor.richTextModule.value(),
          'main_image': image_id,
          'tags': editor.tagsModule.value(),
          'latitude': editor.mapModule.blurredLocation.getLat(),
          'longitude': editor.mapModule.blurredLocation.getLon(),
          'zoom': editor.mapModule.blurredLocation.getZoom()
        },
```

RuthNjeri:
@Sagarpreet I've managed to pass the image and location details via the AJAX call, what is currently pending is the display of the maps and image via the preview tab... I'm not sure if the fact that the preview tag is opened under the URL 'about:blank' is causing a Google maps error... I'm still investigating..

Screenshot 2020-11-04 at 21 08 43
Screenshot 2020-11-04 at 21 09 10

RuthNjeri:
@Sagarpreet The PR is finally ready for review 🎉 ... I've removed the [WIP] tag.
Here's a screenshot of the image and location working in the preview tab
Screenshot 2020-11-04 at 22 39 51

The only current issue is that I'm unable to test location `By Place` ... there seems to be an issue loading it..
Screenshot 2020-11-04 at 22 52 33

app/views/notes/show.html.erb
sagarpreet-chadha: Cool, I think we can remove this comment for now 👍

app/views/editor/rich.html.erb
```diff
@@ -518,4 +522,51 @@
       console.log("Draft:", editor.data.draft);
   }
 
+  var previewWindow = null;
+
+  function preview() {
+    let url = 'notes/preview'
```
sagarpreet-chadha, line 528: We need to add semicolon in Javascript after each statement 😄

app/views/editor/rich.html.erb
```diff
@@ -518,4 +522,51 @@
...
+                    }
+
+    $.ajax({
+      type: "POST",
+      url: url,
+      data:
+        {
+          'title': editor.titleModule.value(),
+          'body': editor.richTextModule.value(),
+          'main_image': image_id,
```
sagarpreet-chadha, line 543: correct 👍

app/views/editor/rich.html.erb
```diff
@@ -518,4 +522,51 @@
...
+        },
+      success: function(data)
+      {
+        previewPopup(data);
+      }
+    })
+  }
+
+  function previewPopup(data) {
+    if(previewWindow == null || previewWindow.closed)
```
sagarpreet-chadha, line 555:
```suggestion
    if(!previewWindow || previewWindow.closed)
```

sagarpreet-chadha (changes requested):

sagarpreet-chadha:
Amazinggg, tested on gitpod 💯 
Screenshot 2020-11-05 at 5 21 26 PM

sagarpreet-chadha (approved):
Awesome, everything seems good 🎉

RuthNjeri:
Thanks! I'm still testing it a bit.. I noticed that after an article has been published and someone clicks on the preview button, it uses the wrong route and prevents the preview window from being opened.. 

I'm currently looking into this.. thanks for polishing up the code 😄 
Screenshot 2020-11-05 at 17 19 24

sagarpreet-chadha:
Awesome 🎉 Great catch 💯

sagarpreet-chadha:
Hey @RuthNjeri , in the last commit I added the time to display instead of `January 01, 1970 00:00` 😄

RuthNjeri:
@Sagarpreet good catch 😄 , I hadn't noticed that! It is weird that it defaults to 1970.. is that intentional?

sagarpreet-chadha:
So as we are not saving our @node object, so its created value is 0. 
So epoctime 0 corresponds to that date 1st Jan, 1970 😄

app/models/node.rb
```diff
@@ -176,6 +176,11 @@ def setup
     save
   end
 
+  def match_preview_tag(key, preview_tags)
+    preview_tag = preview_tags.join(" ").match(key)
```
jywarren, line 180: Could this method in general be replaced with `preview_tags.includes?(key)`? (i think that's the syntax) -- without a separate method? Just looking - this code looks fine, but that might be simpler? Thanks!

app/models/node.rb
```diff
@@ -390,8 +395,9 @@ def response_count(key = 'response')
   end
 
   # power tags have "key:value" format, and should be searched with a "key:*" wildcard
-  def has_power_tag(key)
-    !power_tag(key).blank?
+  def has_power_tag(key, preview_tags = [])
+    return !power_tag(key).blank? if preview_tags.blank?
+    match_preview_tag(key, preview_tags)
```
jywarren, line 400: Might this cause trouble, if another part of the app tries to use `has_power_tag()` in the old way? I just worry we're overloading this pretty simple function with some complicated and relatively limited-scope extra abilities that make it tough to understand and maintain. Let's see lower down...

app/models/node.rb
```diff
@@ -460,21 +466,25 @@ def location_tags
...
+      # tag = Tag.where(name: tagname).try(:first)
+      # if tag && tag.parent
+      #  tags += Tag.includes(:node_tag)
+      #             .references(:community_tags)
+      #             .where("community_tags.nid = ? AND name LIKE ?", self.id, tag.parent)
+      # end
+      tids = tags.collect(&:tid).uniq
+      !NodeTag.where('nid IN (?) AND tid IN (?)', id, tids).empty?
+    else
+      match_preview_tag(tagname, preview_tags)
```
jywarren, line 486: Here, yes, i see the full "if" loop a bit more clearly than in `has_power_tag` above, appreciate that. But again i wonder if we are overloading it with too much complex functionality. Maybe we could do this at a... "shallower" level and not complicate this very standardized method? I'll keep reading though, thanks!

app/models/node.rb
jywarren: Cool!!!

app/models/node.rb
```diff
@@ -715,6 +725,27 @@ def self.new_note(params)
...
+                    latitude: lat.to_f,
+                    longitude: params[:location][:longitude].to_f,
+                    comment: 2,
+                    type:    'note')
+    precision = node.decimals(lat.to_s)
+    node.precision = precision
+    revision = node.new_revision(uid:   author.uid,
+                                title: params[:title],
+                                body:  params[:body])
+    node.status = 4 if author.first_time_poster
```
jywarren, line 742: Maybe this line is not necessary as we're never saving this, so status shouldn't matter. Let's try removing it!
```suggestion
```

app/models/revision.rb
```diff
@@ -143,8 +143,9 @@ def body_raw
   end
 
   def body_extras(content)
-    # inline edit button
-    content = content.gsub('[edit]', '<p class="well" style="padding:6px;"><a class="btn btn-primary" href="' + parent.edit_path + '"><i class="fa fa-white icon-pencil"></i> Edit this page</a> to help complete it!</p>')
+    # inline edit button adds a # if vid is absent showing a preview
+    edit_path = vid ? parent.edit_path : '#'
```
jywarren, line 147: oh wow, complicated, and great catch!

app/views/editor/rich.html.erb
```diff
@@ -518,4 +522,51 @@
       console.log("Draft:", editor.data.draft);
   }
 
+  var previewWindow = null;
```
jywarren, line 525: Really nicely done this whole section. very readable.

app/views/notes/show.html.erb
```diff
@@ -1,25 +1,29 @@
...
     <%= feature("alert-#{@node.power_tag('alert')}") %>
   <% end %>
-  <% if @node.has_power_tag('with') && current_user && @node.coauthors.collect(&:username).include?(current_user.username) %> <div class="alert alert-success"> You are a coauthor of this post. This attributes your contribution, and gives you edit access to the note. <a href="https://publiclab.org/wiki/power-tags#Coauthorship"> Learn more </a></div><% end %>
-  <% if @node.has_power_tag('upgrade') %><div class="alert alert-success"> This is an upgrade for <a href="/<%= @node.power_tag('upgrade') %>"><%= @node.power_tag('upgrade') %></a>. Try building it and <a href="/n/<%= @node.power_tag('upgrade') %>"> report back </a>how it goes.</div><% end %>
+  <% if @node.has_power_tag('with', @preview_tags)  && current_user && @node&.coauthors&.collect(&:username)&.include?(current_user.username) && !@preview  %> <div class="alert alert-success"> You are a coauthor of this post. This attributes your contribution, and gives you edit access to the note. <a href="https://publiclab.org/wiki/power-tags#Coauthorship"> Learn more </a></div><% end %>
```
jywarren, line 14: I see now, the difficulty here. Pretty tough! Was it really hard to try to create temporary `Tag` records too, without saving? I guess, if we don't save them, then the `has_power_tag` functions won't work at all.

So what we're doing is... providing a set of tags to search through /instead/ of the database, right? In a general sense. 

Could we generalize this idea then, to offer as a 2nd parameter, not "preview tags" specifically, but maybe... an `override_tags` parameter, which "overrides" the tags which would be searched? That way, `@node.has_power_tag(tag, override_tags)` would be a little easier to understand, without having to know the specifics of this kind of odd use case? What do you think of that?

We could even add a unit test that demonstrates that `@node.has_power_tag('nonexistent-tag')` returns false, but `@node.has_power_tag('nonexistent-tag', ['nonexistent-tag'])` passes?

app/views/sidebar/_map.html.erb
```diff
@@ -1,17 +1,21 @@
-<% if @node && @node.lat && @node.lon %>
+<% lat = @preview? @node.latitude.to_f : @node.lat %>
+<% lon = @preview? @node.longitude.to_f : @node.lon %>
+
+<% if @node && lat && lon %>
+  <% node_zoom = @preview? @zoom : @node.zoom %>
```
jywarren, line 5: wow, well done here

test/functional/notes_controller_test.rb
```diff
@@ -1061,4 +1061,25 @@ def test_get_rss_feed
...
+   test 'Node is not saved in note preview' do
+    UserSession.create(users(:jeff))
+    title = 'My preview post about balloon mapping'
+    location = {'latitude': "-1",
+                'longitude': "-1",
+                'zoom': "6"
+                }
+
+    assert_no_difference 'Node.count' do  post :preview,
```
jywarren, line 1074: very nice.

app/views/notes/show.html.erb
jywarren: I guess this is just a terminology change to the `has_power_tag` modifications above. So the parameter would be more descriptive as `override_tags` and the name would indicate an easier to understand, more generalized usage. 

The unit test I mention could go here: 

https://github.com/publiclab/plots2/blob/7c365d921f7dcdfe102ba11857569eed70b09a9d/test/unit/node_test.rb#L333-L336

I think it's a relatively simple one to add and would help protect these otherwise kind of obscure changes to the node.rb methods. What do you think!

jywarren (changes requested):
Wow @RuthNjeri (and @Sagarpreet!) this is a really epic and impressive project -- kudos! 

Very excited about it. It seems super thoroughly done. I left a few clarifying questions and a slight change to the node.rb methods as a requested change, but i want to commend you @RuthNjeri for a complex and well developed PR here! We should be able to merge shortly if you can make these small tweaks, and I would certainly appreciate the unit test i mentioned as well. 

Great work. 🎉

RuthNjeri:
Hi @jywarren, thanks! I was able to learn a lot more about the codebase using this task...
I'll have a look at the comments/suggestions you've made and implement them as soon as I can.

app/views/notes/show.html.erb
RuthNjeri: Hi @jywarren so when I tested this locally, it only seemed to display notices related to the power tags method, I don't think it does this 

> `So what we're doing is... providing a set of tags to search through /instead/ of the database, right? In a general sense.`

At some point, I had an internal debate as to whether adding the preview tags within the power methods is really necessary because some of the notices displayed looked like what the author of the note would see and not what a user would see which is the purpose of a preview...

Creating a temporary Tag seems like a great idea.. let me do that first and see if it makes the current implementation less complicated.

jywarren: OK but i think the issue with temporary tag records is that `has_tag` and `has_powertag` both make new database queries, so the temp tags wouldn't appear in such queries. That's why i think your idea of passing a "replacement" or "override" collection of tags as an optional 2nd parameter is a great idea, now that i understood the scope of the problem. Would you like to just rename it to that more general name and we can give that a try? Let me take a look and make a suggestion, but we could confirm it with a couple simple unit tests as well.

app/models/node.rb
jywarren: So here, could we change all instances of `preview_tags` to a more generalized name `override_tags` and add a comment above like `override_tags limits the tags within which to search to a predefined list`?

app/views/notes/show.html.erb
RuthNjeri: Yes, an override is a good idea however, it does not help in displaying that the preview post `is part of` certain tags.. like in a normally posted note...since the `_tagging.html.erb` template still hits the DB(`normal_tags`)
```
<% if @node %>
  <!-- listing of posts in response to this one -->
  <%= render partial: 'tag/replication' %>
  <!-- This is the sidebar tagging display, also renders Subscribe button for multiple subscription -->
  <p style="font-size:0.9em;color:#666;margin-top:5px;">
    <% if @node.normal_tags.count > 0 %>
      This is part of:
    <% end %>
  </p>
  <div id="tags" class="node-tags">
    <%= render partial: 'tag/tags', locals: { power_tag: false, badge_name: 'badge-secondary', tags: @node.normal_tags(:followers).includes(:tag) } %>
    <%= render partial: 'tag/tags', locals: { power_tag: true, badge_name: 'badge-primary', tags: @node.power_tag_objects.includes(:tag) } %>
  </div>
<% else %>
  <div id="tags" class="profile-tags">
    <%= render partial: 'tag/tags', locals: { power_tag: false, badge_name: 'badge-primary', tags: user.user_tags, user: user } %>
  </div>
<% end %>
```
which is another part of displaying the preview tags... 

Currently, even with the override, the code will not display related tags in the preview section...which is another issue to look at...
Screenshot 2020-11-09 at 19 22 59

app/models/node.rb
RuthNjeri: Good idea, let me do that...

RuthNjeri:
@jywarren I have renamed the preview_tag argument in the node.rb to overrride_tags and added the respective test...
The only current issue I have is [this](https://github.com/publiclab/plots2/pull/8620#discussion_r519942535) which I am currently looking at...

> Yes, an override is a good idea however, it does not help in displaying that the preview post `is part of` certain tags.. like in a normally posted note...since the `_tagging.html.erb` template still hits the DB(`normal_tags`)
> 
> ```
> <% if @node %>
>   
>   <%= render partial: 'tag/replication' %>
>   
>   <p style="font-size:0.9em;color:#666;margin-top:5px;">
>     <% if @node.normal_tags.count > 0 %>
>       This is part of:
>     <% end %>
>   </p>
>   <div id="tags" class="node-tags">
>     <%= render partial: 'tag/tags', locals: { power_tag: false, badge_name: 'badge-secondary', tags: @node.normal_tags(:followers).includes(:tag) } %>
>     <%= render partial: 'tag/tags', locals: { power_tag: true, badge_name: 'badge-primary', tags: @node.power_tag_objects.includes(:tag) } %>
>   </div>
> <% else %>
>   <div id="tags" class="profile-tags">
>     <%= render partial: 'tag/tags', locals: { power_tag: false, badge_name: 'badge-primary', tags: user.user_tags, user: user } %>
>   </div>
> <% end %>
> ```
> 
> which is another part of displaying the preview tags...
> 
> Currently, even with the override, the code will not display related tags in the preview section...which is another issue to look at...
> Screenshot 2020-11-09 at 19 22 59

sagarpreet-chadha:
Hey @jywarren , how do you propose we do mocking for related tags? 

Also user cannot control the related tags that appear after post is published, right? Does it makes sense to show these in preview page?

@RuthNjeri new terminologies makes more sense 🙌 

Apart from mocking in this way, the other way would had been to save in DB everything and later delete everything after preview is closed somehow, but that will cause unnecessary DB transactions and make this a little slow as well!

Also ideally we should delete the images that gets saved as part of preview in DB, right @jywarren ?
-> maybe a different route /deleteImages/id that gets called after preview window is closed (we can listen to event listeners)
-> this can be done separately later as well

Thanks 😄

RuthNjeri:
@Sagarpreet I was thinking since the user cannot control related tags that show up after the post is published, what if we display a notice for the user... like this 
Screenshot 2020-11-10 at 20 50 33

jywarren:
Regarding maps, i believe it's just because we don't have an API key for maps when running it from anywhere except publiclab.org/stable.publiclab.org. It should be fine!

RuthNjeri:
Hi @jywarren, that's great feedback! I agree, the tags do make the preview logic a bit complicated.. 
I have updated the files and it now looks like this... let me know what you think cc @Sagarpreet 

Screenshot 2020-11-11 at 20 53 02

sagarpreet-chadha (commented):
This makes implementation very simple 🎉 looks great 😄 💯

RuthNjeri:
I think a rebuild will make the tests pass 😄 ... should I close and reopen the PR or is it something a maintainer should do?

RuthNjeri:
I closed and open the PR to rebuild the tests..

Sagarpreet (approved):
LGTM 🎉 

@jywarren can you review again, maybe we can break separate issues and solve them as follow ups if needed?

jywarren:
This is tremendous. Thank you so much! I'm merging now, please give it a try at https://stable.publiclab.org once it's built there. Thanks again for this big big project!

jywarren (approved):

KarenEfereyan:
This is massive. Well done @RuthNjeri

RuthNjeri:
Thanks, @jywarren! I have tested it on  https://stable.publiclab.org apart from the share and Tweet link which lead to empty posts, everything else works as expected. 

Screenshot 2020-11-18 at 10 32 40

config/routes.rb
```diff
@@ -142,6 +142,7 @@
   get 'notes/liked' => 'notes#liked'
   get 'notes/image/:id' => 'notes#image'
   get 'notes/delete/:id' => 'notes#delete'
+  post 'notes/preview' => 'notes#preview'
```
jywarren, line 145: Hmm, @RuthNjeri - i believe this is live now and i tried going to https://publiclab.org/notes/preview?title=Hello%20there&body=Testing%20this to see if it worked... I got redirected to https://publiclab.org/tag/preview -- i'm sure it's working properly but what's the best way to try it out from a URL? Shouldn't GET params work the same as POST params? Thanks!

sagarpreet-chadha: Hey @jywarren ,
Hitting from browser directly will not work as it is GET, we can define a seperate GET route to handle this case.
For now, we have to use a curl request, I mentioned the template above in this thread.
Also this POST route is authenticated and we need to pass csrf token with the curl.

RuthNjeri: Hi @jywarren, @Sagarpreet is right, the route only accepts POST requests, I have made an edit via this [PR](https://github.com/publiclab/plots2/pull/8755) to make it accept both POST and GET, please have a look

jywarren: Aha, that's awesome, thank you!!! And I see it already works at https://stable.publiclab.org/notes/preview?title=Hello%20there&body=Testing%20this -- that's great!!! Fantastic!!!

jywarren:
I'm going to suggest that we add a button for this to solve #8698!

PR #8683, opened 2020-10-27
test re-filtering moderation digests with user tags

jywarren:
Noting:

It's a bit convoluted because this is an asynchronous scheduled job, but it's initiated in https://github.com/publiclab/plots2/blob/c60aaabb56c018962c76abf3f5f2d114f8492d15/app/jobs/digest_spam_job.rb#L13 and then the mail is triggered in https://github.com/publiclab/plots2/blob/f6f4a6db4ef2fd3a59cdf07023ea2815e006f667/app/models/user.rb#L377, but that then seems to re-fetch the user list and not filter it, in https://github.com/publiclab/plots2/blob/5ed9d5d8f9be585fcd6b853dd680a3437feef95e/app/mailers/admin_mailer.rb#L124-L137. I can make a modification to the last file but i'm a little surprised it's not being sent multiple times.

### DIFF
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -127,11 +127,17 @@ class AdminMailer < ActionMailer::Base
     elsif frequency_digest == User::Frequency::WEEKLY
       @subject = 'Your weekly digest for moderation'
     end
-    moderators = User.where(role: %w(moderator admin)).collect(&:email)
+    all_moderators = User.where(role: %w(moderator admin))
+    moderators = []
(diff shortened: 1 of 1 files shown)


jywarren:
Hi @keshavsethi @gauravano -- just curious if this makes sense to you? I think it's possible we started sending digests to /all moderators/ instead of just those who follow the appropriate moderator digest user tag, over the summer?

jywarren:
And, moderators report getting the digest email only once, but maybe Gmail is filtering out repeated sends? It just seems like we are iterating over the users list twice, you know?

PR #8734, opened 2020-11-11
Return helpful error message when attempting to login/signup w Twitter oauth with no email

### LINKED ISSUE #8325: Trouble with Twitter Signup/Login
Hi there, I just signed up for a new account using the Twitter option. After authorizing via Twitter's consent screen, I was redirected back to publiclab.org and saw this:
2020-08-18_1920x1080

I see the same error when I tried clicking the verification link from the confirmation email and when I tried logging in.

jywarren:
Fixes #8325 with a test - we may need to remove `uid` as well!

## Part 1

We need to modify this logic to account for not having a Twitter email:

Both of these pathways could lead to this same error: 

https://github.com/publiclab/plots2/blob/c60aaabb56c018962c76abf3f5f2d114f8492d15/app/controllers/user_sessions_controller.rb#L75-L99

And if there's no email provided, perhaps we should divert away here too:

https://github.com/publiclab/plots2/blob/c60aaabb56c018962c76abf3f5f2d114f8492d15/app/controllers/user_sessions_controller.rb#L52-L56

I think this section should divert to the final `else` if there is no email in the provided Twitter response.

## Part 2

We then should try solving for the scenario with the allocator undefined for Proc error from #8325:

> if `create(value: "oauth:" + auth['provider'] + ":" + auth['uid'], uid: uid, data: auth.to_hash)` fails, maybe it's the non-existence of auth['uid'] OR it could be the .to_hash and it's failing to hash something that is too complex to hash? That might be a more subtle one.

https://github.com/publiclab/plots2/blob/4b2e5321de6cc0c2341535b722722c1bfa1804f0/app/models/user_tag.rb#L25-L28

### DIFF
--- a/test/functional/user_sessions_controller_test.rb
+++ b/test/functional/user_sessions_controller_test.rb
@@ -178,7 +178,9 @@ class UserSessionsControllerTest < ActionController::TestCase
       assert_not_nil request.env['omniauth.auth']
       #Sign Up for a new user
       assert_difference 'User.count', 0 do
-        post :create
+        assert_difference 'UserSession.count', 0 do
+          post :create
+        end
       end
       assert_equal "You have tried using a Twitter account with no associated email address. Unfortunately we need an email address; please add one and try again, or sign up a different way. Thank you!",  flash[:error]
     end
--- a/test/functional/user_sessions_controller_test.rb
+++ b/test/functional/user_sessions_controller_test.rb
@@ -154,75 +154,74 @@ class UserSessionsControllerTest < ActionController::TestCase
     assert_equal "Successfully logged out.",  flash[:notice]
   end
 
+  test 'sign up and login via provider basic flow for twitter' do
+    assert_not_nil OmniAuth.config.mock_auth[:twitter1]
+    #Omniauth hash is present
+    request.env['omniauth.auth'] =  OmniAuth.config.mock_auth[:twitter1]
+    assert_not_nil request.env['omniauth.auth']
+    #Sign Up for a new user
+    post :create
+    assert_equal "You have successfully signed in. Please change your password using the link sent to you via e-mail.",  flash[:notice]
+    #Log Out
+    post :destroy
+    assert_equal "Successfully logged out.",  flash[:notice]
+    #auth hash is present so login via a provider
+    post :create
+    assert_equal "Successfully logged in.",  flash[:notice]
+  end
 
-    test 'sign up and login via provider basic flow for twitter' do
-      assert_not_nil OmniAuth.config.mock_auth[:twitter1]
-      #Omniauth hash is present
-      request.env['omniauth.auth'] =  OmniAuth.config.mock_auth[:twitter1]
-      assert_not_nil request.env['omniauth.auth']
-      #Sign Up for a new user
+  test 'sign up and login via provider basic flow for twitter user with no email' do
+    assert_not_nil OmniAuth.config.mock_auth[:twitter_no_email]
+    #Omniauth hash is present
+    request.env['omniauth.auth'] =  OmniAuth.config.mock_auth[:twitter_no_email]
+    assert_not_nil request.env['omniauth.auth']
+    #Sign Up for a new user
+    session[:user_session].should be_nil
+    assert_difference 'User.count', 0 do
       post :create
-      assert_equal "You have successfully signed in. Please change your password using the link sent to you via e-mail.",  flash[:notice]
-      #Log Out
-      post :destroy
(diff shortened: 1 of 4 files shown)


test/functional/user_sessions_controller_test.rb
```diff
@@ -170,6 +170,22 @@ class UserSessionsControllerTest < ActionController::TestCase
...
     end
+  
+    test 'sign up and login via provider basic flow for twitter user with no email' do
+      assert_not_nil OmniAuth.config.mock_auth[:twitter_no_email]
+      #Omniauth hash is present
+      request.env['omniauth.auth'] =  OmniAuth.config.mock_auth[:twitter_no_email]
+      assert_not_nil request.env['omniauth.auth']
+      #Sign Up for a new user
+      post :create
+      assert_equal "You have successfully signed in. Please change your password by editing your profile in the upper right menu.",  flash[:notice]
```
jywarren, line 181: OK, we'll have to change this too, from `assert_equal "You have successfully signed in. Please change your password using the link sent to you via e-mail.",  flash[:notice]`

jywarren: https://github.com/publiclab/plots2/blob/4b2e5321de6cc0c2341535b722722c1bfa1804f0/app/controllers/user_sessions_controller.rb#L92-L94

jywarren:
So one big unknown is whether we can create User records with no email in the first place. Many things on the site may break, i suppose... maybe we should enter a dummy email?

jywarren:
I'll write a unit test for it...

jywarren:
Other than the user email issue, i see: 

```
ERROR["test_sign_up_and_login_via_provider_basic_flow_for_twitter_user_with_no_email", #<Minitest::Reporters::Suite:0x0000000010adeb10 @name="UserSessionsControllerTest">, 26.474472665999997]
 test_sign_up_and_login_via_provider_basic_flow_for_twitter_user_with_no_email#UserSessionsControllerTest (26.47s)
Minitest::UnexpectedError:         NoMethodError: undefined method `tr' for nil:NilClass
            app/models/user.rb:445:in `create_with_omniauth'
            app/controllers/user_sessions_controller.rb:77:in `handle_social_login_flow'
            app/controllers/user_sessions_controller.rb:11:in `create'
            test/functional/user_sessions_controller_test.rb:180:in `block in <class:UserSessionsControllerTest>'
```

Indeed, this line means we need an email address! https://github.com/publiclab/plots2/blob/main/app/models/user.rb#L445

```ruby
      email_prefix = auth["info"]["email"].tr('.', '_').split('@')[0]
      email_prefix = auth["info"]["email"].tr('.', '_').split('@')[0] + random_chars until User.where(username: email_prefix).empty?
```

Maybe we should just throw an error saying we require an email in your Twitter account to use it to create an account?

jywarren:
Confirming that removing email address for a test user broke lots of notifications stuff. We'd have to come up with alternative ways for all those features to work...

jywarren:
OK we are clear (from PL staff) to refusing Twitter OAuth account creation from Twitter accounts that don't have emails - and prompting people to create an email to get around this. Thanks!

jywarren:
```
ERROR["test_sign_up_and_login_via_provider_basic_flow_for_twitter_user_with_no_email", #<Minitest::Reporters::Suite:0x00000000105b26a0 @name="UserSessionsControllerTest">, 87.508377427]
 test_sign_up_and_login_via_provider_basic_flow_for_twitter_user_with_no_email#UserSessionsControllerTest (87.51s)
Minitest::UnexpectedError:         NameError: undefined local variable or method `be_nil' for #<UserSessionsControllerTest:0x00000000105ad8a8>
            test/functional/user_sessions_controller_test.rb:179:in `block in <class:UserSessionsControllerTest>'
```

test/functional/user_sessions_controller_test.rb
```diff
@@ -154,61 +154,74 @@ class UserSessionsControllerTest < ActionController::TestCase
...
-      request.env['omniauth.auth'] =  OmniAuth.config.mock_auth[:twitter1]
-      assert_not_nil request.env['omniauth.auth']
-      #Sign Up for a new user
-      post :create
-      assert_equal "You have successfully signed in. Please change your password using the link sent to you via e-mail.",  flash[:notice]
-      #Log Out
-      post :destroy
-      assert_equal "Successfully logged out.",  flash[:notice]
-      #auth hash is present so login via a provider
+  test 'sign up and login via provider basic flow for twitter user with no email' do
```
jywarren, line 173: Noting this is now protected by a specific test for the Twitter-login-with-no-email scenario!

PR #8794, opened 2020-12-01
Try changing search box label on /spam2

jywarren:


jywarren:
This should work, i think!

PR #8795, opened 2020-12-01
Try running Travis jobs in GitHub Actions

jywarren:
starting with db setup, gems, yarn, and unit tests

Fixes #8793 

And lots more stuff related to chromedriver... maybe this is now added in?: 

https://github.com/publiclab/plots2/blob/ff1974a08b39aedc5982ac602e89f9a3c0f6c2d9/.travis.yml#L1-L39

Issue #6887, opened 2019-12-02
Outreachy project planning


jywarren:
First, congratulations again, and thank you for choosing Public Lab for your Outreachy fellowship! :tada: :tada:

I wanted to start to get our respective project plans aligned; first because we will have two fellows working together on this, and second, because there have been and will continue to be some other parallel and overlapping projects with connections to your work. 

This project fits into a broader set of "Geographic systems" (https://github.com/publiclab/plots2/milestone/7) which were also worked on by @ananyaarun (during the last round of Outreachy) and others as well, including @sagarpreet-chadha, @cesswairimu, and more, both here in `plots2` and in [Leaflet Environmental Layers](https://github.com/publiclab/leaflet-environmental-layers)!

@crisner has been doing some great work on the LEL library, while @nstjean has been doing some great work on the `plots2` side of things, so this seems like, potentially, a natural complimentary arrangement. So, let's discuss a bit of each side and where the overlaps are!

### Original proposals

* https://publiclab.org/notes/natalie_stjean/10-29-2019/outreachy-proposal-refine-geographic-ui-and-js-integration-for-publiclab-org
* https://publiclab.org/notes/christie_reni/10-26-2019/outreachy-proposal-refine-geographic-ui-and-js-integration-for-publiclab-org

### Resources

Some starting resources -- it's a good idea to read through our [Style guide](https://publiclab.org/notes/warren/05-07-2019/introducing-a-draft-style-guide-for-public-lab) and [Roadmap](https://publiclab.org/notes/warren/05-22-2019/draft-of-a-public-lab-software-roadmap-comments-welcome) for a solid idea of how new features should be both planned and designed. 

OK I am taking too long to post this so I'll just hit Publish and add more in a bit. 

Welcome @crisner @nstjean and the Outreachy mentors! @ananyaarun @rexagod @cesswairimu @aSquare14 @IshaGupta18 @MayankKashyap!

👋 🙌 🎉 💯 🔥

jywarren:
I've been organizing a more list-style planning issue here, with lots of related issues linked in and grouped; i'll try to organize that today! https://github.com/publiclab/plots2/issues/6801

jywarren:
I also wanted to point to this great resource on testing your work on the `unstable` server -- it's a great way to see how things run in an environment almost identical to the production server! https://github.com/publiclab/plots2/blob/master/doc/TESTING.md#testing-your-work-with-unstable

jywarren:
[...]

PR #8800, opened 2020-12-04
[Outreachy] Add All Posts Route /notes 

RuthNjeri:
Fixes https://github.com/publiclab/plots2/issues/8828

I'm still trying to understand what 'All Posts' means.. this is a first attempt at trying to understand it. 
I'll add a test once feedback is provided.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -38,7 +38,7 @@ class ApplicationController < ActionController::Base
       @maps = Tag.find_nodes_by_type(data, 'map', 20)
     else # type is generic
       # remove "classroom" postings; also switch to an EXCEPT operator in sql, see https://github.com/publiclab/plots2/issues/375
-      hidden_nids = Node.hidden_response_nids
+      hidden_nids = Node.hidden_response_node_ids
       @notes = if params[:controller] == 'questions'
                  Node.questions
                    .joins(:revision)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -38,13 +38,13 @@ class ApplicationController < ActionController::Base
       @maps = Tag.find_nodes_by_type(data, 'map', 20)
     else # type is generic
       # remove "classroom" postings; also switch to an EXCEPT operator in sql, see https://github.com/publiclab/plots2/issues/375
-      hidden_nids = Node.where(type: :note, status: 1).select { |n| n.has_a_tag('hidden:response') }.collect(&:nid)
+      hidden_nids = Node.hidden_response_nids
       @notes = if params[:controller] == 'questions'
                  Node.questions
                    .joins(:revision)
                else
                  Node.research_notes.joins(:revision).order('node.nid DESC').paginate(page: params[:page])
-      end
+               end
 
       @notes = @notes.where('node.nid != (?)', @node.nid) if @node
       @notes = @notes.where('node_revisions.status = 1 AND node.nid NOT IN (?)', hidden_nids) unless hidden_nids.empty?
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -64,10 +64,11 @@ class HomeController < ApplicationController
   end
 
   # Display all posts assuming all posts == activity
-  def all
+  def all_posts
     if current_user
       # underscore used for variables not needed
       # I'm assuming all comments == @comments + @answer_comments
+      # TO DO: Identify what will be pinned at the top of the page. A featured post maybe?
       _, _, @notes, @wikis, _, @comments, @answer_comments = activity
       @questions = Node.questions.where(status: 1)
       render template: 'dashboard/all_posts'
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -63,20 +63,6 @@ class HomeController < ApplicationController
     end
   end
 
-  # Display all posts assuming all posts == activity
-  def all_posts
-    if current_user
-      # underscore used for variables not needed
-      # I'm assuming all comments == @comments + @answer_comments
-      # TO DO: Identify what will be pinned at the top of the page. A featured post maybe?
-      _, _, @notes, @wikis, _, @comments, @answer_comments = activity
-      @questions = Node.questions.where(status: 1)
-      render template: 'dashboard/all_posts'
-    else
-      redirect_to '/research'
-    end
-  end
-
   private
 
   def activity
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -63,6 +63,19 @@ class HomeController < ApplicationController
     end
   end
 
+  # Display all posts assuming all posts == activity
+  def all
+    if current_user
+      # underscore used for variables not needed
+      # I'm assuming all comments == @comments + @answer_comments
+      _, _, @notes, @wikis, _, @comments, @answer_comments = activity
+      @questions = Node.questions.where(status: 1)
+      render template: 'dashboard/all_posts'
+    else
+      redirect_to '/research'
+    end
+  end
+
   private
 
   def activity
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -69,7 +69,6 @@ class HomeController < ApplicationController
     blog = Tag.find_nodes_by_type('blog', 'note', 1).first
     # remove "classroom" postings; also switch to an EXCEPT operator in sql, see https://github.com/publiclab/plots2/issues/375
     hidden_nids = Node.hidden_response_node_ids
-    
     notes = Node.where(type: 'note')
       .where('node.nid NOT IN (?)', hidden_nids + [0]) # in case hidden_nids is empty
       .order('nid DESC')
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -68,12 +68,8 @@ class HomeController < ApplicationController
   def activity
     blog = Tag.find_nodes_by_type('blog', 'note', 1).first
     # remove "classroom" postings; also switch to an EXCEPT operator in sql, see https://github.com/publiclab/plots2/issues/375
-    hidden_nids = Node.joins(:node_tag)
-      .joins('LEFT OUTER JOIN term_data ON term_data.tid = community_tags.tid')
-      .select('node.*, term_data.*, community_tags.*')
-      .where(type: 'note', status: 1)
-      .where('term_data.name = (?)', 'hidden:response')
-      .collect(&:nid)
+    hidden_nids = Node.hidden_response_node_ids
+    
     notes = Node.where(type: 'note')
       .where('node.nid NOT IN (?)', hidden_nids + [0]) # in case hidden_nids is empty
       .order('nid DESC')
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -5,7 +5,7 @@ class NotesController < ApplicationController
 
   def index
     @title = I18n.t('notes_controller.research_notes')
-    @pagy, @notes = pagy(published_notes.order('node.nid DESC'))
+    @pagy, @notes = pagy(published_notes)
   end
 
   def tools
@@ -305,8 +305,8 @@ class NotesController < ApplicationController
   def liked
     @title = I18n.t('notes_controller.highly_liked_research_notes')
     @notes = published_notes
-      .limit(20)
-      .order('cached_likes DESC')
+             .limit(20)
+             .order('cached_likes DESC')
     @unpaginated = true
     render template: 'notes/index'
   end
@@ -450,8 +450,8 @@ class NotesController < ApplicationController
   end
 
   def published_notes
-    hidden_nids = Node.hidden_response_nids
-    notes = Node.research_notes.where('node.status = 1')
+    hidden_nids = Node.hidden_response_node_ids
+    notes = Node.research_notes.where('node.status = 1').order('node.nid DESC')
 
     if hidden_nids.empty?
       notes
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -447,7 +447,7 @@ class NotesController < ApplicationController
 
   def published_notes
     hidden_nids = Node.hidden_response_node_ids
-    
+
     Node.research_notes
         .where('node.status = 1')
         .where.not(nid: hidden_nids)
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -42,7 +42,7 @@ class NotesController < ApplicationController
   # display a revision, raw
   def raw
     response.headers['Content-Type'] = 'text/plain; charset=utf-8'
-    render plain: Node.find(params[:id]).latest.body
+    render plain: Node.find(params[:id]).latest&.body
   end
 
   def show
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -5,7 +5,7 @@ class NotesController < ApplicationController
 
   def index
     @title = I18n.t('notes_controller.research_notes')
-    @pagy, @notes = pagy(published_notes)
+    @pagy, @notes = pagy(published_notes.order('node.nid DESC'))
   end
 
   def tools
@@ -304,8 +304,7 @@ class NotesController < ApplicationController
   # notes with high # of likes
   def liked
     @title = I18n.t('notes_controller.highly_liked_research_notes')
-    @notes = Node.research_notes
-      .where(status: 1)
+    @notes = published_notes
       .limit(20)
       .order('cached_likes DESC')
     @unpaginated = true
@@ -314,7 +313,7 @@ class NotesController < ApplicationController
 
   def recent
     @title = I18n.t('notes_controller.recent_research_notes')
-    @pagy, @notes = pagy(Node.where(type: 'note', status: 1, created: Time.now.to_i - 1.weeks.to_i..Time.now.to_i)
+    @pagy, @notes = pagy(published_notes.where(created: Time.now.to_i - 1.weeks.to_i..Time.now.to_i)
                  .order('created DESC'))
     render template: 'notes/index'
   end
@@ -322,9 +321,8 @@ class NotesController < ApplicationController
   # notes with high # of views
   def popular
     @title = I18n.t('notes_controller.popular_research_notes')
-    @notes = Node.research_notes
+    @notes = published_notes
       .limit(20)
-      .where(status: 1)
       .order('views DESC')
     @unpaginated = true
     render template: 'notes/index'
@@ -459,6 +457,6 @@ class NotesController < ApplicationController
       notes
     else
       notes.where('node.nid NOT IN (?)', hidden_nids)
-    end.order('node.nid DESC')
+    end
   end
 end
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -447,12 +447,9 @@ class NotesController < ApplicationController
 
   def published_notes
     hidden_nids = Node.hidden_response_node_ids
-    notes = Node.research_notes.where('node.status = 1')
-
-    if hidden_nids.empty?
-      notes
-    else
-      notes.where('node.nid NOT IN (?)', hidden_nids)
-    end
+    
+    Node.research_notes
+        .where('node.status = 1')
+        .where.not(nid: hidden_nids)
   end
 end
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -5,7 +5,7 @@ class NotesController < ApplicationController
 
   def index
     @title = I18n.t('notes_controller.research_notes')
-    set_sidebar
+    @pagy, @notes = pagy(published_notes)
   end
 
   def tools
@@ -304,10 +304,6 @@ class NotesController < ApplicationController
   # notes with high # of likes
   def liked
     @title = I18n.t('notes_controller.highly_liked_research_notes')
-    @wikis = Node.limit(10)
-      .where(type: 'page', status: 1)
-      .order('nid DESC')
-
     @notes = Node.research_notes
       .where(status: 1)
       .limit(20)
@@ -318,21 +314,15 @@ class NotesController < ApplicationController
 
   def recent
     @title = I18n.t('notes_controller.recent_research_notes')
-    @wikis = Node.limit(10)
-      .where(type: 'page', status: 1)
-      .order('nid DESC')
-    @notes = Node.where(type: 'note', status: 1, created: Time.now.to_i - 1.weeks.to_i..Time.now.to_i)
-                 .order('created DESC')
-    @unpaginated = true
+    # Not sure why it was unpaginated before
+    @pagy, @notes = pagy(Node.where(type: 'note', status: 1, created: Time.now.to_i - 1.weeks.to_i..Time.now.to_i)
+                 .order('created DESC'))
     render template: 'notes/index'
   end
 
   # notes with high # of views
   def popular
     @title = I18n.t('notes_controller.popular_research_notes')
-    @wikis = Node.limit(10)
-      .where(type: 'page', status: 1)
-      .order('nid DESC')
     @notes = Node.research_notes
       .limit(20)
       .where(status: 1)
@@ -460,5 +450,16 @@ class NotesController < ApplicationController
     longitude = @node.longitude.present? ? @node.longitude.to_f : false
 
     [latitude, longitude]
+  end
+
+  def published_notes
+    hidden_nids = Node.hidden_response_nids
+    notes = Node.research_notes.where('node.status = 1')
+
+    if hidden_nids.empty?
+      notes
+    else
+      notes.where('node.nid NOT IN (?)', hidden_nids)
+    end.order('node.nid DESC')
   end
 end
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -314,7 +314,6 @@ class NotesController < ApplicationController
 
   def recent
     @title = I18n.t('notes_controller.recent_research_notes')
-    # Not sure why it was unpaginated before
     @pagy, @notes = pagy(Node.where(type: 'note', status: 1, created: Time.now.to_i - 1.weeks.to_i..Time.now.to_i)
                  .order('created DESC'))
     render template: 'notes/index'
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -5,7 +5,7 @@ class NotesController < ApplicationController
 
   def index
     @title = I18n.t('notes_controller.research_notes')
-    @pagy, @notes = pagy(published_notes)
+    @pagy, @notes = pagy(published_notes.order('node.nid DESC'))
   end
 
   def tools
@@ -304,10 +304,7 @@ class NotesController < ApplicationController
   # notes with high # of likes
   def liked
     @title = I18n.t('notes_controller.highly_liked_research_notes')
-    @notes = published_notes
-             .limit(20)
-             .order('cached_likes DESC')
-    @unpaginated = true
+    @pagy, @notes = pagy(published_notes.limit(100).order(cached_likes: :desc, nid: :desc))
     render template: 'notes/index'
   end
 
@@ -321,10 +318,9 @@ class NotesController < ApplicationController
   # notes with high # of views
   def popular
     @title = I18n.t('notes_controller.popular_research_notes')
-    @notes = published_notes
-      .limit(20)
-      .order('views DESC')
-    @unpaginated = true
+    @pagy, @notes = pagy(published_notes
+            .limit(100)
+            .order(views: :desc, nid: :desc))
     render template: 'notes/index'
   end
 
@@ -451,7 +447,7 @@ class NotesController < ApplicationController
 
   def published_notes
     hidden_nids = Node.hidden_response_node_ids
-    notes = Node.research_notes.where('node.status = 1').order('node.nid DESC')
+    notes = Node.research_notes.where('node.status = 1')
 
     if hidden_nids.empty?
       notes
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -123,12 +123,13 @@ class Node < ActiveRecord::Base
     path.split('/').last
   end
 
-  def has_a_tag(name)
-    tags.where(name: name).size.positive?
-  end
-
   def self.hidden_response_node_ids
-    Node.where(type: :note, status: 1).select { |n| n.has_a_tag('hidden:response') }.collect(&:nid)
+    Node.joins(:node_tag)
+        .joins('LEFT OUTER JOIN term_data ON term_data.tid = community_tags.tid')
+        .select('node.nid')
+        .where(type: 'note', status: 1)
+        .where('term_data.name = (?)', 'hidden:response')
+        .collect(&:nid)
   end
 
   before_save :set_changed_and_created
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -124,7 +124,11 @@ class Node < ActiveRecord::Base
   end
 
   def has_a_tag(name)
-    return tags.where(name: name).size.positive?
+    tags.where(name: name).size.positive?
+  end
+
+  def self.hidden_response_nids
+    Node.where(type: :note, status: 1).select { |n| n.has_a_tag('hidden:response') }.collect(&:nid)
   end
 
   before_save :set_changed_and_created
@@ -920,8 +924,14 @@ class Node < ActiveRecord::Base
                .group('node.nid')
                .collect(&:nid)
 
-    Node.where(type: 'note')
-                .where('node.nid NOT IN (?)', nids)
+    notes = Node.where(type: 'note')
+
+    # if the nids are empty, it should not be used in the query
+    if nids.empty?
+      notes
+    else
+      notes.where('node.nid NOT IN (?)', nids)
+    end
   end
 
   def body_preview(length = 100)
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -126,7 +126,7 @@ class Node < ActiveRecord::Base
   def self.hidden_response_node_ids
     Node.joins(:node_tag)
         .joins('LEFT OUTER JOIN term_data ON term_data.tid = community_tags.tid')
-        .select('node.nid')
+        .select('node.nid, term_data.tid, term_data.name, community_tags.tid')
         .where(type: 'note', status: 1)
         .where('term_data.name = (?)', 'hidden:response')
         .collect(&:nid)
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -127,7 +127,7 @@ class Node < ActiveRecord::Base
     tags.where(name: name).size.positive?
   end
 
-  def self.hidden_response_nids
+  def self.hidden_response_node_ids
     Node.where(type: :note, status: 1).select { |n| n.has_a_tag('hidden:response') }.collect(&:nid)
   end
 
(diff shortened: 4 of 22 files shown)


RuthNjeri:
The `dashboard/all` route is meant to display all posts. I have used the activity method to derive the details needed(Research note, questions, wiki, comments tab)

The only pending section is what will be displayed at the banner section...should it be a featured post?

app/controllers/home_controller.rb
```diff
@@ -63,6 +63,20 @@ def research
     end
   end
 
+  # Display all posts assuming all posts == activity
+  def all_posts
+    if current_user
+      # underscore used for variables not needed
+      # I'm assuming all comments == @comments + @answer_comments
+      # TO DO: Identify what will be pinned at the top of the page. A featured post maybe?
```
sagarpreet-chadha, line 71: I think this TODO is powered by some config maybe as it keeps changes. @jywarren  can confirm

app/controllers/home_controller.rb
```diff
@@ -63,6 +63,20 @@ def research
...
+      _, _, @notes, @wikis, _, @comments, @answer_comments = activity
```
sagarpreet-chadha, line 72: I think this function is being used some other place also and supports pagination also which is great.

app/views/dashboard/all_posts.html.erb
```diff
@@ -0,0 +1 @@
+<h1>All Posts Template</h1>
```
sagarpreet-chadha, line 1: Just a heads up, you can start taking care of internationalization  from beginning :)

sagarpreet-chadha (commented):
Looks great till now

jywarren:
Ah, my apologies @RuthNjeri - in this new design, we won't be mixing the types, so each of Notes, Questions, Wikis, and Comments will each simply have their own tab, and their own URL (so we won't load them all on a single page load, it'll be more efficient to just reload the page when the tab is clicked, just like at https://publiclab.org/tag/outreachy for example!). 

Make sense? Thanks!!!

app/views/dashboard/all_posts.html.erb
jywarren: This will be great not to miss anything. It is possible to do it at the end, but perhaps a better idea to do it as you go, as Sagarpreet suggests!

app/controllers/home_controller.rb
jywarren: I believe this page would not have anything pinned. The dashboard itself would! Thanks!

app/controllers/home_controller.rb
jywarren: This is where we only need to show "notes" for the initial view, since unlike the current old dashboard we won't be mixing them together. We'll make separate routes, URLs , and tabs for each type. 

Would this perhaps best fit into... a new controller? I'm not sure. If the home controller isn't too full, i guess it's fine here. Your choice! Thank you!!!

jywarren:
So, perhaps we should think of this as a collection of pages that make up "All posts" -- All notes, All questions (would this lead to https://publiclab.org/questions so as not to be redundant?), All wikis (https://publiclab.org/wiki), All comments (this last one might actually link to or replace https://publiclab.org/comments?)

Let's make this PR just about notes, as there is no place where all notes are collected. We can think on the remaining types in the meantime! I'll ping @publiclab/community-reps about this. The main challenge if we link all other types to their corresponding pages, is that if our layout looks like tabs, people may expect those pages to appear like tabs, and be confused if they're led to pages that don't seem to be a sub-tab of the "All posts" page collection. Either this is just OK, and not a big deal, or maybe we would think of some way to indicate that the other types aren't tabs at all, but just links?

app/views/dashboard/all_posts.html.erb
RuthNjeri: I was planning on creating some FTO issues for translations as well. I'll see how I can implement it partially and create some FTOs if possible.

RuthNjeri:
> Ah, my apologies @RuthNjeri - in this new design, we won't be mixing the types, so each of Notes, Questions, Wikis, and Comments will each simply have their own tab, and their own URL (so we won't load them all on a single page load, it'll be more efficient to just reload the page when the tab is clicked, just like at https://publiclab.org/tag/outreachy for example!).
> 
> Make sense? Thanks!!!

This makes perfect sense!

RuthNjeri:
> So, perhaps we should think of this as a collection of pages that make up "All posts" -- All notes, All questions (would this lead to https://publiclab.org/questions so as not to be redundant?), All wikis (https://publiclab.org/wiki), All comments (this last one might actually link to or replace https://publiclab.org/comments?)
> 
> Let's make this PR just about notes, as there is no place where all notes are collected. We can think on the remaining types in the meantime! I'll ping @publiclab/community-reps about this. The main challenge if we link all other types to their corresponding pages, is that if our layout looks like tabs, people may expect those pages to appear like tabs, and be confused if they're led to pages that don't seem to be a sub-tab of the "All posts" page collection. Either this is just OK, and not a big deal, or maybe we would think of some way to indicate that the other types aren't tabs at all, but just links?
> 
> My take is that /questions /wiki /comments are well established and we should try to make the links not appear as tabs, which is a design challenge but not too tough, i think. Let's see what people think!

I'll create a query to collect all notes.. 

About the tabs issue, what if, instead of having a route and page called 'All Posts' we have a dropdown that has a link to all notes, questions, comments, and wiki.. that way, we won't have an entirely new page that looks similar to the topics page.

Since the 'All Posts' link will be on the Dashboard, we can make it a dropdown instead of a link.. unless it is necessary to have a main page for all the posts.

RuthNjeri:
@jywarren I have just discovered that there's an action for all notes in the NotesController. The page is https://publiclab.org/notes

This might mean that we have the pages that display all: notes, questions, wikis, and comments. The only pending issue is the design of the page, I have shared my suggestion of using a dropdown on the Dashboard page, see above...

jywarren:
Hi @RuthNjeri - 

> @jywarren I have just discovered that there's an action for all notes in the NotesController. The page is https://publiclab.org/notes

Whoa! OK yikes yes! Let's use that, and just refine the design to better match the /tag/____ style. And /notes is a good URL!

> Since the 'All Posts' link will be on the Dashboard, we can make it a dropdown instead of a link..

I guess, to me, it makes sense for the /notes page to easily offer links to the other types, in case you go there and then want to "traverse" to other types. What if we instead of tabs, offered some "pill" links in the same order so it looks familiar in placement and order, but doesn't look like they function as tabs. The row of tabs shown here:

image

Could instead be a pill nav: https://getbootstrap.com/docs/4.4/components/navs/#pills

image

That way we can also keep the dashboard simpler with no dropdown. But, we can try this out and re-evaluate once we show it to people -- it would just take some styling changes on the Dashboard to make changes if people really want the dropdown, for example. How does this sound? 

One thing is, I'm not sure if there are functional tests for the /notes route. We may want to add some simple ones. 

Lastly, while we can get rid of all the sidebar content on /notes, we should consider retaining the "by likes" and "by views" options. Maybe those would go in the "sort by" dropdown we had been wanting to get rid of, since they already exist.

Oh-  one more thing -- http://publiclab.org/notes is really slow. I suspect the queries and pagination generating that page are not well optimized. Does anything stand out to you about them? Maybe, we could try to refine them to be closer to the TagController `show` action, but without filtering for tag/topic? I guess it's these three routes:

https://github.com/publiclab/plots2/blob/cbb807ba8e2302f09dafc0060475aa118e34c2c6/app/controllers/notes_controller.rb#L304-L342

RuthNjeri:
Using the /notes URL and refining it seems like a great starting point. I think redesigning the page makes a lot of sense.. I'll break down the tasks needed for this in the Planning issue to keep track of the updates and push updates on this PR once I complete the first task...

I'll update you once I need a review....

jywarren:
Gosh, actually looking at that code i don't see why it's so inefficient. it seems well limited by timestamp. Ah, i see! `config/routes.rb` shows it's actually the `index` action:

https://github.com/publiclab/plots2/blob/cbb807ba8e2302f09dafc0060475aa118e34c2c6/config/routes.rb#L139

So that's this:

https://github.com/publiclab/plots2/blob/cbb807ba8e2302f09dafc0060475aa118e34c2c6/app/controllers/notes_controller.rb#L6-L9

Leading to this sub-method `set_sidebar` which is really old and probably no longer necessary:

https://github.com/publiclab/plots2/blob/cbb807ba8e2302f09dafc0060475aa118e34c2c6/app/controllers/application_controller.rb#L26-L67

I think we should just skip `set_sidebar` and do it ourselves, since we only need that one query for recent nodes!

jywarren:
Ah sorry was still writing into that last comment. But, sounds great, Ruth! Thank you!!!

ebarry:
Really great detective work in this thread @RuthNjeri ! Thank you for finding that https://publiclab.org/notes exists, and planning an update!

I just consulted the @publiclab/community-reps staff, and **there is agreement that can we work with pill navs to the pages where we already present content by type.** There was excitement that wiki updates would be easier/more consistent to navigate to. Several people said that it was really the comments that they needed to browse chronologically.

RuthNjeri:
Thanks for the clarification @ebarry 😄 

About the need to browse through the comments chronologically, we have the comments route https://publiclab.org/comments where the comments are ordered according to the most recent, is that the chronological order that's needed?

Sagarpreet:
> is that the chronological order that's needed

Correct @RuthNjeri 👍

test/functional/notes_controller_test.rb
```diff
@@ -700,12 +702,14 @@ def teardown
     get :liked
     notes = assigns(:notes)
     expected = [nodes(:one)]
+    # Just that using nodes(:question) is misleading because
+    # the /notes page is designed not to return any question related notes
     questions = [nodes(:question)]
```
RuthNjeri, line 707: same as above

test/functional/notes_controller_test.rb
```diff
@@ -690,6 +690,8 @@ def teardown
     get :recent
     notes = assigns(:notes)
     expected = [nodes(:one)]
+    # Just that using nodes(:question) is misleading because
+    # the /notes page is designed not to return any question related notes
     questions = [nodes(:question)]
```
RuthNjeri, line 695: # Just that using nodes(:question) is misleading because
    # the /notes page is designed not to return any question related notes

app/views/notes/index.html.erb
```diff
@@ -39,13 +39,7 @@
...
 
-<% if params[:controller] == "tag" || params[:controller] == "search" %>
-  <%= render :partial => "sidebar/related" %>
-<% elsif params[:action] == "author" || params[:action] == "author_topic" %>
-  <%= render :partial => "sidebar/author" %>
-<% else %>
-  <%= render :partial => "sidebar/user" %>
-<% end %>
+<%# The sidebar is also used in the admin/queue section,
+which reuses the 'notes/index' template %>
```
RuthNjeri, line 45: The sidebar is also used in the admin/queue section, which reuses the 'notes/index' template, not sure if it should be removed.

app/controllers/notes_controller.rb
```diff
@@ -318,21 +314,15 @@ def liked
...
   def recent
     @title = I18n.t('notes_controller.recent_research_notes')
-    @wikis = Node.limit(10)
-      .where(type: 'page', status: 1)
-      .order('nid DESC')
-    @notes = Node.where(type: 'note', status: 1, created: Time.now.to_i - 1.weeks.to_i..Time.now.to_i)
-                 .order('created DESC')
-    @unpaginated = true
+    # Not sure why it was unpaginated before
+    @pagy, @notes = pagy(Node.where(type: 'note', status: 1, created: Time.now.to_i - 1.weeks.to_i..Time.now.to_i)
```
RuthNjeri, line 318: Not sure why it was unpaginated before

app/views/notes/index.html.erb
```diff
@@ -39,13 +39,5 @@
   <%= render :partial => "notes/notes" %>
 
   <hr />
-  
 </div>
```
RuthNjeri, line 43: The sidebar is also used in the admin/queue section, which reuses the 'notes/index' template, not sure if it should be removed.

test/functional/notes_controller_test.rb
```diff
@@ -705,7 +705,7 @@ def teardown
     assert !(notes & questions).present?
```
RuthNjeri, line 705: Just that using nodes(:question) is misleading because
the /notes page is designed not to return any question related notes

RuthNjeri: on line 693 to 695 as well.

RuthNjeri:
Hi! This is ready for review. 😄

app/views/notes/index.html.erb
RuthNjeri: It also removes the write research note button
Screenshot 2020-12-09 at 22 32 57

RuthNjeri:
> Hi! This is ready for review. 😄

Tagging @jywarren, @Sagarpreet, and @cesswairimu.. this is ready for review..

cesswairimu (approved):
Fantastic @RuthNjeri tada: :tada:

Great great job..love the loads of refactoring  and cleanup you did here :rocket: :rocket:
Thanks

cesswairimu:
On *popular* and *liked* notes I feel like the website has a lot of data maybe we should increase the limit to 50 or even 100(we are currently showing top 20) and paginate them. What do people think?


Also @RuthNjeri maybe you could make the checklists into issues and link them back to the planning issue maybe smth like what Renisha did here https://github.com/publiclab/plots2/issues/6887#issuecomment-561287047 so that when reviewing its easier to know what part of the project you are working on..also linking the planning issue on PRs will be closed each time we merge and I believe we would like to keep that opened until the project is done. If you have any other methods for doing this please feel free to use them. Thanks and awesome work :tada:

app/controllers/application_controller.rb
```diff
@@ -38,13 +38,13 @@ def set_sidebar(type = :generic, data = :all, args = {})
       @maps = Tag.find_nodes_by_type(data, 'map', 20)
     else # type is generic
       # remove "classroom" postings; also switch to an EXCEPT operator in sql, see https://github.com/publiclab/plots2/issues/375
-      hidden_nids = Node.where(type: :note, status: 1).select { |n| n.has_a_tag('hidden:response') }.collect(&:nid)
+      hidden_nids = Node.hidden_response_nids
```
Sagarpreet, line 41: Nice 👍

app/controllers/application_controller.rb
```diff
@@ -38,13 +38,13 @@ def set_sidebar(type = :generic, data = :all, args = {})
...
       @notes = if params[:controller] == 'questions'
                  Node.questions
                    .joins(:revision)
                else
                  Node.research_notes.joins(:revision).order('node.nid DESC').paginate(page: params[:page])
-      end
+               end
```
Sagarpreet, line 47: Indentation?

app/controllers/notes_controller.rb
```diff
@@ -318,24 +313,16 @@ def liked
...
   end
 
   # notes with high # of views
   def popular
     @title = I18n.t('notes_controller.popular_research_notes')
-    @wikis = Node.limit(10)
-      .where(type: 'page', status: 1)
-      .order('nid DESC')
-    @notes = Node.research_notes
+    @notes = published_notes
```
Sagarpreet, line 324: `.limit(20)` will take the first 20 records?
Do we need to sort the data in descending order of Node id (nid)? So we are sure we take latest 20 nodes and then find the popular ones?

app/controllers/notes_controller.rb
```diff
@@ -461,4 +448,15 @@ def location
...
+
+  def published_notes
+    hidden_nids = Node.hidden_response_nids
+    notes = Node.research_notes.where('node.status = 1')
+
+    if hidden_nids.empty?
+      notes
+    else
+      notes.where('node.nid NOT IN (?)', hidden_nids)
+    end
```
Sagarpreet, line 460: notes.where('node.nid NOT IN (?)', hidden_nids)
```suggestion
 notes.where('node.nid NOT IN (?)', hidden_nids)
```

app/models/node.rb
```diff
@@ -124,7 +124,11 @@ def slug_from_path
   end
 
   def has_a_tag(name)
-    return tags.where(name: name).size.positive?
+    tags.where(name: name).size.positive?
+  end
+
+  def self.hidden_response_nids
```
Sagarpreet, line 130: Maybe use `hidden_response_node_ids `?

app/models/node.rb
```diff
@@ -920,8 +924,14 @@ def self.research_notes
...
-    Node.where(type: 'note')
-                .where('node.nid NOT IN (?)', nids)
+    notes = Node.where(type: 'note')
+
+    # if the nids are empty, it should not be used in the query
+    if nids.empty?
+      notes
+    else
+      notes.where('node.nid NOT IN (?)', nids)
+    end
```
Sagarpreet, line 934: `notes.where('node.nid NOT IN (?)', nids)` ?

app/models/node.rb
```diff
@@ -920,8 +924,14 @@ def self.research_notes
                .group('node.nid')
                .collect(&:nid)
 
-    Node.where(type: 'note')
-                .where('node.nid NOT IN (?)', nids)
+    notes = Node.where(type: 'note')
+
+    # if the nids are empty, it should not be used in the query
```
Sagarpreet, line 929: Can you also write why we donot have to use `nids` in comment?

app/views/wiki/_wikis.html.erb
```diff
@@ -29,7 +29,7 @@
 <% end %>
 
 <% if @pagy %>
-  <%== pagy_bootstrap_nav @pagy %>
+  <%= pagy_bootstrap_nav @pagy %>
```
Sagarpreet, line 32: Good catch 🎉

Sagarpreet (commented):
Great work 💯

app/controllers/application_controller.rb
RuthNjeri: Thanks!

app/controllers/notes_controller.rb
RuthNjeri: I noticed that when the hidden_nids are empty, the query does not return the rest of the notes.
The previous query was `notes.where('node.nid NOT IN (?)', hidden_nids) unless hidden_nids`

app/models/node.rb
RuthNjeri: That makes sense, I'll modify it.

app/models/node.rb
RuthNjeri: sure!

app/controllers/application_controller.rb
RuthNjeri: Yes, to match the if...

app/controllers/notes_controller.rb
```diff
@@ -318,24 +313,16 @@ def liked
...
 
   # notes with high # of views
   def popular
     @title = I18n.t('notes_controller.popular_research_notes')
-    @wikis = Node.limit(10)
-      .where(type: 'page', status: 1)
-      .order('nid DESC')
-    @notes = Node.research_notes
+    @notes = published_notes
       .limit(20)
```
cesswairimu, line 325: seems like here the `.order` should come before the `limit`

RuthNjeri: Yes, good catch, I'll add that

app/controllers/notes_controller.rb
RuthNjeri: No limit(20) will not automatically take the first 20 records, I should sort the order then limit..you are right..let me change that..

app/controllers/notes_controller.rb
RuthNjeri: I've just realized that the order method is from the [@wiki ](https://github.com/publiclab/plots2/pull/8800/files/c5954b67e9b5c1e72d6e6192a73883f167362c87#diff-c254740097a8c0601832077488db77e572b2c2b926f77508fadc6081b0052804L333)query, but it's true that the order should come before the limit in order to get the most recent ones which is also what @Sagarpreet has suggested.

app/models/node.rb
RuthNjeri: thanks, see comment [here](https://github.com/publiclab/plots2/pull/8800/files/c5954b67e9b5c1e72d6e6192a73883f167362c87#r540164441) on the same

app/controllers/notes_controller.rb
RuthNjeri: Screenshot 2020-12-10 at 17 19 56

RuthNjeri:
> On _popular_ and _liked_ notes I feel like the website has a lot of data maybe we should increase the limit to 50 or even 100(we are currently showing top 20) and paginate them. What do folks think about this? @ebarry @jywarren @sagarpreet-chadha
> 
> Also @RuthNjeri maybe you could make the checklists into issues and link them back to the planning issue maybe smth like what Renisha did here [#6887 (comment)](https://github.com/publiclab/plots2/issues/6887#issuecomment-561287047) so that when reviewing its easier to know what part of the project you are working on..also linking the planning issue on PRs will be closed each time we merge and I believe we would like to keep that opened until the project is done. If you have any other methods for doing this please feel free to use them. Thanks and awesome work 🎉

Thanks @cesswairimu , I'll break down the checklists into issues, it makes much more sense.
I'll wait for @ebarry @jywarren @sagarpreet-chadha to chip in on the limit suggestion as I work on the rest of the feedback.

app/models/node.rb
RuthNjeri: Done!

app/models/node.rb
RuthNjeri: Done!

RuthNjeri:
@Sagarpreet , @cesswairimu about the limit and order methods for the popular which also affects recent and the liked actions..

I've noticed that when I first order all the queries in descending order and then order the cached_liked in DESC as well, the notes are displayed according to the most recent only.

If an older note has the highest like, it will still come after the most recent one..
Screenshot 2020-12-10 at 18 10 38
Screenshot 2020-12-10 at 18 10 23

app/controllers/notes_controller.rb
RuthNjeri: I've added the order in the [published_notes](https://github.com/publiclab/plots2/pull/8800/files#diff-c254740097a8c0601832077488db77e572b2c2b926f77508fadc6081b0052804R454) method

test/functional/notes_controller_test.rb
```diff
@@ -692,7 +692,7 @@ def teardown
     expected = [nodes(:one)]
     questions = [nodes(:question)]
     assert (notes & expected).present?
-    assert (notes & questions).present?
+    assert !(notes & questions).present?
```
RuthNjeri, line 695: This was a failing test before...

app/controllers/notes_controller.rb
```diff
@@ -461,4 +448,15 @@ def location
 
     [latitude, longitude]
   end
+
+  def published_notes
+    hidden_nids = Node.hidden_response_node_ids
+    notes = Node.research_notes.where('node.status = 1').order('node.nid DESC')
```
cesswairimu, line 454: @RuthNjeri I think it could be its because you are already doing an order by nid on `published_notes` on line 454, maybe we could remove the order here and just do it on the methods...what do you think?

RuthNjeri: Hey @cesswairimu in this [version](https://github.com/publiclab/plots2/pull/8800#issuecomment-742583719) (the screenshot), I did that, I only pushed the current one (line 454) after I tested it.. the results are the same.. 

I've attached other screenshots..

Screenshot 2020-12-10 at 20 41 46

Screenshot 2020-12-10 at 20 42 05

Screenshot 2020-12-10 at 20 42 13

RuthNjeri: Maybe the node.nid order takes precedence?

cesswairimu: I believe what the prob is the double `.order`...we could try (with your `published_notes` method as the above screenshot )
```
def liked
    @title = I18n.t('notes_controller.highly_liked_research_notes')
    @notes = published_notes
             .order('cached_likes DESC').limit 20
    @unpaginated = true
    render template: 'notes/index'
  end
```
I don't think we need to order by latest on most liked...then you can have the order on `recent` what do you think?

cesswairimu: Also just seen this for multiple ordering...maybe you could try it out `.order(cached_likes: :desc, nid: :desc)`

RuthNjeri: Thanks @cesswairimu let me run it in the debugger and share the results

RuthNjeri: For this  `I don't think we need to order by latest on most liked...then you can have the order by nid on recent what do you think?` It makes sense, though with the limit of 20 nodes, I think it defaults in picking the earliest 20 nodes...

jywarren:
Just chiming in to say that this is looking great, and I appreciate all the collab and discussion and thought going on. Some refinements like pagination could be done in a follow-up PR if you prefer, but it's your choice! In general we love to do things in discrete smaller PRs but if you're deep into it already, no worries! Thanks, everyone!

Also, I noted some issues with running the tests in GitPod in [this comment](https://github.com/publiclab/plots2/pull/8801#issuecomment-742730049) so i wonder if the same is true here-- but until our Travis tests are fixed in https://github.com/publiclab/plots2/pull/8795 we may need to rely on manually running tests in the standard GitPod environment as a substitute, so if it's possible to confirm that they have passed as a final step before we mark this (and other) PRs for review, that would be great until we finally complete https://github.com/publiclab/plots2/pull/8795. Thank you for your patience with this!!!

app/controllers/notes_controller.rb
RuthNjeri: @cesswairimu the query you shared works as expected 😄 , I think the issue was with the way the arguments within the order params were positioned.. in the first query it was
`Node.research_notes.where('node.status = 1').order('node.nid DESC').order('cached_likes DESC')`
The node.nid order is positioned first so the default order will be according to the` node.nid DESC`

And the second one that you've suggested `Node.research_notes.where('node.status = 1').order(cached_likes: :desc, nid: :desc)`
The node.nid order is positioned second.. and it prioritizes the cached_likes

Screenshot 2020-12-10 at 22 09 27

(the order is based on the most recent with the highest number of likes)


I've tested it with the same query you shared but switched up the position and got the same original problem,
`Node.research_notes.where('node.status = 1').order(nid: :desc, cached_likes: :desc, )`

Screenshot 2020-12-10 at 22 11 26

(the order is based on the most recent and not exactly on the highest number of likes)... still confusing to me but so far this is the only thing that makes sense..


It's now fixed 🎉 , Thanks for the help @cesswairimu 😄

RuthNjeri:
Hi, @jywarren 😄  I'm okay, waiting for the tests to work, I've fixed the order issue and added the limit number.. 
In the meantime, I can begin working on the design from another branch..

app/controllers/notes_controller.rb
RuthNjeri: This has been fixed :)

RuthNjeri:
I've confirmed that all tests running in the `notes_controller_test` are all passing.

app/controllers/notes_controller.rb
Sagarpreet: Awesome 🎉 
I had faced something similar before, so it immediately came to my notice 😄

Sagarpreet (approved):

RuthNjeri:
@jywarren this PR has all tests ready to be merged. The `node_test.rb` and `notes_controller_test.rb` files are all passing.

Please have a look at the hidden_nids query and a test I've just updated [here](https://github.com/publiclab/plots2/pull/8800/files#diff-6997c2ce770b648ac2c34043b6b33b6f659fc3aa24fa82fc6afd7f4ed05c7805R511) to know if it's good to merge or needs a few more revisions... thanks!

app/controllers/notes_controller.rb
jywarren: aha - were you able to fix that behavior? I think you're right that it's incorrect. Thanks!

app/models/node.rb
```diff
@@ -123,8 +123,13 @@ def slug_from_path
     path.split('/').last
   end
 
-  def has_a_tag(name)
-    return tags.where(name: name).size.positive?
+  def self.hidden_response_node_ids
+    Node.joins(:node_tag)
+        .joins('LEFT OUTER JOIN term_data ON term_data.tid = community_tags.tid')
+        .select('node.nid')
```
jywarren, line 129: Looks great! However i think this may need more selected fields -- https://guides.rubyonrails.org/active_record_querying.html#selecting-specific-fields says it actually changes the `SELECT X` section of the SQL query, so we may need all those used in the joins and where fields.

I read your response about this somewhere, i think... but can't find it now? Maybe you deleted it, or was it in Gitter? gosh, my brain isn't working so clearly right now as I'm deep in this Travis thing... sorry!

Just want to confirm this here though after checking the Rails docs!

jywarren:
Ah, and regarding the CodeClimate recommendations, if you click on the CodeClimate check it'll make suggestions, and if you want to ignore them (which is ok! Your judgement!) i can "skip" them, so just tell me if that's what you prefer! Thanks!

app/models/node.rb
RuthNjeri: It was in my other [PR](https://github.com/publiclab/plots2/pull/8848#discussion_r543489655) 😄 

Thanks, I'll have a look at the doc...and update the PR.

RuthNjeri:
> Ah, and regarding the CodeClimate recommendations, if you click on the CodeClimate check it'll make suggestions, and if you want to ignore them (which is ok! Your judgement!) i can "skip" them, so just tell me if that's what you prefer! Thanks!
> 
> (actually i went ahead and bypassed them just to make things simpler for the next step, but if you want you can still address them and i think they're still listed!)

It's okay to ignore the CodeClimate issues, I don't think I can solve them...

RuthNjeri:
The current test failing here is also failing in the Main branch.. https://github.com/publiclab/plots2/runs/1561451524

app/controllers/notes_controller.rb
RuthNjeri: Currently, the if..else statement works as expected.. if we put this line back 
`notes.where('node.nid NOT IN (?)', hidden_nids) unless hidden_nids` we'll also have to specify what happens if the hidden_nids are empty.. maybe 
```suggestion
   return  notes.where('node.nid NOT IN (?)', hidden_nids) unless hidden_nids
   return  notes if hidden_nids.empty?
```
Which is still similar to the current code...unless we should change it...

RuthNjeri:
The force-push is due to the recent rebase...

app/controllers/notes_controller.rb
Tlazypanda: Hey @RuthNjeri This looks amazing!! :tada: :clap: 

Just noting here can we also check for the performance for line - 
`notes = Node.research_notes.where('node.status = 1')` and then again querying within this depending on `hidden_nids`. If they aren't empty this query may take more time for execution. 

For handling this problem with `notes.where('node.nid NOT IN (?)', hidden_nids)` can we try this once with `where.not` and see if it works? something like `notes.where.not(node.nid: hidden_nids)`

RuthNjeri: Hi @Tlazypanda Thanks so much for the suggestion! It's makes the code so much cleaner!

I've combined the query you've suggested and the notes one into one line
```suggestion
    Node.research_notes
        .where('node.status = 1')
        .where.not(nid: hidden_nids)
```

I'm currently testing it and it seems that it's working.. pushing it in a few... it might also solve the performance issue of running two queries separately?

jywarren:
OK, this should be good now if rebased. Thank you for your patience!!!

RuthNjeri:
> OK, this should be good now if rebased. Thank you for your patience!!!

Thanks @jywarren, we should actually merge this[ PR](https://github.com/publiclab/plots2/pull/8800) first, it has all the up to date changes and all the feedback implemented.

This current PR still has some pending feedback I need to implement.

RuthNjeri:
Force pushed due to recent rebase with the main branch

RuthNjeri:
Hi @jywarren, it seems that the database setup is failing when the build is running...
Screenshot 2020-12-16 at 22 04 51

I'm not sure why it's failing...

jywarren:
Strange, ok I'm actually seeing this integration test fail:

```
FAIL["test_should_choose_i18n_for_sidebar/_author_+_sidebar/_post_button", #<Minitest::Reporters::Suite:0x000055bd75e04178 @name="I18nTest">, 129.59562038299998]
 test_should_choose_i18n_for_sidebar/_author_+_sidebar/_post_button#I18nTest (129.60s)
        Expected at least 1 element matching "h4", found 0..
        Expected 0 to be >= 1.
        test/integration/I18n_test.rb:379:in `block (2 levels) in <class:I18nTest>'
        test/integration/I18n_test.rb:365:in `each'
        test/integration/I18n_test.rb:365:in `block in <class:I18nTest>'

=============|
```

I restarted to see if it's intermittent... Any ideas otherwise?

app/views/notes/index.html.erb
```diff
@@ -39,13 +39,5 @@
...
 
-<% if params[:controller] == "tag" || params[:controller] == "search" %>
-  <%= render :partial => "sidebar/related" %>
-<% elsif params[:action] == "author" || params[:action] == "author_topic" %>
-  <%= render :partial => "sidebar/author" %>
-<% else %>
-  <%= render :partial => "sidebar/user" %>
-<% end %>
```
cesswairimu, line 51: Hey @RuthNjeri the removal of this section is the one causing the tests to fail. Also I  don't think  we want to remove this sidebar...it gives access to recent wiki edits ...You might also want to add back the `set_sidebar` under  index in the controller. what do you think?

RuthNjeri: Thanks @cesswairimu 😄 

The removal of that section was part of this [discussion](https://github.com/publiclab/plots2/pull/8800#issuecomment-740052184)  with @jywarren where we replace it with the sort dropdown for the notes...

I think a user will still be able to access the recent wiki edits via the wiki section that will be linked on the notes page, I am working on this [here](https://github.com/publiclab/plots2/pull/8848), therefore, access to the recent edits won't be completely lost, just changed to another page....

I think this can also be open to further discussion before the PR is merged..

RuthNjeri:
> Strange, ok I'm actually seeing this integration test fail:
> 
> ```
> FAIL["test_should_choose_i18n_for_sidebar/_author_+_sidebar/_post_button", #<Minitest::Reporters::Suite:0x000055bd75e04178 @name="I18nTest">, 129.59562038299998]
>  test_should_choose_i18n_for_sidebar/_author_+_sidebar/_post_button#I18nTest (129.60s)
>         Expected at least 1 element matching "h4", found 0..
>         Expected 0 to be >= 1.
>         test/integration/I18n_test.rb:379:in `block (2 levels) in <class:I18nTest>'
>         test/integration/I18n_test.rb:365:in `each'
>         test/integration/I18n_test.rb:365:in `block in <class:I18nTest>'
> 
> =============|
> ```
> 
> I restarted to see if it's intermittent... Any ideas otherwise?

Hi @jywarren the test is related to the sidebar that I removed, so I've removed the test...though there's a discussion happening above whether the sidebar should be removed..

app/views/notes/index.html.erb
cesswairimu: Oh okay, then I think we might need to rewrite that test or remove it

jywarren:
No I think that's correct to remove the sidebar and the test, thank you! This looks awesome and it's so comforting to see the green All checks have passed ✅ message 😅

Merging now! Congratulations!!!! Thanks everyone!!!

jywarren:
This should shortly be visible at stable.publiclab.org!!!

RuthNjeri:
Nice! The notes URL is loading a bit faster now 😄

jywarren:
Awesome to see!

PR #8812, opened 2020-12-09
Hotfix for email sending in production

### LINKED ISSUE #8811: Emails failing to send due to action mailer error

Trying to trigger email sending causes 500 error.
Sentry reports:
```
ArgumentError: wrong authentication type 
  app/controllers/users_controller.rb:261:in `reset'
    PasswordResetMailer.reset_notify(user, key).deliver_now unless user.nil? # respond the same to both successes and failures; security
...
(114 additional frame(s) were not displayed)
```
Possibly the `:authentication` parameter in `production.rb` can't be blank string?

Will attempt hotfix in production by commenting out the `:authentication` line.

icarito:
Fixes #8811 

This fix is not ideal it only comments out the other action mailer parameters because leaving them empty didn't work and caused mails to not be sent. This was tested as a hotfix in production, and I will merge it in order to sync with production.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -69,10 +69,10 @@ Plots2::Application.configure do
   config.action_mailer.smtp_settings = {
   :address => ENV["SMTP_HOST"] || "smtp",
   :port => ENV["SMTP_PORT"] || 25,
-  :user_name => ENV["SMTP_USER"] || '',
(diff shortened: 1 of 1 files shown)


icarito:
Merging to reflect hotfix in production.

icarito:
This is merged and also pushed it to `stable` branch.

Issue #8830, opened 2020-12-11
Implement Mobile Responsiveness on The All Posts Page


RuthNjeri:
This is part of the Dashboard implementation #8513

It will:
- [ ] Implement/Ensure mobile responsiveness in the 'All Posts' page
- [ ] Implement/Ensure mobile responsiveness in the Research notes section
- [ ] Implement/Ensure mobile responsiveness in the Questions section
- [ ] Implement/Ensure mobile responsiveness in the Wiki pages section
- [ ] Implement/Ensure mobile responsiveness in the  Comment section

RuthNjeri:
I think, for now, the all posts page `/notes` is responsive..
The other pages(/comments, /questions, /wiki) I suspect will require a redesign, I think responsiveness will be handled in each individual page redesign, what do you think @jywarren ?

Sagarpreet:
I like the idea 💯

RuthNjeri:
Thanks @Sagarpreet! I'll close this issue.

PR #8848, opened 2020-12-14
[Outreachy] All Posts Design

### LINKED ISSUE #8829: All Posts Design
This issue is part of https://github.com/publiclab/plots2/issues/8513

It will fix:
- [ ] Implement the design for the 'All Posts' page (/notes)..refer to these [notes](https://github.com/publiclab/plots2/pull/8800#issuecomment-740052184).

RuthNjeri:
Fixes #8829 and #8830

* [X] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [X] code is in uniquely-named feature branch and has no merge conflicts 📁
* [X] screenshots/GIFs are attached 📎 in case of UI updation

Screenshot 2020-12-17 at 17 26 54
Screenshot 2020-12-17 at 17 27 10
Screenshot 2020-12-17 at 17 27 35
Screenshot 2020-12-17 at 17 27 46
Screenshot 2020-12-17 at 17 27 58


sagarpreet-chadha (commented):
Code wise everything looks perfect :)
Design wise maybe the dropdown in mobile view leaves a white right margin which can be filled similar to how we see in social media apps.

RuthNjeri:
Thanks @sagarpreet-chadha! Just so I know I've understood, your suggestion is to have a right margin from the dropdown? By maybe pushing the dropdown to the left that will leave some white space on the right?

jywarren:
Hi, this is looking great. One thing we do on the /tag/outreachy - style pages is to use a Bootstrap style to hide the tab text when on narrower screens:

https://github.com/publiclab/plots2/blob/ff3417ab1ad965753a3a159b905a59b2f244b8ea/app/views/tag/show/_nav_tabs.html.erb#L6-L14

Maybe we can do the same here to get the buttons to line up on one line even on narrow screens?

Thanks @RuthNjeri ! This is awesome! Should we be looking to merge anything else before this from your project? 

Great work!

test/functional/notes_controller_test.rb
```diff
@@ -692,7 +690,7 @@ def teardown
     expected = [nodes(:one)]
     questions = [nodes(:question)]
     assert (notes & expected).present?
-    assert (notes & questions).present?
+    assert !(notes & questions).present?
```
jywarren, line 693: nice!

test/functional/notes_controller_test.rb
```diff
@@ -397,8 +396,7 @@ def teardown
 
     assert_response :success
     selector = css_select 'div.note'
-    assert_equal 27, selector.size
-    assert_select 'a[data-test="spam"]','Spam'
+    assert_equal 15, selector.size
```
jywarren, line 399: excellent, due to no longer displaying spam posts!

test/functional/notes_controller_test.rb
```diff
@@ -368,8 +368,7 @@ def teardown
 
     assert_response :success
     selector = css_select 'div.note'
-    assert_equal 27, selector.size
-    assert_select "div p", 'Pending approval by community moderators. Please be patient!'
+    assert_equal 15, selector.size
```
jywarren, line 371: nice!

app/views/notes/index.html.erb
```diff
@@ -15,37 +15,28 @@
...
-    <%= render :partial => "notes/format_toggle" %>
-    <% if params[:action] == "liked" %><a style="margin-right:4px;" class="float-right btn btn-outline-secondary btn-sm" href="/feed/liked.rss"><i class="fa fa-rss"></i></a><% end %>
+  <br/>
+  <% if params[:action] == "liked" %><a style="margin-right:9px;" class="float-right btn btn-outline-secondary btn-sm" href="/feed/liked.rss"><i class="fa fa-rss"></i></a><% end %>
 
-    <ul class="nav nav-tabs" id="tab-bar">
+    <ul class="nav nav-pills">
       <li class="nav-item">
-        <a class="nav-link <% if params[:action] == "recent" %> active <% end %>" href="/notes/recent"> <i class="fa fa-clock-o"></i><span class="hidden-sm hidden-xs"> <%= translation('notes.index.recent') %></span></a>
+        <a class="nav-link <% if params[:controller] == "notes" %> active <% end %>" href="/notes/" target="_blank"> <i class="fa fa-file"></i><span class="hidden-sm hidden-xs"> <%= translation('notes.index.research_notes') %></span></a>
```
jywarren, line 23: Should these be target blank? Perhaps we could have them be regular links? but i'm very open to ideas if you'd like to share your reasoning! Thanks!

app/models/node.rb
```diff
@@ -124,7 +124,11 @@ def slug_from_path
   end
 
   def has_a_tag(name)
-    return tags.where(name: name).size.positive?
+    tags.where(name: name).size.positive?
+  end
+
+  def self.hidden_response_node_ids
```
jywarren, line 130: I really like this! I'm a little worried about performance. Here we are running some Ruby against every `status = 1` note, right? Take a look at this segment; it's not perfect (it also has to collect /all/ hidden nids before we can exclude them, so years from now if there are thousands, that's a problem), but it does build the condition of having the `hidden:response` tag into the initial query so we aren't running `has_a_tag` on all `status = 1` nodes  in the whole database:
 
https://github.com/publiclab/plots2/blob/444fff0f84af196bb4021c50f68f9ae248ea6449/app/controllers/home_controller.rb#L70-L81

That'll get us to at least as efficient as we were before. Later, it may be worth trying to do more, but since we won't be hitting this query as often (since we're moving it off the dashboard) we don't have to stress too much. I can think of a couple ways to optimize further (either an even more complex SQL query to do it all in one query or... maybe... searching for `hidden:response` nids only within the timeframe of the current set of nodes?) but basically i think we shouldn't worry more about it for now. 😅 

Thanks, Ruth!!!

jywarren (changes requested):

app/models/node.rb
jywarren: Oh and one more thing -- often if we add a model method, like `node.hidden_response_node_ids`, we add at least a very minimal unit test to protect it, just to confirm that it is providing the most basic response of returning an array of integers, maybe. Could be worthwhile here!

jywarren:
Awesome work, @RuthNjeri - this is very exciting and we're almost done with this section!

jywarren:
Ah, and it looks like one of the tests changed conflicted with a change from @noi5e -- the notes_controller_test, but I hope it'll be an easy one to resolve. Please reach out if it causes you any trouble though! 🙌

RuthNjeri:
> Hi, this is looking great. One thing we do on the /tag/outreachy - style pages is to use a Bootstrap style to hide the tab text when on narrower screens:
> 
> https://github.com/publiclab/plots2/blob/ff3417ab1ad965753a3a159b905a59b2f244b8ea/app/views/tag/show/_nav_tabs.html.erb#L6-L14
> 
> Maybe we can do the same here to get the buttons to line up on one line even on narrow screens?
> 
> Thanks @RuthNjeri ! This is awesome! Should we be looking to merge anything else before this from your project?
> 
> Great work!

Thanks! This [PR](https://github.com/publiclab/plots2/pull/8800) would need to be merged first, but due to this [feedback](https://github.com/publiclab/plots2/pull/8848#discussion_r542735439) I'll have to update it before merging. I'll let you know once it's ready to be merged.

app/views/notes/index.html.erb
RuthNjeri: Good catch, the `research_notes` link should not be target blank.. but I think for the rest of the links it makes sense for them to be since once you click on either `question, wikis, or comments` there's no way of going back to the notes section from those pages... unless you go back to the dashboard then click the 'all posts' link which is an extra step for the user in loading and reloading the pages... if this is not a problem, I can remove the target blank from the rest of the links too.

app/models/node.rb
RuthNjeri: Thanks, your suggestion to optimize makes sense, it's just occurred to me that the query with the `has_a_tag` method is an N+1 query.. I'll have a look at the one you've shared and use it. I'll be sure to write some tests for it as well.

RuthNjeri: Can we also update the query from:
```
Node.joins(:node_tag) 
   .joins('LEFT OUTER JOIN term_data ON term_data.tid = community_tags.tid') 
   .select('node.*, term_data.*, community_tags.*') 
   .where(type: 'note', status: 1) 
   .where('term_data.name = (?)', 'hidden:response') 
   .collect(&:nid) 
```
To:

```
 Node.joins(:node_tag) 
   .joins('LEFT OUTER JOIN term_data ON term_data.tid = community_tags.tid') 
   .select('node.nid') 
   .where(type: 'note', status: 1) 
   .where('term_data.name = (?)', 'hidden:response') 
   .collect(&:nid)
```

Since the select statement `.select('node.*, term_data.*, community_tags.*') ` gets all the records and we only collect the nids

jywarren: Ah yes, i think that'll work! I forget, does it automatically add conditional columns or do we need to add "type" and "status"?

jywarren: Yes i think we may need to add any column we use in the query... But good idea!

app/views/notes/index.html.erb
jywarren: I think we could go with normal links for all, but i see your point too. Maybe we can brainstorm a bit on the UI design for this in follow-up (like, could we put a compact way to navigate back on the other pages? not sure.); i think your solution is good too, but we needn't let this hold us up. Thanks for thinking carefully and being flexible here!

app/models/node.rb
RuthNjeri: I thought the purpose of the select statement is to determine what is returned? I could be wrong but I tested it with this query that selects the `node.flag`...
Screenshot 2020-12-15 at 19 21 37


It was able to return two nodes I have locally with the tag name `test` and it only returned the node.flag and did not return the node.nid because I did not select it...

RuthNjeri: Hey @jywarren, I've updated the changes for the query and added a new test in this [PR](https://github.com/publiclab/plots2/pull/8800) which should be the first to be merged.

app/views/notes/index.html.erb
RuthNjeri: Alright, I'll remove the target blank and we can follow it up more in the future..

jywarren:
OK, one more rebase and we should be all good. Thank you so much for your patience!!!

RuthNjeri:
> OK, one more rebase and we should be all good. Thank you so much for your patience!!!

Thanks @jywarren! We should actually merge this [PR](https://github.com/publiclab/plots2/pull/8800) first, it has all the feedback implemented. 

This current PR still has some feedback I need to implement.

RuthNjeri:
Hi @sagarpreet-chadha I added some responsive styling on the navbar and the sort dropdown, let me know if it looks much better..
Screenshot 2020-12-17 at 17 26 54
Screenshot 2020-12-17 at 17 27 35
Screenshot 2020-12-17 at 17 27 58

jywarren:
I think this is good for now! I guess I wonder if a follow-up could show
all the buttons on one line, which may be a different bootstrap style, but
it's not essential for this pr unless you want to.

What's left to complete here before we merge? 🎉🎉🎉

RuthNjeri:
Hi @jywarren I tried the inline style but on the mobile responsiveness for the Galaxy Fold it was overlapping with the Sort Dropdown..I can look at it again tomorrow and let you know... I'll fix the conflicting files right now...

jywarren:
Ah ok and sure, that seems like good reason not to do more on the
tabs/buttons in this pr as well! Thanks for being so thorough!!!

jywarren:
This is great. I guess that means we are good to go to merge now? Thanks Ruth! Just ping me in the chatroom if you are ready to go and we can wrap this one up!

RuthNjeri:
Hi @jywarren, I've added a responsive design for the inline issue I had with the Galaxy Fold.. 
The navbar is inline apart from the Galaxy Fold.. 

Once the tests pass, it should be good to go 🚀 

Current Responsive Displays
Screenshot 2020-12-18 at 15 58 49

Screenshot 2020-12-18 at 15 58 34

Galaxy Fold Before

Screenshot 2020-12-18 at 16 02 44


Galaxy Fold After
Screenshot 2020-12-18 at 15 59 19

jywarren:
Fantastic!!!! Hooray!!!

jywarren (approved):

cesswairimu:
:tada: :tada: :tada:

PR #8251, opened 2020-08-04
Set sql_mode for MySQL version in Gitpod

icarito:
Fixes #8177  (<=== Add issue number here)

Make sure these boxes are checked before your pull request (PR) is ready to be reviewed and merged. Thanks!

* [ ] PR is descriptively titled 📑 and links the original issue above 🔗
* [ ] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [ ] code is in uniquely-named feature branch and has no merge conflicts 📁
* [ ] screenshots/GIFs are attached 📎 in case of UI updation
* [ ] ask `@publiclab/reviewers` for help, in a comment below

> We're happy to help you get this ready -- don't be afraid to ask for help, and **don't be discouraged** if your tests fail at first!

If tests do fail, click on the red `X` to learn why by reading the logs.

Please be sure you've reviewed our contribution guidelines at https://publiclab.org/contributing-to-public-lab-software 

Thanks!

PR #8856, opened 2020-12-15
Try setting ONLY_FULL_GROUP_BY in mysql to get gitpod tests running

jywarren:
Rework of #8251 building on success in #8795!

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -5,6 +5,7 @@ ports:
       onOpen: open-preview
 tasks:
     - init: >
+        gem install bundler &&
         bundle install --without production &&
         cp config/database.yml.gitpod config/database.yml &&
         mysql -uroot -proot -e "SET @@global.sql_mode=(SELECT REPLACE(@@global.sql_mode, 'ONLY_FULL_GROUP_BY', ''));" &&
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -5,6 +5,7 @@ ports:
       onOpen: open-preview
 tasks:
     - init: >
+        export CHROME_BIN=/usr/bin/chromium-browser &&
         gem install bundler &&
         bundle install --without production &&
         cp config/database.yml.gitpod config/database.yml &&
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -18,6 +18,7 @@ tasks:
       command: >
         passenger start 
         
+    - command: chromium-browser --no-sandbox
     - command: redis-server
     - command: rails c
     
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -7,6 +7,7 @@ tasks:
     - init: >
         bundle install --without production &&
         cp config/database.yml.gitpod config/database.yml &&
+        mysql -uroot -proot -e "SET @@global.sql_mode=(SELECT REPLACE(@@global.sql_mode, 'ONLY_FULL_GROUP_BY', ''));" &&
         mysql -e "CREATE DATABASE plots;" &&
         cp db/schema.rb.example db/schema.rb &&
         rake db:setup &&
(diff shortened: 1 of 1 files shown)


jywarren:
```

    - init: >
        bundle install --without production &&
        cp config/database.yml.gitpod config/database.yml &&
        mysql -e "CREATE DATABASE plots;" &&
        cp db/schema.rb.example db/schema.rb &&
        rake db:setup &&
        yarn install

rails test test
```

this is working!! init tasks arent running but manually the tests run...

need something for chromedriver though...

jywarren:
And we'll have to get Redis running or skip those -- just like in #8795 !

jywarren:
Ah it looks like bundler is no longer included in default GitPod and this may be why those steps had been failing. I'll add a line for installing it!

jywarren:
Gitpod chrome tips here:

* https://github.com/JesterOrNot/Gitpod-Chrome/blob/master/.gitpod.yml
* https://github.com/publiclab/Leaflet.DistortableImage/pull/750/files

.gitpod.yml
```diff
@@ -15,6 +18,7 @@ tasks:
       command: >
         passenger start 
         
+    - command: chromium-browser --no-sandbox
```
jywarren, line 21: this didn't do anything, maybe not needed?

.gitpod.yml
```diff
@@ -5,8 +5,11 @@ ports:
       onOpen: open-preview
 tasks:
     - init: >
+        export CHROME_BIN=/usr/bin/chromium-browser &&
+        gem install bundler &&
         bundle install --without production &&
         cp config/database.yml.gitpod config/database.yml &&
+        mysql -uroot -proot -e "SET @@global.sql_mode=(SELECT REPLACE(@@global.sql_mode, 'ONLY_FULL_GROUP_BY', ''));" &&
         mysql -e "CREATE DATABASE plots;" &&
```
jywarren, line 13: seeing `App 1912 output: Error: The application encountered the following error: Unknown database 'plots' (ActiveRecord::NoDatabaseError)` here... :-(

jywarren: aha - we got 

`mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)`

jywarren: I'll try removing the root lines... dunno...

jywarren:
Ok, cautiously, i think this works??? @cesswairimu @Sagarpreet @noi5e @RuthNjeri want to give it a try? We do see a range of errors but nothing we haven't seen already. Maybe some would be solved by a rebase, others by #8864?

```

 FAIL["test_that_pinned_notes_with_\"pinned:foo\"_tags_appear_at_the_top_of_[nodes:foo]_inline_tables", #<Minitest::Reporters::Suite:0x0000563ce636f668 @name="NodeSharedTest">, 10.334103200992104]
 test_that_pinned_notes_with_"pinned:foo"_tags_appear_at_the_top_of_[nodes:foo]_inline_tables#NodeSharedTest (10.33s)
        Expected: 1
          Actual: 0
        test/unit/node_shared_test.rb:24:in `block in <class:NodeSharedTest>'
```

jywarren:
Indeed, other tests failing include `test/editor_test.rb` which we are working on in https://github.com/publiclab/plots2/issues/8858

jywarren:
OK, rebased... let's see again...

jywarren:
Great, tests can now be run on GitPod, although some fail!

> 1029 tests, 3446 assertions, 7 failures, 12 errors, 0 skips

jywarren:
I'll merge this and do a follow-up for the remaining tests. Even with some failing this will unblock some folks!

PR #8863, opened 2020-12-16
try triggering github action on pull_request instead of push

jywarren:
Follow-up to #8795 cc @RuthNjeri 

docs: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

Let's also try adding a new commit to test that it re-runs, before merging this!

If adding a new commit doesn't trigger, we could try `[pull_request,push]` maybe?

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#push

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,5 +1,5 @@
 name: continuous-integration
-on: [push]
+on: [pull_request]
 jobs:
   test-runner:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jywarren:
I think we're seeing https://github.com/publiclab/plots2/issues/8864 ...

jywarren:
Confirmed, adding a commit re-starts the tests, so this is great!

jywarren:
Yay!

Issue #8864, opened 2020-12-16
Many revisions fixtures have nid = 1 and are causing intermittent test failures


jywarren:
Follow-up to #8862 --

It looks like MANY of the revisions fixtures have nid = 1 -- that must be the underlying problem. I think we have to given them all unique nids, because notes are not supposed to have more than 1 revision! also... they dont have timestamps set so they are probably getting initialized with almost identical timestamps, which could account for the intermittent failures!

https://github.com/publiclab/plots2/blame/main/test/fixtures/revisions.yml#L91

this may take some untangling of the other tests that use these fixtures... we'll have to see! If we give unique nids we may have to add matching entries to /test/fixtures/nodes.yml to make this work.

_Originally posted by @jywarren in https://github.com/publiclab/plots2/issues/8862#issuecomment-746690241_

noi5e:
Currently working on this, will push changes to the branch in PR #8865

PR #8865, opened 2020-12-16
unique nids in revisions.yml

jywarren:
fixes https://github.com/publiclab/plots2/issues/8864 i hope!

### DIFF
--- a/test/fixtures/nodes.yml
+++ b/test/fixtures/nodes.yml
@@ -517,3 +517,172 @@ wiki_with_multiple_comments:
   type: "page"
   cached_likes: 0
   slug: "wiki-list-of-pokemon"
+
+# /test/unit/revision_test.rb
+checkbox_one:
+  nid: 43
+  uid: 2
+  title: "Should render an unchecked checkbox"
+  path: "/notes/bob/01-04-2021/should-render-an-unchecked-checkbox"
+  created: <%= DateTime.new(2021,1,4).to_i %>
+  changed: <%= DateTime.new(2021,1,4).to_i %>
+  status: 1
+  type: "note"
+  cached_likes: 0
+  slug: "note-should-render-an-unchecked-checkbox"
+
+# /test/unit/revision_test.rb
+checkbox_two:
+  nid: 44
+  uid: 2
+  title: "Should render a checked checkbox"
+  path: "/notes/bob/01-03-2021/should-render-a-checked-checkbox"
+  created: <%= DateTime.new(2021,1,3).to_i %>
+  changed: <%= DateTime.new(2021,1,3).to_i %>
+  status: 1
+  type: "note"
+  cached_likes: 0
+  slug: "note-should-render-a-checked-checkbox"
+
+# /test/unit/revision_test.rb
+hashtag_one:
+  nid: 45
+  uid: 2
+  title: "Hashtag post"
+  path: "/notes/bob/01-02-2021/hashtag-post"
+  created: <%= DateTime.new(2021,1,2).to_i %>
+  changed: <%= DateTime.new(2021,1,2).to_i %>
+  status: 1
+  type: "note"
+  cached_likes: 0
+  slug: "note-hashtag-post"
+
+# /test/unit/revision_test.rb
+hashtag_two:
+  nid: 46
+  uid: 2
+  title: "Ignore headings as hasgtag" 
+  path: "/notes/bob/01-01-2021/ignore-headings-as-hasgtag"
+  created: <%= DateTime.new(2021,1,1).to_i %>
+  changed: <%= DateTime.new(2021,1,1).to_i %>
+  status: 1
+  type: "note"
+  cached_likes: 0
+  slug: "note-ignore-headings-as-hasgtag"
+
+# /test/unit/revision_test.rb
+hashtag_three:
+  nid: 47
+  uid: 2
+  title: "Ignores multiple same hashtags" 
+  path: "/notes/bob/12-31-2020/ignore-multiple-same-hashtags"
+  created: <%= DateTime.new(2020,12,31).to_i %>
+  changed: <%= DateTime.new(2020,12,31).to_i %>
+  status: 1
+  type: "note"
+  cached_likes: 0
+  slug: "note-ignore-multiple-same-hashtags"
+
+# /test/unit/revision_test.rb
+hashtag_four:
+  nid: 48
+  uid: 2
+  title: "Ignores pure numbers" 
+  path: "/notes/bob/12-30-2020/ignores-pure-numbers"
+  created: <%= DateTime.new(2020,12,30).to_i %>
+  changed: <%= DateTime.new(2020,12,30).to_i %>
+  status: 1
+  type: "note"
+  cached_likes: 0
+  slug: "note-ignores-pure-numbers"
+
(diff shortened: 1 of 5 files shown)


jywarren:
OK, so now we probably have to create nodes for nid 39-51, based on copies of the most minimal version of this possible:

https://github.com/publiclab/plots2/blob/5b387b674530afd13110bfc4c6e10eab7ed8a705/test/fixtures/nodes.yml#L1-L11

noi5e:
@jywarren Would it still help to have someone work on this? I'm not sure how urgent this is now that we've switched CI over from Travis... Catching up on this thread just now.

I can probably give this a shot if it makes sense for me to do it (since I'm working with fixtures and adding a bunch of new ones). I'm thinking I would do something like below:
```
one: # vary this title
   nid: 1 # vary these from 39 - 51
   uid: 2 
   title: # probably good to vary this
   path: # will match this with title
   cached_likes: 0 
   slug: # will match this with path and title, just to be safe
```

Let me know if there's anything else good to know!

noi5e:
Just played around with this, pushed some commits. There's still some **new** failing tests to figure out, so more work should be done on this.

I think it makes sense to add new nodes to match the new revisions, if we're going to go this route... I ran the tests on both scenarios, and it seems easier to untangle if there are new nodes matching the revisions.

Honestly, this might be a red herring-- I still basically get the same 7 failures, 12 errors running the tests. So I'm not sure if the fix will solve any of those outstanding testing issues.

noi5e:
Currently test-runner is catching four errors in `notes_controller_test.rb`. They're all kind of similar, so here's just one of them:

https://github.com/publiclab/plots2/blob/9c409ec6bee43b95dfe52191cb5069f4199a4524/test/functional/notes_controller_test.rb#L657-L664

I think what's going on is that get :index is only pulling the first 20 notes due to pagination, maybe? That's why `nodes(:one)` doesn't appear in the array as expected. Possible fix would be to:
- Get `nodes(:one)` to appear in the first 20 results, perhaps by playing around with timestamps?
- Rewrite the test entirely?

Any ideas or suggestions?

jywarren:
I would love help, to be honest, i didn't manage to get to this today! Pls note that we seem to be pretty far from synced with the `main` branch though.

The failure mode here, which may help with your Qs, is that different databases and environments can /order/ fetched records differently, especially if they have the same `created_at` value, and in this case, if they have the same `nid` which otherwise would make their ordering predictable. So, absent something to order them, they get jumbled in order, which could account for your observed issue on the index page, (as it might move in and out of the first 20?).

That said, this may not solve other issues, not sure -- but it hopefully will reduce the incidence of some tests passing in one environment but failing in another due to subtle differences in how databases sort un-sortable data (i.e. how they handle a bunch of records with identical fields). 

Hope this makes sense -- thank you!!!!

jywarren:
OK, this will take a lot of work. But we should also try to incorporate a `home-intro` explanatory feature in nodes.yml and revisions.yml to ensure something more friendly shows up to developers when first booting the front page:

https://github.com/publiclab/plots2/blob/2c6807032309c93d6ff7dfa4e8909db47f6a73ff/app/views/home/home.html.erb#L2

jywarren:
This needs a rebase! But also, let's add comments at the bottom of nodes.yml, revisions.yml, and other relevant files that clearly explain how we need unique nids.

noi5e:
Ugh, the commit history for this PR was awful before! Sorry, that was my bad, it was an attempt gone wrong at `git rebase`. 

But I just did another `git rebase` with the current `main` branch and it looks a lot better now.

Let's see how the tests do. There were four merge conflicts that were a little funny that I undid manually. But I'm sure the tests will be more revealing if I did the merge conflicts wrong.

If I recall correctly, this can't be merged as is. There's a system test that was failing that I mentioned [above here](https://github.com/publiclab/plots2/pull/8865#issuecomment-754299916), that I believed at the time had to do with results pagination on `:index`. But we'll see.

jywarren:
Also noting the functional test failures were:

```

========== FAIL NotesControllerTest#test_should_list_only_research_notes_with_status_1_in_index (81.68s)
        Expected false to be truthy.
        test/functional/notes_controller_test.rb:665:in `block in <class:NotesControllerTest>'

 FAIL NotesControllerTest#test_should_list_only_research_notes_with_status_1_in_liked (81.97s)
        Expected false to be truthy.
        test/functional/notes_controller_test.rb:704:in `block in <class:NotesControllerTest>'

 FAIL NotesControllerTest#test_should_list_only_research_notes_with_status_1_in_popular (82.26s)
        Expected false to be truthy.
        test/functional/notes_controller_test.rb:685:in `block in <class:NotesControllerTest>'

 FAIL NotesControllerTest#test_should_list_research_notes_with_status_1_&_4_in_index_if_admin_is_logged_in (82.65s)
        Expected false to be truthy.
        test/functional/notes_controller_test.rb:675:in `block in <class:NotesControllerTest>'
```

jywarren:
OK let's just rewrite those 4 to actually assert exactly what they say - that in the first one, all results are status == 1, for example. That should solve it and the current assertions aren't very good anyways!

jywarren:
https://github.com/publiclab/plots2/blob/a291fe0fff97d030e7b6f3aab1765bad1bc251a2/test/functional/notes_controller_test.rb#L665-L704

jywarren:
OK, narrowed the test assertions a bit. Let's see. There may be other issues.

jywarren:
We may need to push the "first-time-poster" node to the end so it shows up. But, maybe we are OK with it not appearing on the `/notes` index route anymore anyways.

jywarren:
OK, so the functional test could be fixed by #9321 -- let's rebase over. 

```
======= FAIL NotesControllerTest#test_should_list_research_notes_with_status_1_&_4_in_index_if_admin_is_logged_in (30.38s)
        Expected false to be truthy.
        test/functional/notes_controller_test.rb:669:in `block in <class:NotesControllerTest>
```

System tests...

```
ERROR DashboardTest#test_User_can_flag_a_node_from_dashboard (974.95s)
Minitest::UnexpectedError:         Capybara::ElementNotFound: Unable to find css "#flag_node1"
            test/system/dashboard_test.rb:34:in `block in <class:DashboardTest>'
```

I bet it's a reordering issue related to the nids. Let's check...

jywarren:
https://github.com/publiclab/plots2/blob/04f5093248aa8ffd2e9b5c79d0391f8c93ca44d2/test/system/dashboard_test.rb#L26-L36

jywarren:
I bet that's pagination. Let's adjust that so it uses the last node.

jywarren:
Rebased over the fix...

jywarren:
ALMOSTTTTTT

jywarren:
Even after these NID fixes, this still went wrong:

```
Progress: | FAIL TagSelectionTest#test_graph (0.33s)
        Expected: 18
          Actual: 23
        test/unit/tag_selection_test.rb:14:in `block in <class:TagSelectionTest>'

====================================================================|
```

I wonder if that's a different issue, non-NID? We can follow up in a new issue.

jywarren:
Yes, the tag test seems totally unrelated:

https://github.com/publiclab/plots2/blob/d70efef8b13bdba3e39ab710f4a48b36c1778f12/test/unit/tag_selection_test.rb#L5-L16

PR #8879, opened 2020-12-21
Testing: Post Full Page, Then Immediately Comment

noi5e:
Merge this PR first! (I have another one open at #8881)

- Also made new helper function to retrieve node path in the comment tests.


test/system/comment_test.rb
```diff
@@ -15,27 +15,32 @@ def setup
...
     test "#{page_type_string}: addComment(comment_text, submit_url)" do
-      node_name == :wiki_page ? (visit nodes(node_name).path + '/comments') : (visit nodes(node_name).path)
+      visit get_visit_path(node_name, nodes(node_name).path)
       page.evaluate_script("addComment('#{comment_text}', '/comment/create/#{nodes(node_name).nid.to_s}')")
       assert_selector('#comments-list .comment-body p', text: comment_text)
     end
 
     test "#{page_type_string}: reply to existing comment" do
-      node_name == :wiki_page ? (visit nodes(node_name).path + '/comments') : (visit nodes(node_name).path)
+      visit get_visit_path(node_name, nodes(node_name).path)
```
Sagarpreet, line 42: This is repeated 2 times, is it intentional?

test/system/comment_test.rb
```diff
@@ -80,8 +85,47 @@ def setup
       find("p", text: comment_response_text)
     end
 
+    test "post #{page_type_string}, then comment on FRESH #{page_type_string}" do
+      title_text, body_text = String.new, String.new
+      case page_type_string
```
Sagarpreet, line 90: Wow, this is really cool !!
How is `page_type_string ` getting substituted?

test/system/comment_test.rb
noi5e: Hi Sagarpreet! Which part are you thinking about?
1. `visit get_visit_path` looks a little redundant and confusing, maybe I should rename the method to something like get_path or get_path_string.
  - visit is the testing method that visits the page
  - get_visit_path is a helper function I wrote to retrieve the path for the testing node. mainly because wiki comments are only viewable at /wiki/wiki-name/comments. (i have to add '/comments' to the end of the path string)
2. there were two addComment tests that existed before I was an intern. for some reason, there are two tests for this function. one tests addComment(text, URL), the other just addComment(text).
- I was thinking that maybe these tests aren't necessary, aren't really 'system' tests, and could be removed.

test/system/comment_test.rb
noi5e: It's happening here: https://github.com/publiclab/plots2/blob/9fb5d65ffa744623064a5c6353618c01c6c33ea4/test/test_helper.rb#L43-L50

Helper function that returns the name of the node fixture :)

test/system/comment_test.rb
jywarren: I like `get_path` here, thanks!

If the 2 old tests are redundant and you're confident we test them in other ways, that's fine. Are they not system tests because they test only the functions and not the full-stack interaction? I think there may be an ambiguity because there wasn't another way to test JS functions, and these would address both JS and Ruby, so they are "mostly-full-stack" still... what do you think?

My apologies for missing this PR i see it's been open a while. Thanks!

jywarren: We can merge before or after changing to `get_path` and if you'd still like to eliminate the old tests we can do that in another PR, perhaps? Thank you!!!

jywarren (approved):

test/system/comment_test.rb
noi5e: Changed to `get_path`...

Re: `addComment`, that's a good point, I guess the JS wouldn't be tested in any other way besides here. They're also pretty lightweight and fast-running (compared to mine 😛). 

It seems like `addComment` isn't used anywhere on the site besides in these tests... Yet! I guess that'll be one of the things I do, to try to implement it in most comment forms (it's on my to-dos in the planning issue)... I think once that goes live, then these tests could be removed, because the the manual commenting tests would run via `addComment`.

noi5e:
Ready to merge!

jywarren:
Awesome! Just one conflict to resolve!!

noi5e:
Resolved!

jywarren:
Fantastic!!!!

PR #8881, opened 2020-12-22
Testing: Post Replies with Multiple Comment Boxes Open

noi5e:
I have another PR open at #8879, merge that one first!

- Made new fixtures for question, wiki, note -- each with multiple comments on them.
  - I tried adding comments to the existing fixtures, but the assertions for earlier tests throw errors because Capybara expects to find just one element, but finds several.
  - It seemed a lot easier to just make new fixtures, rather than rewrite the old tests... For now.
- Because of the new fixtures, I also had to alter the assertions in unit tests that count questions, wikis, notes. We ran into this problem before (I'll rewrite these someday!)

### DIFF
--- a/test/fixtures/comments.yml
+++ b/test/fixtures/comments.yml
@@ -162,16 +162,79 @@ comment_by_first_time_user:
   timestamp: <%= Time.now.to_i + 10 %>
   thread: /01
 
-reply_to_this_comment:
+reply_to_this_comment: # comment on question, for comment system tests
   uid: 1
   nid: 37
   status: 1
   comment: Uhh...
   timestamp: <%= Time.now.to_i + 10 %>
 
-note_comment_for_responses:
+note_comment_for_responses: # comment on note, for comment system tests
   uid: 1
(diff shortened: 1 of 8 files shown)


jywarren:
> It seemed a lot easier to just make new fixtures, rather than rewrite the old tests... For now.

Another way to do it is to add the comments within the test, using `@node.add_comment(...)` - esp if they're only going to be used in these tests. But, it's a stylistic or maintainability-related preference and your approach sounds great! This looks good and we can wait on #8879 before merging this but it looks super to me!!

jywarren (approved):

noi5e:
Thank you @jywarren, I didn't know about `@node.add_comment(...)`, will definitely look into that. 

I'll make some time toward the end of the internship to clean up all the tests and fixtures I'm making, will probably use that if these fixtures don't get much use.

noi5e:
Oh yeah this is ready to merge too!

jywarren:
Yay!!!

Issue #8513, opened 2020-10-09
[Planning] New Dashboard Implementation Project (Outreachy)


cesswairimu:
This is an edit update made by @RuthNjeri: 

>  Hi👋🏾  , I'll be working on the new dashboard design implementation. Components from the /tags page will be 
> reused for the new design. Feel free to share any feedback, comments, or concerns. The text below is highly based on my Outreachy proposal with some updates. 

# Break Down 

## All Posts Page

- [X] Optimize the notes page that will display 'All posts' https://github.com/publiclab/plots2/issues/8828
- [X] Write functional and unit tests for the notes page https://github.com/publiclab/plots2/issues/8828
- [x] Implement the design for the 'All Posts' page (/notes)..refer to these [notes](https://github.com/publiclab/plots2/pull/8800#issuecomment-740052184). https://github.com/publiclab/plots2/issues/8829
- [x] Implement/Ensure mobile responsiveness in the 'All Posts' page https://github.com/publiclab/plots2/issues/8830
- [X] Implement/Ensure mobile responsiveness in the Research notes section https://github.com/publiclab/plots2/issues/8830

_These will be addressed individually during each page redesign_
- [X] Implement/Ensure mobile responsiveness in the Questions section https://github.com/publiclab/plots2/issues/8830
- [X] Implement/Ensure mobile responsiveness in the Wiki pages section https://github.com/publiclab/plots2/issues/8830
- [X] Implement/Ensure mobile responsiveness in the  Comment section https://github.com/publiclab/plots2/issues/8830

## Dashboard Page
- [x] Create the new 'Dashboard' URL `/dashboard2` https://github.com/publiclab/plots2/issues/8831
- [x] Implement the new Dashboard design..refer to these [notes](https://publiclab.org/notes/ruthnwaiganjo/10-26-2020/outreachy-winter-proposal-new-dashboard-implementation-project#c27525) https://github.com/publiclab/plots2/issues/8832
- [x] Ensure that the dashboard page is mobile responsive by also collapsing the `Follow more topics` sidebar https://github.com/publiclab/plots2/issues/8833
- [x] Display the Blog topic pinned at the top of the dashboard topics list https://github.com/publiclab/plots2/issues/8834
- [x] Ensure that the Blog topic card does not have an X to dismiss it https://github.com/publiclab/plots2/issues/8834
- [x] Allow user to subscribe to blog updates via the blog topic card by email https://github.com/publiclab/plots2/issues/8835
[...]

PR #8886, opened 2020-12-22
[Outreachy] Dashboard Redesign URL

### LINKED ISSUE #8831: Create Another Dashboard URL
This issue is part of #8513

It will fix:

- [ ] Create the new 'Dashboard' URL `/dashboard2`

RuthNjeri:
Fixes #8831 

I haven't fully completed the queries needed for the new dashboard page.

I would love some clarification.. Please have a look at the Dashboard screenshot on #8513 to get an idea on some of the questions I have raised within the comments. 

* [X] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [X] code is in uniquely-named feature branch and has no merge conflicts 📁

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -47,7 +47,7 @@ class HomeController < ApplicationController
     end
   end
 
-  def dashboard2
+  def dashboard_v2
     # The new dashboard displays the blog and topics list
     if current_user
       @blog = Tag.find_nodes_by_type('blog', 'note', 1).first
@@ -64,7 +64,7 @@ class HomeController < ApplicationController
           # Posts within the topic (Tag.find_research_notes(subscribed_tags[index].tag.name))
             # Each topic card displayes at least 3 posts, would that be the most recent?
 
-      render template: 'dashboard/dashboard2'
+      render template: 'dashboard/dashboard_v2'
     else
       redirect_to '/research'
     end
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -50,7 +50,7 @@ class HomeController < ApplicationController
   def dashboard_v2
     # The new dashboard displays the blog and topics list
     if current_user
-      @blog = Tag.find_nodes_by_type('blog', 'note', 1).first
+      @blog = Tag.find_nodes_by_type('blog', 'note', 1).limit(1).first
       @tags = current_user.subscriptions(:tag).includes(:tag)
       render template: 'dashboard/dashboard_v2'
     else
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -51,19 +51,7 @@ class HomeController < ApplicationController
     # The new dashboard displays the blog and topics list
     if current_user
       @blog = Tag.find_nodes_by_type('blog', 'note', 1).first
-      # Topics a user has subscribed to
-      subscibed_tags = TagSelection.includes(:tag)
-                                    .where(following: true, user_id: current_user.id)
-
-      # Also Needed on the view:
-        # Number of people discussing the tag/topic: Not sure how to get this,
-          # is this the number of people commenting on posts?
-        # Total number of posts (post type: "notes" or wikis)
-          # Assuming that the node type to be displayed is type: "note"
-            # Tag.tagged_node_count(subscribed_tags[index].tag.name)
-          # Posts within the topic (Tag.find_research_notes(subscribed_tags[index].tag.name))
-            # Each topic card displayes at least 3 posts, would that be the most recent?
-
+      @tags = current_user.subscriptions(:tag).includes(:tag)
       render template: 'dashboard/dashboard_v2'
     else
       redirect_to '/research'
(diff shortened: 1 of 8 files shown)


app/controllers/home_controller.rb
```diff
@@ -47,6 +47,29 @@ def dashboard
     end
   end
 
+  def dashboard2
+    # The new dashboard displays the blog and topics list
+    if current_user
+      @blog = Tag.find_nodes_by_type('blog', 'note', 1).first
+      # Topics a user has subscribed to
+      subscibed_tags = TagSelection.includes(:tag)
```
jywarren, line 55: This is looking like a great start, and i like that you're starting with just one query. I bet there are other pages we can look at to find fast standard ways to get this info, too! Here, I think we could use `user.subscriptions` -- 

https://github.com/publiclab/plots2/blob/6505f539cabcb9aa4eb9fc5c00af6cf9584351a2/app/models/user.rb#L191-L196

Then, i search for where this is used in code, to see examples:

https://github.com/publiclab/plots2/search?q=user.subscriptions - and indeed we see a nice example, which also includes the `tag` table, so we get the whole record we need. (`tag_selection` is just a join table, i think!)

```ruby
    <% current_user.subscriptions(:tag).includes(:tag).each do |subscription| %>
```

Actually i wish we had a unit test, as that's usually a great guide. So I'll check there too - aww, looks like https://github.com/publiclab/plots2/blob/main/test/unit/user_test.rb has no example for `user.subscriptions` -- that's too bad! Maybe we make a to-do for ourselves to add a simple unit test there.

As to many of the queries below, the template we'll be using actually does a lot of that work -- it expects a `@tags` variable and will build the whole listing from that:

https://github.com/publiclab/plots2/blob/main/app/views/tag/_topicCard.html.erb

So, if we do this, we might get very close!

```suggestion
      @tags = current_user.subscriptions(:tag).includes(:tag)
```

What do you think?

app/controllers/home_controller.rb
```diff
@@ -47,6 +47,29 @@ def dashboard
     end
   end
 
+  def dashboard2
```
Sagarpreet, line 50: Can we have better name `dashboard2 `?

config/routes.rb
```diff
@@ -227,6 +227,7 @@
   get 'feed/liked' => 'notes#liked_rss'
 
   get 'dashboard' => 'home#dashboard'
+  get 'dashboard2' => 'home#dashboard2'
```
Sagarpreet, line 230: `v2/dashboard` ?
USing v2 we can document also what was missing in v1 maybe, what do you think?

RuthNjeri: Yes, this makes sense! I can use v2

app/controllers/home_controller.rb
RuthNjeri: Yeah, I can use v2 which sounds much better...

app/controllers/home_controller.rb
RuthNjeri: This totally makes sense, thanks @jywarren, I'll do some further research and update the PR..

RuthNjeri: Hi, @jywarren I've added the subscriptions test as well as the query.

app/controllers/home_controller.rb
```diff
@@ -47,6 +47,17 @@ def dashboard
     end
   end
 
+  def dashboard_v2
+    # The new dashboard displays the blog and topics list
+    if current_user
+      @blog = Tag.find_nodes_by_type('blog', 'note', 1).first
```
jywarren, line 53: Here we could add .limit(1) to optimize the query a bit.
```suggestion
      @blog = Tag.find_nodes_by_type('blog', 'note', 1).limit(1).first
```

jywarren (changes requested):
This looks great! Are you interested in adding the partial template to display your list of topics in this pr or are you planning that for a follow-up? Awesome, Ruth!

app/controllers/home_controller.rb
RuthNjeri: Alright 👌🏾

RuthNjeri:
> This looks great! Are you interested in adding the partial template to display your list of topics in this pr or are you planning that for a follow-up? Awesome, Ruth!

Thanks 😄 , I was thinking of doing that with the design task in a follow-up PR...

app/controllers/home_controller.rb
RuthNjeri: Done!

RuthNjeri:
@jywarren I'm curious, in the new next PR, should I create a new topics card partial or should I reuse the one within the tag views?

RuthNjeri:
> @jywarren I'm curious, in the new next PR, should I create a new topics card partial or should I reuse the one within the tag views?

oh, I'll recreate another one, I have just noticed slight differences...

jywarren:
I wonder though if you may want to use conditionals to re-use the existing one to preserve consistency? But I trust your judgement!

OK, this should be ready to merge then!

jywarren:
Let's check it out on stable.publiclab.org once it publishes! Great work!!!! 🎉 🎉 🎉 🎉

PR #8887, opened 2020-12-22
Update Mapbox API to fix map base layers

jywarren:
Noting we have more places to do this too:

https://github.com/publiclab/plots2/search?q=https%3A%2F%2Fa.tiles.mapbox.com%2Fv3%2Fjywarren

Following: https://blog.mapbox.com/deprecating-studio-classic-styles-d8892ac38cb4

I wonder if we could also instead use static tiles, not sure if what i selected are among them? OK, confirmed these are static tiles api: https://docs.mapbox.com/api/maps/static-tiles/

@Sagarpreet i wonder if we need to make this update to LEL as well? 

```js
var baseLayer = L.tileLayer('https://api.mapbox.com/styles/v1/jywarren/ckj06ujnc1nmi19nuelh46pr9/tiles/{z}/{x}/{y}?access_token=pk.eyJ1Ijoianl3YXJyZW4iLCJhIjoiVzVZcGg3NCJ9.BJ6ArUPuTs1JT9Ssu3K8ig', {
        tileSize: 512,
        zoomOffset: -1,
        attribution: '© <a href="https://apps.mapbox.com/feedback/">Mapbox</a> © <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a>'
    }).addTo(map);
```

### DIFF
--- a/app/views/home/nearby.html.erb
+++ b/app/views/home/nearby.html.erb
@@ -12,7 +12,7 @@
 
   <script>
     var map = L.map('map').setView([<%= current_user.lat %>,<%= current_user.lon %>], 8);
-    L.tileLayer('https://api.mapbox.com/styles/v1/mapbox/light-v10/tiles/{z}/{x}/{y}?access_token=pk.eyJ1Ijoianl3YXJyZW4iLCJhIjoiVzVZcGg3NCJ9.BJ6ArUPuTs1JT9Ssu3K8ig', {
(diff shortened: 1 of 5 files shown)


Sagarpreet (approved):

jywarren:
Strangely https://stable.publiclab.org/map is not updating...

jywarren:
its because it uses LEL: https://github.com/publiclab/leaflet-environmental-layers/issues/491 moving there

Issue #2, opened 2013-10-24
404 page has old menubar


btbonval:
I went to delete a spam message and got a 404: "There was an error. Are you sure you have the right address?", but I noticed there was still the "Switch to the old site" button in the upper right. I looked more closely at the menubar and found a number of things missing that is seen from the header on the front page.

Click here to see:
http://publiclab.org/notes/Liuzhuang19880717/10-24-2013/a-good-place-to-frame-nfl-shirt

I'm guessing someone else deleted the spam note before I got to it, leading to the 404. That is irrelevant to this ticket.

PR #8897, opened 2020-12-27
Bug: Fix Image Drag & Drop Cross-Wiring in Edit Comment Form

### LINKED ISSUE #8670: Drag and drop feature of comment editor does not work properly when editing a comment
When a file is uploaded using drag and drop feature it gets added in different text box.
### Steps to reproduce bug :
 * Add file in comment box using drag and drop (Don't post new comment just add the file)
*  Try editing any existing comment, the file will get added in wrong textbox

ezgif com-video-to-gif (8)

noi5e:
This is a PR in progress! Don't merge yet. Eventually it will fix #8670. Right now, just looking for feedback from @jywarren.

Essentially this will fix a bug that occurs when uploading images to the 'Edit Comment' form. Sometimes the image will end up in the main comment form, not the edit comment form. (Read the issue comments for more details)

So far, I've written a system test that consistently reproduces this behavior (see Files Changed).

I'm writing up a comment right now summarizing my research, why I think this bug happens, and questions about where to go from here.


noi5e:
Okay great, the test that I wrote is failing! That means it's catching this bug in action. Here's a summary of my research:

## `$D.selected`
It looks like a 'drop' EventListener is attached to `.dropzone` elements here. I think this is basically where the bug begins:
https://github.com/publiclab/plots2/blob/cd59bfc8b603a200a739fc4d3b043e24148c2630/app/assets/javascripts/dragdrop.js#L34-L38

`$D` contains info on the user and the context. `type` is `comment` here in the comments view, but also `type: 'wiki'`, etc. elsewhere: 
https://github.com/publiclab/plots2/blob/cd59bfc8b603a200a739fc4d3b043e24148c2630/app/views/comments/_form.html.erb#L84-L87

## `handleClick()`
This function in /app/views/comments/_form.html.erb is called in two situations:

1. The 'Preview' button in a comment form is clicked.
2. `'choose one'` is clicked when the user wants to upload an image or file.

Here's the definition of `handleClick()`:
https://github.com/publiclab/plots2/blob/cd59bfc8b603a200a739fc4d3b043e24148c2630/app/views/comments/_form.html.erb#L89-L93

`$E.refresh` is  also called, which might play a role in the bug:
https://github.com/publiclab/plots2/blob/cd59bfc8b603a200a739fc4d3b043e24148c2630/app/assets/javascripts/editor.js#L32-L38

Aha! So `$E.textarea` is assigned as the `textarea` within `$D.selected`. To reiterate, `$D.selected` is the closest `'div.comment-form-wrapper'`. I'm thinking `$D.selected` is used as a kind of marker so the editor will know which comment form has been selected. 

This is interesting because it turns out that the edit comment form is NOT wrapped in a `'div.comment-form-wrapper'`. However, the reply comment form and main comment form both ARE wrapped in `'div.comment-form-wrapper'`. If `handleClick()` is called from an edit comment form, then `$D.selected` is going to find the closest `'div.comment-form-wrapper'` OUTSIDE of the edit comment form. Probably in either a reply form or the main comment form:
question_main_comment


reply_comment


edit_comment

Indeed if I test this out in the browser console, it looks like `$D.selected` looks for the nearest `'div.comment-form-wrapper'`, doesn't find it, and defaults to the main comment form:
Wrong D selected

## Why Doesn't The Bug Appear In Wikis?
Still researching this, but I think these lines in `editor.js` are interesting:
https://github.com/publiclab/plots2/blob/cd59bfc8b603a200a739fc4d3b043e24148c2630/app/assets/javascripts/editor.js#L41-L43

So `$E` (which is basically like an instance of Editor?) is `refresh`ed in questions and notes, but not in wikis.

noi5e:
### Questions
- Do image drag-and-drop upload and image select-from-file upload both run through the same code above? I think they do, but am not sure. Using either one will reproduce the same bug.
- Are there any other parts in the code that would be good for me to research?
- Any strategy suggestions for bug-fixing given all the above?
  - I'm thinking two things so far:
    1. `$D.selected` needs to find some way to reference the edit comment form, without referring to `comment-form-wrapper`. Maybe `#c22edit`? The other way would be to just wrap the form in `comment-form-wrapper`, but I think that could introduce more complications.
    2. The `wrap` function I mentioned at the end of my last comment is interesting. Basically this is the function that's called in order to _wrap_ the image link text. For some reason, it checks to see if it's on a wiki page, and it doesn't refresh if so. This could be why the test passes on wiki pages, but neither on questions nor on notes.

Thanks for reading these long comments! Will continue doing research tomorrow.

noi5e:
Did some more research, I've refined my understanding of what happens here. The questions above still apply so please check them out!

I was mistaken before about what runs where. Looking at `views` I believe that certain JS scripts won't run on certain comment forms. `dragdrop.js` is required for the main comment form and the reply form. However the edit comment form has its own separate image upload script which starts here in `/app/views/_edit.html.erb:
https://github.com/publiclab/plots2/blob/cd59bfc8b603a200a739fc4d3b043e24148c2630/app/views/comments/_edit.html.erb#L72

So I think my assumption before that `dragdrop.js` would run on an edit comment form is wrong I believe.

If I sit down and break down the case where the bug occurs, I think it goes like this:
1. drop an image into the main comment form.
  - `dragdrop.js`'s `drop` eventListener sets `$D.selected` to the main comment form's `div.comment-form-wrapper`.
  - `dragdrop.js`'s callback for successful uploads sets `$E.textarea` to main comment form's `textarea`
2. open an edit comment form for `c22` and drop an image there.
  - `#c22div`'s `drop` eventListener gets called from within the comment edit form partial's embedded JS (`/app/views/_edit.html.erb`)
  - this calls `setInit`, which sets `$E.textarea` to `#c22text textarea`.
3. `_edit.html.erb` has a `done` callback for `fileupload`:
  - it calls `$E.wrap` which is supposed to wrap the image URL and put it in to the `$E.textarea`.
  - `$E.wrap` checks to see if we're currently on a wiki page or not.
  - If it's not a wiki, then it calls `$E.refresh`.
  - `$E.refresh` checks to see if `$D.selected` exists or not.
    - `$D.selected` does exist! It's been set to the main comment form (see step 1). 
    - `$E.textarea` is set to `$D.selected`'s `textarea`, and that's where the image URL string is placed, not the edit comment form like we want.

Those are my working assumptions so far. I think it explains why the error doesn't happen on wikis. Going to map out the success case (where the URL goes into the edit comment form like expected), and then from there think about how to fix this one.

noi5e:
Yeah, I think my scenario for the bug is accurate. Here's the scenario for the success case: 

### Scenario: Success
1. open an edit comment form for `c22` and drop an image there.
  - turns out that `_edit.html.erb`'s `setInit` function gets called from `/app/views/notes/_comment.html.erb`:
  - https://github.com/publiclab/plots2/blob/cd59bfc8b603a200a739fc4d3b043e24148c2630/app/views/notes/_comment.html.erb#L56
  - this means that `$E.textarea` becomes `#c22div textarea` after opening the edit form.
  - `#c22div`'s `drop` eventListener gets called from within the comment edit form partial's embedded JS (`/app/views/_edit.html.erb`)
  - this calls `setInit` AGAIN, which sets `$E.textarea` to `#c22text textarea` again for good measure.
2. `_edit.html.erb` has a `done` callback for `fileupload`:
  - it calls `$E.wrap` which is supposed to wrap the image URL and put it in to the `$E.textarea`.
  - `$E.wrap` checks to see if we're currently on a wiki page or not.
  - If it's not a wiki, then it calls `$E.refresh`.
  - `$E.refresh` checks to see if `$D.selected` exists or not.
    - `$D.selected` does **NOT** exist! 
    - `$E.textarea` remains as `#c22text`, and the image URL string goes into the right comment box. Hooray! 🙌

jywarren:
Just noting that i'll be checking in on this first thing in the morning
tomorrow! If you get especially stuck pls feel free to reach out on Gitter
though and I can check in earlier. Thanks!!

app/views/comments/_edit.html.erb
```diff
@@ -1,153 +1,153 @@
...
-            image_url = data.result.url.split('?')[0];
-            orig_image_url = image_url.replace('medium','original');
-            $E.wrap('[![',']('+image_url+')]('+orig_image_url+')', {'newline': true, 'fallback': data.result['filename']});
-          } else {
-            $E.wrap('<a href="'+data.result.url.split('?')[0]+'"><i class="fa fa-file"></i> ','</a>', {'newline': true, 'fallback': data.result['filename']});
+        });
+
+        $('#c<%= comment.id%>div').bind('drop',function(e) {
+          e.preventDefault();
+          $D.selected = $(e.target).closest('div.comment-form-wrapper').eq(0);
```
noi5e, line 67: Notice this additional line!

noi5e: I wrapped everything in _edit.html.erb in a div.comment-form-wrapper.

The other major change is that now $D.selected is assigned in this eventListener bound to #c1div.

A major part of the drag-and-drop crosswiring is fixed. My tests (including the new one I wrote for this issue) are passing. I had to rewrite some of the system tests in this most recent push: 'edit comment' and 'delete comment' rely on .comment-form-wrapper for some reason.

app/views/comments/_form.html.erb
```diff
@@ -1,13 +1,12 @@
 <div class="comment-form-wrapper" style="background-color:#f8f8f8; border: 1px solid #e7e7e7;padding: 18px;">
   <div class="card-body">
     <form class="comment-form" id="comment-form" data-remote="true" action="/comment/create/<%= @node.nid %><%= "?type=question" if local_assigns[:type]=="question" %>" <% if local_assigns[:aid].blank? %>method="post"<% end %>>
```
noi5e, line 3: Mostly just reorganizing spacing and indentation in this file.

app/views/comments/_edit.html.erb
noi5e: Oops, the tests aren't passing. I'll take a closer look tomorrow.

noi5e:
One last comment before I sleep, it would be really nice if I could just set `$D.selected` to `e.target`'s closest `div.comment-form-wrapper` right here (in fileupload's start listener):
https://github.com/publiclab/plots2/blob/cd59bfc8b603a200a739fc4d3b043e24148c2630/app/assets/javascripts/dragdrop.js#L53-L61

Unfortunately, if I log e.target here, it's pointing to a completely different, and wrong comment. I think it has something to do with the way jQuery fileupload is set up on the site.

Also, I just want to note how many different ways there are to upload an image in a single edit comment:
1. drag-and-drop into #c1div (_edit.html.erb, 'drop' listener)
2. drag-and-drop into the .dropzone in the edit comment form (dragdrop.js, 'drop' listener)
3. clicking on choose one (_edit.html, skip drop, 'start' listener)
4. drag-and-drop _into the .dropzone_ of #c1div (dragdrop.js, 'drop' listener)
5. click on #c1div's dropzone to select a file (dragdrop.js, skip drop, 'start' listener)

app/views/comments/_edit.html.erb
jywarren: this is awesome

jywarren:
This is looking awesome, i am thinking that some of the keys to bear in mind may be:

1. let's try to pragmatically solve the bug you so excellently identified with the test, first, then move on to a more structural change. this lets us trust the test more for showing us that the solution is valid
2. for the later, more structural solution, we might prefer a solid unique ID instead of relying on `.closest('div.comment-form-wrapper')` to get us the correct textarea. Are there cases where we aren't getting, or using, a truly unique ID? For example, is there any re-use of ids, or anywhere we are not able to assign a unique ID for some reason?
3. the MANY ways of choosing an image is in part because people have hugely varying preferences and expectations for image upload. There's actually another -- some people try to copy-paste images 😅 which somehow actually works... sometimes!?? I think that later, once we resolve this particular bug and are more confident in our tests protecting these various scenarios, we might try to consolidate the handling of all of these as much as possible through the same code pathways.
4. Our tests should try to show the 'ideal' way to write code in this area, as well as walk people through all the variations we need to support. You can imagine how the many /types/ of comments, replies, and places to post them, plus AJAX vs. non, plus click to select vs. drag/drop plus clipboard, all contribute to situations where coders not DEEPLY familiar may fix a small issue but inadvertently break other variants of this interaction. Our tests should not only prevent this by testing all the relevant variants, but should model "good usage" such as a) assigning everything a unique ID, b) consistently using that unique ID to address different forms. Hopefully this will make our code more maintainable and legible to folks going forward!
5. It's such a pain to hunt through all the places where code lives here. I think choosing a good pattern and sticking to it could help - like, some minimal code to establish the state (unique ID, is it a wiki, etc) and then passing it into a standard comment form construction function may be a good approach. but you'll know best as you continue... this has simply gotten pretty overgrown over the years!

I hope some of this helps, i know you are very likely already thinking about a lot of this. You're doing an excellent job and unraveling some really subtle and messy threads... and reducing technical debt! No need to tackle every optimization or ideal code structure at once, but i feel that making smaller pragmatic changes, adding tests to document our steps, and then slowly and iteratively working towards a better and more legible structure is probably the best way to go. Good work unraveling this without getting overwhelmed by the scope!!!

jywarren:
> Do image drag-and-drop upload and image select-from-file upload both run through the same code above? I think they do, but am not sure. Using either one will reproduce the same bug.

OK, you found the duplicated code already but I think it could be good to look at the blame view of https://github.com/publiclab/plots2/blob/cd59bfc8b603a200a739fc4d3b043e24148c2630/app/assets/javascripts/dragdrop.js#L44 vs. https://github.com/publiclab/plots2/blob/cd59bfc8b603a200a739fc4d3b043e24148c2630/app/views/comments/_edit.html.erb#L72 and see if it was apparent to the authors that this code existed in 2 places. I tend to think they look similar enough that we should consolidate it in `dragdrop.js` -- but not in this PR, let's think about that as a separate task!

> Are there any other parts in the code that would be good for me to research?

I guess just that given that we use JS to build forms once new comments are added via AJAX, we might start to consider if we want that to be the only way we create comment forms -- that is, consolidate with a `buildCommentForm(options)` method so we can standardize that and test it as well -- just like we did with `addComment()` before!

> Any strategy suggestions for bug-fixing given all the above?

Definitely do things one at a time in separate PRs. And, once you've really grasped what is /supposed/ to happen, write a descriptively-named test for it -- even if it's a pretty small thing! At least add a comment to the test to explain the step -- a good example might be `// $D.selected is used to represent _____` within the tests. That way the test suite both establishes a "known area" you can rely upon, while also generating documentation -- for yourself as much as for any future coder! 

I always feel that it's hard to keep a complex system in my head, so i try not to. If it takes a lot of time to recover a mental state -- get oriented in the code -- that's a sign of code which should be broken down into simpler more descriptive parts, using function names and tests to map things out and become more legible. 

> The wrap function I mentioned at the end of my last comment is interesting. Basically this is the function that's called in order to wrap the image link text. For some reason, it checks to see if it's on a wiki page, and it doesn't refresh if so. This could be why the test passes on wiki pages, but neither on questions nor on notes.

Here, I'd look at the blame view for that wiki switching code and try to trace back who added it and why... i know wiki comments were added much later than others, but i don't totally understand why they'd behave differently in this scenario!

OK, hope this helps!!!! 👍 again, great work!!!

app/views/wiki/edit.html.erb
```diff
@@ -60,8 +60,12 @@
             <input aria-label="Wiki Title" tabindex="1" name="title" type="text" class="form-control" id="title" value="<% if @node && @node.latest %><%= @node.latest.title %><% else %><%= params[:title] %><% end %>">
           </div>
 
-          <script>
+          <script type="application/javascript">
             (function(){
+              $E.initialize({
```
noi5e, line 65: I decided to initialize $E here to point toward the main comment form on /wiki/new once the page loads.

test/application_system_test_case.rb
```diff
@@ -7,45 +7,52 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   # https://web.archive.org/web/20170730200309/http://blog.paulrugelhiatt.com/rails/testing/capybara/dropzonejs/2014/12/29/test-dropzonejs-file-uploads-with-capybara.html
   def drop_in_dropzone(file_path, dropzoneSelector)
+    existing_inputs = page.all('.fakeFileInput')
```
noi5e, line 10: I rewrote this test so that multiple file inputs could be inserted onto a page.

I did notice that this slows down the test considerably, I'm sure this could be written to be more efficient.

I'll put that on my to-dos (will have to brush up on Ruby some more)

app/assets/javascripts/dragdrop.js
```diff
@@ -9,216 +9,132 @@ jQuery(document).ready(function() {
...
+  $('.dropzone').on('dragout',function(e) {
+    $('.dropzone').removeClass('hover');
+  });
+  $('#side-dropzone').on('dragout',function(e) {
+    $('#side-dropzone').removeClass('hover');
+  });
+  $('.dropzone').on('drop',function(e) {
+    $D.selected = $(e.target.closest('div.comment-form-wrapper')).eq(0);
+    e.preventDefault();
+    // $E.initialize({});
```
noi5e, line 37: This is a big change (removing $E.initialize({})).

I tried to trace this in the history, it seems like this was originally done to address an image upload issue in /wiki/new... Ironically, I decided to comment this out now to fix other image upload issues in wikis. I think I'll see the consequences of this in the next few days as I play around with things, but so far it seems all right.

I chose this because I'm seeing that e.target is VERY unreliable in these click handlers (it's such a big issue I'm going to work on it next). I'd rather have $E be stable here, than have both $D and $E be reset with one event.

noi5e:
Linking this PR (which handles **drag & drop** image upload crosswiring) to this issue: #8926 (which describes creating unique HTML IDs to fix **click** image upload crosswiring).

Also @jywarren this PR is ready for review!

app/assets/javascripts/dragdrop.js
jywarren: > multiple handlers being set on multiple .dropzones, with bubbling

yikes 😱 😭 

but, thank you for starting to untangle this!!!

jywarren: OK, just before approving, is this resolved, or what is the latest status here? Would you like help determining what exact purpose `$E.initialize()` played here, that could help to inform or at least thoroughly document this removal? Thank you!!! Mostly looking to leave a good "paper trail" here as to why we choose to remove it.

jywarren (changes requested):
This looks good, but i had one request for info about the `$E.initialize()` removal. 

Also, is there anything in particular you want me to read through extra carefully, or do you feel this code is well protected by tests? 

Noting that, i'm sure you're doing this but, it can be good to boot up the code and manually try it in addition to the tests; there's definitely such thing as over-reliance on automated testing (not that we should do less, of course, but that we shouldn't assume it's alone sufficient - occasional reality checks are healthy!). 

Lastly, how does this PR interact with https://github.com/publiclab/plots2/pull/8927 -- sorry, just to recap -- should it go after https://github.com/publiclab/plots2/pull/8927 or before? Thank you!!!

app/assets/javascripts/dragdrop.js
noi5e: @jywarren Yes, if you have the time and willingness this week, would you mind looking into $E.initialize({}) here? I looked through the blame views, but I think it could help to have a second pair of eyes looking at this... I agree with you that we should be careful about the removal.

I'll take a deeper look again tonight investigating the consequences of removing $E.initialize... I tested it out with both `rails test` and just poking around about a week ago. I remember my impressions were that it was definitely a net positive, but I could write a more detailed comment about that.

#8927 is good to merge before this one! In fact, having that one merged might even help me out here, or maybe even sidestep this conversation.

noi5e:
I'm retracing my steps right now and seeing exactly what I did and why I commented out that line... It's getting very clear. Will explain in a comment later today.

noi5e:
@jywarren Okay, bear with me, but this is going to be another really long comment. 🤓

### `What Happened`
So just to retrace my steps, here's what happened:

1. I made the first four commits in this PR:
  - system tests, then a rewrite of the system tests
  - fixed indentation and spacing
  - `_edit.html.erb`: wrap edit form in `div.comment-form-wrapper`
  - `_edit.html.erb`: change the `#cEDITdiv` comment `drop` eventListener so that `$D.selected` is assigned to the new `div.comment-form-wrapper`
2. Just reiterating these steps to say that I don't think they caused any of the weirdness that happened next.
3. I ran the system tests on CI, and saw that they weren't passing [here in this comment](https://github.com/publiclab/plots2/pull/8897#discussion_r549617362).
4. I took a closer look and saw that there was a separate (but related) bug I had never seen before:
  - It turns out that on **WIKI** pages that have any comments, `$E.textarea` is not initialized to the correct value.
  - Breaking out to explain this in detail:

### `Bug: Wikis with Comment(s)`
You can see this bug in action right now at stable.publiclab.org. Here's a [page I made with a comment on it](https://stable.publiclab.org/wiki/pokemon/comments).

If you log `$E.textarea` on page load, you'll see that it's set to the reply comment form for the first comment. We actually want it to be set to the main comment form:
https://user-images.githubusercontent.com/4361605/103819964-4adb6580-5020-11eb-8a07-cb7efc591cc4.mov

The consequence of this is that if you drop an image into **main**, it's going to end up in the **reply** comment form:
https://user-images.githubusercontent.com/4361605/103820005-5c247200-5020-11eb-8f19-c2a549cccccc.mov

This is caused by there being multiple repeated HTML IDs (again! :scream:). In this case the `textarea` in both main and reply comment forms both have an ID of `text-input`. 

Every `textarea` here needs a unique ID... Unfortunately I overlooked this in the unique HTML IDs PR that got merged yesterday. These unique IDs still need to be added.

### `What I Did`
I saw that maybe a bug fix in wiki/comments would be required before this PR could be merged. But I asked myself, _"Maybe there's something I can do to solve it here and now?"_ Particularly, I was looking at these lines in `dragdrop.js`:
https://github.com/publiclab/plots2/blob/2f7aeacea83bc7d35309ea62675a12b89e86ac3d/app/assets/javascripts/dragdrop.js#L34-L38

I couldn't understand why `$E.initialize({})` was called here. I'd spent a lot of time looking at other parts of the codebase, and thought I had a reasonable grasp about how things worked, but I still couldn't get this one part. (I can't find much info about it in the [original commit](https://github.com/publiclab/plots2/commit/782527b92ed419a7bdba95612e002c381ac3763d)). I do know that in other parts of the codebase, the values in `$E` are refreshed to fix similar kinds of bugs: form cross-wiring, image upload, etc. For example, there's [this line](https://github.com/publiclab/plots2/blob/main/app/assets/javascripts/editor.js#L43) in `editor.js`. That's what I'm guessing is happening above.

It also made my cross-wiring bugfixes difficult and hard to conceptualize, because values in `$E` are changed in `drop`, and then later on in `done`, `$E.textarea` is assigned again to match `$D.selected`. I thought, _if I'm finding this so confusing, it's probably going to also be confusing for people in the future._ (I still think we should find a way to not call initialize here, but I realize that fix might be beyond the scope of this PR)

I discovered that commenting out `$E.initialize({})` does indeed solve this particular bug. Hooray! But there is a big catch.

### `The Catch`
This is where I got a little hasty... The big catch is that **click-to-upload** in reply comment forms breaks... I understand why this breaks, and I could go into more detail about it, but this comment is already long enough.

Writing this now, I'm realizing that's pretty bad, and we definitely don't want to put that into production anywhere. I will also admit that maybe my brain was fatigued from thinking through all these things in pretty intense detail.

But at the time, I was already working on my PR for unique HTML IDs, and I saw that it was very promising. I thought with unique IDs, `e.target` would become accurate. This would mean I could write a fix for the new **click-to-upload** bugs that commenting out `$E.initialize({})` introduced. I'm still pretty confident that this will work. 

In writing this, I'm realizing that there's a more careful way of approaching this, without pushing bugs to production. So I appreciate being able to slow down and look at it.

### `Follow-Up`
What I'm going to do now is restore these things:
- `$E.initialize({})` (I will restore it, meaning, _undo the commenting-out_ I did)
- remove the new `$E.initialize` call I made in `/views/wiki/edit.html.erb`
  - turns out this is totally unnecessary, because this call already happens in the `editor/editor` partial.

I think I'm going to try to solve the wiki bug within this PR and branch by making unique `text-input` IDs. If that's okay! I think there's enough of a paper trail now to do so.

If that works and doesn't re-introduce bugs, I'm going to proceed with the plan (and eventually find a way to rewrite that `$E.initialize({})` call, I hope)

noi5e:
I also thought it might be worth tagging @ujithaperera. No pressure, but do you remember why `$E.initialize({})` is called here?
https://github.com/publiclab/plots2/blob/782527b92ed419a7bdba95612e002c381ac3763d/app/assets/javascripts/dragdrop.js#L26-L29

noi5e:
Okay great! That exact test is failing like I thought it would. Now I'm going to try to fix the bug by assigning unique IDs to for forms with `#text-input`

app/views/comments/_edit.html.erb
```diff
@@ -1,153 +1,151 @@
...
-
-          if (is_image) {
-            image_url = data.result.url.split('?')[0];
-            orig_image_url = image_url.replace('medium','original');
-            $E.wrap('[![',']('+image_url+')]('+orig_image_url+')', {'newline': true, 'fallback': data.result['filename']});
-          } else {
-            $E.wrap('<a href="'+data.result.url.split('?')[0]+'"><i class="fa fa-file"></i> ','</a>', {'newline': true, 'fallback': data.result['filename']});
+        });
+
+        $('#c<%= comment.id%>div').bind('drop',function(e) {
```
sagarpreet-chadha, line 63: Here bind is working?  Or should we use ‘on’

noi5e: I think you're right! I believe bind has been working here, but it is deprecated and should be changed.

I changed it in dragdrop.js, but I'll change these ones soon... Just want to get this PR ready to merge first.

app/assets/javascripts/dragdrop.js
```diff
@@ -1,4 +1,4 @@
-jQuery(document).ready(function() {
```
sagarpreet-chadha, line 1: Does the new syntax also works like before? We want this script to work after document is loaded.

noi5e: Oh yes, this script is still working. I changed the syntax because I believe the old one was deprecated.

noi5e:
Almost there with this one. Changing `#text-input` to `.text-input` is causing the testing errors-- I'll have to rewrite these system tests. I'm also looking carefully through the codebase for places where `#text-input` is selected and changing them... Will work on this more tomorrow.

noi5e:
😌 Phew. I think this is it. Leaving my last few notes.

1. I gave unique IDs to comment textareas. Now there are only two elements with the ID `#text-input`: 
  1. main comment form (at bottom of page) on pages with multiple comments
  2. textarea on pages to create new note, question, and wiki (because we reuse the same form for comments and node creation)
2. unique textarea IDs unfortunately revealed more bugs. i'm going to break out these things into new issues because this PR is gigantic already:
  1. **Bug #1**: I saw for the first time that reply comment forms are cross-wired with each other, in production.
    1. Go to a [page with multiple comments](https://publiclab.org/notes/ruthnwaiganjo/12-28-2020/i-too-struggle) (this is in production), and open the 2nd or 3rd comment (not the 1st one!)
    2. Upload an image by clicking on the dropzone button in the toolbar.
    3. The image will very often end up in the 1st comment reply form (open it by clicking on Reply to this comment...)
    4. Our system tests weren't catching this bug, because they only test for the 1st comment on a page.
  2. **Bug #2**: A bug with the test, image drag & drop into reply form on wiki comments was failing.
    1. This was failing for similar reasons as bug #1. The test used to pass because `$E` was initialized to the first reply comment form on the page. With unique IDs, it has to resolve to the main form so the assertion for image in reply form fails.
  3. **Bug #3**: When I tried to fix bug #2, a new bug manifested. [This test](https://github.com/publiclab/plots2/blob/b3a9aec8d259807619bf9987679470be3bfc2234/test/system/post_test.rb#L107) started to fail.
    1. This was failing with my bugfix for bug #2 because the comment form partial is reused for creating new wikis. 
    2. If $E is initialized with parameters, it has to account for both scenarios: 1) a page with multiple comments on it 2) a create new wiki page.
3. I fixed bugs #2 and #3 in `dragdrop.js`'s `drop` listener by adding a conditional that checks for both scenarios. It is kind of hacky, and I still want to eventually rewrite this so that we don't have to initialize $E here, but for now I think it works really well!

noi5e:
@jywarren This one's ready to merge! It passes all the tests, and I've checked it out locally as well by playing with image upload on comments and /wiki/new. I feel very confident it's not creating any new bugs.

Sorry this is such a big PR with so many long notes. I probably could've broken out the new `textarea` IDs piece into another PR, but I really didn't think that was going to bring up more complications. _(My bugfixes would break a test, which I would then fix, which would break more tests, etc.)_ I will make smaller PRs from hereon out.

Just to summarize, here is what this PR is doing:
- assign unique IDs to comment `textareas`
- rewrite `drop` listener in `dragdrop.js` so that `$E` can be re-initialized for two different scenarios:
  - the appropriate comment form on a multi-comment page
  - `#text-input` on a /wiki/new page
- `/comments/_edit.html.erb`, edit comment form partial:
  - wrap the entire edit comment form in a `div.comment-form-wrapper`
  - assign `$D.selected` to the `div.comment-form-wrapper` when something is dropped onto an edit form's dropzone
- comment system tests:
  - rewrite `drop_in_dropzone` helper so images can be dragged & dropped on multiple `.dropzones`
  - renamed some system tests for clarity
  - rewrote old 'comment image upload' test to focus on the main form, not reply form
    -  i will add the reply form test back in my next PR to fix the bug i mentioned above (Bug 1)
  - add new system test: image DRAG & DROP into EDIT form isn't cross-wired with MAIN form
  - rewrote tests that selected `#text-input` to instead select the appropriate `.text-input`

app/assets/javascripts/dragdrop.js
```diff
@@ -9,216 +9,138 @@ jQuery(document).ready(function() {
...
+  $('.dropzone').on('dragout',function(e) {
+    $('.dropzone').removeClass('hover');
+  });
+  $('#side-dropzone').on('dragout',function(e) {
+    $('#side-dropzone').removeClass('hover');
+  });
+  $('.dropzone').on('drop',function(e) {
+    $D.selected = $(e.target).closest('div.comment-form-wrapper').eq(0);
+    e.preventDefault();
+    let params = {};
```
noi5e, line 37: This part is new, and significant. Rather than just comment out $E, I rewrote this initialization call to account for two different scenarios:
1. pages where there's only one comment form (ie. /wiki/new). we want this to be initialized to #text-input.
2. pages with multiple comments-- here we want $E.textarea to be the closest textarea to the drop event.

Like I've said, I do want to refactor this method and others so that we don't have to $E.initialize, but I feel like this is a very good compromise for now.

app/views/comments/_edit.html.erb
jywarren: sounds good we can consider this in a separate PR! good catch @Sagarpreet!!

jywarren:
OK, i'm going through this comment by comment, so taking notes in my comment as i go... reading through https://github.com/publiclab/plots2/pull/8897#issuecomment-755707745, I wanted to note that some of the confusion from `$E.initialize()` is really because a) it's very very old code, remade many times, and b) it doesn't explain itself well -- what are we initializing? Why is it that we can do it without any explicit parameters? Shouldn't we assign its output to a variable whose state is permanent, like `var replyCommentEditor2 = new CommentEditor({...});` instead of just firing off generic non-specific initialize calls left and right? 😵 

> I will also admit that maybe my brain was fatigued from thinking through all these things in pretty intense detail.

This is definitely a sign that we should try to compartmentalize the logic more. Nobody should be juggling all these things in their head at once -- it's a sign of bad design! 😄 

> The big catch is that click-to-upload in reply comment forms breaks

Did you find this via a test failure? If not, perhaps we should ensure it's covered, because our longer-term desire to redesign the initialization system may mean we're gonna start tearing things out to rebuild... and we'll want good test coverage to do that with confidence. But no worries, we won't do this in this PR most likely!

Basically after this point, you did great... your guess at breaking a test, your docs re: 3 different bugs... all good. 

> Our system tests weren't catching this bug, because they only test for the 1st comment on a page.

great catch! This can be saved for later, let's just be organized about remembering things to circle back to. 

Re: hacky fix, that's 👍 totally fine as you point out we'll come back here to fix more soon... just leave a note to self!

In general, to avoid such large PRs, what if we have the principle that we split out any bugs we find that aren't related to the core function of this PR, into their own issues (or even just to checkboxes on the planning issue for now). 

> not creating any new bugs

exactly!

OK, and finished reading your comments, and all the rest sound super. I really appreciate your including your reasonings here, as I think it'll provide helpful context if we need to circle back at some point. 

And in general, +1000000 rewriting `$E.initialize({});` to be more legible and to manage state better. In general, rather than an ambiguous initialize function, we should have a constructor function which sets things up, creates a persistent state, and returns an instance of a self-managing class that has uniqueness and retains the relationships between different HTML components with high specificity. 

Merging this now, MANY thanks for such careful work, and congrats! This is also bringing into much sharper focus the architectural refinements we hope for down the road, their relationship to existing code, tests, and legibility! 

🎉 🎉 🎉 🎉 🎉

app/assets/javascripts/dragdrop.js
```diff
@@ -9,216 +9,138 @@ jQuery(document).ready(function() {
...
+    if ($D.selected.hasOwnProperty(0)) {
```
jywarren, line 38: This logic is a bit obscure, `hasOwnProperty(0)` but might be improved with an explanatory comment. I think it also shows us that `$D` may also need refactoring along similar lines to `$E`. 

And, apologies as I believe I was originally responsible for both `$D` and `$E` - and wow, i've grown so much as a coder in the ~10 years or so since they were written!!! And, to be frank, so has JavaScript!

test/system/comment_test.rb
```diff
@@ -186,15 +186,70 @@ def get_path(page_type, path)
       page.assert_selector('#preview img', count: 1)
     end
 
+    # sometimes if edit and reply/main comment forms are open,
```
jywarren, line 189: Nice comments here! Thanks!

jywarren (approved):
Awesome. Last review all good. 

> `closest()`

The extra state-management and specificity of a future rewrite of `$E` should make using this kind of DOM searching unnecessary, too!

OK, merging now!!

app/assets/javascripts/dragdrop.js
noi5e: Noted, and agreed, will write some comments here to explain! 

I did notice you wrote a lot of `$D` and `$E`, no worries! 😄 I'm seeing what goes into growing and maintaining a 10+ year codebase, which is such a valuable learning experience. It's a real achievement that this repo has been growing for so long!

It's kind of neat how much of this work is historical. Like it was interesting to realize that the editor code was written for creating new posts first, then repurposed for comments. I understood the code a lot more then.

noi5e:
Thanks @jywarren again for sifting through this very large PR! It was very satisfying to `git merge upstream/main` today. 

> In general, to avoid such large PRs, what if we have the principle that we split out any bugs we find that aren't related to the core function of this PR, into their own issues (or even just to checkboxes on the planning issue for now).

Definitely, yeah, will do this in the future! I'm starting to understand how to break up really complex issues into smaller more manageable PRs. I think before I was a little intimidated at the thought of managing merge conflicts, but now I see that's way more desirable than the alternative.

> Did you find this via a test failure (click-to-upload in reply comment forms breaks)? If not, perhaps we should ensure it's covered, because our longer-term desire to redesign the initialization system may mean we're gonna start tearing things out to rebuild... and we'll want good test coverage to do that with confidence. But no worries, we won't do this in this PR most likely!

Hmm, I don't think so. All the system tests were passing at that point, because we don't have that kind of test coverage.

Definitely noted! This will be covered in the next few PRs, which are focused on click-to-upload issues.

> In general, rather than an ambiguous initialize function, we should have a constructor function which sets things up, creates a persistent state, and returns an instance of a self-managing class that has uniqueness and retains the relationships between different HTML components with high specificity.

Noted!

Will make a small follow-up PR to leave comments, and change the `bind` to `on` that Sagarpreet noticed.

Issue #8926, opened 2021-01-03
Issue: Add Unique HTML IDs to Comment Forms to Fix Image Upload Cross-Wiring Issues


noi5e:
## `Description`
Creating a fresh issue page for this complex bug that appears in many forms. See #8670 and #8478 for two examples of this bug in action. Also see current PR #8897 (awaiting review) that fixes a small piece of this bug.

Comment editor functions are cross-wired:
- Click on a main comment form, and upload an image.
- Open an edit comment form and upload an image.
- **Expected Behavior:** The image link should go into the edit comment form.
- **Actual Behavior:** Instead, the image link ends up in the main comment form _(see image)_. The image upload was 'cross-wired' into the main comment form.

![103166998-428f5900-47dc-11eb-9441-b46ecf6af58a](https://user-images.githubusercontent.com/4361605/103472187-6a803e80-4d3f-11eb-8a6f-1e895ab7672d.gif)

## `Why This Happens`
The JS scripts in `/views/comments/_edit.html.erb` and `/app/assets/dragdrop.js` have variables like `$D.selected` and `$E.textarea` that point toward the comment form that the user last interacted with. However, these are often pointing to the wrong form. They need to be updated every time the user takes particular actions: (toggling a form open, entering text, uploading an image).

## `Comment Forms: Uploading Images`
![Edit Comment](https://user-images.githubusercontent.com/4361605/103472308-f5ae0400-4d40-11eb-8d18-88099b4addfd.jpg)

- **A**: This `div.dropzone` wraps an `input#fileinput`.
  - The user can drag-and-drop (yes, on this tiny area) an image to upload it to the comment, triggering [this 'drop' eventListener in dragdrop.js](https://github.com/publiclab/plots2/blob/9c409ec6bee43b95dfe52191cb5069f4199a4524/app/assets/javascripts/dragdrop.js#L34-L38), then running [this start upload listener here](https://github.com/publiclab/plots2/blob/9c409ec6bee43b95dfe52191cb5069f4199a4524/app/assets/javascripts/dragdrop.js#L53-L62).
  - The user can also just click the button, skipping the 'drop' listener, and going straight to the 'start'.
- **B**: This entire section is wrapped in a `div#c1234div`.
  - The user can drag-and-drop images for upload here.
[...]

Issue #8478, opened 2020-10-07
Wrong text input gets selected while using toolbar on comments


urvashigupta7:
While trying to add new comment  when any of the option is choosen from toolbar the changes are displayed in different text input.

![ezgif com-video-to-gif (6)](https://user-images.githubusercontent.com/41363633/95346113-1c42d700-08d9-11eb-87e9-d018b8cd15bb.gif)

### where to look

https://publiclab.org/questions/Cogreg/09-20-2020/how-do-i-test-home-made-fertilizer

Sagarpreet:
Correct!
Always the first comment editor is changed. This is because if do not have any `id` associated to each comment editor.

nibble0101:
Hi @Sagarpreet . Outreachy applicant here. Can I take a  look at this?

sagarpreet-chadha:
Hey @nibble0101 , yes we would like your help 👍 
So this is a little big issue, you may want to break this down 😄 
Also let’s wait for @jywarren approval before we start because we might want to spin off a separate library for comment editor and solving here may not be useful now.

noi5e:
Starting to work on this one! Just made a PR to add unique IDs to the toolbar buttons. After that, I think it will be a very similar fix to the image upload cross-wiring issues (linked above in this issue)

PR #8927, opened 2021-01-03
Feature: Add Unique HTML IDs to .dropzones, #fileinputs

noi5e:
See issues #8926 #8670 #8478

Adds unique HTML IDs to .dropzones and #fileinputs in comment forms.

Edit Comment

- **A**: _currently_: `.dropzone` (with no ID) wrapping `#fileinput` (which is used twice in the same comment form, or multiple times per page)
  - _after merging_: `#dropzone-small-edit-22`
  - _after merging_: `#fileinput-button-edit-22`
  - _after merging_: `#dropzone-reply-22` for reply forms, `#dropzone-main` for base comment form in notes, questions, and wikis
- **B**: _currently_: `.dropzone` (with no ID) wrapping the form's entire face, same class as A
  - _after merging_: `#dropzone-large-reply-20` in reply forms and `#dropzone-large-main` in main forms
- **C**: _currently_: second `#fileinput` here (distinct from A, but with the same ID)
  - this is now `#fileinput-choose-one-reply-1234` or `#fileinput-choose-one-main`

As usual, had to rewrite some of the old system tests that relied on the old classes and IDs. 

There's probably more tests to rewrite before this is ready-to-merge, thought it would be faster/easier to see in CI rather than run tests locally.

---

### DIFF
--- a/app/views/comments/_edit.html.erb
+++ b/app/views/comments/_edit.html.erb
@@ -8,8 +8,8 @@
       #imagebar {width:100%;}
     </style>
 
-
-    <%= render :partial => "editor/toolbar" %>
+    <%# toolbar needs location & comment_id to make unique element IDs%>
+    <%= render :partial => "editor/toolbar", :locals => { :comment_id => comment.id.to_s, :location => :edit } %>
     
     <div id="c<%= comment.id%>div" class="form-group">
       <textarea aria-label="Edit Comment" onFocus="editing=true" name="body" class="form-control" id="c<%= comment.id%>text" rows="6" cols="40" placeholder="<%= placeholder %>"><%= !(comment.is_a?Answer) ? comment.comment : comment.content %></textarea>
--- a/app/views/comments/_edit.html.erb
+++ b/app/views/comments/_edit.html.erb
@@ -8,7 +8,7 @@
       #imagebar {width:100%;}
     </style>
 
-    <%# toolbar needs location & comment_id to make unique element IDs%>
+    <!-- toolbar needs location & comment_id to make unique element IDs -->
     <%= render :partial => "editor/toolbar", :locals => { :comment_id => comment.id.to_s, :location => :edit } %>
     
     <div id="c<%= comment.id%>div" class="form-group">
--- a/app/views/comments/_form.html.erb
+++ b/app/views/comments/_form.html.erb
@@ -19,16 +19,30 @@
           padding-top: 10px;
         }
       </style>
-
-
-      <%= render :partial => "editor/toolbar" %>
+      <%# if location == reply, toolbar needs reply_to for dropzone IDs %>
+      <% locals = (location == :reply) ? 
+        { :location => location, :reply_to => reply_to } : 
+        { :location => location } 
+      %>
+      <%# location: for creating unique .dropzone IDs %>
+      <%# location can be: edit form, reply form, main comment form, note responses %>
+      <%= render :partial => "editor/toolbar", :locals => locals %>
 
       <% if local_assigns[:reply_to].present? %>
         <%= hidden_field_tag :reply_to, local_assigns[:reply_to] %>
       <% end %>
 
-      <div class="form-group dropzone">
-
+      <%# most comment forms have two dropzones: 1) small button, 2) large form that covers textarea %>
+      <% dropzone_large_id = '' %>
+      <% case location 
+        when :main %>
+        <% dropzone_large_id = '-main' %>
+      <% when :reply %>
+        <% dropzone_large_id = '-reply-' + local_assigns[:reply_to].to_s %>
+      <% when :responses %>
+        <% dropzone_large_id = '-responses' %>
+      <% end %>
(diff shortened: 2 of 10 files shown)


app/views/comments/_edit.html.erb
```diff
@@ -8,8 +8,8 @@
       #imagebar {width:100%;}
     </style>
 
-
-    <%= render :partial => "editor/toolbar" %>
+    <%# toolbar needs location & comment_id to make unique element IDs%>
```
jywarren, line 11: here you could use an HTML comment perhaps? `<!-- comment -->` just to simplify a little?

app/views/comments/_form.html.erb
```diff
@@ -45,24 +59,20 @@
...
-
               <span style="padding-right:4px;float:left;" class="d-none d-md-inline">
                 <%= raw translation('comments._form.drag_and_drop') %>
               </span>
-
               <!-- http://stackoverflow.com/questions/11235206/twitter-bootstrap-form-file-element-upload-button -->
-              <label for="fileinput">
-                <input id="fileinput" type="file" name="image[photo]" style="display:none;" />
+              <label for="fileinput-choose-one<%= dropzone_large_id %>">
+                <input id="fileinput-choose-one<%= dropzone_large_id %>" type="file" name="image[photo]" style="display:none;" />
```
jywarren, line 75: Will this change styling at all? Noting that sometimes adding unique classnames instead of ids can be better since elements are allowed to have more than one class but only one ID. I tend to use classes instead of IDs for that reason, but it's just a convention or habit, so no big deal here, just curious!

app/views/notes/_comment.html.erb
```diff
@@ -166,7 +166,16 @@
         <div id="comment-<%= comment.cid %>-reply-section">
           <% if current_user %>
             <div class="inline" id="question-comment-form">
-              <%= render :partial => "comments/form", :locals => { title: "Reply to this comment", reply_to: comment.cid, comment: false, placeholder: "Help the author refine their post, or point them at other helpful information on the site. Mention users by @username to notify them of this thread by email", url1: '/conduct', author: current_user.title, is_new_contributor:current_user.is_new_contributor? } %>
+              <%= render :partial => "comments/form", :locals => {
```
jywarren, line 169: so much more readable, thank you!

app/views/comments/_form.html.erb
```diff
@@ -19,16 +19,30 @@
...
+      <%# location can be: edit form, reply form, main comment form, note responses %>
+      <%= render :partial => "editor/toolbar", :locals => locals %>
 
       <% if local_assigns[:reply_to].present? %>
         <%= hidden_field_tag :reply_to, local_assigns[:reply_to] %>
       <% end %>
 
-      <div class="form-group dropzone">
-
+      <%# most comment forms have two dropzones: 1) small button, 2) large form that covers textarea %>
```
jywarren, line 35: There's a case possibly to be made that logic doesn't belong in a template, but i see why this may be an exception. But just to explore the possibility, would it be possible to wrap this logic in a function and do it in a single line in the template, with the logic tucked into the `application_helper.rb` (helpers is where template-specific logic often goes, by convention)?

But, i'm not feeling strongly either way, there are pros and cons - it might be more readable in pure Ruby in the helper, but it's also more work to go and find it...

jywarren (approved):
This looks amazing @noi5e -- i think it's pretty much good to merge unless you want to do more; my questions below are clarifications and "just checking" so i defer to you on a final call on any of them. If it's ready to merge just note that in a comment and we can do it! 

Fantastic work here!!! 👍

app/views/comments/_form.html.erb
noi5e: Oh, that's a really good point about styling. I did a simple search in VSCode for `#fileinput` and `input#fileinput` but it didn't come up with anything in `/assets/stylesheets`, so I think this is good. It looks as normal when I load it up in development too.

That's also a good point about classes vs. IDs. Since my ultimate goal is to be able to use `e.target` to set `$D.selected` upon file upload, I thought an ID made the most sense here... Good reminder though that I can use classes in other situations.

I was aware of the one ID per page rule. `dropzone_large_id` includes the comment ID, so it's going to resolve to a unique ID every time this form is rendered (`#fileinput-reply-large-22` or something like that)... Which means that there shouldn't be any conflicting IDs here, at least throughout `comments-list`!

app/views/comments/_form.html.erb
noi5e: Thank you! This is great feedback, and helps me learn more about Rails conventions (which I'm going to be learning for a while). I went ahead and put the logic in `application_helper.rb`, just for the sake of developing good habits.

noi5e:
I just pushed the commits, so this one's ready to merge!

jywarren:
Awesome would you like to take code climates suggestions or just resolve them? Thanks!!!

noi5e:
Just pushed the commits, it's ready (again) now!

jywarren:
Hooray 🎉👏👏

Issue #8945, opened 2021-01-05
Post not appearing on tag cards


jywarren:
https://publiclab.org/tag/classroom-community-science shows 1 note, but https://publiclab.org/tags shows none:

![image](https://user-images.githubusercontent.com/24359/103696493-71d85f80-4f6c-11eb-8a03-156a692c2278.png)

jywarren:
It shows up on the mini tag card here on the note page itself, in the sidebar:

![image](https://user-images.githubusercontent.com/24359/103696879-178bce80-4f6d-11eb-840f-d07779adb95e.png)


Could it be a caching issue?

jywarren:
Solving the incorrect count in #8946, but we still need to get the post to appear in the card...

jywarren:
Aha, never mind - it's not appearing because it was already shown higher on the page. cc @ebarry and @RuthNjeri  that we're correcting the count here but that the post not appearing in a listing is because we prevent posts from appearing twice on a page using this line of code:

https://github.com/publiclab/plots2/blob/71a25ec28bf6c75ad20e101f478cefca7b4354e4/app/views/tag/_topicCard.html.erb#L15-L16

See how it appeared just above in another topic card:

![image](https://user-images.githubusercontent.com/24359/103697880-89184c80-4f6e-11eb-9ac3-d38156934779.png)

PR #8946, opened 2021-01-05
switch from tag.count to Tag.tagged_note_count(tag.name) (Outreachy)

jywarren:
to fix https://github.com/publiclab/plots2/issues/8945

making it match https://github.com/publiclab/plots2/blob/71a25ec28bf6c75ad20e101f478cefca7b4354e4/app/views/tag/_miniCard.html.erb#L5

### DIFF
--- a/app/views/tag/_topicCard.html.erb
+++ b/app/views/tag/_topicCard.html.erb
@@ -32,7 +32,7 @@
       </div>
 
       <div class="card-footer" style="background-color: inherit; border:none;">
-        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= tag.count %> <%= translation('tag.index.posts') %> &raquo;</a>
+        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= Tag.tagged_node_count(tag.name) %> <%= translation('tag.index.posts') %> &raquo;</a>
         <div id="follow-unfollow-column-<%=tag.tid%>" style="float:right;margin:10px 0 10px 10px;">
           <% if current_user %>
             <% if !current_user.following(tag.name) %>


jywarren:
As this touches your code, i thought you might like to review it, @RuthNjeri ! Thank you!!!

RuthNjeri:
About the comment posted on issue #8945

> Aha, never mind - it's not appearing because it was already shown higher on the page. cc @ebarry and @RuthNjeri that we're correcting the count here but that the post not appearing in a listing is because we prevent posts from appearing twice on a page using this line of code:
> 
> https://github.com/publiclab/plots2/blob/71a25ec28bf6c75ad20e101f478cefca7b4354e4/app/views/tag/_topicCard.html.erb#L15-L16
> 
> See how it appeared just above in another topic card:
> 
> image

It's interesting to note that I removed that part of the code from the Dashboard V2 Topics card @jywarren, I only used the where.not to prevent the blog note from being reposted within the topics.. should I revert it to what was there before? It does make sense not to have posts reappearing.. 

https://github.com/publiclab/plots2/blob/d38da1c57cc63227dba142f42538562654822e10/app/views/dashboard_v2/_topicCard.html.erb#L26

app/views/tag/_topicCard.html.erb
```diff
@@ -32,7 +32,7 @@
       </div>
 
       <div class="card-footer" style="background-color: inherit; border:none;">
-        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= tag.count %> <%= translation('tag.index.posts') %> &raquo;</a>
+        <a style="padding-top:15px;text-decoration:underline;color:#808080;display:inline-block;" href="/tag/<%= tag.name %>"><%= Tag.tagged_node_count(tag.name) %> <%= translation('tag.index.posts') %> &raquo;</a>
```
RuthNjeri, line 35: I'm curious about the difference between `tag.count` and `Tag.tagged_node_count(tag.name)` does it mean that tag.count only shows the total number of posts including wikis and `Tag.tagged_node_count(tag.name)` only shows the number of posts?

Is it related to why the classroom-community-science post was showing 0 posts when there was 1?

jywarren: It is related but I think the main difference is that .count is stored in the database and has to be recalculated to be accurate, while the other is dynamically recalculated each time, which is less efficient but more accurate (as long as there are any possible ways to miss recalculating!). I think this could be refactored or renamed for clarity but it isn't a huge deal.

jywarren:
Agreed regarding double display on tag cards! This was because new posts would often appear many times if they had many popular tags.

jywarren:
Ok merged! We can create a follow-up if we want for .count but not urgent. Thanks!!

PR #8962, opened 2021-01-07
Tag sort defaults

### LINKED ISSUE #8941: /tags is (wrongly) alphabetically sorted...? need to better define defaults to see if there are issues
https://publiclab.org/tags?sort=followers&order=asc doesn't seem to sort by # of uses - 

do wiki edits/tags bump these up, or count?

what if we show wiki count?

Possibly related to https://github.com/publiclab/plots2/pull/8935 cc @RuthNjeri

cesswairimu:
Fixes #8941  (<=== Add issue number here)

Thanks!

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -9,7 +9,7 @@ class TagController < ApplicationController
     @title = I18n.t('tag_controller.tags')
     @paginated = true
     @order_type = params[:order].blank? || (params[:order] == "desc") ? "desc" : "asc"
-    @order_name = params[:order].blank? || (params[:order] == "asc") ? "asc" : "desc"
+    @order_type_for_name = params[:order].blank? || (params[:order] == "asc") ? "asc" : "desc"
     powertag_clause = params[:powertags] == 'true' ? '' : ['name NOT LIKE ?', '%:%']
 
     if params[:search]
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -9,6 +9,7 @@ class TagController < ApplicationController
     @title = I18n.t('tag_controller.tags')
     @paginated = true
     @order_type = params[:order].blank? || (params[:order] == "desc") ? "desc" : "asc"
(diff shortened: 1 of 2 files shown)


app/controllers/tag_controller.rb
```diff
@@ -8,7 +8,8 @@ def index
 
     @title = I18n.t('tag_controller.tags')
     @paginated = true
-    @order_type = params[:order].blank? || params[:order] == "desc" ? "asc" : "desc"
+    @order_type = params[:order].blank? || (params[:order] == "desc") ? "desc" : "asc"
```
jywarren, line 11: Hmm, what's the difference between `order_type` and `order_name`? or should `order_name` be like, `name`, `followers` etc?

jywarren: Maybe we need more specific variable names here...

jywarren:
Should we also just be changing the links shown on the page so the initial ones are correctly ordered? Or are we just changing the meaning of omitting the sort order as a param, so... oh, i see; we actually preset them with Ruby variables anyways:

https://github.com/cesswairimu/plots2/blob/47a7ab0af115b1cd38ac44915b34770c74975c52/app/views/tag/index.html.erb#L23-L26

app/controllers/tag_controller.rb
cesswairimu: `order_name` will be the order by  for 'names' so that we can default it to 'asc', while `order_type` will be used for 'uses' and  'followers for these we will default to 'desc' as we discussed...maybe i can use a better name `order_for_name` maybe?

cesswairimu:
The links default links are okay now they are reading from the controller, if not set `uses and `followers` are ordered `desc` and `names` as `asc`

cesswairimu:
@jywarren changed `order_name` to `order_type_for_name`  better?

jywarren:
Sure but... does it have to be a separate variable at all? I'm totally ok with it if there's a good reason, just wasn't sure. Thanks so much, Cess!

cesswairimu:
Yeah we need a different variable for asc ordering...since `order_type`. default is desc

jywarren:
OK, awesome, sounds good! Merged!

PR #8499, opened 2020-10-08
Bypass GitPod SQL .group() error by grouping only in production

jywarren:
Re: #8177 as temp solution:

```
@tags.group(:name) unless Rails.env.development? # for GitPod compatibility; #8117
```

PR #8979, opened 2021-01-08
attempts at solving MYSQL group issue by correcting SQL syntax

### LINKED ISSUE #8177: Duplicate records due to .group() changes for gitpod mysql fixes
I'm concerned that the expansion of `.group()` calls in https://github.com/publiclab/plots2/pull/8152 has created some incorrect outcomes. 

Responding to errors such as:

> Mysql2::Error: Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'plots.node.nid' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by: SELECT  node.nid, node.status, term_data.*, community_tags.* FROM `term_data` INNER JOIN `community_tags` ON `community_tags`.`tid` = `term_data`.`tid` INNER JOIN `community_tags` `node_tags_term_data_join` ON `node_tags_term_data_join`.`tid` = `term_data`.`tid` INNER JOIN `node` ON `node`.`nid` = `node_tags_term_data_join`.`nid` WHERE (node.status = 1) AND (community_tags.date > 1592278988) AND (name NOT LIKE '%:%') GROUP BY `term_data`.`name` ORDER BY count DESC LIMIT 24 OFFSET 0

we expanded the `.group()` terms by a lot: 

https://github.com/publiclab/plots2/pull/8152/files#diff-4e05ad0d64e6100656b63ad1e78f32c5R141

When doing this in #8048 i realized that doing something similar there (`.group('term_data.name, node.nid, term_data.tid, community_tags.nid, community_tags.uid, 
community_tags.date, community_tags.created_at, community_tags.updated_at')` in each instance of `.group()` in `tag_controller.rb`) caused tags to show up multiple times (duh!):

image

jywarren:
Attempt to fix #8177 in rework of #8499; still getting error:

`contains nonaggregated column 'plots.node.nid' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by`

Can't seem to get the `group()` to show up though...


app/controllers/users_controller.rb
```diff
@@ -122,27 +123,23 @@ def list
...
-
+    if params[:id]
+      @users = User.order(order_string)
+        .where('rusers.role = ?', params[:id])
+        .where('rusers.status = 1')
+    elsif @tagname_param
+      @users = User.where(id: UserTag.where(value: @tagname_param).collect(&:uid))
+    else
+      # recently active
+      group_by = Rails.env.development? ? false : 'rusers.id' # for GitPod compatibility; #8117
```
jywarren, line 134: here trying to call `group()` with either `false` or `rusers.id`... but it's not working it seems... or, maybe the `false` is working? Need to try running it carefully to diagnose, but i have to go now...

jywarren:
ok cool so i think that did work, but the reason we used `group()` in the first place was to get past this error:

```
ActionView::Template::Error (Mysql2::Error: In aggregated query without GROUP BY, expression
 #1 of SELECT list contains nonaggregated column 'plots.rusers.id'; this is incompatible wit
h sql_mode=only_full_group_by: SELECT  *, rusers.id, rusers.status, MAX(node_revisions.times
tamp) AS last_updated FROM `rusers` INNER JOIN `node` ON `node`.`uid` = `rusers`.`id` INNER 
JOIN `node_revisions` ON `node_revisions`.`nid` = `node`.`nid` WHERE (node_revisions.status 
= 1) AND (rusers.status = 1) ORDER BY last_updated DESC LIMIT 24 OFFSET 0):
```

jywarren:
This worked!!!

Issue #5650, opened 2019-05-02
Dragging image into wiki pages is broken


jywarren:
I'm having trouble dragging an image onto a wiki edit page; it returns an error on this line of dragdrop.js:

`($D.selected).find('#create_progress').eq(0).show();`

`Uncaught TypeError: Cannot read property 'find' of undefined`

I wonder if `$D.selected` was affected somehow? This is the kind of thing a system test (#5316) will help prevent!

Noting this could be linked to #5229

jywarren:
Yes! It's the same as this: https://github.com/publiclab/plots2/issues/5229#issuecomment-481370671

publiclab-mimi:
I'm currently unable to click to upload or drag an image onto a wiki page. I'm not sure if it's related. I'm not getting an error, just no response no matter how long I let the webpage sit open and "think" about it.

steviepubliclab:
Same issue here on /air-quality

publiclab-mimi:
I also tried this post clearing-cache, in a different browser, and on a different computer. Still no image uploading working for me.

jywarren:
This was fixed in #5657 for comment image uploading, but the fix didn't apply to wiki page image upload. Checking why now; I believe it is failing on `$E.wrap` in:

```js
  // wraps currently selected text in textarea with strings a and b
  wrap: function(a,b,args) {
    this.refresh();
    var len = $E.textarea.val().length;
```

That's here: https://github.com/publiclab/plots2/blob/a3e40c5cb20e94653de20a74e0ac9553c87f0d3a/app/assets/javascripts/editor.js#L43

It's initialized here:

https://github.com/publiclab/plots2/blob/a3e40c5cb20e94653de20a74e0ac9553c87f0d3a/app/views/editor/_editor.html.erb#L39

jywarren:
Yes!!! This changed, re-assigning `$E.textarea` to fix `reply-to-comment` functionality:

https://github.com/publiclab/plots2/pull/5321/files#diff-fd61d47ff16a49707d50fa5f9ae7c5f5R32

We need to make an exemption for wiki image upload.

jywarren:
Filed a fix, which I tested, at #5703!

jywarren:
Thanks, all!

PR #8987, opened 2021-01-11
Bug: Fix CLICK-to-Upload Image Cross-Wiring Issues

### LINKED ISSUE #8926: Issue: Add Unique HTML IDs to Comment Forms to Fix Image Upload Cross-Wiring Issues

noi5e:
Fixes #8926

I think the plan I described in the issue above is actually working! `e.target` is now accurate in the `start` `fileupload` functions. which means we can use it to assign `$D.selected`. That means that every time a user successfully uploads a file by clicking on the upload button, there's a much higher likelihood (if not 100%) that it's going to go into the appropriate box.

**Summary:**
- slight refactoring of `dragdrop.js`'s `drop` eventListener
- assign `$D.selected` to `e.target` in `dragdrop.js`'s `start` `fileupload` function
- some refactoring in `editor.js`'s `wrap` and `refresh` methods
- new system tests:
  - IMMEDIATE image CLICK to upload in **REPLY** comment form works 
  - IMMEDIATE image CLICK to upload in **EDIT** comment form works
  - image CLICK to upload into EDIT form isn't CROSS-WIRED with **MAIN** form
  - image CLICK to upload into EDIT form isn't CROSS-WIRED with **REPLY** form

---

### DIFF
--- a/app/assets/javascripts/dragdrop.js
+++ b/app/assets/javascripts/dragdrop.js
@@ -33,18 +33,19 @@ jQuery(function() {
   });
   $('.dropzone').on('drop',function(e) {
     // this 'drop' listener is also reused for pages with just one form, ie. /wiki/new
-    $D.selected = $(e.target).closest('div.comment-form-wrapper').eq(0);
+    const closestCommentFormWrapper = e.target.closest('div.comment-form-wrapper'); // this returns null if there is no match
     e.preventDefault();
     let params = {};
-    // $D.selected will look different for multi vs. single form pages, because of what $.closest returns:
-    //   multiple comments: { 0: the_closest_element }
-    //   /wiki/new: .comment-form-wrapper doesn't exist!
-    if ($D.selected.hasOwnProperty(0)) { // eg. jQuery finds a .comment-form-wrapper somewhere on the page.
-      params['textarea'] = $D.selected[0].querySelector('textarea').id // assign the ID of the textarea within the closest comment-form-wrapper
+    // there are no .comment-form-wrappers on /wiki/edit or /wiki/new
+    // these pages just have a single text-input form.
+    if (closestCommentFormWrapper) {
+      $D.selected = $(closestCommentFormWrapper);
+      // assign the ID of the textarea within the closest comment-form-wrapper
+      params['textarea'] = closestCommentFormWrapper.querySelector('textarea').id;
     } else {
       // default to #text-input
       // ideally there should only be one per page
-      params['textarea'] = 'text-input' 
+      params['textarea'] = 'text-input';
     }
     $E.initialize(params);
   });
@@ -52,6 +53,7 @@ jQuery(function() {
     e.preventDefault();
   });
 
+  // these functions are also used on /wiki/edit and /wiki/new pages
   $('.dropzone').each(function() {
     $(this).fileupload({
       url: "/images",
@@ -63,6 +65,13 @@ jQuery(function() {
           'nid':$D.nid
         },
         start: function(e) {
+          // 'start' function runs:
+          //   1. when user drag-and-drops image
+          //   2. when user clicks on upload button.
+          // for click-upload-button scenarios, it's important to set $D.selected here, because the 'drop' listener above doesn't run in those:
+          $D.selected = $(e.target).closest('div.comment-form-wrapper');
+          // the above line is redundant in drag & drop, because it's assigned in 'drop' listener too.
+          // on /wiki/new & /wiki/edit, $D.selected will = undefined from this assignment
           $('#imagebar .inline_image_drop').show();
           $('#create_progress').show();
           $('#create_uploading').show();
--- a/app/assets/javascripts/editor.js
+++ b/app/assets/javascripts/editor.js
@@ -1,13 +1,16 @@
 $E = {
   initialize: function(args) {
     args = args || {}
+    // title
+    $E.title = $('#title')
+    // textarea
     args['textarea'] = args['textarea'] || 'text-input'
     $E.textarea = $('#'+args['textarea'])
-    $E.title = $('#title')
+    $E.textarea.bind('input propertychange',$E.save)
+    // preview
     args['preview'] = args['preview'] || 'preview'
     $E.preview = $('#'+args['preview'])
-    $E.textarea.bind('input propertychange',$E.save)
-
+    
     marked.setOptions({
       gfm: true,
       tables: true,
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/dragdrop.js
```diff
@@ -33,25 +33,27 @@ jQuery(function() {
   });
   $('.dropzone').on('drop',function(e) {
     // this 'drop' listener is also reused for pages with just one form, ie. /wiki/new
-    $D.selected = $(e.target).closest('div.comment-form-wrapper').eq(0);
+    const closestCommentFormWrapper = e.target.closest('div.comment-form-wrapper'); // this returns null if there is no match
     e.preventDefault();
     let params = {};
```
noi5e, line 38: I rewrote this part for greater readability, using JavaScript's `closest` (as opposed to jQuery's method with the same name).

JavaScript's `closest` returns `null`, whereas jQuery's returns an object. I thought this would be more intuitive.

app/assets/javascripts/dragdrop.js
```diff
@@ -63,6 +65,13 @@ jQuery(function() {
           'nid':$D.nid
         },
         start: function(e) {
+          // 'start' function runs:
+          //   1. when user drag-and-drops image
+          //   2. when user clicks on upload button.
+          // for click-upload-button scenarios, it's important to set $D.selected here, because the 'drop' listener above doesn't run in those:
+          $D.selected = $(e.target).closest('div.comment-form-wrapper');
+          // the above line is redundant in drag & drop, because it's assigned in 'drop' listener too.
+          // on /wiki/new & /wiki/edit, $D.selected will = undefined from this assignment
```
noi5e, line 74: This is a major change: assign `$D.selected` in the `start` upload function.

As a reminder, this function runs on both `click` and `drop` image upload events. It's important to assign `$D.selected` here for `click` events, because the `drop` listener doesn't run for those.

I left a comment in the code about this, but I'm thinking it would be good down the road to not have to assign `$D.selected` twice in a row for `drop` events. Not sure how to do that yet.

app/assets/javascripts/editor.js
```diff
@@ -1,13 +1,16 @@
...
     args = args || {}
+    // title
+    $E.title = $('#title')
+    // textarea
     args['textarea'] = args['textarea'] || 'text-input'
     $E.textarea = $('#'+args['textarea'])
-    $E.title = $('#title')
+    $E.textarea.bind('input propertychange',$E.save)
+    // preview
     args['preview'] = args['preview'] || 'preview'
```
noi5e, line 11: No additions here, just rearranged stuff for greater readability. Sometime soon, I'd like to rewrite this part to use JavaScript ES6's default parameters: `initialize(textarea = 'text-input')`

app/assets/javascripts/editor.js
```diff
@@ -23,24 +26,25 @@ $E = {
...
-          $E.textarea = ($D.selected).find('textarea').eq(0);
-          $E.preview = ($D.selected).find('#preview').eq(0);
-          $E.textarea.bind('input propertychange',$E.save);
-      }
+    // textarea
+    $E.textarea = ($D.selected).find('textarea').eq(0);
+    $E.textarea.bind('input propertychange',$E.save);
+    // preview
+    $E.preview = ($D.selected).find('#preview').eq(0);
   },
```
noi5e, line 39: I didn't think these lines needed to be inside of a conditional checking for `$D.selected` (it made it little less readable). I moved the conditional into `$E.wrap`

app/assets/javascripts/editor.js
```diff
@@ -23,24 +26,25 @@ $E = {
...
   wrap: function(a,b,args) {
-    var isWiki = (window.location + '').includes('wiki');
-    if (!isWiki) this.refresh();
+    // this RegEx is: /wiki/ + any char not "/" + /edit
+    const isWikiCommentPage = (/\/wiki\/[^\/]+\/edit/).test(window.location.pathname);
+    // we don't need to refresh $E's values if we're on a page with a single comment form, ie. /wiki/new or /wiki/edit
+    if (window.location.pathname !== "/wiki/new" && !isWikiCommentPage && $D.selected) {
+      this.refresh();
+    }
     var len = $E.textarea.val().length;
```
noi5e, line 48: This is a pretty big change too. The `(window.location + '').includes('wiki')` part was originally done in #5650 to fix image upload in `/wiki/new` and `/wiki/{wiki name}/edit`... 

Unfortunately this had the side effect of running `refresh` on wiki comments pages (which `include('wiki')` in the `window.location`). I finally figured out this doesn't need to happen, because the `wiki#show` view just loads comments through `note#comments`. This used to make writing system tests more difficult because things ran differently for wiki comments.

The last note is that I'm thinking that `this.refresh();` may not need to be called at all. We could remove it entirely if we assign `$D.selected` in all the right places. But that's for another PR.

test/system/comment_test.rb
```diff
@@ -321,21 +448,11 @@ def get_path(page_type, path)
...
 
     test "#{page_type_string}: edit comment" do
+      nodes(node_name).add_comment({
+        uid: 2,
+        body: comment_text
+      })
       visit get_path(page_type, nodes(node_name).path)
-      # Create a comment
-      main_comment_form =  page.find('h4', text: /Post comment|Post Comment/).find(:xpath, '..') # title text on wikis is 'Post comment'
-      # fill out the comment form
```
noi5e, line 327: I'm trying to prune the system tests bit-by-bit. Every time we post a comment manually where we don't need to do so is extra runtime. Here `node.add_comment` can save some time.

jywarren (approved):
Looks great! Ready to merge? Thanks!!!

noi5e:
@jywarren Yep, ready to merge!

sagarpreet-chadha (approved):
Fantastic !!

jywarren:
🎉🎉🎉

PR #8995, opened 2021-01-12
Feature: Add Unique Toolbar Button IDs

noi5e:
Partial fix for #8478 

I think this will be relatively simple to review 🤞 

Adding unique IDs and common classes to the toolbar buttons used in comment forms.

My eventual goal is to assign click eventHandlers to each button. Having unique IDs will ensure that `e.target` will be accurate, as we've seen with bug fixes for image upload cross-wiring.

---

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -77,6 +77,20 @@ module ApplicationHelper
     end
   end
 
+  # used in views/editor/_toolbar.html.erb
+  def get_toolbar_element_id(location, reply_to, comment_id)
+    case location
+    when :main 
+      'main' 
+    when :reply 
+      'reply-' + reply_to.to_s
+    when :edit 
+      'edit-' + comment_id 
+    when :responses 
+      'responses' 
+    end
+  end
+
   # used in views/comments/_form.html.erb
   def get_textarea_id(location, reply_to)
     case location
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -77,6 +77,20 @@ def get_large_dropzone_id(location, reply_to)
     end
   end
 
+  # used in views/editor/_toolbar.html.erb
```
jywarren, line 80: thanks for the comment here, these can be tough to trace! There is the ability to make view-specific helpers like `editor_helper.rb` but often if they only contain a few lines, it can make them harder to find!

app/views/editor/_toolbar.html.erb
```diff
@@ -46,12 +83,40 @@
...
+      class="recover-button btn btn-outline-secondary btn-sm"
+    >
+      Recover
+    </a>
   </div>
 
   <div class="btn-group">
-    <a aria-label="Help" data-toggle="tooltip" title="Help" data-placement="bottom" class="btn btn-outline-secondary btn-sm" target="_blank" href="/wiki/authoring-help"><i class="fa fa-question-circle" ></i></a>
+    <a 
+      id="help-button-<%= toolbar_element_id %>"
```
jywarren, line 110: appreciate the formatting work here. Thanks!

jywarren (approved):

jywarren:
Awesome, thanks!!!

PR #9001, opened 2021-01-12
Try splitting GitHub Actions into parallel jobs

jywarren:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobs

example: https://github.com/jeffeb3/hello-world-composite-run-steps-action/blob/master/.github/workflows/main.yml

https://github.com/actions/runner/issues/646

Update: working!

image


jywarren:
Nice! Ok we may have to change branch protection rules.

image

jywarren:
It's not persisting the results of the setup step for subsequent steps to use. Maybe we need to make the setup step into an action like in 
https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-composite-run-steps-action

jywarren:
It looks like there may not be a way to persist dependencies across jobs... https://github.community/t/same-directory-for-multiple-jobs/18173/7

However maybe we can use gem caching for this? https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#caching-dependencies

jywarren:
ok, here is a truly parallel job setup, it repeats a lot of gem setup: https://boringrails.com/articles/building-a-rails-ci-pipeline-with-github-actions/

jywarren:
need to re-integrate: 

```

# Trying to cache yarn deps but they're in public/assets/**/**, but application.js is also 
# compiled there so we have to create an exception for that for any in-app js changes
#     - name: Cache yarn
#       uses: actions/cache@v2
#       with:
#         path: public/assets/**/**
#         key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
#         restore-keys: |
#           ${{ runner.os }}-yarn-
```

jywarren:
example: https://github.com/jeffeb3/hello-world-composite-run-steps-action/blob/master/.github/workflows/main.yml

jywarren:
Noting we can't yet use `uses:` within the local action: https://github.com/actions/runner/issues/646#issuecomment-749118072

jywarren:
also i think ruby gems are installed twice??

jywarren:
Looking good, fingers 🤞 -- i added Yarn back to functional tests as that was required too:

image

jywarren:
Functional tests need Redis -- i guess that's not accessible from the sub-action? I'll try moving it back...

```
=================================ERROR["test_test_digest_emails_to_moderators", #<Minitest::Reporters::Suite:0x000055d2436b4020 @name="AdminControllerTest">, 39.392088379000015]
 test_test_digest_emails_to_moderators#AdminControllerTest (39.39s)
Minitest::UnexpectedError:         Redis::CannotConnectError: Error connecting to Redis on 127.0.0.1:6379 (Errno::ECONNREFUSED)
            app/controllers/admin_controller.rb:403:in `test_digest_email_spam'
            test/functional/admin_controller_test.rb:644:in `block in <class:AdminControllerTest>'
```

jywarren:
This is running quite slow - 21m for system tests - but this is actually better than non-parallel runs in https://github.com/publiclab/plots2/actions/runs/494481028 for example. Merging and we can optimize further now!

jywarren:
About 30% faster, i think!

PR #9030, opened 2021-01-18
[Outreachy]Add Featured Topics on The Dashboard

### LINKED ISSUE #8956: Add Featured Topics On The Dashboard
This issue is part of #8513

It will fix:

- [ ] Add featured topics at the bottom of the page and at the Follow More Topics Section.

RuthNjeri:
Fixes #8956 

PR #9030 needs to be merged first.

Adding the featured tags
Screenshot 2021-01-18 at 20 03 18
Displaying the feature tags
Screenshot 2021-01-18 at 20 04 56
Screenshot 2021-01-18 at 20 15 06

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -55,6 +55,8 @@ class HomeController < ApplicationController
       exclude_tids = Tag.where(name: %w(blog everything)).pluck(:tid)
       # Not all tags have notes, some are wikis. A text will be displayed to show this.
       @pagy, @tag_subscriptions = pagy(current_user.subscriptions(:tag).includes(:tag).where.not(tid: exclude_tids))
+      # Trending tags exculding subscriptions
+      @trending_tags = trending
       render template: 'dashboard_v2/dashboard'
     else
       redirect_to '/research'
@@ -156,5 +158,10 @@ class HomeController < ApplicationController
       answer_comments
     ]
     response
+  end
+
+  def trending
+    exclude_tids = @tag_subscriptions.pluck(:tid)
+    Tag.trending.where.not(tid: exclude_tids)
   end
 end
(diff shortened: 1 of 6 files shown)


app/controllers/home_controller.rb
```diff
@@ -157,4 +159,9 @@ def activity
     ]
     response
   end
+
+  def trending
+    exclude_tids = @tag_subscriptions.pluck(:tid)
+    Tag.trending.where.not(tid: exclude_tids)
```
jywarren, line 165: very nice!!! so `@tag_subscriptions` is assumed, rather than using a parameter? It might be good to be explicit since this would fail without an existing value for `@tag_subscriptions`, right? But it's fine with me for now; maybe this could be filed as a follow-up PR?

test/system/dashboard_v2_test.rb
```diff
@@ -0,0 +1,22 @@
+require "application_system_test_case"
+
+class DashboardV2Test < ApplicationSystemTestCase
+  Capybara.default_max_wait_time = 60
+
+  test 'view trending topics' do
```
jywarren, line 6: This is great and it's awesome to see a system test suite as a testing space for the whole project showing up here!

jywarren (approved):

app/controllers/home_controller.rb
RuthNjeri: Yes, that's a valid point...I'll create a follow up Issue/PR for this

RuthNjeri: Hi @jywarren 👋🏾 

I tested this out locally and it seems that it does not break if the `@tag_subscriptions` is empty... it returns an empty array...
Tested below...
Screenshot 2021-01-20 at 21 22 03

However, I am assuming that maybe it would be best to have a sort of 'fail safe' if `@tag_subscriptions` is empty...
I came up with this code...let me know if this makes sense... I am not sure if it is redundant...because we know what will happen if the `@tag_subscriptions` is empty

```
  def trending
    exclude_tids = @tag_subscriptions.present? ? @tag_subscriptions.pluck(:tid) : []
    Tag.trending.where.not(tid: exclude_tids)
  end
```

Issue #9004, opened 2021-01-13
[Discussion] Refactor Comment Editor Using OOP


noi5e:
I wanted to open up an issue to collect ideas/observations I've had about the comment editor's state management. I'll leave this issue open for a few weeks to collect thoughts as they come up (if that's okay). I'm hoping this will help us to conceptualize a code overhaul. 

To start with, I wanted to give a bug as an example: 

**[Video of this bug in action](https://user-images.githubusercontent.com/4361605/104402558-6a382e00-550b-11eb-9278-368cd549a67d.mov)**
1. Go to a note/question/wiki page with multiple comments. 
2. Open up a _"Reply to this comment..."_ form.
3. Open up a separate _Edit Comment_ form.
4. Scroll back to the _Reply Comment_ form.
5. Press rich-text buttons in the _Reply Comment_ form.

✅ **Expected Behavior:** Rich-text changes go in the _Reply Comment_ form.
❌ **Actual Behavior:** They go in the _Edit Comment_ form.

### `Why This Happens`
It's because we set state (for the focused comment form) here in the edit comment form partial:
https://github.com/publiclab/plots2/blob/a3c87969663f482432cd4cbb43b95cd70726bfd4/app/views/comments/_edit.html.erb#L46-L52

But we don't do a similar thing when we open a reply comment form:
https://github.com/publiclab/plots2/blob/a3c87969663f482432cd4cbb43b95cd70726bfd4/app/views/notes/_comment.html.erb#L213-L217

### `The Bug Fix`
I currently have a PR open that fixes the above bug at #9003.

I intentionally decided not to fix the bug by putting lines to `set state` when opening a reply comment form.

After all, just because a user opens up a comment form, doesn't mean that they necessarily want the focus to be there.

I thought a better fix would be to `set state` when the rich-text button was clicked. That seems to signal stronger user intention.

### `State Management Idea`
So the idea is to start thinking about which user actions should set state, and which should not.

✅ **Set State**
- click on toolbar buttons
- drag-and-drop an image
- type text into a form
- load a new page (initial state should focus on the main comment form)

❌ **Don't Set State**
- open a reply form
- open an edit form

### `The Goal`
Just to summarize: some sort of object-oriented, React-like refactoring where as much of this code can be inside a single script, creating modular comment form objects, where the state can be clearly transparent at each step of the way. 🎈 

---
(This issue is part of the larger Comment Editor Overhaul Project with Outreachy. Refer to Planning Issue #9069 for more context)

[...]

PR #9035, opened 2021-01-18
Feature: Add setState Method in editor.js

noi5e:
This is a small part of a larger object-oriented refactoring for `editor.js`, see #9004 for details.

Basically the major change here is that I am creating a setState method for use in cases like this:
https://github.com/publiclab/plots2/blob/f0be879e5e0fb57590f4207a6ca059d92b203bdf/app/views/comments/_edit.html.erb#L42-L48

I think `$E.initialize` is used for two different cases, and I'm trying to make a greater distinction between the two:
1. `$E.initialize` when the document loads, and the editor **must** point toward the main comment form (or the only editor form on the page, in the case of `/wiki/new`, `/wiki/edit`, and `/features/new`
2. `$E.initialize` when we want the editor `textarea` to change on a multi-comment page, as in the above example. For this case, we can use the new `setState` method instead.

I have confidence this is going to help!!!

---

### DIFF
--- a/app/assets/javascripts/dragdrop.js
+++ b/app/assets/javascripts/dragdrop.js
@@ -29,12 +29,10 @@ jQuery(function() {
     });
 
     $(this).on('drop',function(e) {
-      const params = getEditorParams(e.target); // defined in editorHelper.js
+      const { textArea, preview, dSelected } = getEditorParams(e.target); // defined in editorHelper.js
       e.preventDefault();
-      if (params.hasOwnProperty('dSelected')) {
-        $D.selected = params['dSelected'];
-      }
-      $E.initialize(params);
+      if (dSelected) { $D.selected = dSelected; }
+      $E.setState(textArea, preview);
     });
 
     $(this).fileupload({
--- a/app/assets/javascripts/editor.js
+++ b/app/assets/javascripts/editor.js
@@ -71,7 +71,7 @@ $E = {
     var end = $E.textarea[0].selectionEnd;
     const fallbackParameterExists = args && args['fallback'];
     const newlineParameterExists = args && args['newline'];
-    var sel = fallbackParameterExists ? $E.textarea.val().substring(start, end) : args['fallback']; // // fallback if nothing has been selected, and we're simply dealing with an insertion point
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/dragdrop.js
```diff
@@ -29,12 +29,10 @@ jQuery(function() {
     });
 
     $(this).on('drop',function(e) {
-      const params = getEditorParams(e.target); // defined in editorHelper.js
+      const { textArea, preview, dSelected } = getEditorParams(e.target); // defined in editorHelper.js
```
noi5e, line 32: aka [ES6 destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

jywarren: Oohhhh!!! 🤗

jywarren (approved):
Looks great! I like the separation of functions!

app/assets/javascripts/editor.js
```diff
@@ -54,14 +53,19 @@ $E = {
...
-    if (window.location.pathname !== "/wiki/new" && !isWikiCommentPage && $D.selected) {
-      this.refresh();
-    }
+  wrap: function(a, b, args) {
+    // we only refresh $E's values if we are on a page using the rich-text editor (most pages).
+    // the legacy editor pages only have one editor form, unlike pages with multiple comments.
+    if (this.isRichTextEditor(window.location.pathname)) { this.refresh(); }
     var len = $E.textarea.val().length;
     var start = $E.textarea[0].selectionStart;
     var end = $E.textarea[0].selectionEnd;
```
noi5e, line 71: @jywarren codeclimate won't let this pass because of 'cognitive complexity' in the wrap function... I think the rules are stricter now since the recent changes in CI.

Honestly, I have a feeling that it's the older stuff later on in the method that's holding this back (conditionals like `args && args['fallback']`, a nested if statement inside `if (args && args['newline'])`. I'm not very familiar with this code yet, can you take a look and let me know if you have any ideas to simplify things here so this can pass?

noi5e: @jywarren Never mind! I got it to work. Tests are passing now too.

Ready to merge!

cesswairimu (approved):
Wow :rocket: Great job

jywarren:
Awesome work!!!!!

PR #9043, opened 2021-01-19
These changes to be able to run on GCLOUD

icarito:
This were changes required to deploy on GCLOUD.

They are pointing to environment variables mostly.
Beware that the SMTP_* variables break in production, we've tried this before.
We should agree if we can try to deliver directly to Google SMTP, and possibly build the array procedurally (the issue is that empty string changes the actionmailer default and causes issues with current setup).

Thanks!

### DIFF
--- a/config/database.yml.example
+++ b/config/database.yml.example
@@ -26,6 +26,7 @@ production:
   username: <%= ENV['DB_USER'] || 'plots' %>
   password: <%= ENV['DB_PASS'] || 'plots' %>
   database: <%= ENV['DB_NAME'] || 'plots' %>
+  socket: <%= ENV['DB_SOCKET'] || '' %>
   port: <%= ENV['DB_PORT'] || 3306 %>
   host: <%= ENV['DB_HOST'] || 'db' %>
   encoding: utf8mb4
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -68,12 +68,13 @@ Plots2::Application.configure do
 
   config.action_mailer.smtp_settings = {
   :address => ENV["SMTP_HOST"] || "smtp",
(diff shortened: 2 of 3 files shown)


jywarren:
This looks good - do you think we should do this in stages, coordinated so we're watching this in production? Or are you OK merging this and monitoring the deployment asynchronously? Thank you, Sebastian!!!

icarito:
> This looks good - do you think we should do this in stages, coordinated so we're watching this in production? Or are you OK merging this and monitoring the deployment asynchronously? Thank you, Sebastian!!!

I expect the SMTP variables will cause some trouble if deployed like this.
The default values of SMTP_USER, SMTP_PASS, SMTP_AUTH, and SMTP_STLS break current configuration.

We have two options:
 - figure out correct values of the above for current configuration (plain local MTA)
 - point directly to deliver mail thru google (which our MTA ultimately does too)

The tradeoff is our [`/smtp_test`](https://github.com/publiclab/plots2/blob/main/app/controllers/admin_controller.rb#L377) monitoring route, which we'll loose as it's currently implemented because it doesn't support TLS/SSL.

jywarren:
Would you like me to try to refactor the smtp route for SSL? Is that at the Rails application level?

icarito:
Hi Jeff,
I think what needs to be refactored is the way we build the `config.action_mailer.smtp_settings` hash.
The reason is that if we set `user_name` and `password`, `authentication` hash values to `''` the result is not the same as when we omit these hash values.
So something like pseudocode:
```
if ENV[key]:
  config.action_mailer.smtp_settings[key] = ENV[key]
```
For each instance.
About the SMTP test, I believe it is not very effective to be honest so I would just drop it. Sentry report *should* be enough.

jywarren (approved):

PR #9056, opened 2021-01-20
Incoming mail monitoring/logging

jywarren:
To monitor uptime for processing incoming mail, which is later processed by:

https://github.com/publiclab/plots2/blob/38e37eb20602d799ab935e4e281f32782e856fda/app/models/comment.rb#L235

This logs incoming messages to `log/mailman.log`

OK! Now it also logs /after/ processing the comment, to confirm it was received. 

And if it receives any email entitled exactly `Incoming mail parsing test`, it logs that specially as well with the timestamp.

### DIFF
--- a/script/mailman_server
+++ b/script/mailman_server
@@ -28,6 +28,7 @@ Mailman::Application.run do
 #       retries ||= 0
       # do something with the database that does not fetch mails yet (let's try to ensure we have a MySQL connection first)
       puts Comment.last.inspect
+      Mailman.logger.info "Mail received from #{mail.from.first}: '#{mail.subject}' - '#{mail&.html_part}'"
       Comment.receive_mail(message)
 #     rescue Exception => e
 #       Mailman.logger.error "Exception occurred while receiving message:\n#{message}"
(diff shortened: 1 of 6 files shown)


script/mailman_server
```diff
@@ -28,6 +28,7 @@ Mailman::Application.run do
 #       retries ||= 0
       # do something with the database that does not fetch mails yet (let's try to ensure we have a MySQL connection first)
       puts Comment.last.inspect
+      Mailman.logger.info "Mail received from #{mail.from.first}: '#{mail.subject}' - '#{mail&.html_part}'"
       Comment.receive_mail(message)
```
jywarren, line 32: we can add more after the mail is processed, too, perhaps, to confirm when it's competed. Could we automatically periodically send mail in and filter it with a special subject line, say, once per day? 

```suggestion
      if message.subject == "Incoming mail parsing test"
        Mailman.logger.info "Received incoming mail parsing test at #{Time.now.to_s}"
      end
      last_comment = Comment.last
      Comment.receive_mail(message)
      if Comment.last.id != last_comment.id
        Mailman.logger.info "Comment created from #{mail.from.first}: '#{mail.subject}''"
      end
```

script/mailman_server
```diff
@@ -28,6 +28,7 @@ Mailman::Application.run do
 #       retries ||= 0
       # do something with the database that does not fetch mails yet (let's try to ensure we have a MySQL connection first)
       puts Comment.last.inspect
+      Mailman.logger.info "Mail received from #{mail.from.first}: '#{mail.subject}' - '#{mail&.html_part}'"
```
jywarren, line 31:
```suggestion
      Mailman.logger.info "Mail received from #{message.from.first}: '#{message.subject}' - '#{message&.html_part}'"
```

PR #9062, opened 2021-01-21
Add Unique IDs to Comment Forms

noi5e:
Related to #9004 (Object-Oriented Refactor of Comment Editor)

See this comment for more context: https://github.com/publiclab/plots2/issues/9004#issuecomment-765004416

---
(This issue is part of the larger Comment Editor Overhaul Project with Outreachy. Refer to Planning Issue #9069 for more context)

PR #9067, opened 2021-01-23
Integrate Edit Comment Form JavaScript with editor.js

noi5e:
Part of an object-oriented refactoring of `editor.js`. See #9004 for more details.

Please merge #9062 first!

**SUMMARY:**
In the past, there were two separate kinds of comment form partials:
1. REPLY and MAIN comment forms rendered at `/app/views/comments/_form.html.erb`
2. EDIT comment forms rendered at `/app/views/comments/_edit.html.erb`

To boot, these two different comment form types relied on two separate sources for JavaScript editor functions like image upload, and rich-text editing:
1. REPLY and MAIN comment forms rely on `editor.js`.
2. EDIT comment forms rely on a `<script>` tag within the partial [here](https://github.com/publiclab/plots2/blob/38e37eb20602d799ab935e4e281f32782e856fda/app/views/comments/_edit.html.erb#L41-L114)

This made code maintainability more difficult because if a change was made in `editor.js`, it had to account for the `<script>` tag.

This PR integrates the two JS sources. Now `_edit.html.erb` relies on `editor.js`! Hooray!

**CHANGES:**
1. I changed CSS IDs in `_edit.html.erb` so that `editor.js` can access them:
  - `#c123preview` is now `#comment-preview-edit-123`
  - `#c123text` is now `#text-input-edit-123`
2. Deleted `<script>` tag in `_edit.html.erb` so now it relies on `editor.js` instead.

**Slightly unrelated changes:**
1. I deleted `setInit` function definition and calls in `_edit.html.erb`. `setInit` was kind of like a way that we used to setState in EDIT comment forms before. The problem is that we called it whenever the user opened up an edit comment form, or clicked on Preview, which I don't think is a good place to setState. See [this comment](https://github.com/publiclab/plots2/issues/9004#issue-784759413) for an explanation. 
2. Add a `.dropzone` class to `#c123div`, mainly so `editor.js` can toggle preview (Yes, we toggle on `.dropzone` when the user clicks on "Preview")
3. Move `$E.initialize()` call from `_form.html.erb` (REPLY and MAIN comment form partial) into higher level partials. This is so that `editor.js` isn't initialized 5 times on a page with 5 comments (this was happening before! :scream:)

---

### DIFF
--- a/app/views/comments/_edit.html.erb
+++ b/app/views/comments/_edit.html.erb
@@ -12,119 +12,48 @@
     <!-- toolbar needs location & comment_id to make unique element IDs -->
     <%= render :partial => "editor/toolbar", :locals => { :comment_id => comment.id.to_s, :location => :edit } %>
     
-    <div id="c<%= comment.id%>div" class="form-group">
-    <textarea aria-label="Edit Comment" onFocus="editing=true" name="body" class="form-control" id="c<%= comment.id%>text" rows="6" cols="40" placeholder="<%= placeholder %>"><%= !(comment.is_a?Answer) ? comment.comment : comment.content %></textarea>
-    <div class="imagebar">
-      <div id="c<%= comment.id%>progress" style="display:none;" class="progress progress-bar-container active pull-right">
-        <div id="c<%= comment.id%>progress-bar" class="progress-bar progress-bar-striped progress-bar-animated" style="width: 0%;"></div>
-      </div>
-      <p>
-        <span id="c<%= comment.id%>uploading" class="uploading uploading-text" style="display:none;">
-          <%= translation('comments._edit.uploading') %>
-        </span>
-        <span id="c<%= comment.id%>prompt" class="prompt choose-one-prompt-text">
-          <span style="padding-right:4px;float:left;" class="hidden-xs">
-          <%= raw translation('comments._edit.drag_and_drop') %>
+    <div id="c<%= comment.id%>div" class="form-group dropzone">
+      <textarea aria-label="Edit Comment" onFocus="editing=true" name="body" class="form-control" id="text-input-edit-<%= comment.id%>" rows="6" cols="40" placeholder="<%= placeholder %>"><%= !(comment.is_a?Answer) ? comment.comment : comment.content %></textarea>
+      <div class="imagebar">
+        <div id="c<%= comment.id%>progress" style="display:none;" class="progress progress-bar-container active pull-right">
+          <div id="c<%= comment.id%>progress-bar" class="progress-bar progress-bar-striped progress-bar-animated" style="width: 0%;"></div>
+        </div>
+        <p>
+          <span id="c<%= comment.id%>uploading" class="uploading uploading-text" style="display:none;">
+            <%= translation('comments._edit.uploading') %>
           </span>
-          <label id="c<%= comment.id%>input_label" for="c<%= comment.id%>input">
-            <input id="c<%= comment.id%>input" type="file" name="image[photo]" style="display:none;" />
-            <a class="hidden-xs"><%= translation('comments._edit.choose_one') %></a>
-            <span class="visible-xs">
-              <i class="fa fa-upload"></i>
-              <a><%= translation('comments._edit.upload_image') %></a>
+          <span id="c<%= comment.id%>prompt" class="prompt choose-one-prompt-text">
+            <span style="padding-right:4px;float:left;" class="hidden-xs">
+            <%= raw translation('comments._edit.drag_and_drop') %>
             </span>
-          </label>
-        </span>
-      </p>
+            <label id="c<%= comment.id%>input_label" for="c<%= comment.id%>input">
+              <input id="c<%= comment.id%>input" type="file" name="image[photo]" style="display:none;" />
+              <a class="hidden-xs"><%= translation('comments._edit.choose_one') %></a>
+              <span class="visible-xs">
+                <i class="fa fa-upload"></i>
+                <a><%= translation('comments._edit.upload_image') %></a>
+              </span>
+            </label>
+          </span>
+        </p>
+        </div>
       </div>
-    </div>
-      <script type="application/javascript">
-        function setInit(id) {
-          const textArea = 'c'+id+'text';
-          const preview = 'c'+id+'preview';
-          $E.setState(textArea, preview);
-        }
-
-        $('#c<%= comment.id%>div').on('dragover',function(e) {
-          e.preventDefault();
-          $('#c<%= comment.id%>div').addClass('hover');
-        });
-
-        $('#c<%= comment.id%>div').on('dragout',function(e) {
-          $('#c<%= comment.id%>div').removeClass('hover');
-        });
-
-        $('#c<%= comment.id%>div').on('drop',function(e) {
-          e.preventDefault();
-          $D.selected = $(e.target).closest('div.comment-form-wrapper').eq(0);
-          setInit(<%= comment.id %>);
-        });
-
-        $('#c<%= comment.id%>div').fileupload({
-          url: "/images",
-          paramName: "image[photo]",
-          dropZone: $('#c<%= comment.id%>div'),
-          dataType: 'json',
-          formData: {
-            'uid':<%= current_user.uid %>,
-            'nid':<%= comment.uid %>
-          },
-          start: function(e) {
-            $('#c<%= comment.id%>progress').show()
-            $('#c<%= comment.id%>uploading').show()
-            $('#c<%= comment.id%>prompt').hide()
-            $('#c<%= comment.id%>div').removeClass('hover');
-          },
-          done: function (e, data) {
-            $('#c<%= comment.id%>progress').hide()
-            $('#c<%= comment.id%>progress #c<%= comment.id%>progress-bar').css('width', 0);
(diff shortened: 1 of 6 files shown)


app/views/notes/_comment.html.erb
```diff
@@ -53,7 +53,7 @@
         <i data-toggle="tooltip" title="This comment was posted by email." class="fa fa-envelope"></i>
       <% end %>
       <% if current_user && comment.uid == current_user.uid %>
-        <a aria-label="Edit comment" class="btn btn-outline-secondary btn-sm" id="edit-comment-btn" href="javascript:void(0)" onClick="$('#c<%= comment.cid %>edit').toggle();$('#c<%= comment.cid %>show').toggle();setInit(<%= comment.cid %>);">
+        <a aria-label="Edit comment" class="btn btn-outline-secondary btn-sm edit-comment-btn" href="javascript:void(0)" onClick="$('#c<%= comment.cid %>edit').toggle();$('#c<%= comment.cid %>show').toggle();">
```
noi5e, line 56: Small, unrelated change: I changed this to a class instead of an ID.

app/views/comments/_edit.html.erb
```diff
@@ -12,119 +12,48 @@
...
-              $('#c<%= comment.id %>edit .preview-btn').button(this.previewing ? 'previewing' : 'reset');
-              $E.generate_preview('c<%= comment.id %>preview',$('#c<%= comment.id %>text').val())">
+          <a 
+            class="btn btn-default preview-btn"  
+            data-previewing-text="Hide Preview" 
+            onClick="
+              $E.setState(
+                'text-input-edit-<%= comment.id%>', 
+                'comment-preview-edit-<%= comment.id %>'
+              );
```
noi5e, line 53: Because `editor.js` has a new `setState` method, I can call it here instead of the old inline JavaScript.

app/views/comments/_edit.html.erb
```diff
@@ -12,119 +12,48 @@
...
-        <div id="c<%= comment.id%>progress-bar" class="progress-bar progress-bar-striped progress-bar-animated" style="width: 0%;"></div>
-      </div>
-      <p>
-        <span id="c<%= comment.id%>uploading" class="uploading uploading-text" style="display:none;">
-          <%= translation('comments._edit.uploading') %>
-        </span>
-        <span id="c<%= comment.id%>prompt" class="prompt choose-one-prompt-text">
-          <span style="padding-right:4px;float:left;" class="hidden-xs">
-          <%= raw translation('comments._edit.drag_and_drop') %>
+    <div id="c<%= comment.id%>div" class="form-group dropzone">
```
noi5e, line 15: Add the class `.dropzone` here because this is what `editor.js` uses to toggle preview.

app/views/comments/_edit.html.erb
```diff
@@ -12,119 +12,48 @@
...
-            $('#c<%= comment.id%>prompt').show()
-            var is_image = false
-            if (data.result['filename'].substr(-3,3) == "jpg") is_image = true
-            if (data.result['filename'].substr(-4,4) == "jpeg") is_image = true
-            if (data.result['filename'].substr(-3,3) == "png") is_image = true
-            if (data.result['filename'].substr(-3,3) == "gif") is_image = true
-            if (data.result['filename'].substr(-3,3) == "JPG") is_image = true
-            if (data.result['filename'].substr(-4,4) == "JPEG") is_image = true
-            if (data.result['filename'].substr(-3,3) == "PNG") is_image = true
-            if (data.result['filename'].substr(-3,3) == "GIF") is_image = true
```
noi5e, line 91: I kind of just deleted this code, but can definitely save some of it by either:
1. placing it back in the partial, but commenting it out
2. saving choice lines and moving it to `editorToolbar.js` (which used to be `dragdrop.js`)

Like, here's a similar thing happening in `editorToolbar.js`:
https://github.com/publiclab/plots2/blob/d2e249dcc9bc66f9b4523caeb9b345409bcff062/app/assets/javascripts/editorToolbar.js#L103-L107

Which JS is preferable in this instance?

app/views/comments/_edit.html.erb
```diff
@@ -12,119 +12,48 @@
...
-
-            if (is_image) {
-              image_url = data.result.url.split('?')[0];
-              orig_image_url = image_url.replace('medium','original');
-              $E.wrap('[![',']('+image_url+')]('+orig_image_url+')', {'newline': true, 'fallback': data.result['filename']});
```
noi5e, line 96: For example Jeffrey you were saying in our meeting today that these lines that are resizing the image display to be smaller might be useful to keep.

app/views/comments/_edit.html.erb
jywarren: This seems a little odd sounding -  isn't `dropzone` supposed to mean the area we can drag a file/image into? Coudl this be an artifact from when preview and drag-drop were implemented at the same time, and might we think of a more legible classname or of separating preview and drag-drop? 

It kind of makes sense that we remove `.dragdrop` when we turn on preview, maybe... so you can't dragdrop onto the preview... but I don't know, shouldn't it just hide the dragdrop and not remove the class? This suggests to me that we should separate the functions and namings to make it more legible and less interdependent, whether or not I'm really understanding what's happening in the current model. What do you think?

Thanks, @noi5e !!!

app/views/comments/_edit.html.erb
jywarren: ` if (['gif', 'GIF', 'jpeg', 'JPEG', 'jpg', 'JPG', 'png', 'PNG'].includes(extension)) `

this seems like a MUCH more compact way to do this. And, `editorToolbar.js` has the CSV detection too! Nice! Let's just rely on that more refined code! 👍

app/views/comments/_edit.html.erb
jywarren: Looking at `editorToolbar.js` i think this already exists there so we're all good!

https://github.com/publiclab/plots2/blob/d2e249dcc9bc66f9b4523caeb9b345409bcff062/app/assets/javascripts/editorToolbar.js#L108-L112

jywarren:
Looks like a system test failure -- comment system tests -- actually really reassuring that these are getting triggered!!!

```

ERROR CommentTest#test_note:_IMMEDIATE_image_SELECT_upload_into_EDIT_comment_form (433.92s)
Minitest::UnexpectedError:         NoMethodError: undefined method `[]' for nil:NilClass
            test/system/comment_test.rb:344:in `block (2 levels) in <class:CommentTest>'

==============[Screenshot]: tmp/screenshots/failures_test_question:_IMMEDIATE_image_SELECT_upload_into_EDIT_comment_form.png
ERROR CommentTest#test_question:_IMMEDIATE_image_SELECT_upload_into_EDIT_comment_form (687.53s)
Minitest::UnexpectedError:         NoMethodError: undefined method `[]' for nil:NilClass
            test/system/comment_test.rb:344:in `block (2 levels) in <class:CommentTest>'

====[Screenshot]: tmp/screenshots/failures_test_wiki:_IMMEDIATE_image_SELECT_upload_into_EDIT_comment_form.png
ERROR CommentTest#test_wiki:_IMMEDIATE_image_SELECT_upload_into_EDIT_comment_form (833.77s)
Minitest::UnexpectedError:         NoMethodError: undefined method `[]' for nil:NilClass
            test/system/comment_test.rb:344:in `block (2 levels) in <class:CommentTest>'
```

app/views/comments/_edit.html.erb
noi5e: Good to know! Finished resolving merge conflicts about an hour ago. I restored the code and left it commented out.

noi5e:
> Looks like a system test failure -- comment system tests -- actually really reassuring that these are getting triggered!!!

Yep! My first PR got merged into `main` this morning and created some merge conflicts here. I just resolved some of them, but have to update some selectors in the system tests. This was caused because `#c123preview` had become `#comment-preview-edit-123`. There might be a few more instances of this; doing my best to update everything.

noi5e:
@jywarren Tests passing again, ready to merge!

jywarren (approved):

app/views/comments/_edit.html.erb
noi5e: @jywarren I am totally in agreement that `.dropzone` is not really the best class to refer to in this instance. I think the reason this was done in the past has to do with this:
103472308-f5ae0400-4d40-11eb-8d18-88099b4addfd

What we're toggling on is `B` in the above image. I think in the past, `.dropzone` was toggled for preview because it was just a way of referring to _"The visible, largest part of the comment form"_. I acknowledge it's weird, but I was just following the precedent of older code.

For example, it's happening here in this legacy code in `editor.js`:
https://github.com/publiclab/plots2/blob/b3009791076b15c9da81d58aec3de2a7a2b0f784/app/assets/javascripts/editor.js#L141-L146

I think the best thing to do moving forward is to add another CSS ID/class 😢 that has the same coverage as `.dropzone` in this instance, and toggle preview on **that**. Adding it to my planning issue so I don't forget to do this.

jywarren: Oh, thanks for the explanation! This makes total sense and is the kind of thing that is hard to plan for in the moment, but is really helpful down the road later. A second class sounds like just the right solution!

Issue #9069, opened 2021-01-23
Comment Editor Overhaul Project [Outreachy]


noi5e:
### **_(Last updated Saturday January 23rd)_**

Here's a planning issue that I made to cover my 13-week internship with Outreachy and Public Lab!

I will update this comment at least once a week to show my progress. I'm also compiling issues relevant to the Comment Editor in this already existing [Milestone](https://github.com/publiclab/plots2/milestone/28). 

## `Week Eight`
I'm starting to work on refactoring comment editor code:

### Open PRs (merge from top-to-bottom)
1. Add Unique IDs to Comment Forms ❗ **[open PR: #9062]** **(merge first!)**
  - [x] standardize `<form>` and `preview` IDs in reply and main comment forms
2. Integrate Edit Comment Form JS with editor.js ❗ **[open PR: #9067]**
  - [x] delete `<script>` in `_edit.html.erb` and point it toward `editor.js` instead
  - [x] standardize edit form's `textarea` and `preview` IDs for `editor.js`
  - [x] remove `setInit` call on `#edit-btn` click
  - [x] move `$E.initialize` to higher-level views
 
### Start Refactoring Comment Editor Code:
- object-oriented refactor of `editor.js`: 💬 **[issue #9004]**
- [ ] assign `$D.selected` using unique IDs, not `div.comment-form-wrapper`
  - [ ] remove `$D.selected` altogether?
  - [ ] possible to remove `$E.refresh` call in `$E.wrap`?
  - [ ] remove `handleClick()` function and call when opening reply forms.
- [ ] `$E.initialize({ default main editor form ID })`
- [ ] refactor save & recover methods
- [ ] remove functionality for DRAG onto edit form's `.dropzone` to upload image

## `Coming Up Soon`
- continue filling out comment system tests 🔍 **[issue #8810]**:
  - [ ] **priority:** rich-text changes on freshly posted comments
  - [ ] **priority:** image upload for freshly posted comments
  - [ ] editing existing comments
  - [ ] editing existing comment replies
  - [ ] editing freshly posted comment replies
  - [ ] preview buttons working correctly?
  - [ ] unit tests for new wiki, note, question: `$E.textarea` must initialize correctly
  - [ ] image progress bars show on `/wiki/new` and `/wiki/edit` routes
- responsive design: `choose one` upload in comment forms gets set to `display: none` on window resizing
  - this causes inconsistencies with image upload
- toggle preview in edit comments isn't selecting the whole form
[...]

PR #9104, opened 2021-01-31
Refactor editor.js with Object-Oriented Principles

noi5e:
See #9004

I think that went a lot easier than I expected! This seems to work okay testing in the local environment, lets see how the system tests do. Will update this comment if there's anything significant.

**UPDATE:** 

Ready to merge! 🙌 

I know this looks like a lot of commits, but most of them were just to pass codeclimate. Is it just me or did it get really strict all of a sudden?

---

### DIFF
--- a/app/assets/javascripts/editor.js
+++ b/app/assets/javascripts/editor.js
@@ -1,13 +1,21 @@
-// jQuery (document).ready function:
-
-$E = {
-  initialize: function() {
-    // call setState with no parameters, aka. default parameters.
-    // default parameters point toward either:
-    //   1. the comment form at the bottom of multi-comment wikis/questions/research notes
-    //   2. the only editor form on /wiki/new and /wiki/edit
-    $E.setState();
-    
+class Editor {
+  // default parameters reference the IDs of:
+  //   1. main comment form in multi-comment wikis, questions, & research notes.
+  //   2. the only editor form on /wiki/new and /wiki/edit
+  constructor(textarea = "text-input", preview = "comment-preview-main", title = "title") {
+    this.textarea = textarea;
+    this.preview = preview;
+    this.title = title;
+    this.previewing = false;
+    this.previewed = false;
+    this.templates = {
+      'blog': "## The beginning\n\n## What we did\n\n## Why it matters\n\n## How can you help",
+      'default': "## What I want to do\n\n## My attempt and results\n\n## Questions and next steps\n\n## Why I'm interested",
+      'support': "## Details about the problem\n\n## A photo or screenshot of the setup",
+      'event': "## Event details\n\nWhen, where, what\n\n## Background\n\nWho, why",
+      'question': "## What I want to do or know\n\n## Background story"
+    };
+      
     marked.setOptions({
       gfm: true,
       tables: true,
@@ -23,36 +31,36 @@ $E = {
         return code;
       }
     });
-  },
-  setState: function(textarea = 'text-input', preview = 'comment-preview-main', title = 'title') {
+  }
+  setState = function(textarea = 'text-input', preview = 'comment-preview-main', title = 'title') {
     $E.title = $('#' + title + 'title'); // not sure why this exists? seems like $E.title is always #title
     $E.textarea = $('#' + textarea);
     $E.textarea.bind('input propertychange', $E.save);
     $E.preview = $('#' + preview);
-  },
-  is_editing: function() {
-    return ($E.textarea[0].selectionStart == 0 && $E.textarea[0].selectionEnd == 0)
-  },
-  refresh: function() {
+  };
+  // code seems unused, commenting out for now.
+  // is_editing = function() {
+  //   return ($E.textarea[0].selectionStart == 0 && $E.textarea[0].selectionEnd == 0)
+  // };
+  refresh = function() {
     // textarea
     $E.textarea = ($D.selected).find('textarea').eq(0);
     $E.textarea.bind('input propertychange',$E.save);
     // preview
     $E.preview = ($D.selected).find('.comment-preview').eq(0);
-  },
-  isRichTextEditor: function(url) {
-    // this RegEx matches three different cases where the legacy editor is still used:
+  };
+  isSingleFormPage = function(url) {
+    // this RegEx matches three different pages where only one editor form is present (instead of multiple comment forms):
     //   1. /wiki/new
     //   2. /wiki/{wiki name}/edit
     //   3. /features/new
-    const legacyEditorPath = RegExp(/\/(wiki|features)(\/[^\/]+\/edit|\/new)/);
-    return !legacyEditorPath.test(url); // if we're not on one of these pages, we are using the rich-text editor.
-  },
+    const singleFormPath = RegExp(/\/(wiki|features)(\/[^\/]+\/edit|\/new)/);
+    return singleFormPath.test(url);
+  };
   // wraps currently selected text in textarea with strings a and b
-  wrap: function(a, b, args) {
-    // we only refresh $E's values if we are on a page using the rich-text editor (most pages).
-    // the legacy editor pages only have one editor form, unlike pages with multiple comments.
-    if (this.isRichTextEditor(window.location.pathname)) { this.refresh(); }
+  wrap = function(a, b, args) {
+    // we only refresh $E's values if we are on a page with multiple comments
+    if (this.isSingleFormPage(window.location.pathname) === false) { this.refresh(); }
     var len = $E.textarea.val().length;
     var start = $E.textarea[0].selectionStart;
     var end = $E.textarea[0].selectionEnd;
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/editor.js
```diff
@@ -1,13 +1,21 @@
...
-    // default parameters point toward either:
-    //   1. the comment form at the bottom of multi-comment wikis/questions/research notes
-    //   2. the only editor form on /wiki/new and /wiki/edit
-    $E.setState();
-    
+class Editor {
+  // default parameters reference the IDs of:
+  //   1. main comment form in multi-comment wikis, questions, & research notes.
+  //   2. the only editor form on /wiki/new and /wiki/edit
+  constructor(textarea = "text-input", preview = "comment-preview-main", title = "title") {
```
noi5e, line 5: These default params are just a stepping stone for now. 

Since the element IDs are all standardized now, it's easier to do a rewrite so that constructor takes just **one** parameter. For example: 

`$E = new Editor('reply-1234');` means that the editor object has access to both `#comment-preview-**reply-1234**` and `#text-input-**reply-1234**`. 

Much simpler.

app/assets/javascripts/editor.js
```diff
@@ -23,36 +31,36 @@ $E = {
...
-  },
-  is_editing: function() {
-    return ($E.textarea[0].selectionStart == 0 && $E.textarea[0].selectionEnd == 0)
-  },
-  refresh: function() {
+  };
+  // code seems unused, commenting out for now.
+  // is_editing = function() {
+  //   return ($E.textarea[0].selectionStart == 0 && $E.textarea[0].selectionEnd == 0)
+  // };
```
noi5e, line 44: Couldn't find this code used anywhere doing a search for `is_editing` in VSCode. Will delete in the next few PRs if the tests keep passing.

app/assets/javascripts/editor.js
```diff
@@ -23,36 +31,36 @@ $E = {
         return code;
       }
     });
-  },
-  setState: function(textarea = 'text-input', preview = 'comment-preview-main', title = 'title') {
+  }
+  setState = function(textarea = 'text-input', preview = 'comment-preview-main', title = 'title') {
```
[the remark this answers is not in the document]
noi5e: I'm a little stumped about this error here. Maybe the `constructor` function before it needs a semicolon after the curly brace, like `};`? Will try that in the morning.

app/assets/javascripts/editor.js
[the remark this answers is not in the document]
noi5e: I think I will have to delete templates to get this PR to pass this check.

app/assets/javascripts/editor.js
```diff
@@ -1,13 +1,21 @@
...
-
-$E = {
-  initialize: function() {
-    // call setState with no parameters, aka. default parameters.
-    // default parameters point toward either:
-    //   1. the comment form at the bottom of multi-comment wikis/questions/research notes
-    //   2. the only editor form on /wiki/new and /wiki/edit
-    $E.setState();
-    
+class Editor {
```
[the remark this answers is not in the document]
noi5e: I could get around this by commenting out the `h1`, and `h3` - `h7` methods that aren't currently used anywhere in production.

On the other hand, I could also delete template functionality, which is on my to-dos.

app/assets/javascripts/editor.js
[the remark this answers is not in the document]
noi5e: Figured it out... In OOP JS, object methods are written like `setState()` instead of `setState = function()`.

app/assets/javascripts/editor.js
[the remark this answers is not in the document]
noi5e: Just decided to remove `\n` in the templates definition (it's going to get deleted anyway) to get this to pass.

app/assets/javascripts/editor.js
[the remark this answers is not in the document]
noi5e: Commented out the header methods to pass this. I have a bad habit of making PRs too big and out-of-scope.

app/assets/javascripts/editor.js
```diff
@@ -16,43 +19,43 @@ $E = {
...
-  },
+    const singleFormPath = RegExp(/\/(wiki|features)(\/[^\/]+\/edit|\/new)/);
+    return singleFormPath.test(url);
+  }
   // wraps currently selected text in textarea with strings a and b
-  wrap: function(a, b, args) {
-    // we only refresh $E's values if we are on a page using the rich-text editor (most pages).
-    // the legacy editor pages only have one editor form, unlike pages with multiple comments.
-    if (this.isRichTextEditor(window.location.pathname)) { this.refresh(); }
+  wrap(a, b, args) {
```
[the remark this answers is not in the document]
noi5e: Yikes, I remember codeclimate was giving me a lot of grief [in this PR](https://github.com/publiclab/plots2/pull/9035#discussion_r559910489) as well.

I did a little bit of work in the `wrap` method (mostly renaming methods and writing new comments), which started triggering this error. Then I completely scaled it back, removing all the changes, and still got the error. Weird.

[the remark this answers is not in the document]
noi5e: This is passing now.

app/assets/javascripts/editor.js
Sagarpreet: Nice !!

app/assets/javascripts/editor.js
```diff
@@ -23,117 +26,110 @@ $E = {
...
-    if (newlineParameterExists) {
-      replace = replace + "\n\n";
-    }
-    if (newlineParameterExists && $E.textarea[0].selectionStart > 0) { 
-      replace = "\n" + replace; 
-    }
-    $E.textarea.val($E.textarea.val().substring(0,start) + replace + $E.textarea.val().substring(end,len));
-  },
-  bold: function() {
+  wrap(a, b, newlineDesired = false, fallback) {
```
Sagarpreet, line 56: can we name it better than `a` and `b` for new contributors to understand this better?

app/assets/javascripts/editor.js
Sagarpreet: Let's remove and clean this then 💯

app/assets/javascripts/editor.js
```diff
@@ -23,117 +26,110 @@ $E = {
...
+    // we only refresh $E's values if we are on a page with multiple comments
+    if (!this.isSingleFormPage(window.location.pathname)) { this.refresh(); }
+
+    const selectionStart = $E.textarea[0].selectionStart;
+    const selectionEnd = $E.textarea[0].selectionEnd;
+    const selection = fallback || $E.textarea.val().substring(selectionStart, selectionEnd); // fallback if nothing has been selected, and we're simply dealing with an insertion point
+
+    let newText = a + selection + b; // ie. ** + selection + ** (wrapping selection in bold)
+    if (newlineDesired) { newText = newText + "\n\n"; }
```
Sagarpreet, line 65:
```suggestion
    let newText = newlineDesired ?  a + selection + b + "\n\n" : a + selection + b; // ie. ** + selection + ** (wrapping selection in bold)
```

app/assets/javascripts/editor.js
```diff
@@ -23,117 +26,110 @@ $E = {
...
-  italic: function() {
+  }
+  italic() {
     $E.wrap('_','_')
-  },
-  link: function(uri) {
+  }
+  link(uri) {
     uri = prompt('Enter a URL');
     if (uri === null) { uri = ""; }
```
Sagarpreet, line 82:
```suggestion
    if (!!uri) { uri = ""; }
```

app/assets/javascripts/editorToolbar.js
```diff
@@ -108,21 +108,17 @@ $(function() {
...
               orig_image_url = file_url + '?s=o' // size = original
-              $E.wrap('[![', '](' + file_url + ')](' + orig_image_url + ')', {'newline': true, 'fallback': data.result['filename']}) // on its own line; see /app/assets/js/editor.js
+              $E.wrap('[![', '](' + file_url + ')](' + orig_image_url + ')', true, data.result['filename']);
               break;
             case 'csv':
-              $E.wrap('[graph:' + file_url + ']', '', {'newline': true})
+              $E.wrap('[graph:' + file_url + ']', '', true);
               break;
             default:
-              $E.wrap('<a href="'+data.result.url.split('?')[0]+'"><i class="fa fa-file"></i> ','</a>', {'newline': true, 'fallback': data.result['filename'].replace(/[()]/g , "-")}) // on its own line; see /app/assets/js/editor.js
```
Sagarpreet, line 117: Why was fallback used? and why is it removed now?

app/assets/javascripts/post.js
```diff
@@ -2,7 +2,7 @@ jQuery(document).ready(function() {
 
   $('.datepicker').datepicker()
   
-  $E.initialize()
+  $E = new Editor();
```
Sagarpreet, line 5: 🎉

app/views/questions/show.html.erb
```diff
@@ -1,10 +1,14 @@
 <%= javascript_include_tag('comment_expand') %>
 <%= javascript_include_tag('notes') %>
 <%= javascript_include_tag('textbox_expand') %>
-<script> var comments_length = <%= @node.comments.length %>; $(function(){
-  $E.initialize();
-  $("img").lazyload();
-});</script>
+<script> 
+  var comments_length = <%= @node.comments.length %>;
```
Sagarpreet, line 5: This is not getting used anywhere, right?

app/assets/javascripts/editor.js
noi5e: I totally agree about renaming `a` and `b`!!! I'll do it soon.

app/assets/javascripts/editor.js
noi5e: Deleted it in another PR!

app/assets/javascripts/editor.js
noi5e: Thank you!!!!!

app/assets/javascripts/editorToolbar.js
noi5e: This is the only place where `fallback` is still used (3 times):
https://github.com/publiclab/plots2/blob/97247e30289dfb27f93446b84d036ad9811e7983/app/assets/javascripts/editorToolbar.js#L108-L118

I didn't delete `fallback` here, but I just separated it out into two parameters instead of one object:
- **Before**: `$E.wrap(a, b, { fallback: "fallback", newline: "true" });`
- **After**: `$E.wrap(a, b, true, "fallback")`;`

In other words, `wrap` now takes four parameters, instead of three. And none of the parameters are objects.

I know it might be a little harder to read but I really needed to do this to pass codeclimate. It was failing this PR saying that the old code in `$E.wrap` had too much 'cognitive complexity'. Part of the reason why was because the parameter was an object, and we had to do a check like `const fallBackExists = args && args['fallback']`&mdash;which adds to cognitive complexity.

**tldr;** I didn't delete fallback, the same functionality is still there. Just changed parameters.

app/views/questions/show.html.erb
noi5e: `comments_length`? I think it still is here:
https://github.com/publiclab/plots2/blob/97247e30289dfb27f93446b84d036ad9811e7983/app/assets/javascripts/question.js#L13-L23

noi5e: But I don't know what this code does 😅

app/assets/javascripts/editor.js
noi5e: Interesting, I've never heard of `!!` to check for `null`!

sagarpreet-chadha: Hey it is single !

sagarpreet-chadha: So !! gives false for undefined, null, 0 and “”

noi5e: Okay! I think it makes sense to check for `undefined` too, so will commit this. Thank you!

noi5e:
@jywarren @Sagarpreet ready to merge again!

sagarpreet-chadha (approved):

app/assets/javascripts/editor.js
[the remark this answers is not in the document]
jywarren: Great! We don't ALWAYS have to agree with CC on every suggestion. But i think long term it makes for better code, so I really appreciate it!

app/assets/javascripts/post.js
jywarren: 🎉 🎉 🎉  niceeeeee

app/views/questions/show.html.erb
jywarren: This looks like old code related to the "Answers" system which we've deprecated; answers were back turned into comments. It seems it was for showing only the first... 3 answers, hiding the rest behind a "show more" kind of thing? I think that's all gone now, so we should be good to remove it.

jywarren (approved):
> I know this looks like a lot of commits, but most of them were just to pass codeclimate. Is it just me or did it get really strict all of a sudden?

Hmm. i wonder if this is just an old and messy code file and we had previously dismissed CC suggestions, but re-editing the file is bringing them all back up?

This looks super though, thank you! Let's remember to circle back to other thoughts here like removing segments, in a follow-up PR; @noi5e if you can drop them into an issue for now that'd be great!

PR #9115, opened 2021-02-02
Try speeding up yarn install in CI using yarn install --frozen-lockfile

jywarren:
https://stackoverflow.com/questions/58482655/what-is-the-closest-to-npm-ci-in-yarn

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,7 @@ jobs:
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically (not working?)
     - uses: ./.github/actions/setup-test-environment
     - name: Install JavaScript dependencies with Yarn
-      run: yarn check || yarn install;
+      run: yarn check || yarn install --frozen-lockfile;
     - name: "Functional Tests"
       env:
         RAILS_ENV: test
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically (not working?)
     - uses: ./.github/actions/setup-test-environment
     - name: Install JavaScript dependencies with Yarn
-      run: yarn check || yarn install;
+      run: yarn check || yarn install --frozen-lockfile;
```
Sagarpreet, line 43: I think we do not add modules so frequently, right? Can we cache node_modules folder somehow and do `yarn install` only if package.json or yarn.lock changes to speed things up?

jywarren: I looked this up and found this was the most common and "standard" way to speed up yarn installs, but yeah, i think we can use one of these:

https://dev.to/mpocock1/how-to-cache-nodemodules-in-github-actions-with-yarn-24eh

https://github.com/actions/cache/blob/main/examples.md#node---yarn

Let's try in a different PR? Thanks!

jywarren:
Lets compare to times here: 

https://github.com/publiclab/plots2/actions/runs/534731427

image

jywarren:
OK, here's one run: 

image

I'll re-run again to check consistency.

jywarren:
wow that's a big speed increase!!! 

image

jywarren:
Let's keep an eye on these speeds. I hope this step has really helped!!!

sagarpreet-chadha:
Wow 🎉

PR #9126, opened 2021-02-04
Bounce non-user email replies

### LINKED ISSUE #3322: Bounce non-user email replies and errors so they don't "disappear silently"
We should bounce back a reply email with either: 

`Your reply did not get posted because we couldn't find an account on PL.org` (phrased more friendly and with a prompt to sign up?)

or if some other error occurred, just:

`There was an error and your comment couldn't be posted`

This could be broken into 2 emails -- and two issues.

Currently, the logic for processing the email is here:

https://github.com/publiclab/plots2/blob/60c95f0764036de8a6cac9732fb24896260e517b/script/mailman_server#L26-L31

That already has a `rescue` block, where we could send the "there was an error" message!

For where there is no user found (for example if the sending email doesn't match the one on file for this user), we have a conditional loop here:

https://github.com/publiclab/plots2/blob/60c95f0764036de8a6cac9732fb24896260e517b/app/models/comment.rb#L209-L212

That could have an `else` clause which returns an email. 

We could develop 2 email templates with mailer files here: https://github.com/publiclab/plots2/tree/master/app/mailers

Or we could try to send a shorter plain text response without a template, following some docs here (they're old so please test this out or double-check!):

* https://stackoverflow.com/questions/4916557/how-can-i-send-mail-with-rails-without-a-template#12974849
* `ActionMailer::Base.mail(from: "me@example.com", to: "you@example.com", subject: "test", body: "test").deliver`
* Up to date docs here: https://api.rubyonrails.org/classes/ActionMailer/Base.html#method-i-mail

Manasa2850:
Fixes #3322 (<=== Add issue number here)

Thanks!


Manasa2850:
@jywarren I've used regex to get the email ID and sending an email using `ActionMailer`. Is this fine?
Also, what should be the sender's email ID... Do we have a public lab email ID that I need to put there?
Thank you!

Manasa2850:
@jywarren @cesswairimu the failing test is related to `home_controller`. But I haven't made any changes related to `home_controller`. Could you please check it?
Thank you!

app/models/comment.rb
```diff
@@ -246,6 +246,14 @@ def self.new_comment_from_email(mail)
         end
       end
       return node_id
+    else
+      email = mail.select { |s| s.match(/.*@.*/) }
+      ActionMailer::Base.mail(
+        from: "me@example.com", # @jywarren what should this be?
+        to: email,
+        subject: "Could not post your reply",
+        body: "Your reply wasn't posted since we couldn't find your account on publiclab.org. Please sign up on publiclab.org"
```
jywarren, line 255: Thank you! However, I'm not sure about this one, would it potentially attract spammers? Or, is this a good opportunity to recruit new users? But, how would they know what address to email if they aren't already a user? Thanks for thinking on this!!

Manasa2850: Hi @jywarren. Even I'm not too sure how they would get our email ID.
Going by this [comment](https://github.com/publiclab/plots2/issues/3322#issuecomment-472016501), I thought people reply to the email we send them...but again I'm not sure how we know their email ID if they're not a user.
Please let me know if I'm missing out on something. Thanks!

Manasa2850:
@jywarren regarding sending back an email in case of other errors, I checked the `rescue Exception` code in `script/mailman_server`. But it seems like we commented it out since it caused the site to slowdown. Should we have it back or keep it as it is for now?
Thank you!

app/models/comment.rb
jywarren: OK I think this makes sense. I appreciate it!
```suggestion
        body: "Your reply wasn't posted since we couldn't find your account on publiclab.org. Please sign up at https://publiclab.org or try sending email from the account matching your username. Thank you!"
```

jywarren:
Hi, unfortunately the `rescue` is bad for memory management it seems, so let's keep it. Thank you @Manasa2850 and apologies for the slow reply!

app/models/comment.rb
```diff
@@ -246,6 +246,14 @@ def self.new_comment_from_email(mail)
         end
       end
       return node_id
+    else
+      email = mail.select { |s| s.match(/.*@.*/) }
+      ActionMailer::Base.mail(
+        from: "me@example.com", # @jywarren what should this be?
```
jywarren, line 252:
```suggestion
        from: "do-not-reply@publiclab.org",
```

jywarren (approved):

jywarren:
Awesome work, thank you!!!

Issue #9125, opened 2021-02-04
Filter Out Power Tags


RuthNjeri:
![image](https://user-images.githubusercontent.com/24359/106818309-1c17d580-6646-11eb-9879-135bfdfd7201.png)

Filter out the power tags (i.e. those with `key:value` format) showing up in trending. And also, make a special case for filtering out `first-time-poster` since that by definition will likely be the most used but not that useful in a search?

For such filtering, perhaps we could base on some of this work -- #6924 or #4232

_Originally posted by @jywarren in https://github.com/publiclab/plots2/issues/8513#issuecomment-772873938_

PR #9130, opened 2021-02-04
[Outreachy]Filter out Power Tags

RuthNjeri:
Fixes https://github.com/publiclab/plots2/issues/9125

Before 
Screenshot 2021-02-04 at 22 37 37

After
Screenshot 2021-02-04 at 22 37 51

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -163,7 +163,7 @@ class HomeController < ApplicationController
     # Trending tags exculding subscriptions and tag first time poster tag
     # The first time poster tag is not useful in search
     exclude_tids = @tag_subscriptions.pluck(:tid)
-    first_time_poster_tid =  Tag.where(name: "first-time-poster").first.tid
+    first_time_poster_tid = Tag.where(name: "first-time-poster").first.tid
     exclude_tids << first_time_poster_tid
     Tag.trending.where('term_data.name NOT LIKE (?)', '%:%').where.not(tid: exclude_tids)
   end
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -164,7 +164,7 @@ class HomeController < ApplicationController
     # The first time poster tag is not useful in search
     exclude_tids = @tag_subscriptions.pluck(:tid)
     exclude_tids << Tag.where(name: %w(everything first-time-poster)).pluck(:tid)
-    exclude_tids.flatten.uniq
+    exclude_tids.flatten!.uniq
     Tag.trending.where('term_data.name NOT LIKE (?)', '%:%').where.not(tid: exclude_tids)
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/home_controller.rb
```diff
@@ -161,7 +160,11 @@ def activity
   end
 
   def trending
+    # Trending tags exculding subscriptions and tag first time poster tag
+    # The first time poster tag is not useful in search
     exclude_tids = @tag_subscriptions.pluck(:tid)
-    Tag.trending.where.not(tid: exclude_tids)
+    exclude_tids << Tag.where(name: %w(everything first-time-poster)).first&.tid
```
RuthNjeri, line 166: This query looks fishy...I'll fix it.

jywarren: Ok! Thanks! Ping me when it's ready to merge! 🎉

test/functional/home_controller_test.rb
```diff
@@ -141,13 +141,4 @@ def setup
...
   end
-
-  test 'trending tags are returned when a user has not subscribed to any topics' do
-    current_user = users(:user_without_tag_subscriptions)
-    expected_tag = Tag.trending.pluck(:name).sample
-    UserSession.create(current_user)
-    get :dashboard_v2
-    assert_includes response.body, expected_tag
-  end
-
```
RuthNjeri, line 152: The reason this test keeps [failing](https://github.com/publiclab/plots2/pull/9133/checks?check_run_id=1839068000) is because of how I first hard code the expected tags on line 147 and pick out one of the tag names "randomly"

This query is then run separately in the home controller https://github.com/publiclab/plots2/blob/2188e0a4d9430e0c8f9f60db00cb54bd74d6ef94/app/controllers/home_controller.rb#L163-L166

Therefore, if line 147 in the test file returns a result, there is no guarantee that it will be the same as when line 165 in the home controller file is run, that is why the expected, is not the same as the returned value, therefore the test fails(https://github.com/publiclab/plots2/pull/9133/checks?check_run_id=1839068000) 

The trending tags will always be different for all users depending on the subscriptions they have on the site. 

In the meantime, I am trying to identify what part of the trending query would return different results each time... 
```
  def self.trending(limit = 5, start_date = DateTime.now - 1.month, end_date = DateTime.now)
    Tag.select('term_data.tid, term_data.name, term_data.count') # ONLY_FULL_GROUP_BY, issue #8152 & #3120
       .joins(:node_tag, :node)
       .where('node.status = ?', 1)
       .where('node.created > ?', start_date.to_i)
       .where('node.created <= ?', end_date.to_i)
       .distinct
       .order('count DESC')
       .limit(limit)
  end
```

RuthNjeri: This test is better off as a count > 0 and not suited for specificity since the tag names returned are not consistent.

app/controllers/home_controller.rb
RuthNjeri: @jywarren, The tests are now passing and the issue above has been fixed. You can merge once you've reviewed the changes I have made, thanks!

jywarren (approved):
Thank you Ruth!!!!! Looks awesome.

Issue #9160, opened 2021-02-08
Update Tag Model Timestamp if An Activity Is Created or Modified


RuthNjeri:
This is part of https://github.com/publiclab/plots2/issues/8513

This task completes part of  https://github.com/publiclab/plots2/issues/8966

A new timestamp column should be created in the Tag model. This timestamp will be updated each time an activity: comments, research notes,  or wikis have been created or modified.

The new timestamp column will be used to order the topic cards in dashboard v2 according to the most recently updated

PR #9165, opened 2021-02-09
[Outreachy] Tag Model Activity

RuthNjeri:
Fixes https://github.com/publiclab/plots2/issues/9160

If a comment, note, or wiki is created within a topic, the topic latest_activity_nid and activity_timestamp attributes get updated with the comment id "c#{comment_id}", note id, wiki id.

This also happens when a wiki is edited.

### DIFF
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -675,7 +675,7 @@ class Node < ActiveRecord::Base
   def update_tag_timestamp(cid)
     tids = NodeTag.where(nid: nid).pluck(:tid)
     comment_id = "c#{cid}"
-    Tag.update_tag_activity(tids, comment_id)
+    Tag.update_tags_activity(tids, comment_id)
   end
 
   def new_revision(params)
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -668,11 +668,11 @@ class Node < ActiveRecord::Base
                     message_id: params[:message_id],
                     tweet_id: params[:tweet_id])
     c.save
-    update_tag_timestamp(c.cid) if c.valid?
+    update_tag_activity(c.cid) if c.valid?
     c
   end
 
-  def update_tag_timestamp(cid)
+  def update_tag_activity(cid)
     tids = NodeTag.where(nid: nid).pluck(:tid)
     comment_id = "c#{cid}"
     Tag.update_tags_activity(tids, comment_id)
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -668,11 +668,11 @@ class Node < ActiveRecord::Base
                     message_id: params[:message_id],
                     tweet_id: params[:tweet_id])
     c.save
-    update_tag_activity(c.cid) if c.valid?
+    tag_activity(c.cid) if c.valid?
     c
   end
 
-  def update_tag_activity(cid)
+  def tag_activity(cid)
     tids = NodeTag.where(nid: nid).pluck(:tid)
     comment_id = "c#{cid}"
     Tag.update_tags_activity(tids, comment_id)
@@ -878,7 +878,6 @@ class Node < ActiveRecord::Base
 
             if node_tag.save
               saved = true
-              tag.run_count # update count of tag usage # REMEMBER The node_tag model has a callback that updates run_count
               # send email notification if there are subscribers, status is OK, and less than 1 month old
               isStatusValid = status == 3 || status == 4
               isMonthOld = created < (DateTime.now - 1.month).to_i
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -673,9 +673,9 @@ class Node < ActiveRecord::Base
   end
 
   def update_tag_timestamp(cid)
-    tags_ids = NodeTag.where(nid: nid).pluck(:tid)
+    tids = NodeTag.where(nid: nid).pluck(:tid)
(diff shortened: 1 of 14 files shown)


db/migrate/20210208151828_add_activity_timestamp_to_tag.rb
```diff
@@ -0,0 +1,5 @@
+class AddActivityTimestampToTag < ActiveRecord::Migration[5.2]
+  def change
+    add_column :term_data, :activity_timestamp, :datetime
```
jywarren, line 3: OK, so to get this working for comments as well, we would have to add updaters to each of the different triggers for nodes, wikis, and comments, right?

RuthNjeri: Yes.. I do it here for comments https://github.com/publiclab/plots2/pull/9165/files#diff-d521fcf96c3de2b8c810e0e8da2c12824cd1de1a7999f8b600284e8f358703c1R671

and here for nodes https://github.com/publiclab/plots2/pull/9165/files#diff-d521fcf96c3de2b8c810e0e8da2c12824cd1de1a7999f8b600284e8f358703c1R671

I only do it for creation... I only need to add the update one for Wikis...

jywarren: Awesome Ruth, as i said you are several steps ahead of me! I'm glad I understood this one -- thanks for explaining!

app/models/node.rb
```diff
@@ -871,7 +878,6 @@ def add_tag(tagname, user)
 
             if node_tag.save
               saved = true
-              tag.run_count # update count of tag usage
```
RuthNjeri, line 874: it is already updated with a callback in the node_tag.rb file 

https://github.com/publiclab/plots2/blob/a29514a4378d2eed2ff99167153da3becf915d24/app/models/node_tag.rb#L11

test/functional/wiki_controller_test.rb
```diff
@@ -173,7 +178,10 @@ def teardown
     newtitle = 'New Title'
 
     post :update, params: { id: wiki.nid, uid: users(:bob).id, title: newtitle, body: 'Editing about Page' }
+    # latest_activity_nid should be updated in all tags related to the wiki
```
jywarren, line 181: This all looks great. Would there be any extra benefit to some tests that assert both before and after we take an action that bumps up a topic? We could then test against the `@tag_subscriptions` variable in the home controller action?

What do you think? Thanks Ruth this is exciting!

RuthNjeri: Hi Jeff, thanks!

That's a test I thought about but wasn't sure how to approach it, I'll definitely try using the `@tag_subscriptions`. I'll let you know once it is complete.

jywarren: We might also want to add an assertion before and after the count changes, as that'll be one way to protect the code while removing the redundant update you've taken out!

jywarren: I think you can directly access `@tag_susbcriptions` with:

```ruby
assert_equal SOME_KNOWN_VALUE, assigns[:tag_susbcriptions]
assert_equal SOME_KNOWN_VALUE_2, assigns[:tag_susbcriptions].length
assert_equal SOME_KNOWN_VALUE_3, assigns[:tag_susbcriptions].first.some_attribute
```

RuthNjeri:
Hi @jywarren 😄  I think the tests are now complete...you can have another look and let me know

test/functional/comment_controller_test.rb
```diff
@@ -44,9 +44,18 @@ def teardown
...
+      post :create, params: { id: node.nid, body: '[notes:awesome]' }, xhr: true
     end
+
+    # latest_activity_nid for the node should be updated with the comment id in the format 'c_comment_id'
+    cid = Comment.where(comment: '[notes:awesome]').first.cid
+    expected_activity_nid = "c#{cid}"
+    node_tid = node.tag.first.tid
+    returned_activity_nid = Tag.where(tid: node_tid).first.latest_activity_nid
+
+    assert_equal expected_activity_nid, returned_activity_nid
```
jywarren, line 58: Very nice! Thanks!

jywarren: This could even be done in a unit test, which would test this behavior in isolation outside the context of a controller, but this is great as well!

test/functional/home_controller_test.rb
```diff
@@ -141,4 +141,79 @@ def setup
...
+    Revision.create(nid: node.id, uid: current_user.id, title: node.title,  body: 'Topic order Note')
+    node.add_tag(expected_first_topic.name, current_user)
+
+    get :dashboard_v2
+    expected_first_topic.reload
+    new_first_topic = assigns[:tag_subscriptions].first.tag
+
+    # In some cases, the previous first tag might still be the first tag because nodes can have multiple tags.
+    # Checking the latest_activity_node ensures that the most recent activity was bumped up.
+    assert_equal expected_first_topic.latest_activity_nid, new_first_topic.latest_activity_nid
```
jywarren, line 161: Nice!

jywarren (approved):
These are great tests. Thanks @RuthNjeri !!!! Great work!

Issue #1502, opened 2017-07-05
Please add this tag visualization to the top of /tags


ebarry:
This is a request for someone with access to editing special pages to add this visualization of tags from the beginning of time to November 2016 to the top of publiclab.org/tags

https://www.dropbox.com/s/s78g3ufhsav5xzo/plots_tag_graph_256_filtered.png?dl=0
![plots_tag_graph_256_filtered](https://user-images.githubusercontent.com/161439/27880158-04412cd4-6192-11e7-8aef-84a3841fda2f.png)

CC:
@gretchengehrke 
@skilfullycurled

jywarren:
Hi, Liz - i'm a bit reluctant to put a static graphic like this in our permanent codebase, but maybe a suggestion could be that we display a "feature" (like our banners) on the top of that page, and then admins could display whatever they want there. Would that work?

jywarren:
It would go above or below this line: https://github.com/publiclab/plots2/blob/master/app/views/tag/index.html.erb#L4

And look like:

```ruby
	  <% cache('feature_tag-page-header') do %>
	    <%= feature('tag-page-header') %>
	  <% end %>
```

ebarry:
Well, i don't so much want to decorate that page as i want to add "insight at a glance" . 
A different point, but maybe relevant as to why i'd suggest adding a graphic visualization is that this tag page still doesn't have any sorting capabilities to see "recent" or "popular" much less to see either of those by geography.

skilfullycurled:
There are actually python gephi bindings which we could use to generate it dynamically.  I'm actually working on a javascript network visualization right now, so let me see how that works out.  If it goes well, then I can translate what I did into a python script which can generate the data structure to then be visualized in javascript.

jywarren:
Hi, all - i think a generated graph would be great, and is something we could put in the permanent code. 

@ebarry i'm not saying this is decoration and not content, i'm more saying this would go out of date quickly, and also our goal is to store /no/ content in our codebase -- only infrastructure. So this is just a way to implement it -- does my proposed solution sound OK?

re ` this tag page still doesn't have any sorting capabilities to see "recent" or "popular" much less to see either of those by geography.` I'd be happy to work with you to come up with some feature requests to get contributors building to solve this if it's a priority for you. Could be some easy `first-timers-only` issues if you can help get them in the queue!

ebarry:
Let's go back to basics on this issue :)
**What is the goal of visualizing tags?**

[...]

PR #9169, opened 2021-02-09
fixed jlouvain tag viz

jywarren:
Work on #1502 to properly colorize communities

image

image

This'll work quite differently in stable, so the fix will be more apparent there. Thanks!


ebarry:
ah huge win! can't wait to see it live and play around with it

Issue #8810, opened 2020-12-09
System Tests: Comments [Outreachy]


noi5e:
@jywarren @cesswairimu @sagarpreet-chadha I could use a little guidance on how to write tests for async comment testing. I realized that I don't fully understand what _async_ means in this context.

[This planning Google Doc](https://docs.google.com/document/d/1hOqa4tC0jqEuSaur1zKMtnxnknkARui3HTKqO3AUFMA/edit) mentions it on page 14:
> _(Test for)_ Comment boxes created asynchronously (like, reply form on a just-posted comment) rather than on a freshly loaded page

That's talking about this kind of action _(except for the weird email that opens in the tab, I think that only happens locally for some reason)_:
![Untitled 5](https://user-images.githubusercontent.com/4361605/101602225-a29b9800-39b2-11eb-9ecb-d214a5c5facd.gif)

I guess this would actually be an **asynchronous comment response**. Would a good test for this basically be:
- navigate to page, and stay on it
- post a comment
- wait asynchronously for the DOM to update with the comment
- then post a reply to the comment you just posted(?)

**Second Question:**
- What's an example of **posting an asynchronous comment** (different from the comment _response_ I posted above)? So far I know [two](https://github.com/publiclab/plots2/blob/c069cfbec093ede01d8731d3b60b91b452b5f84e/app/assets/javascripts/comment.js#L62-L70) different [methods](url) for posting a comment... Would the tests for either of these be considered async? If not what?

jywarren:
Yes, this sounds like a good workflow. One thing that may simplify things
(worth trying) is that Capybara, which runs our system tests, automatically
waits a bit until it can find a CSS match for an assertion in a test. So,
if you do the action and then do a CSS assertion, it'll wait until the
action has completed and the page is finished loading/reloading to try the
assertion. I think there's a timeout but as everything is local it should
be a pretty fast response from the server in any case.

I think we've even left some comments in the code of the system tests
noting this technique!

jywarren:
perhaps the 2 levels of testing here are: 

1. the one you've proposed -- posting a comment and without a page refresh, replying to it
2. trying some other action upon a newly created (by AJAX/JS) comment form - like uploading an image - and ensuring that it appears in the correct box and not in any others

My thought is that it's good to keep in mind that we want to be confirming the "normal" behavior of the form and comment editor that was just created -- so using it to post a reply is just one way we can confirm this. What do you think?

[...]

PR #9183, opened 2021-02-11
Merge comments/edit & comments/form Partials

noi5e:
These two partials are basically very close copies of each other... Even more so with all the recent refactoring of the Comment Editor.

- [comments/_edit](https://github.com/publiclab/plots2/blob/main/app/views/comments/_edit.html.erb)
- [comments/_form](https://github.com/publiclab/plots2/blob/main/app/views/comments/_form.html.erb)

In the past, if I made a change in one partial, I had to do it in the other to match.

Merging these two partials into one, to increase maintainability and DRYness!

---


app/views/comments/_form.html.erb
```diff
@@ -1,104 +1,156 @@
+<!-- comment is false in reply & main comment forms; defined for edit forms -->
+<% comment_id_string = comment ? comment.id.to_s : nil %>
 <!-- get_comment_form_id defined in app/helpers/application_helper.rb: -->
-<% comment_form_id = get_comment_form_id(location, local_assigns[:reply_to], nil) %>
-<div class="comment-form-wrapper card-body" style="background-color:#f8f8f8; border: 1px solid #e7e7e7;padding: 18px;">
-  <form id="comment-form-<%= comment_form_id %>" class="comment-form" data-remote="true" action="/comment/create/<%= @node.nid %><%= "?type=question" if local_assigns[:type]=="question" %>" <% if local_assigns[:aid].blank? %>method="post"<% end %>>
-    <% if current_user && current_user.id == @node.uid %>
-      <span class="pull-right" style="color:#ccc;">
-        <a target="_blank" href="https://publiclab.org/comment-templates">Add a template</a>
-      </span>
```
noi5e, line 8: Unrelated change: I'm deleting the template functionality in this PR.

app/views/comments/_form.html.erb
```diff
@@ -1,104 +1,156 @@
...
+    padding: 18px;
+    <%= "display: none" if location == :edit  %>
+  "
+>
+  <form 
+    id="comment-form-<%= comment_form_id %>" 
+    class="comment-form" 
+    <% if location != :edit %>
+      action="/comment/create/<%= @node.nid %>"
+      data-remote="true"
```
noi5e, line 20: This `data-remote="true"` (along with `comment.js`) seems to play a crucial role in form submission in edit forms vs. the others.

Edit comment submissions refresh the page, while reply & main comment form submissions don't!

Funnily enough, this was something that was stumping me in [this comment here](https://github.com/publiclab/plots2/issues/8810#issuecomment-747867020) way back a the beginning of the internship.

noi5e: Like, if you're just searching in VSCode for places where [submitFormAjax](https://github.com/publiclab/plots2/blob/main/app/assets/javascripts/submitFormAjax.js) and [addComment](https://github.com/publiclab/plots2/blob/main/app/assets/javascripts/submitFormAjax.js) are called, you don't find anything. For the longest time, I thought that these functions weren't actually being called.

But now I think they are! They work based on attaching extra classes to particular comment forms. I think. This intrigues me because a similar thing could be done to edit comment forms so they are submitted with AJAX and don't refresh the page.

jywarren: OK, i think i can help here -- the `data-remote` attribute is used by the obscure `rails-ujs` library -- https://guides.rubyonrails.org/working_with_javascript_in_rails.html#remote-elements

However this is so obscure that I think it's just not worth using this technique; we don't have to remove it but let's add an explanatory comment to help future coders.

noi5e: Ahh, very good to know!

app/views/comments/_form.html.erb
```diff
@@ -1,132 +1,217 @@
+<!-- comment is false in reply & main comment forms; defined for edit forms -->
+<% comment_id_string = comment ? comment.id.to_s : nil %>
 <!-- get_comment_form_id defined in app/helpers/application_helper.rb: -->
-<% comment_form_id = get_comment_form_id(location, local_assigns[:reply_to], nil) %>
-<div class="comment-form-wrapper card-body" style="background-color:#f8f8f8; border: 1px solid #e7e7e7;padding: 18px;">
-  <form id="comment-form-<%= comment_form_id %>" class="comment-form" data-remote="true" action="/comment/create/<%= @node.nid %><%= "?type=question" if local_assigns[:type]=="question" %>" <% if local_assigns[:aid].blank? %>method="post"<% end %>>
```
noi5e, line 4: Deleting this `?type=question` part here.

noi5e:
Just did a lot of double-checking of the now-deleted edit partial. This one is ready to merge!

cesswairimu (approved):
This is looking great :rocket:  thanks for working on this

jywarren (approved):

jywarren:
This looks great. Thank you! This will also be easy to track if there are any errors, as the PR touches both files. So even if we catch something later this will be traceable. Thanks a lot!!!

PR #9193, opened 2021-02-13
Added wiki modal popup

### LINKED ISSUE #4932: Pop up a modal menu with templates when starting a new wiki page
Going to https://publiclab.org/wiki/new opens a new blank wiki page, but it's a bit bare, and hard for people to know where to begin. Let's offer a Bootstrap modal with some template options:

screen shot 2019-03-06 at 3 51 27 pm

The actual content of the modal we could set in a Feature ([learn about this system here](https://github.com/publiclab/plots2/blob/master/doc/DATA_MODEL.md#features)) which means we don't have to put the HTML for this into the codebase - we'll store it in the database, and I can help with that. For now, we can put that code into a Gist: https://gist.github.com

But the code for the modal which will contain the templates will have to go in our codebase. 

Here's the template for the "new wiki page" page; but note that we use this for editing the wiki page too:

https://github.com/publiclab/plots2/blob/master/app/views/wiki/edit.html.erb

To distinguish, we might need to set a variable in the controller that we can use to turn the modal on: 

https://github.com/publiclab/plots2/blob/master/app/controllers/wiki_controller.rb#L121

Manasa2850:
Fixes #4932 (<=== Add issue number here)

Thanks!

### DIFF
--- a/app/assets/stylesheets/wiki.css
+++ b/app/assets/stylesheets/wiki.css
@@ -83,15 +83,3 @@ a.navbar-offset {
   background: white;
 }
 
-.modal{
-  margin-top: 300px;
-}
-
-.modal-body{
-  margin-top: 10px;
-  padding: 0 80px;
-}
-.modal-footer{
-  padding-bottom: 10px;
-  text-align:center;
-}
--- a/app/assets/stylesheets/wiki.css
+++ b/app/assets/stylesheets/wiki.css
@@ -83,3 +83,15 @@ a.navbar-offset {
   background: white;
 }
 
+.modal{
+  margin-top: 300px;
+}
+
+.modal-body{
+  margin-top: 10px;
+  padding: 0 80px;
+}
+.modal-footer{
+  padding-bottom: 10px;
+  text-align:center;
+}
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -106,7 +106,6 @@ class WikiController < ApplicationController
   end
 
   def edit
-    @is_new = false
     @revision = Revision.new
     @node = if params[:lang]
               Node.find_wiki(params[:lang] + '/' + params[:id])
@@ -134,7 +133,6 @@ class WikiController < ApplicationController
   end
 
   def new
-    @is_new = true
     @revision = Revision.new
     if current_user &.first_time_poster
       flash[:notice] = "Please post a question or other content before editing the wiki. Click <a href='https://publiclab.org/notes/tester/04-23-2016/new-moderation-system-for-first-time-posters'>here</a> to learn why."
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -106,6 +106,7 @@ class WikiController < ApplicationController
   end
 
   def edit
+    @is_new = false
     @revision = Revision.new
     @node = if params[:lang]
               Node.find_wiki(params[:lang] + '/' + params[:id])
@@ -133,6 +134,7 @@ class WikiController < ApplicationController
   end
 
   def new
+    @is_new = true
     @revision = Revision.new
     if current_user &.first_time_poster
       flash[:notice] = "Please post a question or other content before editing the wiki. Click <a href='https://publiclab.org/notes/tester/04-23-2016/new-moderation-system-for-first-time-posters'>here</a> to learn why."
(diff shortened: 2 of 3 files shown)


Manasa2850:
@jywarren this is how the modal popup looks. Could you please help me out with the feature? I'm not really sure where I need to add the content. Thanks!

ezgif com-video-to-gif

app/assets/stylesheets/wiki.css
```diff
@@ -83,3 +83,15 @@ a.navbar-offset {
   background: white;
 }
 
+.modal{
```
jywarren, line 86: Aha! it looks like some custom styles. I wonder though if we can just use the standard "Bootstrap" styles as described at http://getbootstrap.com/docs/4.4/components/modal/ -- would you be able to make that change? 

And, would you be able to share a screenshot of what it looks like with your work added? Thank you so much!

app/controllers/wiki_controller.rb
```diff
@@ -133,6 +134,7 @@ def edit
   end
 
   def new
+    @is_new = true
```
jywarren, line 137: This works -- but there's maybe a more compact way to do it; we could skip this line, and in the view, just do

```ruby
if params[:action] == "new" && params[:controller] == "wiki"
```

What do you think?

app/views/editor/wikiRich.html.erb
```diff
@@ -128,6 +147,10 @@
 
 
 <script>
+  $(window).on('load',function(){
+    $('#wikiModal').modal('show');
```
jywarren, line 151: Is it possible that the modal without the `fade` class, or aria-hidden=true, might display already by default?

app/views/editor/wikiRich.html.erb
```diff
@@ -15,6 +15,25 @@
 
 <script src="/lib/publiclab-editor/dist/PublicLab.Editor.js"></script>
 
+<% if @is_new %>
+  <div class="container">
+    <div class="modal fade" id="wikiModal" tabindex="-1" role="dialog" aria-labelledby="exampleModalLongTitle" aria-hidden="true">
```
jywarren, line 20: Let's try for a more descriptive title, maybe "New Wiki Template chooser"?

app/views/editor/wikiRich.html.erb
```diff
@@ -15,6 +15,25 @@
...
+      <div class="modal-dialog" role="document">
+        <div class="modal-content">
+        
+          <div class="modal-body"> 
+            <h1> Content from the feature will be added here! </h1>
```
jywarren, line 25: What if we skip the footer and just use a "Feature" as described in https://github.com/publiclab/plots2/blob/main/doc/DATA_MODEL.md#features -- that way site admins can just put in whatever text they need here. Actually couldn't we just put the entire container div in here, and just have a complete feature which just gets displayed on new wiki pages? That way it's totally customizable! From line 19-34 would be replaced by the "feature" code as described in https://github.com/publiclab/plots2/blob/main/doc/DATA_MODEL.md#features

jywarren (changes requested):
Thanks! Made some suggestions and this is a complex issue so if you need help i'm happy to offer input! Thank you!!!

app/views/editor/wikiRich.html.erb
Manasa2850: @jywarren that's needed to display the modal every time the page is loaded. 
I tried removing the `fade` class and `aria-hidden=true` but it doesn't display by default :cry:

app/views/editor/wikiRich.html.erb
Manasa2850: Sure @jywarren! Even I was thinking of using a `feature` there but I'm not really sure where I need to add the HTML code for the feature. Could you please help me out with that?

Manasa2850:
@jywarren I realised that the styling was unnecessary. This is how it looks by default
modal
Please let me know if any changes need to be made to the styling of the modal.
Thanks!

Manasa2850:
@jywarren could you please review this? Thanks in advance!

app/views/editor/wikiRich.html.erb
```diff
@@ -128,6 +144,10 @@
 
 
 <script>
+  $(window).on('load',function(){
```
jywarren, line 147: Hi! My apologies for the slowness here. Could we move this up into the same block of code as above, so it's all driven from the same place? Let me try making a code suggestion. Thanks!

jywarren:
OK so the modal content itself, which we'd put in the database, would become:

```html
<div class="modal-content">
  <div class="modal-body"> 
    <h1> Content from the feature will be added here! </h1>
  </div>
</div>
```

jywarren:
Thank you for your help on this!!!!!

PR #9221, opened 2021-02-20
Fix for preview button

### LINKED ISSUE #7122: Wiki "Preview" state does not show a way to go back to editing

imajit:
Fixes #7122  
On clicking the Preview button the text in button was not changing as intended 
previewnotwrk

[here](https://github.com/publiclab/plots2/blob/cc98fd8a2b325fef7c7b7ab1dcc08ed83d76b04e/app/assets/javascripts/editor.js#L138) the previewBtn is defined based on id of tag but [here](https://github.com/publiclab/plots2/blob/cc98fd8a2b325fef7c7b7ab1dcc08ed83d76b04e/app/views/wiki/edit.html.erb#L77) there is no such tag .So I declared the id same as the one being used in the .js file and it solves the issue.
After changes
previewworking

I also found some other files having similar declaration so I changed them too.

### DIFF
--- a/app/views/editor/question.html.erb
+++ b/app/views/editor/question.html.erb
@@ -108,8 +108,7 @@
           <% end %>
 
           <a id="publish" tabindex="5" class="publish btn btn-primary btn-lg">Publish</a>
-          <a tabindex="6" data-previewing-text="Previewing (click to edit)" class="btn btn-outline-secondary btn-lg preview-btn" onClick="$E.toggle_preview()">Preview</a>
-
+          <a tabindex="6" id="toggle-preview-button-main" data-previewing-text="<%= t('wiki.edit.previewing_text') %>" class="btn btn-outline-secondary btn-lg preview-btn" onClick="$E.toggle_preview()"><%= t('wiki.edit.preview') %></a>
           <%= render partial: 'editor/event' %>
 
         </div>
--- a/app/views/editor/post.html.erb
+++ b/app/views/editor/post.html.erb
@@ -108,7 +108,7 @@
           <% end %>
 
           <a id="publish" tabindex="5" class="publish btn btn-primary btn-lg">Publish</a>
-          <a tabindex="6" data-previewing-text="Previewing (click to edit)" class="btn btn-outline-secondary btn-lg preview-btn" onClick="$E.toggle_preview()">Preview</a>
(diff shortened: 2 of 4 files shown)


noi5e:
@imajit Wow, great catch! Thank you. I made changes in the past few months to the code in `editor.js`, but I forgot to update the code to match in `/wiki/edit` and `/wiki/new`.

noi5e:
It also looks like the form should be toggled when the preview button is clicked. Like how the comment form disappears on notes when comment preview is toggled.

My bad! I completely overlooked this bug when I was refactoring preview in `editor.js`.

I'll add this to my to-dos, and make an issue for the bug. The fix will have to pass comment system tests since `editor.js` serves both `/wiki/new` and node comments.

app/views/editor/question.html.erb
```diff
@@ -108,8 +108,7 @@
           <% end %>
 
           <a id="publish" tabindex="5" class="publish btn btn-primary btn-lg">Publish</a>
-          <a tabindex="6" data-previewing-text="Previewing (click to edit)" class="btn btn-outline-secondary btn-lg preview-btn" onClick="$E.toggle_preview()">Preview</a>
-
+          <a tabindex="6" id="toggle-preview-button-main" data-previewing-text="<%= t('wiki.edit.previewing_text') %>" class="btn btn-outline-secondary btn-lg preview-btn" onClick="$E.toggle_preview()"><%= t('wiki.edit.preview') %></a>
```
noi5e, line 111: I wonder if this phrase could be different. Like, `Hide Preview` (that's how it is on comments) instead of `Previewing (click to edit)`. The latter is kind of confusing.

noi5e (approved):

app/views/editor/question.html.erb
imajit: @noi5e `Hide Preview` is used in some parts in comment preview I guess, but as initial code had `Previewing (click to edit)` I didn't change it

noi5e: @imajit, Thanks, yes, I know you didn't write that piece of text, but it's worth saying :)

cesswairimu (approved):
Great!! thanks for fixing this @imajit :tada:

PR #9234, opened 2021-02-24
Hide <textarea> When Toggling Preview in Legacy Editor

noi5e:
Fix for this bug mentioned in this comment: https://github.com/publiclab/plots2/pull/9221#issuecomment-782782858

---

### DIFF
--- a/app/assets/javascripts/editor.js
+++ b/app/assets/javascripts/editor.js
@@ -3,12 +3,13 @@ class Editor {
   //   defaultForm - when the Editor is initialized, there needs to be a default editor form:
   //     1. the main comment form in multi-comment wikis, questions, & research notes.
   //     2. the only editor form on /wiki/new and /wiki/edit
-  //   elementIDPrefixes - the ID naming conventions are different depending on context:
+  //   isSingleFormPage - to distinguish between a) pages with multiple comments b) pages like /wiki/new, and /features/new with only one comment form
+  //     elements have different ID naming conventions on the two kinds of pages:
   //     1. multi-form pages with multiple comments: #comment-preview-123
   //     2. /wiki/new and /wiki/edit: #preview-main
-  constructor(defaultForm = "main", elementIDPrefixes = {}) {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/editor.js
```diff
@@ -3,12 +3,13 @@ class Editor {
   //   defaultForm - when the Editor is initialized, there needs to be a default editor form:
   //     1. the main comment form in multi-comment wikis, questions, & research notes.
   //     2. the only editor form on /wiki/new and /wiki/edit
-  //   elementIDPrefixes - the ID naming conventions are different depending on context:
+  //   isSingleFormPage - to distinguish between a) pages with multiple comments b) pages like /wiki/new, and /features/new with only one comment form
+  //     elements have different ID naming conventions on the two kinds of pages:
   //     1. multi-form pages with multiple comments: #comment-preview-123
   //     2. /wiki/new and /wiki/edit: #preview-main
-  constructor(defaultForm = "main", elementIDPrefixes = {}) {
+  constructor(defaultForm = "main", isSingleFormPage = false) {
```
noi5e, line 10: This is hopefully a lot less confusing than `elementIDPrefixes`.

jywarren (approved):
Looks great, do you think it's worth adding a test for the various places this is used?

noi5e:
> Looks great, do you think it's worth adding a test for the various places this is used?

I'm not sure, I don't really have a good sense for that. I can if you think it's worth it. `/wiki/new` and `/wiki/edit` probably get the most traffic out of all the possibilities right?

jywarren:
I guess i think it would be worthwhile esp if we already have a test set up
for those views and can just add an extra "preview pressing" step? That way
this won't recur ever.

Thanks, Will!!!

noi5e:
Added a new test for `/wiki/new` preview. This one's ready to merge!

jywarren:
Nice, thank you @noi5e !!!

PR #9269, opened 2021-03-04
Filter out moderators for moderator email notifications

### LINKED ISSUE #9114: Filter out moderators who've checked "Do you want to be notified by a email for moderation emails?" in /settings
Looks like we missed a couple email notifications in  https://github.com/publiclab/plots2/issues/6282 -- easy fix though!

Any line which collects moderator emails in this file (and elsewhere) we should go through and filter out any one which has a tag opting out of moderator email notifications!

Those are the ones which have a tag like `notifications:noemail` or `no-moderation-emails` (confirmed this toggle works!)

That said, we have a setting in /settings for this:

https://github.com/publiclab/plots2/blob/876d0fc084064aaecc23f8003630d7d1ab858fa1/app/views/users/settings.html.erb#L50-L54

image

So let's filter for at least `no-moderation-emails` and then maybe also  `notifications:noemail` when collecting emails on these lines:

https://github.com/publiclab/plots2/blob/97247e30289dfb27f93446b84d036ad9811e7983/app/mailers/admin_mailer.rb#L74

Here's a code example for how this works:

https://github.com/publiclab/plots2/pull/7183/files

```ruby
    all_moderators = User.where(role: %w(moderator admin))
    moderators = []
    all_moderators.each do |mod_user|
      moderators << mod_user.email unless mod_user.has_tag('no-moderation-emails')
    end
```

****

cc https://github.com/publiclab/plots2/issues/6867 https://github.com/publiclab/plots2/issues/6282

17sushmita:
Fixes #9114

### DIFF
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -13,7 +13,7 @@ class AdminMailer < ActionMailer::Base
     all_moderators = User.where(role: %w(moderator admin))
     moderators = []
     all_moderators.each do |mod_user|
-      moderators << mod_user.email unless mod_user.has_tag('no-moderation-emails')
+      moderators << mod_user.email unless mod_user.has_tag('no-moderation-emails') || mod_user.has_tag('notifications:noemail')
     end
     if node.status == 4 # only if it remains unmoderated
       mail(
@@ -74,7 +74,7 @@ class AdminMailer < ActionMailer::Base
     all_moderators = User.where(role: %w(moderator admin))
     moderators = []
     all_moderators.each do |mod_user|
-      moderators << mod_user.email unless mod_user.has_tag('no-moderation-emails')
+      moderators << mod_user.email unless mod_user.has_tag('no-moderation-emails') || mod_user.has_tag('notifications:noemail')
     end
     mail(
       to: "comment-moderators@#{ActionMailer::Base.default_url_options[:host]}",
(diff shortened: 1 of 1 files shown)


17sushmita:
@jywarren, I could not find any tag named `notifications:noemail`. So I've filtered out only tag named `no-moderation-emails`. Can you review this PR?

app/mailers/admin_mailer.rb
```diff
@@ -71,7 +71,11 @@ def notify_moderators_of_comment_spam(comment, moderator)
     @moderator = moderator
     @comment = comment
     @footer = feature('email-footer')
-    moderators = User.where(role: %w(moderator admin)).collect(&:email)
+    all_moderators = User.where(role: %w(moderator admin))
+    moderators = []
+    all_moderators.each do |mod_user|
+      moderators << mod_user.email unless mod_user.has_tag('no-moderation-emails')
```
jywarren, line 77: Ah, `notifications:noemail` tag was a proposed new tag, so none will exist yet. But if we could add that too, that'd be great just as a more general case than only for moderators. Could we include it with an `||` OR conditional?

jywarren (changes requested):
This looks great, just the one tweak in the conditional and we can merge! I'm going to resolve/ignore the CodeClimate requests as the code looks pretty tidy. Thanks!!

app/mailers/admin_mailer.rb
17sushmita: I have added tag `notifications:noemail` with or condition

keshavsethi (approved):

jywarren (approved):
This looks great. Thank you so much!!!!

jywarren:
Excellent work!!!! Thanks everyone for the reviews as well! 🎉

PR #9381, opened 2021-03-27
Hide first-time or spam replied-comments

### LINKED ISSUE #8854: Hide replies to comments that have been spammed or aren't yet moderated
It looks like replies to comments are not filtered for spam. We should see this message: 

“Are you sure? The user will no longer be able to log in or publish, and their content will be hidden except comments.” seems wrong

Example of this happening: https://publiclab.org/notes/bhamster/09-02-2020/public-lab-virtual-event-on-all-things-microplastics 

See this reply to a comment:

image

Instead it should have a message like this, which is working for regular comments (those that reply to a post, not another comment:

image

### ISSUE
(This might be fixed by now...lets write a test to confirm that and fix any failures that may arise from the test):

Replies to comments aren’t filtering out spam comments. We use this `replied_comments` method in several places, so we may need to make this fix several times. 

https://github.com/publiclab/plots2/search?q=replied_comments

We can't do it in the definition because it's actually an ActiveRecord relation, not a custom method we can add filters to: https://github.com/publiclab/plots2/blob/828321c2a8fd91e7a55140a511ed4e6dda718b61/app/models/comment.rb#L10

We should filter out `status = 0` in the definition of `comment.replied_comments`, and also for `status = 4` show the message above. 

Then, we ought to write some basic tests: 

https://github.com/publiclab/plots2/blob/876d0fc084064aaecc23f8003630d7d1ab858fa1/test/functional/comment_controller_test.rb#L57-L61

17sushmita:
Fixes #8854 

- Hide spam or first time replies to comments
- Added tests to confirm the if the spam or first-time replied comment is visible to different group of users or not.

### DIFF
--- a/test/system/comment_test.rb
+++ b/test/system/comment_test.rb
@@ -491,6 +491,92 @@ class CommentTest < ApplicationSystemTestCase
       assert_selector("#{edit_preview_id} img", count: 1)
     end
 
+    test "#{page_type_string}: should list first time replied comment to moderator" do
+      comment = nodes(node_name).add_comment({
+        uid: 19,
+        body: "This is a first time reply",
+        status: 4,
+        reply_to: 1,
+      })
+      visit get_path(page_type, nodes(node_name).path)
+      page.find("#c#{comment.id}")
+      assert_equal comment.status, 4
+      assert_selector("#c#{comment.id} div p", text: "Moderate first-time comment:")
+    end
+
+    test "#{page_type_string}: should not list first time replied comment to non-registered user" do
+      comment = nodes(node_name).add_comment({
+        uid: 19,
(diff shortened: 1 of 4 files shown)


cesswairimu:
restarting build

test/system/comment_test.rb
```diff
@@ -491,6 +491,92 @@ def get_path(page_type, path)
       assert_selector("#{edit_preview_id} img", count: 1)
     end
 
+    test "#{page_type_string}: should list first time replied comment to moderator" do
```
jywarren, line 494: Gosh this is fantastic!!! Thank you and great work!!!

jywarren (approved):

PR #9384, opened 2021-03-29
Added animation to save icon when user enters a comment text

### LINKED ISSUE #9362: Animate Save Icon When User Enters Comment Text
## `Description`
From a user interaction standpoint, it would be helpful to have the save icon blink whenever a user enters text into a comment form.

The save icon is the 💾 in the comment toolbar below:

Screen Shot 2021-03-22 at 7 11 24 PM

See #9131, particularly [this comment](https://github.com/publiclab/plots2/issues/9131#issuecomment-776245431).

Currently the comment form saves every single time the input is changed. This isn't very useful to the user, so we'll also want to implement some kind of [debouncing](https://www.freecodecamp.org/news/javascript-debounce-example/) or `setTimeout` delay.

## `Summary of Changes to Be Made`
These changes can be done in separate PRs, and by different contributors. But they probably have to be done in sequence:

1. Add debouncing to delay the `save` method in `editor.js`, either wait 1000ms, or only run `save` when user input has stopped for 500ms (you choose the time).
2. Add some animation to the 💾 save icon when the `save` method runs, so that user knows that it's running. Darken the icon background. Replace the 💾 with one of [these FontAwesome spinner icons](https://fontawesome.com/how-to-use/on-the-web/styling/animating-icons).
3. Swap out the `Drag and drop to add an image or file, or choose one` with `Saving...` to absolutely spell it out for the user.

## `Useful Links & Images`
What the end goal looks like:
Screen Shot 2021-03-23 at 12 04 39 PM

These changes are going to be made in the comment/legacy editor, which lives at `app/assets/javascripts/editor.js`:
https://github.com/publiclab/plots2/blob/62fdcd0dd54b47727a046a0129ba138710082f0b/app/assets/javascripts/editor.js#L116-L119

Leave a comment below if you want to claim this issue, and post here if you need any help getting oriented to the codebase! Good luck! 🍀

waridrox:
Fixes #9362 

Final working version - 

https://user-images.githubusercontent.com/58583793/112806713-b7508880-9094-11eb-9d2a-76de29f88b53.mp4

_I've removed the `console.log` statements 😄._ 

@noi5e kindly review this. Thanks :)

### DIFF
--- a/app/assets/javascripts/editor.js
+++ b/app/assets/javascripts/editor.js
@@ -104,16 +104,55 @@ class Editor {
   // h7() {
   //   this.wrap('#######','')
   // }
+
+  //debounce function addition
+  debounce(func, wait, immediate) {
+    var timeout;
+    return function () {
+      var context = this,
+        args = arguments;
+      var later = function () {
+        timeout = null;
+        if (!immediate) func.apply(context, args);
+      };
+      var callNow = immediate && !timeout;
+      clearTimeout(timeout);
+      timeout = setTimeout(later, wait);
+      if (callNow) func.apply(context, args);
+    };
+  }
+
   // this function is dedicated to Don Blair https://github.com/donblair
   attachSaveListener() {
     // remove any other existing eventHandler
     $("textarea").off("input.save"); // input.save is a custom jQuery eventHandler
     const thisEditor = this; // save a reference to this editor, because inside the eventListener, "this" points to e.target
-    this.textAreaElement.on("input.save", function() {
-      thisEditor.save(thisEditor);
-    });
-  }
-  save(thisEditor) {
+    //implementing a debounce function on save method
+    this.textAreaElement.on(
+    "input.save",
+    debounce(function () {
+      //changing styles and text
+      let saving_text = $('<p id="saving-text"> Saving... </p>');
+      $(".imagebar").prepend(saving_text);
+      $(".imagebar p").not("#saving-text").hide();
+
+      $("#save-button-main").find("i").removeClass("fa fa-save");
+      $("#save-button-main").find("i").addClass("fas fa-sync fa-spin");
+
+      //adding delay and reverting the styles
+        setTimeout(() => {
+          $("#save-button-main").find("i").removeClass("fas fa-sync fa-spin");
+          $("#save-button-main").find("i").addClass("fa fa-save");
+  
+          $(".imagebar").find("#saving-text").remove();
+          $(".imagebar p").not("#saving-text").show();
+        }, 400);
+  
+        thisEditor.save(thisEditor);
+      }, 700)
+    );
+  }  
+    save(thisEditor) {
     const storageKey = "plots:" + window.location.pathname + ":" + thisEditor.commentFormID;
     localStorage.setItem(storageKey, thisEditor.textAreaValue);
   }
--- a/app/assets/javascripts/editor.js
+++ b/app/assets/javascripts/editor.js
@@ -132,23 +132,42 @@ class Editor {
     "input.save",
     debounce(function () {
       //changing styles and text
-      let saving_text = $('<p id="saving-text"> Saving... </p>');
-      $(".imagebar").prepend(saving_text);
-      $(".imagebar p").not("#saving-text").hide();
-
-      $("#save-button-main").find("i").removeClass("fa fa-save");
-      $("#save-button-main").find("i").addClass("fas fa-sync fa-spin");
-
-      //adding delay and reverting the styles
+      //explicitly handling main comment form
+      if ($('#text-input-main').is(':focus')) {
+       
+       $("#comment-form-main .btn-toolbar #save-button-main").find("i").removeClass("fa fa-save").addClass("fas fa-sync fa-spin");
+     
+       let saving_text = $('<p id="saving-text"> Saving... </p>');
+       $("#comment-form-main .imagebar").prepend(saving_text);
+       $("#comment-form-main .imagebar p").not("#saving-text").hide();
+        
+       //adding delay and revering the styles
         setTimeout(() => {
-          $("#save-button-main").find("i").removeClass("fas fa-sync fa-spin");
-          $("#save-button-main").find("i").addClass("fa fa-save");
-  
-          $(".imagebar").find("#saving-text").remove();
-          $(".imagebar p").not("#saving-text").show();
+          $("#comment-form-main .btn-toolbar #save-button-main").find("i").removeClass("fas fa-sync fa-spin").addClass("fa fa-save");;
+          
+          $("#comment-form-main .imagebar").find("#saving-text").remove();
+          $("#comment-form-main .imagebar p").not("#saving-text").show();
         }, 400);
-  
-        thisEditor.save(thisEditor);
+    }
+    else { 
+        //handling other comment forms
+        let comment_temp_id = (document.activeElement.parentElement.parentElement.id);
+        let imager_bar = (document.activeElement.nextElementSibling.className);
+
+        $('#'+comment_temp_id).find('.btn-toolbar').find(".save-button").find("i").removeClass("fa fa-save").addClass("fas fa-sync fa-spin");
+
+        let saving_text = $('<p id="saving-text"> Saving... </p>');
+        $('#'+comment_temp_id).find('.'+imager_bar).prepend(saving_text);
+        $('#'+comment_temp_id).find('.'+imager_bar).find("p").not("#saving-text").hide();
+
+        setTimeout(() => {
+          $('#'+comment_temp_id).find('.btn-toolbar').find(".save-button").find("i").removeClass("fas fa-sync fa-spin").addClass("fa fa-save");
+          
+          $('#'+comment_temp_id).find('.'+imager_bar).find("#saving-text").remove();
+          $('#'+comment_temp_id).find('.'+imager_bar).find("p").not("#saving-text").show();
+        }, 400);
+    }
+      thisEditor.save(thisEditor);
       }, 700)
     );
   }  
--- a/app/assets/javascripts/editor.js
+++ b/app/assets/javascripts/editor.js
@@ -137,7 +137,7 @@ class Editor {
        
        $("#comment-form-main .btn-toolbar #save-button-main").find("i").removeClass("fa fa-save").addClass("fas fa-sync fa-spin");
      
-       let saving_text = $('<p id="saving-text"> Saving... </p>');
+       let saving_text = $('<p id="saving-text" style="padding-bottom: 8px"> Saving... </p>');
        $("#comment-form-main .imagebar").prepend(saving_text);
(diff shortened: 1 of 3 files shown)


waridrox:
> One thing to note is that the save & recover system test needs to be updated (that's what's failing right now).

Oh!😅 that's what is failing...

waridrox:
> > One thing to note is that the save & recover system test needs to be updated (that's what's failing right now).
> 
> Oh!😅 Will take a look at this soon...

waridrox:
My apologies, got a little busy 😅, @noi5e can you point me to the right file to implement the tests for debounce function. Further researching upon, I came across the `sleep` method in ruby which waits the program for a certain amount of time.

RuthNjeri:
Hi @waridrox, from the system tests [here](https://github.com/publiclab/plots2/pull/9384/checks?check_run_id=2216654085) the failing test is 
Screenshot 2021-04-15 at 22 14 49

`comment_test.rb`

waridrox:
Just when I though about adding tests, I realised we missed an edge case wherein due to the similar functional elements like the toolbar and the image upload section, the current changes also reflect if a `Reply to a Comment` is open along with `Post a Comment` window. Here's an example for that case and I'm working on first fixing that! 


https://user-images.githubusercontent.com/58583793/115920687-2a80cb00-a498-11eb-9b46-9b3cba1d9f1c.mp4

waridrox:
Here's the fix for the edge case: reason being that each comment has its own unique id associated so earlier code was targeting all ids instead of the specific one that the user is currently making changes on.


https://user-images.githubusercontent.com/58583793/116014772-927c1080-a654-11eb-97e8-14bf5da79322.mp4


@noi5e could you give some insights on testing the new changes since there are now ids associated with each comment section 😅 Thanks :)

waridrox:
The tests have been updated to account for debounce function and tests for save and recover button pass now :)
Screenshot 2021-04-27 at 6 11 41 PM

app/assets/javascripts/editorToolbar.js
```diff
@@ -90,9 +90,52 @@ $(function() {
     })
     // for save & recover buttons
     .on("click", ".save-button", function(e) {
-      $E.setState(e.currentTarget.dataset.formId); // string that is: "main", "reply-123", "edit-123" etc.
-      $E.save($E);
-    })
+      //preventing multiple clicks on the button
+      if(!e.detail || e.detail == 1) {
```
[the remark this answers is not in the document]
waridrox: Isn't `===`  supposed to check type as well 🤔. I think here `==` should suffice...

RuthNjeri:
Hi @waridrox, what do you think about the other comments made by CodeClimate? Do you think you can address them? I see you responded to one...would you like to address the rest? We usually don't implement everything suggested by CodeClimate depending on the trade offs.

waridrox:
> Hi @waridrox, what do you think about the other comments made by CodeClimate? Do you think you can address them? I see you responded to one...would you like to address the rest? We usually don't implement everything suggested by CodeClimate depending on the trade offs.

Yes, @RuthNjeri, I've looked at the comments by CodeClimate, most of them being a repetition of one another 😅. But I don't see any problem in the issues that it is suggesting to be fixed - 

- Function attachSaveListener has 32 lines of code (exceeds 25 allowed). - This function also houses comments for explanations, but CodeClimate is also counting those in the number of lines.
- Similar blocks of code found in 2 locations. Consider refactoring. - The code do start similar but eventually do different things at the completion.
- Expected '===' and instead saw '=='. - It means that '===' is being enforced unnecessarily when '==' can take of that.

waridrox:
@noi5e @RuthNjeri @jywarren, pls review the same, this seems complete from my end. Thanks :)

waridrox:
@noi5e @jywarren pls review the changes. Thanks :)

test/system/comment_test.rb
```diff
@@ -359,6 +363,8 @@ def get_path(page_type, path)
       page.find('#text-input-edit-' + edit_id_num)
         .click
         .fill_in with: comment_text_edit
+        # taking into account the time for debounce function
+        sleep(0.7)
```
RuthNjeri, line 367: @waridrox, what is the purpose of the sleep? Will the tests fail if the sleep is removed? And why 0.7 seconds?

app/assets/javascripts/editorToolbar.js
```diff
@@ -90,9 +90,52 @@ $(function() {
...
+      if(!e.detail || e.detail === 1) {
+          //explicitly handling main comment section
+          if ((this.id) === "save-button-main") {
+            // toggling the favicon save icon class to add a spinner icon
+            $(this).find("i").toggleClass("fa fa-save fas fa-sync fa-spin");
+      
+            //changing the text from "Upload an image" to "Saving..."
+            let saving_text = $('<p id="saving-text"> Saving... </p>');
+            $("#comment-form-main .imagebar").prepend(saving_text);
+            $("#comment-form-main .imagebar p").not("#saving-text").hide();
```
RuthNjeri, line 103: Is this change of text to make it re-usable with the comment save? Meaning that if it stays `upload an image` it would mean that this text would appear when saving a text?

test/system/comment_test.rb
waridrox: @RuthNjeri, here the purpose of `sleep` is to wait the tests for 0.7 seconds in order for the save function to take effect due to the debounce function, whenever new content is put in the comment text area: 
```javascript
    }, 700)
     );
   }  
     save(thisEditor) {
```

The tests would fail because this would be a new addition to the comment area hence the use of `sleep` is to update the tests.

app/assets/javascripts/editorToolbar.js
waridrox: Yes, indeed! The text `upload an image` would be hidden and in turn the text `Saving...` would be displayed to the user at the image upload text section of the comment form.

waridrox: The `upload an image` text is simply not interchanged with the `Saving...` text because the `upload an image` text also houses a hidden text, i.e. `Uploading` when an image is dragged on the section...

app/assets/javascripts/editor.js
```diff
@@ -104,16 +104,74 @@ class Editor {
...
+      var context = this,
+        args = arguments;
+      var later = function () {
+        timeout = null;
+        if (!immediate) func.apply(context, args);
+      };
+      var callNow = immediate && !timeout;
+      clearTimeout(timeout);
+      timeout = setTimeout(later, wait);
+      if (callNow) func.apply(context, args);
```
RuthNjeri, line 121: I have less experience using JS than Rails, is there a reason for using var over the latest let here?

waridrox: `var` is used for global scope but I guess here it doesn't matter since the values it holds are accessed only by the debounce function. Thanks for the catch!!!

app/assets/javascripts/editor.js
```diff
@@ -104,16 +104,74 @@ class Editor {
...
+    let timeout;
+    return function () {
+      let context = this,
+        args = arguments;
+      let later = function () {
+        timeout = null;
+        if (!immediate) func.apply(context, args);
+      };
+      let callNow = immediate && !timeout;
+      clearTimeout(timeout);
```
RuthNjeri, line 119: Why, clear timeout here, if the variable has a null value in line 115, why not just reassign it on line 120 and skip doing line 119?

app/assets/javascripts/editorToolbar.js
RuthNjeri: Why not have a specific class or id for the text functionality and then hide the imagebar class/text? Is it easier to just reuse the imagebar class?

app/assets/javascripts/editor.js
waridrox: @RuthNjeri, at line 115 - 
```javascript
timeout = null;
```
the timeout variable gets cleared to make the next execution run as soon as the function is called, i.e. at line 120:
```javascript
      if (callNow) func.apply(context, args);
```

On the other hand, `clearTimeout` at line 119:
```javascript
clearTimeout(timeout);
```
 is used to clear out the previous created timeout, so clearTimeout should be done before setTimeout at line 120: 
```javascript
      timeout = setTimeout(later, wait);
```
in order to cancel out previous invocation of setTimeout function. Thus, `clearTimeout` is essential for the invocation the the `setTimeout` function 😅...

app/assets/javascripts/editorToolbar.js
waridrox: That may involve refactoring text functionality class code / styles associated with the current imagebar id in different places...That's why imagebar class was reused 😅

waridrox:
Sorry the tests did not run and instead failed...

pydevsg:
Refactor the code @waridrox. 
This will pass the test 😅

jywarren:
If you'd like we can bypass the code climate suggestions! They are only suggestions but if you'd like to address any of them we can also wait, no rush!

waridrox:
@jywarren @pydevsg, I was talking about all the tests instead of just the code climate ones😅I've tried to address them but all of them seem fine to me...

RuthNjeri (approved):
Thanks for working on this @waridrox, it looks good to me, I think one more reviewer should approve this(@jywarren or @cesswairimu)...then it will be okay to merge 🚀

waridrox:
After updating the branch with upstream, unit tests seem to be failing...🤔

jywarren:
Aha - for some reason i also saw this on https://github.com/publiclab/plots2/pull/9665#issuecomment-848231223 and suggested a fix there - adding a sort to the test or to the main codebase, as i believe it's an ordering issue that can happen sometimes if we don't explicitly send a sort to the database. If you're able to open a new PR then possibly this and https://github.com/publiclab/plots2/pull/9665#issuecomment-848231223 can both be rebased over it once we resolve that. In the meantime, i'll restart the test to see if we get past it that way. Thanks!

jywarren:
Ok sorry @waridrox but now the unit tests upstream are fixed and just one more rebase should work. Many thanks for your patience!!!

waridrox:
@jywarren Done!

jywarren:
OK, merging now, and many thanks!!!

Regarding the revert button, i wonder if we should have saving add the text to an array of recent texts that have been saved. Then revert could just cycle back through that list.

jywarren:
@waridrox @noi5e we could break out that idea of the array of comments saves into a new issue too! I think we'd have to be careful to flush the list... maybe only save 5 or 10 states, and/or delete much older ones or something? For example, what if someone types in something private, it saves, and they change their mind before posting it, they log out, someone else logs in, and they are able to recover the text.

jywarren:
i.e. we could make it auto-delete texts older than an hour to mitigate this risk maybe.

jywarren:
I know @waridrox is now working on other projects, though! So perhaps someone else from the @publiclab/soc team would be interested in taking this idea of fixing the comment revert button up?

PR #9393, opened 2021-03-30
Unstable seems to work [DO_NOT_MERGE_DIRECTLY]

icarito:
This is to compare what is different in unstable branch and main, because `unstable` builds webpacker correctly and main doesn't.


icarito:
Not needed anymore

Issue #9416, opened 2021-04-01
"Save" not working on /features


ebarry:
### Please describe the problem (or idea)

> What happened just before the problem occurred? Or what problem could this idea solve?
As an Admin, I edited text in a Feature, then clicked Save. The page reloaded with a "This page does not exist" screen, see screenshot. I also tested clicking "Save" without editing, with the same result. 

<img width="998" alt="Screen Shot 2021-04-01 at 10 08 36 AM" src="https://user-images.githubusercontent.com/161439/113306835-7e055c00-92d2-11eb-8845-98919d05ded7.png">

> What did you expect to see that you didn't?
I expected to see my edits successfully saved. 

### Please show us where to look

https://publiclab.org/features/17211
https://publiclab.org/features/11796/
and for Jeff: https://publiclab.slack.com/archives/CV7KG8MFT/p1617198609007000

### What's your PublicLab.org username?

bhamster
liz

Thank you!

daemon1024:
Noting some finding that may assist in solving this.

When we try to edit and save in local development environment, the route seems to throwing off an exception.

```No route matches [POST] "/features/<id>"```

It should be calling `update` features controllers. 
https://github.com/publiclab/plots2/blob/9e66b0f1a7c19733388d233308d2b090111bc627/app/views/features/edit.html.erb#L1
https://github.com/publiclab/plots2/blob/9e66b0f1a7c19733388d233308d2b090111bc627/app/controllers/features_controller.rb#L68 the update controller hasn't been updated for quite a while so not exactly sure what changed that is throwing off this exception now.

jywarren:
Wow very much appreciated! I got drawn into an adjacent bug 😅

jywarren:
I'll check the routes.rb file!

jywarren:
https://github.com/publiclab/plots2/blame/main/config/routes.rb

jywarren:
This line is 6 years old, so i doubt it's related! 

https://github.com/publiclab/plots2/blame/9e66b0f1a7c19733388d233308d2b090111bc627/app/views/features/edit.html.erb#L1

jywarren:
I really don't see much in routes.rb...

daemon1024:
I checked the the routes, there's no `POST` method for `features/:id`

daemon1024:
https://stackoverflow.com/questions/19501374/rails-4-routingerror-no-route-matches-post This might be related

jywarren:
[...]

PR #9469, opened 2021-04-06
Starting minimal features test

jywarren:
re #9416

This doesn't actually test the routing/form issue in that issue but provides some basic protection for features.

### DIFF
--- a/test/integration/features_test.rb
+++ b/test/integration/features_test.rb
@@ -1,0 +1,23 @@
+require 'test_helper'
+
+class FeaturesTest < ActionDispatch::IntegrationTest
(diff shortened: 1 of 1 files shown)


jywarren:
This test might be redundant with https://github.com/publiclab/plots2/blob/features-tests/test/functional/features_controller_test.rb, honestly, and we might want to move to a system test to test the views too... though it's more expensive.

jywarren:
Just the new test failed. Considering closing this if no other tests fail, as we've fixed the underlying routes/form bug and this new test may be redundant with our functional tests... we really need a system test to catch the bug that just occurred. Thanks all!

PR #9483, opened 2021-04-09
Fix text overflow bug in editor

### LINKED ISSUE #9476: Text overflows in editor (rich mode)
### Problem

When using the editor's Rich mode (i.e. wysiwyg) the text does not wrap around inside of the text area, but instead overflows and causes visual bugs.

image

### Where to find this bug

https://publiclab.org/post

### How to reproduce

Use the editor's "Rich" mode and type in a long string of characters (no new lines)

eightballocto:
Fixes #9476 (text did not wrap before addition of CSS rule), thanks @aliciapaz

### DIFF
--- a/app/assets/stylesheets/wiki.css
+++ b/app/assets/stylesheets/wiki.css
@@ -81,5 +81,6 @@ a.navbar-offset {
 
 .wk-wysiwyg {
   background: white;
+  word-break: break-all;
 }
 
--- a/app/assets/stylesheets/wiki.css
+++ b/app/assets/stylesheets/wiki.css
@@ -81,6 +81,5 @@ a.navbar-offset {
 
 .wk-wysiwyg {
   background: white;
-  word-break: break-all;
 }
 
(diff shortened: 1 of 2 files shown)


sahilsaha7773 (commented):
Hi @eightballocto! Thanks for opening this! Can you please attach a screenshot?

eightballocto:
@sahilsaha7773 Sure, here's the result after adding the rule 
image

However, it seems like this only *partially* fixes the issue, because I notice that long strings of something like say, ellipses,
..results in this.
image

My initial theory is that this is an issue with margins, and so I will further investigate when I can.

eightballocto:
Turns out I had added the rule to the wrong selector, latest commits should fix this definitively :) 
image

jywarren:
Awesome work!!!

jywarren (approved):

jywarren:
Congrats!!! Thank you!!!

Issue #9694, opened 2021-05-28
[Outreachy Planning] Tag / Topic System Enhancement Project - Manasa


Manasa2850:
These are the tasks for my Outreachy project. This is my original proposal https://publiclab.org/notes/Manasa2850/04-10-2021/gsoc-2021-proposal-tag-topic-system-refinements

1. Tag Locking

- [ ] Create a 'locked' tag which can be added to content by mods / admin.
- [ ] Change UI of the page after adding the 'locked' tag.
- [ ] Modify permissions of pages having the 'locked' tag.
- [ ] Functional tests + System tests.

2. Delete Tags from users who are banned

- [ ] Delete tags of banned users retroactively.
- [ ] Functional tests.

3. Differentiate between power tags and normal tags

- [ ] Change the location of power tags.
- [ ] Have different colors for normal tags and power tags.
- [ ] Prompt users to enter the 'type' of the power tag before publishing.
- [ ] System tests.

Tlazypanda:
Hey @Manasa2850 great breakdown 🎉  here is a great resource by @jywarren on how we can further use break-me-up tags to break these issues into even smaller ones https://publiclab.org/notes/warren/01-18-2018/software-outreach-modularity-is-great-for-collaboration 

Probably the next steps for now can be to start with the first phase first task at hand and write a detailed issue for that (once done it might be really great if you can come back to this issue and tag the issue number here for reference for example): 
- Create a 'locked' tag which can be added to content by mods / admin. (#0000) 

This will really help to get a quick glance of all the issues and its progress.  Happy coding 👩‍💻 💯

PR #9709, opened 2021-05-31
Created a locked tag which can be added to notes by mods / admin

Manasa2850:
Part of #9694 

Thanks!

### DIFF
--- a/app/assets/javascripts/tagging.js
+++ b/app/assets/javascripts/tagging.js
@@ -140,7 +140,7 @@ function promptTag(val) {
       break;
 
     default:
-      addTag(expr);
+      addTag(val);
       break;
 
   }
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -206,12 +206,6 @@ class NotesController < ApplicationController
     render template: 'notes/show'
   end
 
-  def lock_note
-    @node = Node.find(params[:id])
-    @node.add_tag('locked', current_user)
-    redirect_to URI.parse(@node.path).path
-  end
-
   def edit
     @node = Node.find_by(nid: params[:id], type: 'note')
 
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -206,6 +206,12 @@ class NotesController < ApplicationController
     render template: 'notes/show'
   end
 
+  def lock_note
+    @node = Node.find(params[:id])
+    @node.add_tag('locked', current_user)
+    redirect_to URI.parse(@node.path).path
+  end
+
   def edit
     @node = Node.find_by(nid: params[:id], type: 'note')
 
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -386,7 +386,7 @@ class TagController < ApplicationController
         end
       end
     elsif node_tag.name == "locked"
-      flash[:error] = "You cannot delete the locked tag."
+      flash[:error] = "Only admins can delete the locked tag."
       redirect_to Node.find_by(nid: params[:nid]).path
     else
       flash[:error] = I18n.t('tag_controller.must_own_tag_to_delete')
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -359,7 +359,7 @@ class TagController < ApplicationController
     node_tag = NodeTag.where(nid: params[:nid], tid: params[:tid]).first
     node = Node.where(nid: params[:nid]).first
     # only admins, mods, and tag authors can delete other peoples' tags
-    if node_tag.uid == current_user.uid || logged_in_as(['admin', 'moderator']) || node.uid == current_user.uid
+    if node_tag.uid == current_user.uid || logged_in_as(['admin', 'moderator']) || (node.uid == current_user.uid && node_tag.name != "locked")
 
       tag = Tag.joins(:node_tag)
                    .select('term_data.name')
@@ -385,6 +385,9 @@ class TagController < ApplicationController
           end
         end
       end
+    elsif node_tag.name == "locked"
+      flash[:error] = "You cannot delete the locked tag."
+      redirect_to Node.find_by(nid: params[:nid]).path
     else
       flash[:error] = I18n.t('tag_controller.must_own_tag_to_delete')
       redirect_to Node.find_by(nid: params[:nid]).path
--- a/app/views/notes/show.html.erb
+++ b/app/views/notes/show.html.erb
@@ -86,6 +86,10 @@
     </table>
   <% end %>
 
+  <% if current_user && ( current_user.role == "admin" || current_user.role == "moderator") && !(@node.has_tag('locked')) %>
+    <%= link_to "Lock Note", "/notes/lock_note/#{@node.id}" %>
+  <% end %>
+
   <% if current_user && 
     ( current_user.role == "admin" || current_user.role == "moderator" || 
       current_user.id == @node.id || current_user.is_coauthor?(@node) ) && 
--- a/app/views/notes/show.html.erb
+++ b/app/views/notes/show.html.erb
@@ -87,7 +87,7 @@
   <% end %>
 
   <% if current_user && ( current_user.role == "admin" || current_user.role == "moderator") && !(@node.has_tag('locked')) %>
-    <%= link_to "Lock Note", "/notes/lock_note/#{@node.id}" %>
+    <a href="javascript:window.location.reload(true)" onClick="addTag('locked');">Lock Note</a>
   <% end %>
 
   <% if current_user && 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -334,7 +334,6 @@ Plots2::Application.routes.draw do
   get 'admin/publish_comment/:id' => 'admin#publish_comment'
   get 'admin/mark_comment_spam/:id' => 'admin#mark_comment_spam'
   get 'smtp_test' => 'admin#smtp_test'
-  get 'notes/lock_note/:id' => 'notes#lock_note'
 
   get 'post' => 'editor#post', :as => :editor_post
   post 'post' => 'editor#post', :as => :editor_path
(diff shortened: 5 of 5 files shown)


Manasa2850:
ezgif com-gif-maker
I would like some suggestions regarding how the 'Lock Note' button should look. Thanks!

Manasa2850:
@jywarren @RuthNjeri this is how it looks when a normal user tries to delete the 'locked' tag on their content.
ezgif com-gif-maker

Another way of implementing this, is by not giving the `delete tag` option to normal users for the 'locked' tag. But this might cause some confusion to users as to why they can't delete tags on their own notes. Please let me know your thoughts regarding this.

app/controllers/notes_controller.rb
```diff
@@ -206,6 +206,12 @@ def preview
     render template: 'notes/show'
   end
 
+  def lock_note
```
jywarren, line 209: Here, i wonder if we could simply use the existing new tag route at tag_controller.rb, instead of a custom route. What do you think?

config/routes.rb
```diff
@@ -334,6 +334,7 @@
   get 'admin/publish_comment/:id' => 'admin#publish_comment'
   get 'admin/mark_comment_spam/:id' => 'admin#mark_comment_spam'
   get 'smtp_test' => 'admin#smtp_test'
+  get 'notes/lock_note/:id' => 'notes#lock_note'
```
jywarren, line 337: Re-using the existing route could also remove this line.

app/views/notes/show.html.erb
```diff
@@ -86,6 +86,10 @@
     </table>
   <% end %>
 
+  <% if current_user && ( current_user.role == "admin" || current_user.role == "moderator") && !(@node.has_tag('locked')) %>
+    <%= link_to "Lock Note", "/notes/lock_note/#{@node.id}" %>
```
jywarren, line 90: So i think it would need to be a JavaScript call instead of a direct route, in this case; see how it's done in this menu:

image

The functionality is buried quite deep here, but i think fundamentally it uses a JS function called `add_tag()`

Here's the menu itself:

https://github.com/publiclab/plots2/blob/main/app/views/tag/_advanced_tagging.html.erb

And aha - here it is in JavaScript:

https://github.com/publiclab/plots2/blob/e646cfd248e46fe9cf11a2eb7860bbf29f949b7d/app/assets/javascripts/tagging.js#L108

This is great to see, @Manasa2850 - thank you! What do you think about simplifying and taking this approach?

jywarren:
I love the "can't delete" message you've added. Great work!

Manasa2850:
Thanks @jywarren for reviewing! I'll take a look at these suggestions and try implementing them.

app/assets/javascripts/tagging.js
```diff
@@ -140,7 +140,7 @@ function promptTag(val) {
       break;
 
     default:
-      addTag(expr);
+      addTag(val);
```
Manasa2850, line 143: This change is not directly related to the 'locked' tag but is required in the `promptTag` function.

jywarren: Ah, was it a bug from before? Cool. Thanks!

app/views/notes/show.html.erb
```diff
@@ -86,6 +86,10 @@
     </table>
   <% end %>
 
+  <% if current_user && ( current_user.role == "admin" || current_user.role == "moderator") && !(@node.has_tag('locked')) %>
+    <a href="javascript:window.location.reload(true)" onClick="addTag('locked');">Lock Note</a>
```
jywarren, line 90: I think this may be an issue in Firefox, but I could be wrong. I noted this in https://github.com/publiclab/plots2/pull/9726 too!

jywarren (changes requested):
Just a note on Firefox usage. This looks really good and a bit simpler too! Love it!!!

app/views/notes/show.html.erb
Manasa2850: @jywarren I use Firefox and it works fine!

Manasa2850:
@jywarren I use Firefox and it works fine!

jywarren:
Sounds great. Merging this! Excellent work!

app/views/notes/show.html.erb
```diff
@@ -86,6 +86,10 @@
     </table>
   <% end %>
 
+  <% if current_user && ( current_user.role == "admin" || current_user.role == "moderator") && !(@node.has_tag('locked')) %>
```
jywarren, line 89: What if, in follow up, we moved this button into the advanced tagging menu? This would then appear on wikis too!

app/controllers/tag_controller.rb
```diff
@@ -359,7 +359,7 @@ def delete
     node_tag = NodeTag.where(nid: params[:nid], tid: params[:tid]).first
     node = Node.where(nid: params[:nid]).first
     # only admins, mods, and tag authors can delete other peoples' tags
-    if node_tag.uid == current_user.uid || logged_in_as(['admin', 'moderator']) || node.uid == current_user.uid
+    if node_tag.uid == current_user.uid || logged_in_as(['admin', 'moderator']) || (node.uid == current_user.uid && node_tag.name != "locked")
```
RuthNjeri, line 362: From line 359 above, we have the `node_tag` variable derived from this query `node_tag = NodeTag.where(nid: params[:nid], tid: params[:tid]).first` ,

 I am curious as to whether checking to see if `node_tag.name != "locked"` will work if the first tag derived is not `"locked"` but contains a list of tags like `[air-quality, locked]`....the node itself does contain a locked tag but it may not be the first to be picked hence someone else might delete a tag.

Should checking if the tags returned contain `"locked"`, before picking the first one in line 359 help?

RuthNjeri: lol I was just reviewing this and panicked that I had merged it @jywarren 😅 

Are the concerns I have above okay? Or will the tagging work as expected... @jywarren?

jywarren: Oh you're totally right, my bad! Yes, let's change this, we can use `@node.has_tag('locked')` instead!

jywarren: @Manasa2850 what do you say? Thanks all!

Manasa2850:
@jywarren I like your idea of moving the lock button to the advanced menu, but I feel it should be there on the main page too since that's a more obvious place for admins / mods to see. Let's have the button in both places. What do you think?

@RuthNjeri @jywarren sure I'll put a condition for the 'locked' tag so that it works even when it's not the first one.
Thanks!!

Manasa2850:
@RuthNjeri @jywarren sorry I might be wrong but line 359 `node_tag = NodeTag.where(nid: params[:nid], tid: params[:tid]).first` refers to the 'locked' tag when we click the delete tag option on the 'locked' tag right? Because `tid` is unique for every tag we create. So when we click on the delete tag option on the 'locked' tag, `node_tag` will correspond to the 'locked' tag itself. I don't think it will make a difference whether 'locked' is the first tag in the list of tags or not.

I tried it out when 'locked' is not the first tag and it seems to be working fine.
Please do let me know if I'm missing out on something here. Thanks!

RuthNjeri:
> @RuthNjeri @jywarren sorry I might be wrong but line 359 `node_tag = NodeTag.where(nid: params[:nid], tid: params[:tid]).first` refers to the 'locked' tag when we click the delete tag option on the 'locked' tag right? Because `tid` is unique for every tag we create. So when we click on the delete tag option on the 'locked' tag, `node_tag` will correspond to the 'locked' tag itself. I don't think it will make a difference whether 'locked' is the first tag in the list of tags or not.
> 
> I tried it out when 'locked' is not the first tag and it seems to be working fine.
> Please do let me know if I'm missing out on something here. Thanks!

Hi Manasa, yes you are right, if someone clicks on the `locked` tag to try and delete it, the tid in the query will be that of the locked tag. However, let us say a user is trying to delete another tag, lets say `air quality` and that node, for instance a research note, has the `locked` tag present, the tag retrieved from the delete is the `air quality` and it might get deleted because we have not checked if the current node has an additional locked tag...

From my understanding, a research note can have many tags and in order to freely delete any of the tags, it must not contain the `locked` tag...

app/views/notes/show.html.erb
RuthNjeri: @Manasa2850 I just tested the `Lock Note` option on the stable site, and on the first click, the reload does not add the tag, I have to click it again or manually reload the page. I am using a chrome browser. 


https://user-images.githubusercontent.com/7622875/121544523-7836a000-ca12-11eb-85cb-579019d695d6.mov

Manasa2850: @RuthNjeri I'm not really sure why that's happening on the stable site. I tried it out on localhost both on Chrome as well as Firefox and the tag gets added on reload. There's no need to refresh the page. 
ezgif com-gif-maker

I'll take a deeper look at this to find out why that's happening on the stable site.

RuthNjeri: I also think you need to be an admin to test this out on the stable site, I don't know how to make someone an admin on the stable site, but if you share your username with Jeff, he will be able to do that.

Manasa2850: My username on the stable site is Manasa2850.

Manasa2850: @RuthNjeri I tried it out on stable and this is what I observer
1. On Firefox - it works perfectly fine. We don't need to refresh the page for the locked tag to get added.
stable-firefox

2. On Chrome - sometimes it works fine, without requiring a refresh. But few other times, on clicking the `Lock Note` button, the note gets locked (i.e the lock favicon appears and the Lock Note button goes away) but the 'locked' tag card doesn't appear in the sidebar. We need to refresh the page in order to see it on the sidebar.
stable-chrome

I will try to find out why this is happening just on Chrome. Thanks!

RuthNjeri: Thanks for the update @Manasa2850

PR #9752, opened 2021-06-07
Prune Comment System Tests for Faster CI

noi5e:
This should have happened a long time ago...

This is all up for debate, but I'm pruning the comment system tests so that System Test CI doesn't take 20 minutes. Locally, when I run this PR, the comment system tests take 4 minutes. Big improvement!

I know the structure of `comment_test.rb` is confusing, so I'll break it down here. The system tests are divided into 3 parts:

1. Posting comments, posting comment replies, deleting comments, and editing comments. These four basic CRUD-type functions are tested for the Rails commenting system, and then run a **second** time on the React commenting system.
2. The vast majority of comment tests are run on research notes.
3. A select few tests are run on research notes, **AND** wikis, **AND** questions. 

This PR does the following:
- Comments out tests for cross-wiring image upload bugs (see #8670 for more context). The comment system rewrite reduces the likelihood of these bugs occurring. But if they ever come back, we can un-comment these again.
- Moves comment moderation tests from Part 3 to Part 2, so they're only run on research notes, and not the other two page types.
- Moves image upload and rich-text tests from Part 3 to Part 2 as well.

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -171,7 +171,7 @@ jobs:
         DB_PASSWORD: root
         DB_PORT: ${{ job.services.mysql.ports[3306] }}
       run: bundle exec rails g webpacker:install && bundle exec rails g webpacker:install:react && bundle exec rails g react:install
-    - name: "System Tests"
+    - name: "Comment System Tests"
       env:
         RAILS_ENV: test
         DB_PASSWORD: root
@@ -181,7 +181,7 @@ jobs:
       run: |
         export DISPLAY=:99
         chromedriver --url-base=/wd/hub &
-        bundle exec rails test:comment_system_tests
+        bundle exec rails test test/comment_system_tests
     - name: Archive system test screenshots
       uses: actions/upload-artifact@v2
       with:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -141,6 +141,53 @@ jobs:
         name: system-test-screenshots
         path: tmp/screenshots/*
 
+  comment-system-tests:
+    runs-on: ubuntu-latest
+    services:
+      redis:
+        image: redis
+        # Set health checks to wait until redis has started
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          # Maps port 6379 on service container to the host
+          - 6379:6379
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.6.6 # Not needed with a .ruby-version file
+        bundler-cache: true # runs 'bundle install' and caches installed gems automatically (not working?)
+    - uses: nanasess/setup-chromedriver@v1.0.1
+    - uses: ./.github/actions/setup-test-environment
+    - name: Install JavaScript dependencies with Yarn
+      run: yarn check || yarn install --frozen-lockfile;
+    - name: Run webpacker setup
+      env:
+        RAILS_ENV: test
+        DB_PASSWORD: root
+        DB_PORT: ${{ job.services.mysql.ports[3306] }}
+      run: bundle exec rails g webpacker:install && bundle exec rails g webpacker:install:react && bundle exec rails g react:install
+    - name: "System Tests"
+      env:
+        RAILS_ENV: test
+        DB_PASSWORD: root
+        DB_PORT: ${{ job.services.mysql.ports[3306] }}
+        REDIS_HOST: localhost
+        REDIS_PORT: 6379
+      run: |
+        export DISPLAY=:99
+        chromedriver --url-base=/wd/hub &
+        bundle exec rails test:comment_system_tests
+    - name: Archive system test screenshots
+      uses: actions/upload-artifact@v2
+      with:
+        name: system-test-screenshots
+        path: tmp/screenshots/*
+
   remove-labels:
     runs-on: ubuntu-latest
     needs: [add-label, unit-tests, functional-tests, integration-tests, system-tests]
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,7 +3,7 @@
 comment:
   layout: "reach, diff, flags, files"
   behavior: default
-  after_n_builds: 4
+  after_n_builds: 5
   require_changes: true  # if true: only post the comment if coverage changes
   require_base: yes       # [yes :: must have a base report to post]
   require_head: yes       # [yes :: must have a head report to post]
@@ -27,4 +27,4 @@ codecov:
   # https://docs.codecov.io/v4.3.6/docs/comparing-commits
   allow_coverage_offsets: true
   notify:
-    after_n_builds: 4
+    after_n_builds: 5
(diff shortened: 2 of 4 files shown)


noi5e:
Down to 12 minutes in this PR, compared to 24 in the other one I have open.

RuthNjeri:
This is pretty neat @noi5e...my question is about

> Comments out tests for cross-wiring image upload bugs (see #8670 for more context). The comment system rewrite reduces the likelihood of these bugs occurring. But if they ever come back, we can un-comment these again.

In my opinion, those tests should not be commented out. In the event they fail, it would signal to us that the bug has re-occured. It is better for us to find out from the tests first than the user...let me know your thoughts...

RuthNjeri (approved):

jywarren:
Hi, this is a great conversation and an impressive speed-up!

Just so we know all the possibilities, is it at all possible to parallelize these tests, for example to run some of the system tests concurrently? Either by splitting up the tests into 2 suites, or by using some switch to run half the tests in one job and half in another? I totally hear @RuthNjeri's input on keeping them online, i just wonder if we can have both?

noi5e:
@RuthNjeri and @jywarren thanks for the reviews and thoughts. I definitely see the importance of keeping the tests just in case they pop up again in the future.

@jywarren Do you have any links for documentation on how to run these tests in parallel? Is it something like [what's described here](https://dev.to/arturt/how-to-run-minitest-parallel-tests-on-github-actions-bmd)? Or would it be something that's done in Capybara (or whatever our testing suite is exactly, still not sure 😅)

jywarren:
Yes exactly, in different jobs most likely in GitHub actions. This is relatively low priority though, and I worry that you're taking on more than your scope, so please only take this on if it brings you joy!

jywarren:
What if we used this PR to try moving the commented-out tests into a separate job? Does it work to simply make a folder and file called `/commenting/full_stack_comment_tests.rb` and then invoke it via `rails test test/commenting/full_stack_comment_tests.rb` as described in https://github.com/publiclab/plots2/blob/main/doc/TESTING.md#running-just-one-test ?

Then we could run it with a copy of the system tests job, changing only this last line:

https://github.com/publiclab/plots2/blob/cf6f6818562e960a772fef610610e647b9a7e2a3/.github/workflows/tests.yml#L130

to:

`rails test test/commenting/full_stack_comment_tests.rb`

noi5e:
> What if we used this PR to try moving the commented-out tests into a separate job? Does it work to simply make a folder and file called `/commenting/full_stack_comment_tests.rb` and then invoke it via `rails test test/commenting/full_stack_comment_tests.rb` as described in https://github.com/publiclab/plots2/blob/main/doc/TESTING.md#running-just-one-test ?

I'll try experimenting with this during this week. Just `git rebase`d!

jywarren:
Just a note that due to our codecov thresholds config, we'll have to change `after_n_builds` to 5 in two spots in this file, to merge this!

https://github.com/publiclab/plots2/blob/a65efa664aadf756601a39b4a96f973fa099ee16/codecov.yml#L6-L30

Looks good though @noi5e !!

noi5e:
Just pushed a commit increasing `after_n_builds` to 5, from 4.

noi5e:
The new `comment-system-tests` workflow won't start for some reason, here's the most recent error. I got this same error even before I made the last commit to change `after_n_builds`:

Screen Shot 2021-07-27 at 2 22 39 PM

noi5e:
I changed the `bundle exec rails test:comment_system_tests` command to `bundle exec rails test test/comment_system_tests`. I think this will still fail.

My guess is that for whatever reason Rails needs extra instruction in order to run tests in a custom directory (that's not `unit`, `functional`, `integration`, or `system`). But how to get it to do that? Something like [this from StackOverflow](https://stackoverflow.com/questions/48465072/run-tests-in-custom-test-folder-in-rails-5)?

noi5e:
Again, the test doesn't start, screenshot of the most recent failed test build:

Screen Shot 2021-07-28 at 11 03 48 AM

jywarren:
Hm, i think the "unexpected end" was from a mistaken extra "end" on line 480, maybe? I tried removing it and was able to run `rails test test/comment_system_tests` in GitPod (although without chromedriver installed there the tests didn't actually run). Let's see if that gets them to run in Github Actions?

jywarren:
🎉 🎉 🎉

jywarren:
Cool but whoa -- `tests / integration-tests (pull_request) Successful in 26m`

Screen Shot 2021-07-28 at 8 51 12 PM

vs in https://github.com/publiclab/plots2/pull/9958 -- 

Screen Shot 2021-07-28 at 8 51 44 PM

Should we re-trigger them to try to get better times?

jywarren:
Re-running to see...

jywarren:
Wow whats going on with integration tests?

Screen Shot 2021-07-29 at 10 03 02 AM

jywarren:
Running again. Could it have to do with time of day or something? Or a GitHub Actions slowdown? The integration tests themselves really did clock 1390 seconds or 23 minutes.

jywarren:
Hmm. It's pretty consistent. What do you think is going on here?

jywarren:
OK - speed issues seem to be across the board. Here's an unrelated recent PR: https://github.com/publiclab/plots2/pull/9965

Huh, yes, they're all this way: https://github.com/publiclab/plots2/actions/runs/1072626662

Screen Shot 2021-07-29 at 11 09 59 AM

jywarren:
This, 9 days ago affected integration tests for the first time in a few months: https://github.com/publiclab/plots2/pull/9878/files

jywarren:
But it was only a whitespace change, actually. So it could be a change to our actual application code that has caused integration tests to slow significantly.

jywarren:
This, 2 days ago, ran fast: https://github.com/publiclab/plots2/actions/runs/1071468860

Screen Shot 2021-07-29 at 11 14 44 AM

Hmm.

jywarren:
OK, the change was between:

`change webpacker.yml to see public/lib`
#9959 merged 2 days ago

`Update merge-pr-sentry-release.yml to change sentry environment to "staging"`
#9956 merged 2 days ago

So it was likely the change to `webpacker.yml` in response to the tocbot fix:

```diff
-  additional_paths: []
+  additional_paths: ['public/lib']
```

Since we found that this didn't actually solve the problem, and merged https://github.com/publiclab/plots2/pull/9918 instead, let's just revert #9959 for now to speed things back up.

jywarren:
Now watching https://github.com/publiclab/plots2/pull/9966 for faster load times. Let's merge that and then rebase this on top of it to get a final speed check (assuming this works).

jywarren:
That worked. They failed due to an intermittent issue, but went back to their original load times:

Screen Shot 2021-07-29 at 12 11 28 PM

jywarren:
Now rebasing over #9966 having just merged it. Fingers X and we should be able to merge this then!

jywarren:
Let's also update the branch protection rules to include this additional set "comment-system-tests" once we merge this!

jywarren:
Yay!!!

image

I'll merge this now!!!

noi5e:
Nice! Thanks for the assist on this @jywarren

PR #9772, opened 2021-06-10
Tests for tag locking

Manasa2850:
Part of #9694

Thanks!

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -359,7 +359,7 @@ class TagController < ApplicationController
     node_tag = NodeTag.where(nid: params[:nid], tid: params[:tid]).first
     node = Node.where(nid: params[:nid]).first
     # only admins, mods, and tag authors can delete other peoples' tags
-    if node_tag.uid == current_user.uid || logged_in_as(['admin', 'moderator']) || (node.uid == current_user.uid && node_tag.name != "locked")
+    if (node_tag.uid == current_user.uid && !node.has_tag('locked')) || logged_in_as(['admin', 'moderator']) || (node.uid == current_user.uid && !node.has_tag('locked'))
 
       tag = Tag.joins(:node_tag)
                    .select('term_data.name')
@@ -385,8 +385,8 @@ class TagController < ApplicationController
           end
         end
       end
-    elsif node_tag.name == "locked"
-      flash[:error] = "Only admins can delete the locked tag."
+    elsif node.has_tag('locked')
+      flash[:error] = "Only admins can delete tags on locked pages."
       redirect_to Node.find_by(nid: params[:nid]).path
     else
       flash[:error] = I18n.t('tag_controller.must_own_tag_to_delete')
--- a/app/models/node.rb
+++ b/app/models/node.rb
@@ -1062,7 +1062,7 @@ class Node < ActiveRecord::Base
       errors ? "Only admins may create raw pages." : false
     elsif tagname[0..4] == 'rsvp:' && user.username != one_split
       errors ? I18n.t('node.only_RSVP_for_yourself') : false
-    elsif tagname == 'locked' && user.role != 'admin'
+    elsif tagname == 'locked' && !user.can_moderate?
       errors ? I18n.t('node.only_admins_can_lock') : false
     elsif tagname == 'blog' && user.role != 'admin' && user.role != 'moderator'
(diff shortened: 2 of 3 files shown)


test/functional/tag_controller_test.rb
```diff
@@ -48,6 +48,16 @@ def setup
     assert_response :success
   end
 
+  test 'only admins can added the locked tag' do
```
RuthNjeri, line 51: Thanks for creating the test @Manasa2850.. this looks good, the `added` is in past tense, can it be in the present like `'only admins can add the locked tag'`

Also, I think another test that confirms an admin and a moderator can add a locked tag will help.

Manasa2850: Oh sorry my bad :sweat_smile: 
Sure, I'll add the other test too! Thanks!

test/functional/tag_controller_test.rb
```diff
@@ -48,6 +48,25 @@ def setup
...
+         params: {
+            name: 'locked',
+            nid: nodes(:one).nid
+         }
+    assert_equal 'Error: only admins can lock pages.', assigns[:output][:errors][0]
+  end
+
+  test 'admin and moderators can add the locked tag' do
+    UserSession.create(users(:admin))
+    UserSession.create(users(:moderator))
```
RuthNjeri, line 63: Thanks for adding this test @Manasa2850!

Is the UserSession being used from line 64-68? It seems to me that line 64-68 is interacting with the models directly and looks more like a unit test. There is no post request using line 62 or 63...

This test suite can run without the UserSession, but I wonder if that makes sense since this is a controller test? Which is testing the actions?

My suggestion would be to split this test into 2, one for the admin and another for the moderator, where you can use the UserSession and hit the endpoint after, similar to what you have done in line 51

Manasa2850: Sure @RuthNjeri I'll split it into 2 tests. I had combined them into one test to avoid repetition but what you mentioned about using UserSession makes more sense. Thank you!

app/models/node.rb
```diff
@@ -1062,7 +1062,7 @@ def can_tag(tagname, user, errors = false)
       errors ? "Only admins may create raw pages." : false
     elsif tagname[0..4] == 'rsvp:' && user.username != one_split
       errors ? I18n.t('node.only_RSVP_for_yourself') : false
-    elsif tagname == 'locked' && user.role != 'admin'
+    elsif tagname == 'locked' && !user.can_moderate?
```
Manasa2850, line 1065: This was initially only for admins. I've changed it to include moderators too.

app/controllers/tag_controller.rb
```diff
@@ -359,7 +359,7 @@ def delete
     node_tag = NodeTag.where(nid: params[:nid], tid: params[:tid]).first
     node = Node.where(nid: params[:nid]).first
     # only admins, mods, and tag authors can delete other peoples' tags
-    if node_tag.uid == current_user.uid || logged_in_as(['admin', 'moderator']) || (node.uid == current_user.uid && node_tag.name != "locked")
+    if (node_tag.uid == current_user.uid && !node.has_tag('locked')) || logged_in_as(['admin', 'moderator']) || (node.uid == current_user.uid && !node.has_tag('locked'))
```
Manasa2850, line 362: This is with reference to [this](https://github.com/publiclab/plots2/pull/9709#discussion_r644721788) in PR #9709.

jywarren (approved):
Looks great! I'll leave it to @RuthNjeri to merge as you've reviewed more closely than I. Thank you both!

app/controllers/tag_controller.rb
RuthNjeri: Thanks for adding this @Manasa2850, this looks good to me. I am curious about the comment in line 361...

`# only admins, mods, and tag authors can delete other peoples' tags` 
from the code, it seems that even a node_tag author or the author of the node cannot delete a tag as long as a node contains the `locked` tag?

So maybe the comment should be `# only admins, mods can delete other peoples' tags if the tags contain the locked tag`

RuthNjeri (approved):
Everything looks great @Manasa2850, I only had one comment about the comment in the tag controller here https://github.com/publiclab/plots2/pull/9772/files#r654912815 which is not that urgent...

I am merging it now. Great job 🎉

Manasa2850:
Thanks @RuthNjeri. I'll create an FTO for the changes in line 361.

Issue #9667, opened 2021-05-25
[Planning] Improving Education Related Usability and Functionality Project (Outreachy)


TildaDares:
This issue lists a breakdown of my project tasks. There's still a lot to be added but I'll start with this first.

[My Outreachy proposal](https://publiclab.org/notes/mathildaudufo/04-26-2021/outreachy-education-related-usability-upgrades)
[Original Project Description](https://github.com/publiclab/plots2/issues/9257)

- [ ] Write unit test for method that finds user's drafts
- [ ] Add listing of drafts posts on profile page
- [ ] Display drafts tab in author's page
- [ ] Create FTO issue to replace 'with' with comma in co-authored posts
- [ ] Write unit test for method that checks if a tag is a location tag
- [ ] Add bootstrap tooltips explaining longitude and latitude tags
- [ ] Implement design for co-authored posts page
- [ ] Create FTOs for moving the new v2 dashboard redesign to be the new dashboard
- [ ] Add table of contents on wiki/markdown pages
- [ ] Cache prefetched names for mentions either daily or weekly
- [ ] Intermix prefetched usernames and API responses, de-duplicating and client side filtering for alphanumeric match
> The two tasks above still need breaking down. I'm still trying to understand it.

I won't be including the design discussions in my planning.
Also more tasks will be included from this [milestone](https://github.com/publiclab/plots2/milestone/13)

jywarren:
Nice! i linked the milestone and see this issue in its description. Great work!

Tlazypanda:
@TildaDares Awesome first draft 🎊  And I see the FTOs 👀  amazing!

TildaDares:
> * figure out how to intermix prefetched usernames and API responses, de-duplicating and client-side filtering for alphanumeric match

@RuthNjeri Can you break this down for me? It's mentioned in the project description.

jywarren:
Hi @TildaDares -- i think i wrote this line 😅 

What i was thinking was, if we fetch some recently used (we'll call them "prioritized") usernames and we want them to float to the top of the results list for convenience, we don't want them to later show up a 2nd time further down the list. So this has a few parts:

1. figure out how to blend the two lists, both as a matter of API (do we do this on the client side or the server side? pros/cons of each?) 
[...]

PR #9773, opened 2021-06-10
Autocompletion of usernames in comment boxes

### LINKED ISSUE #9755: [Discussion] Autocompletion in comment boxes
I've been looking through At.js docs https://github.com/ichord/At.js/wiki/Callbacks, https://github.com/ichord/At.js/wiki and https://github.com/publiclab/plots2/blob/main/app/assets/javascripts/atWhoAutoComplete.js. From what I understand I think the `data` key is what is supposed to hold the pre-fetched usernames

```
data: ["one", "two", "three"]
```

But instead of hardcoded names, the names are generated from recently active users on the site. I looked through the repo for the query that gets the recently active users on the site and I found this https://github.com/publiclab/plots2/blob/700fdf41a08d52bfd51d7774902dc9217ed225b5/app/controllers/users_controller.rb#L137-L142
I'm a little confused on how to display the results of that query in json format in https://github.com/publiclab/plots2/blob/700fdf41a08d52bfd51d7774902dc9217ed225b5/app/api/srch/search.rb

@jywarren @cesswairimu @RuthNjeri

TildaDares:
Fixes #9755

Part of larger planning issue in #9667

Test here https://unstable.publiclab.org/

https://user-images.githubusercontent.com/63427719/121555040-8ed8f900-ca0a-11eb-82fe-e58b37a24c5d.mov

Thanks!

### DIFF
--- a/app/assets/javascripts/atWhoAutoComplete.js
+++ b/app/assets/javascripts/atWhoAutoComplete.js
@@ -1,8 +1,17 @@
 (function() {
 
   // settings at https://github.com/ichord/At.js/wiki/Base-Document#settings
+
+  // checks if the 'name' key in the JSON data is named 'username' or 'name' and then returns the
+  // correct key-value
+  const displayName = (item) => item.username ? item.username : item.name;
+
   var at_config = {
     at: "@",
+    displayTpl: (item) => `<li>${displayName(item)}</li>`,
+    insertTpl: (item) => `@${displayName(item)}`,
+    // loads and saves remote JSON data by URL
+    data: '/users/active',
     delay: 400,
     callbacks: {
       remoteFilter: debounce(function(query, callback) {
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -505,6 +505,18 @@ class User < ActiveRecord::Base
     recent_locations.last
   end
 
+  def self.recently_active_users(limit = 5, order = 'last_updated DESC')
+    Rails.cache.fetch('users/active', expires_in: 24.hours) do
+      User.select('rusers.username, rusers.status, rusers.id, MAX(node_revisions.timestamp) AS last_updated')
+        .joins("INNER JOIN `node_revisions` ON `node_revisions`.`uid` = `rusers`.`id` ")
+        .where("node_revisions.status = 1")
+        .where("rusers.status = 1")
+        .group('rusers.id')
+        .order(order)
+        .limit(limit)
+    end
+  end
+
   private
 
   def decrease_likes_banned
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -412,6 +412,11 @@ class UsersController < ApplicationController
     redirect_to "/login", flash: { notice: action_msg }
   end
 
+  def recently_active_users
+    active_users = User.recently_active_users
+    render json: active_users, root: false
+  end
+
   private
 
   def subscribe_multiple_tag(tag_list)
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -53,6 +53,7 @@ Plots2::Application.routes.draw do
   get 'signup' => 'users#new'
   get 'home' => 'home#front'
   get 'verify/:token' => 'users#verify_email'
+  get 'users/active' => 'users#recently_active_users'
   resources :relationships, only: [:create, :destroy]
 
   get '/wiki/:id/comments', to: 'wiki#comments'


app/models/user.rb
```diff
@@ -505,6 +505,17 @@ def latest_location
     recent_locations.last
   end
 
+  def self.recently_active_users(limit = 5, order = 'last_updated DESC')
+    Rails.cache.fetch('users/active', expires_in: 24.hours) do
+      User.select('rusers.username, rusers.id, MAX(node_revisions.timestamp) AS last_updated')
+        .joins("INNER JOIN `node_revisions` ON `node_revisions`.`uid` = `rusers`.`id` ")
+        .where("node_revisions.status = 1")
```
jywarren, line 512: I wonder if it would make things more efficient to add `.where("rusers.status = 1") as well here, and to the `.select()` above? We do have tens of thousands of user records. Just an idea, not sure if the join already makes this efficient!

app/assets/javascripts/atWhoAutoComplete.js
```diff
@@ -1,8 +1,14 @@
...
 
   // settings at https://github.com/ichord/At.js/wiki/Base-Document#settings
+
+  const displayName = (item) => item.username ? item.username : item.name;
+
   var at_config = {
     at: "@",
+    displayTpl: (item) => `<li>${displayName(item)}</li>`,
+    insertTpl: (item) => `@${displayName(item)}`,
+    data: '/users/active',
```
jywarren, line 11: How will this interact with the request down on line 15, of `$.getJSON("/api/srch/profiles?query=" + query + "&sort_by=recent&field=username", {}, function(data) {`? Does this override it?

jywarren: Maybe even if both can co-exist, we should add a comment clarifying the difference because it's a little hard to follow.

jywarren: aha! So `data` is the initial query after an `@`, and the `remoteFilter` is when you begin typing letters?

jywarren (changes requested):
This is looking great and the video is awesome. Just had a question about the difference between 2 queries; maybe a comment would help! Thanks!!!

app/assets/javascripts/atWhoAutoComplete.js
TildaDares: @jywarren Exactly! I'll add a comment to clarify this.

app/models/user.rb
TildaDares: @jywarren I'll add that and see if it works.

Tlazypanda (approved):
@TildaDares This looks great!! :tada: another consideration can be to play around a bit with the ordering of the clauses in our query and benchmark the results :v:

jywarren:
> @TildaDares This looks great!! 🎉 another consideration can be to play around a bit with the ordering of the clauses in our query and benchmark the results ✌️

Yes and you can also watch the rails log when running locally, and see if all the chained conditions are rolled up into a single query, and it should provide a timestamp of how long it took in development mode. But, if your local is a pretty small database, it may not really show performance issues on the scale we're looking at in the live site. 

Great ideas! Thanks!

app/models/user.rb
```diff
@@ -505,6 +505,18 @@ def latest_location
...
 
+  def self.recently_active_users(limit = 5, order = 'last_updated DESC')
+    Rails.cache.fetch('users/active', expires_in: 24.hours) do
+      User.select('rusers.username, rusers.status, rusers.id, MAX(node_revisions.timestamp) AS last_updated')
+        .joins("INNER JOIN `node_revisions` ON `node_revisions`.`uid` = `rusers`.`id` ")
+        .where("node_revisions.status = 1")
+        .where("rusers.status = 1")
+        .group('rusers.id')
+        .order(order)
+        .limit(limit)
```
jywarren, line 516: Here, i wonder -- is 5 enough? What if we changed the limit to 15? Or, alternatively, what if there were a way to narrow this list based on something specific to the current user? But, that might increase the complexity of the query and therefore slow it down a lot. 

Finally, i wonder if we could add caching to this query. That could make this really speedy and it seems OK for there to be at least an hour long cache here, maybe? Look at this example - 

https://github.com/publiclab/plots2/blob/cfba9ef309198ca0731f9c0d54214816c8915278/app/controllers/stats_controller.rb#L5

This can be added in followup, so I'll go ahead and merge this now, but these are maybe some more refinements we can make -- and they're simple enough that maybe they could be an FTO too!

jywarren (approved):

jywarren:
Just noting that I think this is ready to go -- we can test it in stable, and we won't be publishing to the live site for at least a few days or maybe longer -- and the extra ideas I suggested above could be done in another PR!

Thanks, everyone! @TildaDares this is great work!

app/models/user.rb
TildaDares: > What if we changed the limit to 15?

@jywarren Setting the limit to 15 sounds like a good idea. 


> Finally, i wonder if we could add caching to this query. That could make this really speedy and it seems OK for there to be at least an hour long cache here, maybe? Look at this example -

The query already has a cache or is there somewhere else I'm supposed to add it? The cache has an expiration time of 24 hours so I'll change it to an hour.

https://github.com/publiclab/plots2/blob/c76e4859dcb69090ccf31ed8fe6140e4eaa567d2/app/models/user.rb#L509

jywarren:
ah, ok, perfect! Thank you!

Issue #9686, opened 2021-05-26
[Planning] Translation System Refinement


imajit:
I have not completely planned this out yet but these are some parts I'll work on : 
- [x] UI Issues with Translation icon
    - [x] Fix position of icon
    - [x] Improve pop-up 
    - [x] Fix UI 
- [ ] Translation function improvement
    - [x] Handle inline function calls
    - [x] Positioning and visibility of icon
    - [x] Reduce Icon Count
    - [ ] Dialect Support [Discussion #9809  ]
- [ ] Tests for translation function
    - [ ] Tests for new changes in Translation function
    - [ ] Tests for UI changes
    - [ ] Tests for FTO related changes
- [x] Setting up workflow to import translations 
    - [x] Create a dummy project to safely test integration
    - [x] Export Strings from Github Repo to Transifex
    - [x] Importing Reviewed strings from Transifex
- [ ] Translation Dashboard and Community box section 
    - [ ] Add FAQ section to existing wiki page 
- [ ] Test for FTOs issue 
    - [x] Verify UI changes in Gitpod
    - [ ] FTO issue template 
- [ ] Add Documentation
    - [x] New Translation Function
    - [ ] UI verification in Gitpod
    - [x] Transifex integration
I'll add subtasks and other tasks. Feel free to add new ideas/tasks :smile:

jywarren:
These are looking good! Many may break out into subtasks, agreed! As you go, you might convert some into headings instead of tasks. Cool!

jywarren:
I also linked you to https://github.com/publiclab/plots2/milestone/22, and if you want we can make this issue the new "lead" issue there by reordering and mentioning it in the description!

Tlazypanda:
Hey @imajit Nice breakdown 💯  Here is a comment that I made on @Manasa2850 planning issue that might be helpful - https://github.com/publiclab/plots2/issues/9694#issuecomment-850962289 

Also It might be helpful for both of you to go through each other's planning issue and help brainstorm. Brownie points for finding overlap if any and handling that or expanding your proposals a bit to get some new ideas onboard. Thank you! Have fun 🎊

imajit:
![Screenshot from 2021-06-03 01-57-37](https://user-images.githubusercontent.com/38528640/121190289-86f14b80-c888-11eb-9186-dba3c9e56dcf.png)
[...]

PR #9787, opened 2021-06-12
Fixed inline translation function call with translate(key, {}, html)

imajit:
Part of #9686 
Translation function needs to be called with `false` parameter in cases where it is called inside HTML tags.


app/views/dashboard_v2/_blog.html.erb
```diff
@@ -7,17 +7,17 @@
         <div aria-hidden = 'true' class="img" style="background-image: url('<%= @blog.scraped_image %>')">
       <% end %>
       <%# subscribe to blog updates possibly via email %>
-      <a class="ellipsis pull-right pl-2" data-toggle="dropdown" rel="tooltip" title="<%= translation('tag.show.blog_updates') %>" style="cursor:pointer; background-color: #444" aria-expanded="true"><i class="fa fa-ellipsis-h" style="color:#ccc;font-size:18px;margin-right:10px;"></i></a>
+      <a class="ellipsis pull-right pl-2" data-toggle="dropdown" rel="tooltip" title="<%= translation('tag.show.blog_updates',{},false) %>" style="cursor:pointer; background-color: #444" aria-expanded="true"><i class="fa fa-ellipsis-h" style="color:#ccc;font-size:18px;margin-right:10px;"></i></a>
       <div class="dropdown-menu">
         <div class="dropdown-item">
-          <a style="margin-top:-8px;" rel="tooltip" title="<%= translation('tag.show.blog_updates') %>" href="/subscribe/tag/blog">
+          <a style="margin-top:-8px;" rel="tooltip" title="<%= translation('tag.show.blog_updates',{},false) %>" href="/subscribe/tag/blog">
```
jywarren, line 13: Hi @imajit this is looking really nice! I wonder if it's possible to simplify the parameters and omit the `{}` by checking if `options == true` OR by using `translation(key, {html: true})` for some slightly shorter cleaner syntax?  This could definitely be done in follow-up, as this PR looks good. Are there any other places where this is used and where it would need to be updated?

jywarren (approved):
Just a note on the parameter simplicity. I was also wondering -- can you add to the translation documentation in the README to clarify the usage of this new parameter? Again it's fine to do this in another PR -- but if you can add an item to your planning issue in https://github.com/publiclab/plots2/issues/9686 that would be great. Thank you!!!

app/views/dashboard_v2/_blog.html.erb
imajit: Yeah, I'll change the syntax to make it more clear. 
I think I have changed all the occurrences. If I find any new ones I'll make ftos to change them.

imajit:
> Just a note on the parameter simplicity. I was also wondering -- can you add to the translation documentation in the README to clarify the usage of this new parameter? Again it's fine to do this in another PR -- but if you can add an item to your planning issue in #9686 that would be great. Thank you!!!

Sure, will add documentation about the updates in the function.

Issue #9365, opened 2021-03-23
[Phase 2] React Rewrite of Comment System


noi5e:
The first phase of the React rewrite of the Comment/Legacy System was merged in #9176. Yay! 🥳 

To anyone who wants to see the React system in production, visit any research note, and add the following parameter to the end of the URL:

`?react=true`

I'm going to do a lot of follow-up work (in multiple PRs) in the next two weeks to try to get it as close to the fully functioning system as possible.

Including:
- system tests comparing performance with the Rails system
- emoji reaction system
- comment preview
- moderated comments

### `To-Do List` _(last edited Tuesday March 23rd)_
- [ ] ❗ system tests showing that React can properly do CRUD with comments
- [ ] ❗ assign propTypes to nested elements
- [ ] ❗ implement useReducer hook
- [ ] ❗ guest browsing: show link to `/login` instead of the main comment form
- [ ] ❗ temporarily disable user submission & clicking for comment publishing, updating, and deleting; add loading spinners
- [ ] functional tests for the controller making React's initial props
- [ ] add prompt before deleting comment
- [ ] get ESLint for React & React Hooks working again, and working with codeclimate CI too
- [ ] move `get_initial_react_props` method from the controller to the `Comment` model
- [ ] make initial Component on `document.load` a spinner; AJAX request afterward for initial props

noi5e:
I'm back!

Starting to add system tests ensure that basic functionality for React comments exists. See #9665, which tests posting comments, and comment replies.

An added benefit of system tests for React comments is that we can use test runtime as a rough measure of whether or not React comments are faster than the Rails comment system.

**React**:
<img width="381" alt="Screen Shot 2021-05-24 at 2 15 11 PM" src="https://user-images.githubusercontent.com/4361605/119410525-d71fb980-bc9d-11eb-8213-82015f576b70.png">

**Rails**:
<img width="379" alt="Screen Shot 2021-05-24 at 2 16 20 PM" src="https://user-images.githubusercontent.com/4361605/119410561-e6066c00-bc9d-11eb-8dc1-90764254671d.png">

A difference of .707 seconds, or 3% faster!

jywarren:
This is awesome. Thanks @noi5e !!!

noi5e:
Opened PR for editing comments.

Here's the new speed breakdown for all 3 new tests:

**React:**
[...]

PR #9801, opened 2021-06-15
Implement useReducer for Comment State Management

noi5e:
Part of a React rewrite of the commenting system, see #9365 for more context.

In this PR, I begin to switch state management from React's [useState](https://reactjs.org/docs/hooks-state.html) hook to [useReducer](https://reactjs.org/docs/hooks-reference.html#usereducer) instead. `useReducer` is modeled after other React state management libraries like [Redux](https://redux.js.org/).

Why the switch? The app state is starting to get very complex, and will only get more complex as features are added. For example, let's take the example of "_disable submit comment button after user clicks on it, and before server sends a response_". This is a feature which I'm going to start work on soon, and it will require adding more variables into state: `waitingForServerResponse`, `submitButtonIsDisabled`, and the like.

Switching over to `useReducer` will make this codebase easier to understand and debug in the long run.

I've created a new `reducers.js` file. Everything the app does that changes state will be described in this file. For now, it looks like this: 
```
/* eslint-disable complexity */
import { makeDeepCopy } from "./helpers";

const reducer = (state, action) => {
  switch(action.type) {
    case "CREATE COMMENT":
      return {
        ...state,
        comments: [
          ...state.comments, 
          action.newComment
        ]
      };
    case "UPDATE COMMENT":
      for (let i = 0; i < state.comments.length; i++) {
        // find the comment in state
        if (state.comments[i].commentId === action.newComment.commentId) {
          let newComment = makeDeepCopy(state.comments[i]);
          newComment.htmlCommentText = action.newComment.htmlCommentText; // update comment text
          newComment.rawCommentText = action.newComment.rawCommentText;
          return {
            ...state,
            comments: Object.assign(
              [],
              state.comments,
              { [i]: newComment }
            )
          }; // keep the rest of state.comments, but replace comment at index i with newComment
        }
      }
      break;
    case "DELETE COMMENT":
      for (let i = 0; i < state.comments.length; i++) {
        // find the comment in state by ID
        if (state.comments[i].commentId === action.commentId) {
          return {
            ...state,
            comments: state.comments.filter(comment => action.commentId !== comment.commentId)
          };
        }
      }
      break;
    default:
      throw new Error(); // default should never be called
  }
}

export {
  reducer
}
```

---

### DIFF
--- a/app/javascript/components/CommentsContainer.js
+++ b/app/javascript/components/CommentsContainer.js
@@ -1,8 +1,8 @@
-import React, { useState } from "react";
+import React, { useReducer, useState } from "react";
 import PropTypes from "prop-types";
 
 import { UserContext } from "./user-context";
-import { makeDeepCopy } from "./helpers";
+import { reducer } from "./reducers";
 
 import CommentForm from "./CommentForm";
 import CommentsHeader from "./CommentsHeader";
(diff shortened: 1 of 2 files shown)


noi5e:
Ready to merge! Tests pass. Asking for an override on the codeclimate issues :)

jywarren:
Fantastic work, @noi5e, thank you!

PR #9810, opened 2021-06-15
Fixed globe icon placement in buttons by adding left padding

imajit:
Part of #9686 
In my last PR I had by reduced the button to the red part in the nav bar
Screenshot from 2021-06-15 16-55-32
I have corrected that part now the entire button is clickable in cases where globe icon is not there , here entire green part is now clickable 
Screenshot from 2021-06-15 16-56-09

For buttons I have added some left padding in the function,now the icons don't render inside the buttons and in cases where the button occupies the entire width globe icon is rendered below the button
**Admin Page Buttons** 
Screenshot from 2021-06-15 17-23-28

**Dashboard Buttons**
Screenshot from 2021-06-16 01-39-38

**Subscriptions Page Buttons**
Screenshot from 2021-06-16 01-48-07
The add button here is a `submit` button and has some tests linked with it. If you feel it is okay with the nested globe icon I can just leave it unchanged, else I'll change it with the globe icon outside the button and change the tests for it in a later PR.


jywarren:
Hi @imajit this looks lovely! Thank you for your attention to detail as well!

I think we used to limit this to 5 max appearing on the screen so that it's not overwhelming. Did we ever implement that? I think this is fine but that's the only thing I worry a little about here. Do you perhaps want to address that soon too? I don't think it has to block merging this one. 

Thank you!!!

app/views/sidebar/_dashboard.html.erb
```diff
@@ -7,12 +7,31 @@
         <span class="sr-only">Toggle Dropdown</span>
       </a>
       <div class="dropdown-menu dropdown-menu-right mt-2">
-        <a class="dropdown-item" href="/post?n=15322&title=How%20to%20...&tags=activity:new,draft"><%= translation('dashboard._header.dropdown.share_activity') %></a>
-        <a class="dropdown-item" href="/post"><%= translation('dashboard._header.dropdown.share_research_note') %></a>
-        <a class="dropdown-item" href="/questions/new?tags=question:general"><%= translation('dashboard._header.dropdown.ask_question') %></a>
-        <a class="dropdown-item" href="/post?template=event&tags=event"><%= translation('dashboard._header.dropdown.post_event') %></a>
-        <div class="dropdown-divider"></div>
-        <a class="dropdown-item" href="/post?tags=blog-submission"><%= translation('dashboard._header.dropdown.tell_story_blog') %></a>
+        <% if current_user&.has_tag('translation-helper') && I18n.locale != :en %>
```
jywarren, line 10: Ooh, actually... this is interesting. This repetition does make for a lot more code... i wonder... what do you think, is there any way to reduce repetition in this while still achieving what we want? What are some ideas to try to brainstorm this? 

Sorry to pause a moment, but I hope my hesitation makes sense, @imajit -- still very much appreciate this work!

imajit: Yeah, I agree there is a repetition in the code as I'm wrapping the button inside a div so that the globe icon stays inline with the text. 
-  If I keep the `if` part only, then the button size reduces for all users
- If I keep the `else` part, then the globe icon overflows to the next line
- I can't store it in a new file as partial as the repetition  is kind of nested inside each `div`
So, I did this bit ugly implementation to satisfy all cases.  One way to avoid this is if I reduce the number of globe icons in the navbar, in that case, there will be less code repetition and fewer number of globe icons on the Dashboard. 
For now, I can't see a different way of implementation though :confused:

imajit:
> Hi @imajit this looks lovely! Thank you for your attention to detail as well!
> 
> I think we used to limit this to 5 max appearing on the screen so that it's not overwhelming. Did we ever implement that? I think this is fine but that's the only thing I worry a little about here. Do you perhaps want to address that soon too? I don't think it has to block merging this one.
> 
> Thank you!!!

I did read about it in a discussion https://github.com/publiclab/plots2/issues/5651#issuecomment-706574770 but I don't think it has been implemented, there is a way using jquery to do this, 
`       jQuery(".translationClass").children(":gt(6)").hide();`
this will show only 5 globe icons, but I have never tested it on a live server with multiple users.

app/views/sidebar/_dashboard.html.erb
jywarren: Hmm.... I totally appreciate this is a tough one. Thanks for thinking hard on it. What if... we did it in JavaScript, after rendering? Could that work?

jywarren:
> this will show only 5 globe icons, but I have never tested it on a live server with multiple users.

Did you want to try it in the JavaScript dev console on stable.publiclab.org to get a sense for it?

app/views/sidebar/_dashboard.html.erb
imajit: I added a helper function to render the dropdown button, this reduces some repetition

app/helpers/application_helper.rb
```diff
@@ -160,11 +160,20 @@ def translation(key, options = {}, html = true)
...
-      raw(%(<span>#{translated_string} <a href="https://www.transifex.com/publiclab/publiclaborg/translate/#de/$?q=text%3A#{translated_string}">
+      raw(%(<span>#{translated_string} <a style='padding-left:3px;' href="https://www.transifex.com/publiclab/publiclaborg/translate/#de/$?q=text%3A#{translated_string}">
           <i data-toggle='tooltip' data-placement='top' title='Needs translation? Click to help translate this text.' style='position:relative; right:2px; color:#bbb; font-size: 15px;' class='fa fa-globe'></i></a>
        </span>))
     else
       raw(translated_string)
     end
   end
+
+  def create_dropdown(href, text)
```
jywarren, line 171: Ah, this is much cleaner. Could we rename this though so it's a little more specific? Let me think --- 
```suggestion
  def create_nav_dropdown_item(href, text)
```

jywarren (changes requested):
The helper function does look good. I suggested a slight renaming. Does this approach work for the lines in this partial too, like lines 11-26?

jywarren:
Also, this will need conflicts resolve as well. Thanks for sticking with this @imajit -- much appreciated!

imajit:
Hmm, I think this is due to some recently added tests , looking into

jywarren:
Ah indeed. You can try resolving directly or rebasing!

app/helpers/application_helper.rb
jywarren: OK awesome. I think this is one of the last things before we merge, right? Thank you!!

imajit: Yeah

jywarren (approved):

jywarren:
Nice! Thanks for doing this so thoroughly, @imajit !!

PR #9811, opened 2021-06-15
Changed pop-up statement

imajit:
Part of #9686 
Screenshot from 2021-06-16 02-38-19
I think instead of `this text` we can have pop-up with `the text "string to be translated" `, it makes pop-up bit more clear but makes the pop-up a bit longer.  

Screenshot from 2021-06-16 02-34-19


RuthNjeri (approved):
This is awesome!! It looks great to me @imajit

PR #9820, opened 2021-06-16
intermix prefetched usernames and API responses, de-duplicating and c…

TildaDares:
Blending the list on the server side would be complex and inflexible and since both APIs return different JSON objects manipulating their objects would be a lot harder which is why I chose to do it on the client side. 
 
No matter how I order the arrangement of both lists, the callback in the `remoteFilter` function always displays the usernames that match the first strings in the username at the top. In the video, you'll see this demonstrated when I type `@a`.

https://user-images.githubusercontent.com/63427719/122276759-07452b80-cedd-11eb-8349-303db6b3bc51.mov

You can test it here https://unstable.publiclab.org/

### DIFF
--- a/app/assets/javascripts/atWhoAutoComplete.js
+++ b/app/assets/javascripts/atWhoAutoComplete.js
@@ -2,13 +2,62 @@
 
   // settings at https://github.com/ichord/At.js/wiki/Base-Document#settings
 
+  let userObj = null;
+
+  // stores recently active users in userObj
+  (function storeRecentlyActiveUsers() {
+    fetch('/users/active').then(res => res.json()).then(data => { userObj = data })
+  })();
+
   // checks if the 'name' key in the JSON data is named 'username' or 'name' and then returns the
   // correct key-value
   const displayName = (item) => item.username ? item.username : item.name;
 
+  // displays a recently active text for prioritised users
+  const displayPriorityUsers = item => {
+    const username = displayName(item);
+    if(item.priority) {
+      return username + ' <small>recently active</small>'
+    }
+    return username
+  }
+
+  // returns a list of users that match the query
+  const filterUsers = query => {
+    const users = []
+    userObj.forEach(str => {
+      if(str.username.includes(query)){
+        users.push(str.username)
+      }
+    });
+
+    return users
+  };
+
+  // remove duplicates of prioritised usernames
+  const removeDuplicates = (userArr, prioritisedUsers) => {
+    const newData = []
+    userArr.forEach(i => {
+      if(!prioritisedUsers.includes(i.doc_title)){
+        newData.push({name: i.doc_title, priority: false})
+      }
+    });
+
+    return newData;
+  }
+
+  // merges normal and prioritised usernames
+  const mergeUsers = (recentlyActiveUsers, normalUsers) => {
+    const prioritisedUsers = recentlyActiveUsers.map(user => { 
+      return { name: user, priority: true }
+    })
+
+    return prioritisedUsers.concat(normalUsers);
+  }
+
   var at_config = {
     at: "@",
-    displayTpl: (item) => `<li>${displayName(item)}</li>`,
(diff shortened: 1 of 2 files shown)


jywarren:
> Blending the list on the server side would be complex and inflexible and since both APIs return different JSON objects manipulating their objects would be a lot harder which is why I chose to do it on the client side.

Great justification - awesome!

Looking now!

jywarren:
This looks super! 

> No matter how I order the arrangement of both lists, the callback in the remoteFilter function always displays the usernames that match the first strings in the username at the top. In the video, you'll see this demonstrated when I type @a.

Do you think this is a problem? And if not, do you think this is ready to merge?

jywarren:
image

The issue you mentioned about ordering is here:

image

So, are you saying there really isn't a way to intercept the ordering /after/ `remoteFilter`? If that's the case, i think it's OK. What do you think?

jywarren (approved):

jywarren:
Finally, is there a system test you'd like to add to this, which would protect the functionality? It could search for the `recently active` text, perhaps?

jywarren:
And, it would be nice to have it in the same PR, so that the test and the functionality are linked in the history!

TildaDares:
> image
> 
> The issue you mentioned about ordering is here:
> 
> image
> 
> So, are you saying there really isn't a way to intercept the ordering /after/ `remoteFilter`? If that's the case, i think it's OK. What do you think?

I'm still looking through At.js docs and issues for how to override the remoteFilter. For now, I think the merge can wait.

TildaDares:
> Finally, is there a system test you'd like to add to this, which would protect the functionality? It could search for the `recently active` text, perhaps?

I'm definitely going to write a test. I just needed to confirm I was going in the right direction.

TildaDares:
There's an open issue about the default filtering in https://github.com/ichord/At.js/issues/448 but I've not found a solution.

TildaDares:
The test I wrote is failing even though it passes locally. I'm trying to figure out what the issue is.

RuthNjeri:
Hi @TildaDares 👋🏾 , looking at this now, you can respond tomorrow.

What makes a user marked as recently active? should something be done with the users in the test case to make one active? Also, something else that could be an option is maybe adding a few seconds with before checking the assertion, with sleep or something else I have seen in the tests is a function called ` wait_for_ajax` that I think waits for the response of an API call to complete...

TildaDares:
@RuthNjeri I've tried the `wait_for_ajax` method and the `sleep` but it's not working. And a user is marked as active with the `recently active` text. Is there a way to look at the screenshots taken when running a system test? Locally, I can look at screenshots in the `tmp/screenshots` folder.

RuthNjeri:
> @RuthNjeri I've tried the `wait_for_ajax` method and the `sleep` but it's not working. And a user is marked as active with the `recently active` text. Is there a way to look at the screenshots taken when running a system test? Locally, I can look at screenshots in the `tmp/screenshots` folder.

Hi @TildaDares, I am not sure that there is a way to access the tmp/screenshot, currently I don't see any artefacts stored in the workflow

Screenshot 2021-06-29 at 15 39 25

There is a process to enable it, maybe @jywarren can have a look at it? (the process requires access to the Repo settings), https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-repository

I see you've updated the test to check for the list of names instead of the `recently active` text, that's okay too and a great alternative test. Maybe once we have access to the artefacts we can see what's happening...In the meantime, I'll have one final look and merge the PR 🚀

RuthNjeri (approved):

jywarren:
Great problem solving and yes, i'll look at the artifacts! We never had time to get those working in GitHub Actions yet. It'd be great!

jywarren:
OK, cool, i opened a PR for the screenshot saving here, let's see: https://github.com/publiclab/plots2/pull/9868

PR #9856, opened 2021-06-25
UI test for Translation Function

imajit:
Part of #9686 
Screenshot from 2021-06-26 01-23-06
I'm trying to check if `<span>` occurs in placeholder but it doesn't seem to work, when I try to search the word `Search` then the test works.
Not able to figure out what is the mistake in this  :confused:  @jywarren can you please help me out with this one ?

### DIFF
--- a/test/integration/I18n_test.rb
+++ b/test/integration/I18n_test.rb
@@ -2,18 +2,7 @@ require 'test_helper'
 require 'sanitize'
 
 class I18nTest < ActionDispatch::IntegrationTest
-
-  test 'should translate search input bar' do
-    available_testing_locales.each do |lang|
-      get '/change_locale/es'
-      follow_redirect!
-      get '/dashboard'
-      follow_redirect!
-      #assert_select '#searchform_input:match("placeholder",?)',I18n.t('layout._header.search')
-      assert_select '#searchform_input:match("placeholder",?)',/span/
-    end
-  end
-
+  
   test 'should choose i18n-locale for header partial' do
     available_testing_locales.each do |lang|
       # This test could use a rewrite. Two `get`s in succession?
--- a/test/integration/I18n_test.rb
+++ b/test/integration/I18n_test.rb
@@ -3,6 +3,17 @@ require 'sanitize'
 
 class I18nTest < ActionDispatch::IntegrationTest
 
+  test 'should translate search input bar' do
+    available_testing_locales.each do |lang|
+      get '/change_locale/es'
+      follow_redirect!
+      get '/dashboard'
+      follow_redirect!
+      #assert_select '#searchform_input:match("placeholder",?)',I18n.t('layout._header.search')
+      assert_select '#searchform_input:match("placeholder",?)',/span/
+    end
+  end
+
   test 'should choose i18n-locale for header partial' do
     available_testing_locales.each do |lang|
       # This test could use a rewrite. Two `get`s in succession?
--- a/test/system/dashboard_test.rb
+++ b/test/system/dashboard_test.rb
@@ -3,6 +3,23 @@ require "application_system_test_case"
 class DashboardTest < ApplicationSystemTestCase
   Capybara.default_max_wait_time = 60
 
+  test 'should translate search input bar' do
+    visit '/'
+    click_on 'Login'
+    fill_in("username-login", with: "admin")
+    fill_in("password-signup", with: "password")
+    click_on "Log in"
+    visit '/dashboard'
+    uid = users(:admin).id
+    visit '/profile/tags/create/'+uid.to_s+'?translationswitch=yes&name=translation-helper"'
+    available_testing_locales.each do |lang|
+      visit '/change_locale/es'
+      visit '/dashboard'
+      assert_selector(:xpath,'//*[@id="searchform_input"]')
+      #assert_selector(:xpath,'/html/body/nav/div/form/div/i')
+    end
+  end
+
   test 'viewing the dashboard' do
     visit '/'
 
--- a/test/system/dashboard_test.rb
+++ b/test/system/dashboard_test.rb
@@ -1,22 +1,22 @@
 require "application_system_test_case"
+require "i18n"
 
 class DashboardTest < ApplicationSystemTestCase
   Capybara.default_max_wait_time = 60
 
   test 'should translate search input bar' do
-    visit '/'
-    click_on 'Login'
-    fill_in("username-login", with: "admin")
-    fill_in("password-signup", with: "password")
-    click_on "Log in"
-    visit '/dashboard'
-    uid = users(:admin).id
-    visit '/profile/tags/create/'+uid.to_s+'?translationswitch=yes&name=translation-helper"'
     available_testing_locales.each do |lang|
-      visit '/change_locale/es'
+      visit '/'
+      visit '/change_locale/'+lang.to_s
+      click_on I18n.t('layout._header.login.login_title', locale: lang)
+      fill_in("username-login", with: "Bob")
+      fill_in("password-signup", with: "secretive")
+      click_on I18n.t('user_sessions.new.log_in', locale: lang)
       visit '/dashboard'
-      assert_selector(:xpath,'//*[@id="searchform_input"]')
-      #assert_selector(:xpath,'/html/body/nav/div/form/div/i')
+      uid = users(:bob).id
+      visit '/profile/tags/create/'+uid.to_s+'?translationswitch=yes&name=translation-helper'
+      assert_selector(:xpath, './/input[@id="searchform_input"][contains(@placeholder,"Search")]')
+      visit '/logout'
     end
   end
 
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -17,7 +17,7 @@
     <div class="collapse navbar-collapse" id="header-navbar-collapse">
       <form class="form-inline mr-3" id="searchform" autocomplete="off">
         <div class="input-group">
-          <input aria-label="Enter Search Query" type="text" id="searchform_input" class="form-control search-query typeahead" role="search" qryType="tags" placeholder="<%= translation('layout._header.search') %>"  value="<%= params[:query] %>" required>
+          <input aria-label="Enter Search Query" type="text" id="searchform_input" class="form-control search-query typeahead" role="search" qryType="tags" placeholder="<%= translation('layout._header.search', {}, false) %>"  value="<%= params[:query] %>" required>
           <div class="input-group-append">
             <button class="btn btn-light" type="submit" title="Search button" aria-label="Search Button"><i class="fa fa-search"></i></button>
           </div>
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -17,7 +17,7 @@
     <div class="collapse navbar-collapse" id="header-navbar-collapse">
       <form class="form-inline mr-3" id="searchform" autocomplete="off">
         <div class="input-group">
-          <input aria-label="Enter Search Query" type="text" id="searchform_input" class="form-control search-query typeahead" role="search" qryType="tags" placeholder="<%= translation('layout._header.search',{},false) %>"  value="<%= params[:query] %>" required>
+          <input aria-label="Enter Search Query" type="text" id="searchform_input" class="form-control search-query typeahead" role="search" qryType="tags" placeholder="<%= translation('layout._header.search') %>"  value="<%= params[:query] %>" required>
           <div class="input-group-append">
             <button class="btn btn-light" type="submit" title="Search button" aria-label="Search Button"><i class="fa fa-search"></i></button>
           </div>


jywarren:
Looks like the error is:

```
========== FAIL I18nTest#test_should_translate_search_input_bar (278.62s)
        Expected at least 1 element matching "#searchform_input:match("placeholder",/span/)", found 0..
        Expected 0 to be >= 1.
        test/integration/I18n_test.rb:13:in `block (2 levels) in <class:I18nTest>'
        test/integration/I18n_test.rb:7:in `each'
        test/integration/I18n_test.rb:7:in `block in <class:I18nTest>'
```

noi5e:
@imajit, I'm playing around with this right now in GitPod. I don't know very much about the areas you're working on. I'm doing a lot of catch-up to try and understand what this test is supposed to do.

It looks like `/change_locale/es` 404s in GitPod if I visit this URL. Not sure if that's how it's supposed to work:

Screen Shot 2021-07-12 at 12 54 12 PM

```
assert_select '#searchform_input:match("placeholder",?)',/span/
```

I'm not familiar with this `:match("placeholder",?)` selector, is that a CSS or Capybara thing? Can you point me to documentation for it? 

I'm guessing that this line is looking up the `placeholder` attribute on the `#searchform_input` element, and checking to see if the phrase `span` is included in it. It might be an encoding issue or something? `<span>` would appear in the raw HTML, but not as it displays on the page. 🤔 

Is there an equivalent of `take_screenshot` in an integration test? It would help to see what `#searchform_input` actually looks like on this page.

imajit:
> @imajit, I'm playing around with this right now in GitPod. I don't know very much about the areas you're working on. I'm doing a lot of catch-up to try and understand what this test is supposed to do.
> 
> It looks like `/change_locale/es` 404s in GitPod if I visit this URL. Not sure if that's how it's supposed to work:
> 
> Screen Shot 2021-07-12 at 12 54 12 PM
> 
> ```
> assert_select '#searchform_input:match("placeholder",?)',/span/
> ```
> 
> I'm not familiar with this `:match("placeholder",?)` selector, is that a CSS or Capybara thing? Can you point me to documentation for it?
> 
> I'm guessing that this line is looking up the `placeholder` attribute on the `#searchform_input` element, and checking to see if the phrase `span` is included in it. It might be an encoding issue or something? `<span>` would appear in the raw HTML, but not as it displays on the page. thinking
> 
> Is there an equivalent of `take_screenshot` in an integration test? It would help to see what `#searchform_input` actually looks like on this page.

I wrote this test to check if UI breaks for Translation team members. It is able to fetch the `Search ...` string correctly( I tried that in the commented line) but when I try to do the same for `<span>` it fails

noi5e:
> I wrote this test to check if UI breaks for Translation team members. It is able to fetch the Search ... string correctly( I tried that in the commented line) but when I try to do the same for <span> it fails

Ah, okay, so this line does work:
```
assert_select '#searchform_input:match("placeholder",?)',I18n.t('layout._header.search')
```

That means the page and element are rendering correctly, and you can search for the `placeholder` attribute.

Why don't you just `puts` or `p` to the console, and see what the output for `placeholder` is? That way you can see if it includes `span`.

jywarren:
> Is there an equivalent of take_screenshot in an integration test? It would help to see what #searchform_input actually looks like on this page.

I don't think there is. Integration tests are the closest thing to system tests but still don't run in a "real" browser. They're still simulated just by parsing the response from the application in text, rather than actually rendering it. But good thought... we could alternatively try a system test if that's going to be easier, though it "costs" a little more compute time.

jywarren:
I do recall that the parsing and matching in integration tests is notoriously fickle, one reason why it made sense to move to system tests where it really is just a browser. :-////

Tlazypanda:
Hey @imajit @jywarren @noi5e I tried this locally by hitting the `change_locale` endpoint and I can't see this span text in the textbox actually which can justify why the test is failing but  then I am not sure if the way to run this was right. Can you exactly comment on what are the steps to reproduce or check this locally?

imajit:
> Hey @imajit @jywarren @noi5e I tried this locally by hitting the `change_locale` endpoint and I can't see this span text in the textbox actually which can justify why the test is failing but then I am not sure if the way to run this was right. Can you exactly comment on what are the steps to reproduce or check this locally?

Hi! This is only visible to users having `translation-helper` user tag. The issue is not there at present, I have resolved it by adding `false` parameter in the translation function here https://github.com/publiclab/plots2/blob/3d3de3f43256c8195c575ec07060d3b783cd3c62/app/views/layouts/_header.html.erb#L20 . I wanted to add this test make sure the function does not break with later changes . 
To reproduce the error you need to add `translation-helper` tag and remove the `false` parameter. You can see the UI error in `/change_locale/es`

noi5e:
I put a console log into the test:

`puts assert_select '#searchform_input'`

Which has this output:

`<input aria-label="Enter Search Query" type="text" id="searchform_input" class="form-control search-query typeahead" role="search" qrytype="tags" placeholder="Search on Public Lab" value="" required>`

Looks like the `placeholder` attribute doesn't have `span` in it, which would make this test fail(?) Sorry if this confuses things, I'm still trying to figure out what the expected behavior here is. 

In [the original comment's screenshot](https://github.com/publiclab/plots2/pull/9856#issue-678151452), it looks like `<span>` is visible onscreen in the search placeholder. That's not what we want... Right?

imajit:
> I put a console log into the test:
> 
> `puts assert_select '#searchform_input'`
> 
> Which has this output:
> 
> `<input aria-label="Enter Search Query" type="text" id="searchform_input" class="form-control search-query typeahead" role="search" qrytype="tags" placeholder="Search on Public Lab" value="" required>`
> 
> Looks like the `placeholder` attribute doesn't have `span` in it, which would make this test fail(?) Sorry if this confuses things, I'm still trying to figure out what the expected behavior here is.
> 
> In [the original comment's screenshot](https://github.com/publiclab/plots2/pull/9856#issue-678151452), it looks like `<span>` is visible onscreen in the search placeholder. That's not what we want... Right?

Yes, that is true, we do not want the `span` in the placeholder. I wanted to keep this test with the logic that if we change any other function that breaks the UI this test will fail. For now, I have written the reverse logic that when UI breaks the test will pass just to see if the test actually finds the break or not. If this passes then I will just switch the condition so that it fails when the UI breaks. 

I also tried to write a system test for this based on changes in https://github.com/publiclab/plots2/pull/9878 , it works with `search` but fails with `span`, I guess tests work differently than actual browsers in rendering nested HTML tags.

imajit:
```
  test 'should translate search input bar' do
    visit '/'
    click_on 'Login'
    fill_in("username-login", with: "admin")
    fill_in("password-signup", with: "password")
    click_on "Log in"
    visit '/dashboard'
    uid = users(:admin).id
    visit '/profile/tags/create/'+uid.to_s+'?translationswitch=yes&name=translation-helper"'
    available_testing_locales.each do |lang|
      visit '/change_locale/es'
      visit '/dashboard'
      assert_selector(:xpath,'.//input[@id="searchform_input"][contains(@placeholder,"search")]')
      #assert_selector(:xpath,'.//input[@id="searchform_input"][contains(@placeholder,"span")]')
    end
  end
```

noi5e:
@imajit Nice! Does that system test work?

As to why the integration test wasn't able to find it... `<span>` is being inserted into the search form input via the DOM. Therefore, searching for it as a raw HTML attribute wouldn't find it. But I'm not 100 on that 🤷

imajit:
> @imajit Nice! Does that system test work?
> 
> As to why the integration test wasn't able to find it... `<span>` is being inserted into the search form input via the DOM. Therefore, searching for it as a raw HTML attribute wouldn't find it. But I'm not 100 on that shrug

No even this doesn't seem to work :disappointed:  , it needs changes made in https://github.com/publiclab/plots2/pull/9878 so won't work for now in local dev.

imajit:
I too think as the `span` part is inserted by the helper, the integration test won't be able to search it in the raw HTML

noi5e:
@imajit, I was thinking about this today, and I remembered there are some system tests that run JavaScript within the headless browser. Check out: https://github.com/publiclab/plots2/blob/b73594fb7d2b1a78f9f0afb074c55ac92bfd8e43/test/system/comment_test.rb#L222-L265

Basically I think you would be able to get this test to work if you did that, and used JS methods that look through the DOM for the value you want. Something like `document.getElementById("searchform_input").getAttribute("placeholder")`

test/integration/I18n_test.rb
```diff
@@ -3,6 +3,17 @@
...
 class I18nTest < ActionDispatch::IntegrationTest
 
+  test 'should translate search input bar' do
+    available_testing_locales.each do |lang|
+      get '/change_locale/es'
+      follow_redirect!
+      get '/dashboard'
+      follow_redirect!
+      #assert_select '#searchform_input:match("placeholder",?)',I18n.t('layout._header.search')
+      assert_select '#searchform_input:match("placeholder",?)',/span/
```
jywarren, line 13: now that https://github.com/publiclab/plots2/pull/9878 is merged did you want to rebase this? Just linking relevant lines here:

https://github.com/publiclab/plots2/blob/b73594fb7d2b1a78f9f0afb074c55ac92bfd8e43/config/locales/en.yml#L631

https://github.com/publiclab/plots2/blob/b73594fb7d2b1a78f9f0afb074c55ac92bfd8e43/app/views/layouts/_header.html.erb#L20

So since we don't have a spanish equivalent, you're looking for the helper now:

https://github.com/publiclab/plots2/blob/main/config/locales/es.yml

Is it inserting this via JavaScript after the page is rendered, or in Ruby before?

jywarren: Since we're passing html = false, it should show line 167, which is just the translated string if it exists, or the English if it doesn't, right? This is all in Ruby so the only JS needed is when the placeholder is built in JavaScript. That said, are we sure the Bootstrap UI scripts (which build the placeholder) run in the limited "simulation" of a browser in integration tests? If not, best move this over to a system test as you've discussed.

imajit:
Screenshot from 2021-07-25 22-16-44
Screenshot from 2021-07-25 22-46-39

In my previous commits, I was not writing the `xpath` correctly and when I saw the web browser console, I found out that the `i` tag is rendered outside so I tried to check it the `i` tag is rendered or not, when the UI breaks but somehow it still does not work. 
I'm not sure but can this be a possibility that the tests are running before the helper function renders the HTML ?

noi5e:
I think having this as a system test definitely is the way to go.

You and Jeffrey were both mentioning today in the meeting about tests not finding elements because the page isn't fully loaded. Maybe inserting more `page.find` calls is the way to go... I noticed that the system test you posted in a comment doesn't have `page.find` calls, so that could be why the test isn't finding the element. The [comment system tests](https://github.com/publiclab/plots2/blob/main/test/system/comment_test.rb) have a lot of examples of this kind of thing.

Regarding JS in system tests, I also want to point out this code which inserts a new HTML element onto the page after it's loaded:
https://github.com/publiclab/plots2/blob/f41f8e75c7674d5cad14b0a4afccb19df6b7c9af/test/application_system_test_case.rb#L9-L20

Then the system test uses it to simulate someone drag & dropping an image for upload:
https://github.com/publiclab/plots2/blob/f41f8e75c7674d5cad14b0a4afccb19df6b7c9af/test/system/comment_test.rb#L666-L671

I know this is unrelated, but it's just a for-example showing that it's possible for system tests to manipulate the DOM, and get the true and current attributes of HTML elements, like the `<span>` element if it's being inserted by a helper after page load. 

Definitely try `page.find` stuff, and I still think it's doable with JS!

noi5e:
I think having this as a system test definitely is the way to go.

I also agree that maybe the test is not finding elements because the page isn't fully loaded. Maybe inserting more `page.find` calls is the way to go... I noticed that the system test you posted in a comment doesn't have `page.find` calls, so that could be why the test isn't finding the element. The [comment system tests](https://github.com/publiclab/plots2/blob/main/test/system/comment_test.rb) have a lot of examples of this kind of thing.

Regarding JS in system tests, I also want to point out this code which inserts a new HTML element onto the page after it's loaded:
https://github.com/publiclab/plots2/blob/f41f8e75c7674d5cad14b0a4afccb19df6b7c9af/test/application_system_test_case.rb#L9-L20

Then the system test uses it to simulate someone drag & dropping an image for upload:
https://github.com/publiclab/plots2/blob/f41f8e75c7674d5cad14b0a4afccb19df6b7c9af/test/system/comment_test.rb#L666-L671

I know it's unrelated, but it's just a for-example showing that it's possible for system tests to manipulate the DOM, and get the true and current attributes of HTML elements, like the `<span>` element if it's being inserted by a helper after page load. 

Definitely try `page.find` stuff to wait for the page to fully load, and if that doesn't work I still think it's doable with JS!

jywarren:
So the key here i THINK is not that the helper won't have rendered yet, since that's all on the server side in ruby, but that the key markup is within a Bootstrap "popover" - https://getbootstrap.com/docs/4.4/components/popovers/ - which is created after Bootstrap JavaScript executes - it pulls out the `popover` attribute and then generates new HTML and inserts it into the page to create the popover. Integration tests don't run JavaScript at all, actually! So we have to do this as system test as @noi5e points out. 

After that you may still have some timing issues and can try other techniques that @noi5e suggested. But I hope simply running this in system tests will fix most of it!

imajit:
This test will fail when the translation function fails to render UI correctly.

test/system/dashboard_test.rb
```diff
@@ -1,8 +1,25 @@
...
+      visit '/'
+      visit '/change_locale/'+lang.to_s
+      click_on I18n.t('layout._header.login.login_title', locale: lang)
+      fill_in("username-login", with: "Bob")
+      fill_in("password-signup", with: "secretive")
+      click_on I18n.t('user_sessions.new.log_in', locale: lang)
+      visit '/dashboard'
+      uid = users(:bob).id
+      visit '/profile/tags/create/'+uid.to_s+'?translationswitch=yes&name=translation-helper'
+      assert_selector(:xpath, './/input[@id="searchform_input"][contains(@placeholder,"Search")]')
```
jywarren, line 18: Very nice!!!

jywarren (approved):

jywarren:
Well done, @imajit !!!

PR #9878, opened 2021-07-02
Verify Translation in Development

### LINKED ISSUE #9686: [Planning] Translation System Refinement

imajit:
Fixes #9686 
To review of FTOs easily, now the contributor can quickly change the required files and run it on Gitpod. As some of the FTOs related to translation need UI checks, in PR contributors can add the Gitpod UI screenshots.
ezgif com-gif-maker
I read about this method to show the button in **development mode only**, not sure if this is the best practice, also some numbers are attached to the link whenever the button is pressed, it doesn't affect the result though not sure how it is getting added each time.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -47,6 +47,16 @@ class HomeController < ApplicationController
     end
   end
 
+  def switch_on_translation
+    UserTag.create_if_absent(current_user.uid,'translation-helper')
+    redirect_to '/change_locale/zh-CN'
+  end
+
+def switch_off_translation
+    UserTag.remove_if_exists(current_user.uid,'translation-helper')
+    redirect_to '/change_locale/en'
+  end
+
   def dashboard_v2
     # The new dashboard displays the blog and topics list
     if current_user
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -47,16 +47,6 @@ class HomeController < ApplicationController
     end
   end
 
-  def switch_on_translation
-    UserTag.create_if_absent(current_user.uid, 'translation-helper')
-    redirect_to '/change_locale/zh-CN'
-  end
-
-  def switch_off_translation
-    UserTag.remove_if_exists(current_user.uid, 'translation-helper')
-    redirect_to '/change_locale/en'
-  end
-
   def dashboard_v2
     # The new dashboard displays the blog and topics list
     if current_user
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -48,12 +48,12 @@ class HomeController < ApplicationController
   end
 
   def switch_on_translation
-    UserTag.create_if_absent(current_user.uid,'translation-helper')
+    UserTag.create_if_absent(current_user.uid, 'translation-helper')
     redirect_to '/change_locale/zh-CN'
   end
 
-def switch_off_translation
-    UserTag.remove_if_exists(current_user.uid,'translation-helper')
+  def switch_off_translation
+    UserTag.remove_if_exists(current_user.uid, 'translation-helper')
     redirect_to '/change_locale/en'
   end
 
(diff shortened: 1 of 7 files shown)


noi5e (approved):

noi5e:
@imajit Nice work!

imajit:
The tests commit fails , is this because the code the tests are checking is not there in the code-base yet ?

app/controllers/home_controller.rb
```diff
@@ -47,6 +47,16 @@ def dashboard
     end
   end
 
+  def switch_on_translation
```
jywarren, line 50: maybe we should name this
```suggestion
  def switch_on_translation_helper
```

Is this only for development mode? If so, couldn't we use an existing route somehow -- https://github.com/publiclab/plots2/blob/main/test/functional/user_tags_controller_test.rb#L8-L15 maybe? To add the tag? though i guess it's 2 steps.

jywarren:
> The tests commit fails , is this because the code the tests are checking is not there in the code-base yet ?

If the code is not there yet, then yes! But you can then add the code and the test would then begin to pass. That would be "Test Driven Development" - a nice pattern!

config/routes.rb
```diff
@@ -378,6 +378,8 @@
   post 'comment/react/delete/:id' => 'comment#react_delete'
   post 'comment/react/update/:id' => 'comment#react_update'
 
+  get 'switch_on_translation' => 'home#switch_on_translation'
```
jywarren, line 381: Yes, if possible if we could use existing routes, we'd be able to skip some of the tests because we wouldn't be introducing a whole new route, you know? That would keep codebase complexity down a bit, which is really helpful in the long run on a big project like this.

imajit:
ezgif com-gif-maker(1)
I initially kept this button in the welcome message I could not see it anymore so I just placed it in the top part. I added an extra parameter to the existing route and changed the controller so now no new routes or controllers are needed. Thanks @jywarren for the suggestion it was a bit more complex but yes it keeps the codebase simple now  :smile: . I'm not sure why codecov is failing though for codeclimate I guess I'll add a commit later.
@noi5e @jywarren should I move the button to some other place ?

jywarren:
This is looking great! As to another place, it's only in development mode, but maybe it'd confuse people? What if we put it in the navbar menu? And don't worry about CodeCov, i'm working on that in https://github.com/publiclab/plots2/pull/9905

Thanks!

imajit:
The top Nav-bar right ? the one with the search bar

Tlazypanda:
Ohh @imajit can't say for sure but I think yes @jywarren meant the top nav bar :sweat_smile: looping him here in case I missed it

imajit:
This is how it looks in top-nav 
ezgif com-gif-maker(2)

jywarren:
Would it be possible to put it in the same dropdown as under the profile
picture (the small grey circle to the right side)? Thank you!

imajit:
ezgif com-gif-maker(3)

noi5e:
Looks good to me! @jywarren, the `Switch on Translation` button is now in the nav, so I believe this one is merge-ready

jywarren:
Excellent, thank you so much!!!

Issue #9873, opened 2021-06-29
Periodic test failures in GitHub Actions CI


jywarren:
I'm seeing occasional errors that seem to self-resolve when we re-run the tests. I'm not sure the cause. Here's one log I found:

```
2021-06-29 20:46:08 +0000 Rack app ("GET /barnstar/givestar=basic&nid=1" - (127.0.0.1)): Mysql2::Error
=[Screenshot]: tmp/screenshots/failures_test_awarding_barnstar_functions_correctly.png
ERROR PostTest#test_awarding_barnstar_functions_correctly (873.23s)
Minitest::UnexpectedError:         Mysql2::Error: 
            app/models/tag.rb:430:in `update_tags_activity'
            app/models/node.rb:678:in `tag_activity'
            app/models/node.rb:671:in `add_comment'
            app/controllers/tag_controller.rb:284:in `barnstar'
```

These can sometimes happen due to timing issues -- because tests run all at once with no time passing in between. We can sometimes resolve them by adding an order clause to the SQL, or by using TimeCop to manipulate the time during tests. Let's collect more data to see what tests are inconsistent (paste them here before restarting them, please) and go from there. cc @publiclab/soc

jywarren:
Confirmed the exact same bug blocked https://github.com/publiclab/plots2/pull/9848 - and is blocking lots of dependabot PRs too, so that's a good place to look for this happening!

jywarren:
https://github.com/publiclab/plots2/pull/9864/ as well! So, what is the actual error? 

It seems related to this line from https://github.com/publiclab/plots2/pull/9165 by @RuthNjeri --

https://github.com/publiclab/plots2/blob/78bf29f75865cb6d0dd5e5f55e3ff24ea707a65a/app/models/tag.rb#L430

No worries, Ruth, it seems super obscure. But we're doing an `update_all` -- could it be blocked by some other database lock? We are running tests in parallel, which kind of simulates how this could run in production with many users at once, you know? 

Could we do this differently?

daemon1024:
I wonder if introducing pessimistic locking will help us prevent this race condition. 

https://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html

jywarren:
Oh, interesting! So how would that look? Like this?

```ruby
Tag.transaction do
 Tag.lock
  .where(tid: tids)
  .update_all(activity_timestamp: DateTime.now, latest_activity_nid: activity_id) 
end
```

[...]

PR #9879, opened 2021-07-02
Order tags based on best match to search query

### LINKED ISSUE #9623: Existing tags not showing up as an exact match in search bar autosuggestions

Not quite a bug, but more of an observation that I didn't realize was limiting the results I'd see when using the search bar: when previously searching for mapping content, I entered `mapping` into the search bar, and that tag didn't appear in the autosuggested list of tags (see screenshot). But today I saw the `mapping` tag on a post and clicked on it, and saw from the tag page that there's quite a lot of content there tagged with `mapping`! 

mapping results

I expected to see `mapping` appear in the autosuggested list of tags as an exact match when entering that term into the search bar, given that the `mapping` tag page exists and there's lots of content attached to that tag.

Existing content with the `mapping` tag: https://publiclab.org/tag/mapping

bhamster

17sushmita:
Fixes #9623

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -78,19 +78,16 @@ class SearchService
   # chained to the notes that are tagged with those values
   def search_tags(query, limit = 10)
     suggestions = []
+    # order tag autosuggestions by placing exact match of the tag on the top, then tags with query and some other text as suffix, 
+    # then tags with some text as prefix or suffix and lastly the tags with some text as prefix
+    tags_order = 'CASE WHEN name LIKE ? OR name LIKE ? THEN 1 WHEN name LIKE ? OR name LIKE ? THEN 2 WHEN name LIKE ? OR name LIKE ? THEN 4 ELSE 3 END', "#{query}", "#{query.to_s.gsub(' ', '-')}", "#{query}%", "#{query.to_s.gsub(' ', '-')}%", "%#{query}", "%#{query.to_s.gsub(' ', '-')}"
     # filtering out tag spam by requiring tags attached to a published node
     # also, we search for both "balloon mapping" and "balloon-mapping":
     Tag.where('name LIKE ? OR name LIKE ?', "%#{query}%", "%#{query.to_s.gsub(' ', '-')}%")
       .includes(:node)
       .references(:node)
       .where('node.status = 1')
-      .order('
-        CASE
-          WHEN name LIKE "' + query + '" THEN 1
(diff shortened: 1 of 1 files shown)


app/services/search_service.rb
```diff
@@ -84,6 +84,13 @@ def search_tags(query, limit = 10)
       .includes(:node)
       .references(:node)
       .where('node.status = 1')
+      .order('
+        CASE
```
namangupta01, line 88: Hey!
Just a quick suggestion can we store the `case` string in a variable and then just use that inside order?
For example: tags_order = "CASE WHEN name like #{query} THEN 1 WHEN name LIKE #{query}% THEN 2 WHEN name LIKE %{query} THEN 4 ELSE 3"
and then just use `order(tags_order)`.

Also, should we include one more case: `WHEN name like %#{query}% `

namangupta01 (changes requested):
Just a couple of suggestions. Other than this everything looks great. 👍

app/services/search_service.rb
17sushmita: @namangupta01, the first suggestion seems to be a good idea to store the string as  a variable and using it. I'm adding a commit with the changes. 
As per `WHEN name like %#{query}%` case is concerned, I think it will show up in the ELSE condition as it is the only condition left for else ig. What do you think?

jywarren:
Trying to resolve in https://github.com/publiclab/plots2/issues/9873

jywarren:
Sorry this is so annoying! I actually just merged #9881 so if you'd like to try rebasing over it it might help!

cesswairimu (approved):
This is looking good :tada: :tada: 

We can ignore code climate warnings...let me know if we are good to merge...Thanks @17sushmita  :tada: :tada:

app/services/search_service.rb
cesswairimu: Hi @namangupta01, kindly give this another look when you get some time. Thanks and nice to you see around!

17sushmita:
> This is looking good  
> 
> We can ignore code climate warnings...let me know if we are good to merge...Thanks @17sushmita

Yes, we are good to merge @cesswairimu. Thanks 😄

jywarren:
Looks great! I resolved the CodeClimate and for some reason the CodeCov error just went away instantly. 🤷

ebarry:
this is so great!

bhamster07:
Hooray, thanks so much everyone!! 😀

PR #9890, opened 2021-07-08
Corrected Greek YAML file langauge code 

imajit:
Part of #9686 
To sync language files with transifex, the language code in YAML files should same as they are in https://www.transifex.com/explore/languages/ . This file has Greek translations but was labeled Italian (it) in the YAML file.
w1

### DIFF
--- a/config/locales/gr.yml
+++ b/config/locales/gr.yml
@@ -1,5 +1,5 @@
 ---
-it:
+el:
   activerecord:
     errors:
       messages:


Tlazypanda (approved):
@imajit nice catch :100: can we maybe modify the commit message to something more specific so that the commit history looks more defined? Thanks :v:

RuthNjeri (approved):
This looks great, thanks @imajit and @Tlazypanda 🎉

PR #9893, opened 2021-07-08
created drafts page

### LINKED ISSUE #9798: [Discussion] Refine design of co-authored posts on profile pages
Presently coauthored posts are displayed separately from normal posts as seen here https://publiclab.org/profile/warren. I suggested changing the query that gets a user's posts (notes, questions and wikis) to also include co-authored posts.

```
def find_posts_by_type(type = 'note')
  coauthored_tag = "with:" + name.downcase
  
  Node.where(type: type).where('node.uid = ? OR term_data.name = ? OR term_data.parent = ?', uid, coauthored_tag.to_s, 
  coauthored_tag.to_s).order('created DESC')
end

```
What does the community think?

@jywarren @RuthNjeri @cesswairimu

TildaDares:
Fixes #9798
Part of larger planning issue in #9667

Screenshot 2021-07-09 at 00 14 56

## Responsiveness

https://user-images.githubusercontent.com/63427719/125001931-b17f2180-e04b-11eb-9f22-936c1d4f8b28.mov

When a user who is neither a moderator or the drafts author tries to visit the drafts page, a flash notice is displayed and the user is redirected to the homepage

Screenshot 2021-07-09 at 00 23 25

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -427,6 +427,17 @@ class NotesController < ApplicationController
     end
   end
 
+  def drafts
+    @user = User.find_by(name: params[:id])
+    if current_user&.can_moderate? || current_user == @user
+      @pagy, @drafts = pagy(@user.drafts, items: 24)
+      render template: 'notes/drafts'
+    else
+      flash[:warning] = "This page is only visible to the author and moderators."
+      redirect_to '/'
+    end
+  end
+
   private
 
   def set_node
--- a/app/views/notes/_nav_tabs.html.erb
+++ b/app/views/notes/_nav_tabs.html.erb
@@ -1,4 +1,4 @@
-<div class="node-types" style="margin-bottom:8px;">
+<div class="node-types mr-3" style="margin-bottom:8px;">
   <ul class="nav nav-tabs" aria-labelledby="dropdownMenuButton">
 
     <li class="nav-item">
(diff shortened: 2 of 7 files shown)


Tlazypanda (commented):
@TildaDares wow this looks great! 🎉  The responsive design and all the tests 💯 

Maybe we can have the route under notes only like `notes/drafts` but this is just my suggestion @jywarren can comment better

RuthNjeri (approved):
This looks great, good job @TildaDares! 🎉 

Aside from @Tlazypanda's suggestion which makes sense...I think the PR is good to go 🚀

noi5e:
@TildaDares Looks good to me! Great work!

noi5e (approved):

noi5e:
This is unrelated and I'm sure it doesn't affect this PR at all, but I'm just noting that codecov has a new feature displaying on this PR which I can't make sense of:

Screen Shot 2021-07-12 at 10 51 58 AM

TildaDares:
@Tlazypanda @RuthNjeri You're right about using the `notes/drafts` route but using that route would mean that moderators wouldn't be able to see other users drafts page. I'm fine with other option though.

jywarren:
> noting that codecov has a new feature

whoa... codecov had been a bit broken. Maybe they introduced a new system upstream to fix that but it came with some additional features or settings defaults?

jywarren:
Looks awesome. Thanks @TildaDares !!!!!

PR #9909, opened 2021-07-16
add wait_for_ajax for barnstar system test

jywarren:
Re: #9873


jywarren:
I think we figured this out in https://github.com/publiclab/plots2/pull/9917

PR #9914, opened 2021-07-16
Make gh actions labeling conditional on event type

jywarren:
re #9873


jywarren:
Seems unnecessary now due to #9915 and #9916!!

PR #9932, opened 2021-07-21
Plurals Correction in YAML files

imajit:
Part of  #9686 

https://community.transifex.com/t/plurals-rules-error-on-uploading-translation-file-to-transifex-project/2370/2 
I have tried to fix the plurals error in the source file based on the Transifex community response.


jywarren:
Look ok here at least - should we merge? Thanks!

imajit:
yes, I'll check `Sync with github` from Project integration and see if it works after this

PR #9940, opened 2021-07-22
Reduced response time in nearby People search API

### LINKED ISSUE #7556: Research why loading nearbyPeople Srch API is so slow to load
In our maps we are now loading data from `https://publiclab.org/api/srch/nearbyPeople?nwlat=10.0&selat=0.0&nwlng=0.0&selng=10.0`
This code is located here: https://github.com/publiclab/plots2/blob/master/app/api/srch/search.rb#L333

Accessing this API takes a lot longer than it should. For example, it takes 19.43s to load:

FireShot Capture 304 - API Tester - apitester com

Meanwhile another API we use in our map layers has far more results in only 83ms!

FireShot Capture 305 - API Tester - apitester com

Some of our other Search API requests perform much better than nearbyPeople. Why is that particular one so slow? In order for our maps to be useful we can't have such a long delay in loading every time we refresh or move the map.

17sushmita:
Fixes #7556 

* [x] code is in a uniquely-named feature branch and has no merge conflicts 📁

### Before - 
GET: https://publiclab.org/api/srch/nearbyPeople?nwlat=10.0&selat=0.0&nwlng=0.0&selng=10.0
Time: 39.36
image
 
### After - 
GET: https://unstable.publiclab.org/api/srch/nearbyPeople?nwlat=10.0&selat=0.0&nwlng=0.0&selng=10.0
Time : 2.12s
image

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -204,9 +204,7 @@ class SearchService
     items = user_locations
 
     # selects the items whose node_tags don't have the location:blurred tag
-    
     items = items.where('user_tags.value <> "location:blurred"')
-
 
     # Here we use period["from"] and period["to"] in the query only if they have been specified,
     # so we avoid to join revision table
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -204,11 +204,9 @@ class SearchService
     items = user_locations
 
     # selects the items whose node_tags don't have the location:blurred tag
-    items.select do |item|
-      item.user_tags.none? do |user_tag|
-        user_tag.name == "location:blurred"
-      end
-    end
+    
+    items = items.where('user_tags.value <> "location:blurred"')
+
 
     # Here we use period["from"] and period["to"] in the query only if they have been specified,
     # so we avoid to join revision table


daemon1024:
Wow 👀 So we don't need to deal with n+1 problem at all?

17sushmita:
> Wow  So we don't need to deal with n+1 problem at all?

Ah! I didn't see the discussion on the main issue. Let's think of improving the performance furthermore. Currently looping over every node by the user to check if it has `location: blurred` was really slow so this can be simply improved by using a rails query.

daemon1024:
Yeah we need to analyse further but in my opinion maybe we should proceed with merging this, since this is a significant improvement ( Significant as in Huge :O ) Can work on others in subsequent PRs as we proceed with discussion at #7556

app/services/search_service.rb
```diff
@@ -204,11 +204,7 @@ def tagNearbyPeople(coordinates, tag, field, period = nil, sort_by = nil, order_
     items = user_locations
 
     # selects the items whose node_tags don't have the location:blurred tag
-    items.select do |item|
-      item.user_tags.none? do |user_tag|
-        user_tag.name == "location:blurred"
-      end
-    end
+    items = items.where('user_tags.value <> "location:blurred"')
```
jywarren, line 207: awesomeeeee

jywarren (approved):

jywarren:
Very excited for this!

PR #9918, opened 2021-07-18
moved tocbot require statement to javascripts/application.js

TildaDares:
The wiki menu is not displaying on the main site and this error is being returned.
`Uncaught ReferenceError: tocbot is not defined`

Changing the location of the require statement for the tocbot library should fix this error. I think it's also important to note that this feature works perfectly on local and stable envs.

PR #9959, opened 2021-07-27
change webpacker.yml to see public/lib

jywarren:
as per https://github.com/publiclab/plots2/blob/a65efa664aadf756601a39b4a96f973fa099ee16/config/application.rb#L79

Fixes #9918


jywarren:
Looks like this worked, we think! But, it's not certain. It's not clear if this works after the 2nd time anyways, or if our change here was the solution. Let's merge it, note that this may recur, and leave the issue open. We'll know when we deploy again soon.

PR #9980, opened 2021-08-03
Refactor legacy gallery code into node revision body

### LINKED ISSUE #4074: Refactor legacy gallery code into node revision body
We have a very old leftover feature which shows a collection of images in a "gallery" at the top of a note:

https://github.com/publiclab/plots2/blob/5c0d96274bd0be15feba69d8f30093082402ace5/app/views/notes/show.html.erb#L68

Full code for the display is here; https://github.com/publiclab/plots2/blob/5c0d96274bd0be15feba69d8f30093082402ace5/app/views/notes/_gallery.html.erb

It's so simple, we could just insert the images at the top of the node body (actually the node revision body) and have it be part of the content. 

We'd use this code to display each image, just like in our gallery:

```
  <a target="_blank" href="<%= image.image.path(:original) %>"><img rel="tooltip" data-title="<%= image.description %>" style="margin-bottom:4px;" class="img-rounded" src="<%= image.image.path(:thumb) %>" /></a>
```

1. collect all nodes with image galleries, so, 
2. in [a migration](https://github.com/publiclab/plots2/blob/master/db/migrate/20180707065151_change_status_of_comments.rb) -- insert the above code (but with proper substitutions not in IRB style) once per image at the top of the node `revision.body`
3. we'd delete this model: https://github.com/publiclab/plots2/blob/5c0d96274bd0be15feba69d8f30093082402ace5/app/models/drupal_content_field_image_gallery.rb
4. drop the corresponding database table in the same migration as in step 2
5. remove the "gallery" code from node.rb

This is not a super complex change, but has a lot of moving parts. We'd love help with it, and are happy to support! Thanks!

Relates to #956 as well!

TildaDares:
Fixes #4074


TildaDares:
@jywarren A lot of tests are failing because I removed the gallery code. I think I have to remove this too
https://github.com/publiclab/plots2/blob/5c0d96274bd0be15feba69d8f30093082402ace5/app/views/notes/show.html.erb#L68

And also delete the `gallery` partial in https://github.com/publiclab/plots2/blob/5c0d96274bd0be15feba69d8f30093082402ace5/app/views/notes/_gallery.html.erb#L2-L4

jywarren:
```
=ERROR NotesControllerTest#test_first-timer_moderated_note_(status=4)_shown_to_moderator_with_notice_and_approval_prompt_in_full_view (100.13s)
Minitest::UnexpectedError:         ActionView::Template::Error: undefined method `gallery' for #<Node:0x000055ef31b76ad0>
            app/views/notes/_gallery.html.erb:2:in `_app_views_notes__gallery_html_erb___2709518038458471136_47242931054120'
```

db/migrate/20210803085827_add_images_to_revision_body.rb
```diff
@@ -0,0 +1,10 @@
+class AddImagesToRevisionBody < ActiveRecord::Migration[5.2]
+  def change
```
jywarren, line 2: what about moving it here?

```ruby
  def gallery(node)
    if !node.drupal_content_field_image_gallery.empty? && node.drupal_content_field_image_gallery.first.field_image_gallery_fid
      node.drupal_content_field_image_gallery
    else
      []
    end
  end
```

PR #9995, opened 2021-08-05
sort published drafts by last updated

### LINKED ISSUE #9985: Refine chronological display of published notes that were previously saved as drafts

Returning to a comment from @jywarren from https://github.com/publiclab/plots2/issues/2667#issuecomment-389886458 that was not taken up at that time. It's about updating the date when publishing posts that were drafted, and saved as drafts, on an earlier date: 

> Second, we set a final url with the publish date in it. I wonder if we should reset the url at the moment of final publication, and before then we can use a temporary one, resetting this on publication? This is a little complex but seems workable, no? There's a generate_path method to recalculate the path with the current title and date. That way the title could be changed as well and this would reflect in the final URL.

This would be very helpful as currently if someone publishes a draft even from a week ago, there's too much newer activity for it to even show up on the first page of latest results. 

cc @steviepubliclab

TildaDares:
Fixes #9985

### DIFF
--- a/test/functional/notes_controller_test.rb
+++ b/test/functional/notes_controller_test.rb
@@ -865,6 +865,29 @@ class NotesControllerTest < ActionController::TestCase
     end
    end
 
+   test "drafts are sorted by last updated" do
+    UserSession.create(users(:jeff))
+    saved, node, revision = Node.new_note( title: 'Sort Drafts', draft: 'true', uid: 2, body: 'I have sorted drafts' )
+    revision_timestamp = revision.timestamp
+
+    assert_equal 3, node.status
+    assert_equal DateTime.now.to_i, revision_timestamp
+
+    updated_timestamp = nil
+
+    Timecop.freeze(Date.today + 1.day) do
+      get :publish_draft, params: { id: node.id }
+
+      updated_timestamp = DateTime.now.to_i
+      assert_response :redirect
+      node = assigns(:node)
+      assert_equal 1, node.status
+    end
+    
+    assert_equal updated_timestamp, node.latest['timestamp']
+    assert_redirected_to '/notes/' + users(:jeff).username + '/' + (Date.today + 1.day).strftime('%m-%d-%Y') + '/' + node.title.parameterize
+   end
+
    test 'co-author can publish the draft' do
     perform_enqueued_jobs do
        UserSession.create(users(:test_user))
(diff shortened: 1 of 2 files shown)


TildaDares:
@jywarren None of the tests are failing

noi5e:
Nice @TildaDares! 

Is there a way we can get this tested in some way before merging? Screenshots, or even with a scrubbed database? That seems important for the original issue at #9985.

ebarry:
Thank you for writing this @TildaDares !!!!

jywarren:
We discussed using Timecop on the weekly call, i think that could be a nice test -- maybe a functional test? Or integration test?

jywarren:
This is awesome though, thank you!!!

TildaDares:
> We discussed using Timecop on the weekly call, i think that could be a nice test -- maybe a functional test? Or integration test?

@jywarren Is it how notes on the activity feed are sorted I'm supposed to test? Currently, I'm testing this by changing the timestamp of an old note to a time far in the future so that it will appear as the first item in the activity feed but this line sorts the items in the activity feed by newly created.
https://github.com/publiclab/plots2/blob/cff3581e627f852a6a1ab1dfbd2ea54a01a9631e/app/controllers/home_controller.rb#L155

jywarren:
So as to the test, it looks like we actually have one here! 

https://github.com/publiclab/plots2/blob/cff3581e627f852a6a1ab1dfbd2ea54a01a9631e/test/functional/notes_controller_test.rb#L840-L866

But are we not testing enough? We want to:

1. create a draft
2. confirm the draft has current timestamp
3. move time forward more than 1 day
4. publish the draft
5. confirm draft has new (current day + one day) timestamp
6. confirm draft has new URL too

It does look like that line 155 is sorting a second time by created_at, but for nodes, created_at is just a function which formats the timestamp, i believe? see def created_at in node.rb to confirm.

Hope this helps!

test/functional/notes_controller_test.rb
```diff
@@ -865,6 +865,26 @@ def test_get_rss_feed
...
+
+    Timecop.freeze(Date.today + 1.day) do
+      get :publish_draft, params: { id: node.id }
+
+      assert_response :redirect
+      node = assigns(:node)
+      assert_equal 1, node.status
+    end
+    
+    assert_not_equal revision_timestamp, node.latest['timestamp']
```
jywarren, line 884: Here could we assert what it should be equal to? Also, we could assert the wrong (original) timestamp before we publish the draft, to show that it was different before vs after. 

This looks great! Thanks Tilda!!!

jywarren (changes requested):

test/functional/notes_controller_test.rb
TildaDares: @jywarren I've made the changes

jywarren:
This is fantastic. Great job!!!!

Issue #9983, opened 2021-08-03
Speed up GitHub Actions test runs/checks by caching dependencies


jywarren:
I suspect we may not be caching properly and wanted to suggest we try this out:

https://dev.to/github/caching-dependencies-to-speed-up-workflows-in-github-actions-3efl

We have gem caching supposedly installed here, so let's confirm it's working:

https://github.com/publiclab/plots2/blob/a578843b8127c1bf96c5bc023b673a652b0cd074/.github/workflows/tests.yml#L53-L59

Likewise we should be able to cache javascripts from the `yarn install` line. Let's see what we can do to improve our load times!

Currently here is where we stand:

![image](https://user-images.githubusercontent.com/24359/128056426-1559eab2-5fa3-4b1e-a101-1874fb70d0b4.png)

You can read through our logs here:

https://github.com/publiclab/plots2/actions/runs/1090845487

noi5e:
Reading up on this now. Here's what [GitHub Actions's docs](https://github.com/actions/cache/blob/main/examples.md#ruby---bundler) say about caching Ruby gems:

> Caching gems with Bundler correctly is not trivial and just using actions/cache is not enough.

> Instead, it is recommended to use ruby/setup-ruby's bundler-cache: true option whenever possible:

```
- uses: ruby/setup-ruby@v1
  with:
    ruby-version: ...
    bundler-cache: true
```

That seems to be what we're doing here:
https://github.com/publiclab/plots2/blob/5482fc868ac96a9b0622b50aaaee7b3003033b72/.github/workflows/tests.yml#L51-L59

🤔

noi5e:
What are the indications that Ruby gems are NOT being cached correctly?

Here's the [unit test run](https://github.com/publiclab/plots2/runs/3222445809?check_suite_focus=true) from the most recent logs @jywarren posted:
```
bundle install
  /opt/hostedtoolcache/Ruby/2.6.6/x64/bin/bundle config --local path /home/runner/work/plots2/plots2/vendor/bundle
  /opt/hostedtoolcache/Ruby/2.6.6/x64/bin/bundle config --local deployment true
  Cache key: setup-ruby-bundler-cache-v3-ubuntu-20.04-ruby-2.6.6-Gemfile.lock-c3c1488c69b07d2f3f5aff65476c89408928cde98964b01410ba5c94aa79b8fa
  Received 134217728 of 252020325 (53.3%), 128.0 MBs/sec
[...]

PR #9998, opened 2021-08-07
Cache Yarn Dependencies in GitHub Actions

noi5e:
See #9983 particularly [this comment](https://github.com/publiclab/plots2/issues/9983#issuecomment-894716371).

Work in progress, first I'm going to try to echo the yarn cache directory.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -55,6 +55,9 @@ jobs:
         ruby-version: 2.6.6 # Not needed with a .ruby-version file
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically (not working?)
     - uses: ./.github/actions/setup-test-environment
+    - name: Get yarn cache directory path
+      id: yarn-cache-dir-path
+      run: echo "::set-output name=dir::$(yarn cache dir)"
     - name: Install JavaScript dependencies with Yarn
       run: yarn check || yarn install --frozen-lockfile;
     - name: Run webpacker setup
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,8 +58,13 @@ jobs:
     - name: Get yarn cache directory path
       id: yarn-cache-dir-path
       run: echo "::set-output name=dir::$(yarn cache dir)"
-    - name: Install JavaScript dependencies with Yarn
-      run: yarn check || yarn install --frozen-lockfile;
+    - uses: actions/cache@v2
+      id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+      with:
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+        key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-yarn-
     - name: Run webpacker setup
       env:
         RAILS_ENV: test
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -61,8 +61,8 @@ jobs:
       id: yarn-cache-dir-path
       outputs:
         dir:
-        description: "yarn cache directory"
-        value: ${{ steps.random-number-generator.outputs.dir }}
+          description: "yarn cache directory"
+          value: ${{ steps.random-number-generator.outputs.dir }}
       run: echo "::set-output name=dir::$(echo yarn cache dir)"
     - uses: actions/cache@v2
       id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -59,7 +59,11 @@ jobs:
       run: yarn check || yarn install --frozen-lockfile;
     - name: Get yarn cache directory path
       id: yarn-cache-dir-path
-      run: yarn cache dir
+      outputs:
+        dir:
+        description: "yarn cache directory"
+        value: ${{ steps.random-number-generator.outputs.dir }}
+      run: echo "::set-output name=dir::$(echo yarn cache dir)"
     - uses: actions/cache@v2
       id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
       with:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -55,22 +55,9 @@ jobs:
         ruby-version: 2.6.6 # Not needed with a .ruby-version file
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically (not working?)
     - uses: ./.github/actions/setup-test-environment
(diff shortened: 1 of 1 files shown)


noi5e:
Okay, that made this happen:
Screen Shot 2021-08-07 at 4 21 16 PM

I was expecting the directory to be printed in the run.

Not really sure what I'm doing, maybe I should try failing harder. I'll try implementing this part from [the docs](https://github.com/actions/cache/blob/main/examples.md#node---yarn):

```
- uses: actions/cache@v2
  id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
  with:
    path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
    key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
    restore-keys: |
      ${{ runner.os }}-yarn-
```

noi5e:
Functional tests failed after my last changes:
Screen Shot 2021-08-07 at 4 34 47 PM

Here's the error:
Screen Shot 2021-08-07 at 4 34 12 PM

`Cache not found for input keys`. 🤔 

- There's a bit `key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}` that I added. Maybe this needs to be a more specific key?
- I also removed the `yarn install` part, maybe I'm not supposed to do that?

noi5e:
@jywarren Whenever it makes sense, can you take a look at this? 😄

jywarren:
Thanks so much for looking into this! As you can tell i'm a little behind on reviews, will catch up here ASAP!

noi5e:
Latest commit: trying to run `yarn cache dir` according to [these docs](https://classic.yarnpkg.com/en/docs/cli/cache/#toc-yarn-cache-dir) instead of the other command.

.github/workflows/tests.yml
```diff
@@ -57,6 +57,16 @@ jobs:
     - uses: ./.github/actions/setup-test-environment
     - name: Install JavaScript dependencies with Yarn
       run: yarn check || yarn install --frozen-lockfile;
+    - name: Get yarn cache directory path
+      id: yarn-cache-dir-path
+      run: yarn cache dir
+    - uses: actions/cache@v2
+      id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+      with:
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
```
jywarren, line 66: Huh, seeing an error here:
 
```
Run actions/cache@v2
  with:
    key: Linux-yarn-08fd7bab35005d3e8e92ed89e54f7a999ee6535e35e43d54deb4a9a5efd3bf37
    restore-keys: Linux-yarn-
  
Error: Input required and not supplied: path
```

jywarren: I'm a little stumped, to be honest, sorry! 

But, your `yarn cache dir` seems to have output something useful:

```
Run yarn cache dir
  yarn cache dir
  shell: /usr/bin/bash -e {0}
/home/runner/.cache/yarn/v6
```

jywarren: Ah so do we need to define an output of name `dir` for `yarn-cache-dir-path`?

jywarren: aha -- https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputs-for-composite-actions

.github/workflows/tests.yml
```diff
@@ -57,6 +57,20 @@ jobs:
     - uses: ./.github/actions/setup-test-environment
     - name: Install JavaScript dependencies with Yarn
       run: yarn check || yarn install --frozen-lockfile;
+    - name: Get yarn cache directory path
+      id: yarn-cache-dir-path
+      outputs:
```
jywarren, line 62: Tried this!

jywarren: Huh... an hour later tests haven't started...?

jywarren:
Is something here blocking GitHub action runs? Hmm...

jywarren:
Aha - under Checks, i see this error:

```
Invalid workflow file : .github/workflows/tests.yml#L62
The workflow is not valid. .github/workflows/tests.yml (Line: 62, Col: 7): Unexpected value 'outputs'
```

jywarren:
Could we just use this? https://github.com/marketplace/actions/yarn-node-install-cache

jywarren:
Well, that ran! I see:

```

/usr/bin/uname
Linux
/usr/local/bin/yarn cache dir
/home/runner/.cache/yarn/v6
Cache keys: Linux-yarn-node-cache-2fee04878e10de47453d7a3168291c15, Linux-yarn-node-cache-
Cache not found
```

but, the first time, the cache would be generated, right? I'm not sure the cache persistence rules. I'll try running this again to see if it finds the cache from this last run?

jywarren:
Unfortunately it doesn't save the previous run if we don't make a new commit (or maybe i just don't know how): https://github.com/publiclab/plots2/actions/runs/1195643414

jywarren:
Wait! Yay! It worked!!!

```
Run egordm/gha-yarn-node-cache@v1
/usr/bin/uname
Linux
/usr/local/bin/yarn cache dir
/home/runner/.cache/yarn/v6
Cache keys: Linux-yarn-node-cache-2fee04878e10de47453d7a3168291c15, Linux-yarn-node-cache-
Cache Size: ~242 MB (253387137 B)
/usr/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/9a3235da-b161-4a2e-a4fd-a1baac79bd2e/cache.tzst -P -C /home/runner/work/plots2/plots2
Cache restored from key: Linux-yarn-node-cache-2fee04878e10de47453d7a3168291c15
```

Wow i almost didn't expect that to work. 

Does this mean we should be OK now? 

Hmm. When i look at the next step, "Install JavaScript dependencies with Yarn" i see that the latest run took only 21s vs. the first commit in this PR taking 46s. So that seems good although maybe not as good as I'd have guessed. 

But, when I look closer, I see that both end with similar times:

* the first commit cites 14.94s - https://github.com/publiclab/plots2/runs/3271177757#step:7:2243
* the latest cites 16.20s to completion - https://github.com/publiclab/plots2/pull/9998/checks?check_run_id=3500056847#step:7:91

However, the first commit has about 2000 lines of errors in the `yarn check` output, while the latest has no errors. So `yarn check` accounts for the discrepancy actually, not `yarn install --frozen-lockfile`!

The command for install is, for reference: `yarn check || yarn install --frozen-lockfile;`

jywarren:
Overall here are the times: 

First commit: 

image

Latest commit:

image

I'm re-running now for comparison.

jywarren:
Hmm. I mean, ~22 seconds for yarn install consistently. But, that's not really making a dent in our overall run time: 

image

jywarren:
I mean, i guess that step is just not that lengthy anyways. But nice to have the cache running if I read this right. @noi5e what do you think?

jywarren:
I guess let's copy this into the other steps too. 

Overall our startup and setup aren't too bad: 

image

because we're parallelized, shaving 20s didn't have a huge effect.

PR #10005, opened 2021-08-09
Updated helper function

imajit:
Part of #9686 
The previous update was failing in some nested string translation cases.
Screenshot from 2021-08-10 03-10-04


Tlazypanda (commented):
Hey @imajit this looks good :tada: can we possibly add some tests in this or in a follow up pr?

Issue #9946, opened 2021-07-25
Optimise `tagLocations`


daemon1024:
When we first open up the map endpoint, first and default markers that load on maps are tag locations.

But when we start interacting with the map, it starts sending loads of requests to the API.

https://user-images.githubusercontent.com/47106543/126897921-6025596f-11ea-4534-9673-53805122b83a.mp4

Which causes throttles eventually

We could debounce queries on these events. ( cc @jywarren )

Potential source for these event triggers

https://github.com/publiclab/plots2/blob/f3ec8ff3d8d73d55e791a4eb7db84d1aeb6c106d/app/assets/javascripts/leafletHelper.js#L154

https://github.com/publiclab/plots2/blob/f3ec8ff3d8d73d55e791a4eb7db84d1aeb6c106d/app/assets/javascripts/leafletHelper.js#L54

Ref #9698

daemon1024:
Well, it may help somewhat but I don't think debounce is helping as much as I thought it would.

On a wider map, the endpoint itself is slow.

![image](https://user-images.githubusercontent.com/47106543/126904108-8d337447-3172-47c1-96ba-c272069bbd77.png)

But unlike our findings in #7556 , 
![image](https://user-images.githubusercontent.com/47106543/126904254-b4d98ee4-4f95-4bb5-8dee-f0325d71a8ed.png)
I don't see any particular SQL bottlenecks. More like I don't see anything useful in Skylight.

On switching to allocation insights,

![image](https://user-images.githubusercontent.com/47106543/126904334-ecbc3e30-5925-4c2b-aadd-983eb15e7ebd.png)

I see a lot of new events which are hogging up a lot of time, will check it out.

Link for skylight info of the screenshots - https://oss.skylight.io/app/applications/GZDPChmcfm1Q/1627225620/5m/endpoints/Srch::Search%20%5BGET%5D%20srch%2Ftaglocations?mode=memory

jywarren:
One clue to figure out is, when you change the `limit` does it vastly affect the speed, as @17sushmita noted in the call today? That might make sense, but if it doesn't much help, a question is, does `limit` actually reduce query time or are we just adding a limit near the end of a chain of queries and we aren't getting much benefit from it?

[...]

Issue #9698, opened 2021-05-30
[Planning] Geographic Features Refinement


daemon1024:
- Standardize Adding New Spreadsheet-based layers https://github.com/publiclab/leaflet-environmental-layers/issues/488
	- Add widget on the PL website
	- Set up script(s) to generate content for new layers
	- Integrate the widget and script through API calls and GH Actions
- Increase Amount of Content on `/map`
	- Develop microservice over the current API
		- TBD
	- Benchmark and Investigate the current People API https://github.com/publiclab/plots2/issues/7556

Succeeds https://github.com/publiclab/plots2/issues/1416

Should we keep the tasks to the existing milestone i.e. https://github.com/publiclab/plots2/milestone/7 or create a separate milestone for the soc project?

This is currently broad planning points for the project. I will be detailing out, break it further, adding more relevant tasks.

ebarry:
Let's keep this milestone, I"ll update it with this planning issue. Thanks!

daemon1024:
https://github.com/publiclab/leaflet-environmental-layers/issues/121 I was wondering if spinning it out the component as Leaflet Plugin, would ease out our scripts and workflow.

What are your thoughts about this @jywarren ?

daemon1024:
https://github.com/publiclab/leaflet-environmental-layers/issues/480 Bumping this issue and adding to planning task.

Have we consider adding a guide/docs which doesn't require people to download/install things? Using CDNs like [jsdelivr](https://www.jsdelivr.com/package/npm/leaflet-environmental-layers?path=dist) or [unpkg](https://unpkg.com/browse/leaflet-environmental-layers@2.4.3/dist/). This would ease out for someone who is wanting to try it out without needing to externally add anything.

jywarren:
> [publiclab/leaflet-environmental-layers#121](https://github.com/publiclab/leaflet-environmental-layers/issues/121) I was wondering if spinning it out the component as Leaflet Plugin, would ease out our scripts and workflow.
> 
> What are your thoughts about this @jywarren ?

I like this idea a lot! It'd be a nice output of the fellowship too!

jywarren:
[...]

PR #10059, opened 2021-08-18
Increase Limit and Debounce Queries

daemon1024:
Ref #9946 
#9698

### DIFF
--- a/app/assets/javascripts/leafletHelper.js
+++ b/app/assets/javascripts/leafletHelper.js
@@ -51,7 +51,7 @@
        });
    }
 
-   function contentLayerParser(map, markers_hash, map_tagname) {
+   function contentLayerParser(map, markers_hash, map_tagname, limit) {
       var NWlat = map.getBounds().getNorthWest().lat ;
       var NWlng = map.getBounds().getNorthWest().lng ;
       var SElat = map.getBounds().getSouthEast().lat ;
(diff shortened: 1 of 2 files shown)


jywarren (approved):

daemon1024:
https://user-images.githubusercontent.com/47106543/129971172-de37ff32-ecfc-4d93-b83b-dae71bd1505e.mp4

Working as intended I believe :)
Try it out at https://unstable.publiclab.org/map

daemon1024:
I will try implementing the map clustering here, probably will look nicer? instead of overlapping markers. Wdyt @jywarren

jywarren:
This looks great, why dont we try this as in and do clustering in another pr? Could I merge this now?

Great work today!!

jywarren:
Hmm, the UI freezes up a bit for me on mobile. Is it trying to crunch too much data?

jywarren:
Also I worry a bit about overloading the server. What if we enabled these higher limits only if there's a secret GET parameter like limit= in /map?limit=50 for example. Then we can get a controlled sense of the server load in a live environment.

daemon1024:
It's doable, fetching the limit from query param. 
But,
Just wondering why would it overload the server 🤔

We have been dealing with a lot heavier queries and while having the number of heavy queries higher too. 

From what I can estimate currently limit=100 would be equivalent to earlier limit=5. 

Let me know if I misunderstood something.

jywarren:
I'm just always cautious and would like to monitor a little bit in the production environment. Would you mind making it a parameter? Thanks!!

jywarren:
You know, just don't want to make assumptions!

jywarren (approved):

jywarren:
Ok awesome. We can do some load tests and then change the default.

jywarren:
Very nice! Let's try to stress this for a week after publishing to the live site on Monday and then we'll lift the limit!

image

PR #10076, opened 2021-08-22
[WIP] Screenshot UI changes

imajit:
Part of #9686 
This test will run only on PRs that have 'Translation System' as a label. They are system tests only which basically will screenshot web page corresponding to file changed. 
To Do: send a list of files changed to tests, create a mapping of view file and corresponding route to take screenshot of only files changed, comment screenshot in PR.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -264,3 +264,5 @@ jobs:
       with:
         name: ui-test-screenshots
         path: tmp/screenshots/ui-test/*
+
+        
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -257,7 +257,7 @@ jobs:
       run: |
         export DISPLAY=:99
         chromedriver --url-base=/wd/hub &
-        $${{needs.changed-files.outputs.views}} > test/reports/changed-files.txt &
+        ${{needs.changed-files.outputs.views}} > test/reports/changed-files.txt &
         bundle exec rails test:ui-tests
     - name: Archive ui test screenshots
       uses: actions/upload-artifact@v2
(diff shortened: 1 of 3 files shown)


imajit:
@jywarren this one is still not complete yet, just wanted to know if I'm on the right track ?

.github/workflows/tests.yml
```diff
@@ -202,3 +202,65 @@ jobs:
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Get changed files
+      id: changes
+      run:  |
+        echo "::set-output name=views::$(git diff --name-only --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | grep app/views/ | xargs))"
+
+  translation-screenshot:
+    if: ${{ github.event.label.name == 'Translation System' }}
```
jywarren, line 219: This looks interesting! So, this is triggered only by that label. What gives it that label? 

My only caution here is that this extra script lengthens and complicates our standard test runs a bit. It might make it a little harder to read and maintain. For example, what if something changes in the test setup. We'd have to make sure this part also gets updated accordingly. But I'm not against it. Is there any way, though, to reduce the repetition and to rely on common reusable code?

Looking very good though!!

imajit:
@jywarren @noi5e  is there some way to restart these tests, it seems they are stuck

jywarren:
Noting we could trigger this by using the label: https://github.com/publiclab/plots2/labels/needs-screenshot

imajit:
Really sorry this PR is pending for a long time. I'm not sure why the tests are not running. I'm not familiar with GitHub workflow so was checking how things go after adding small changes. 
I think opening a new PR might rerun the tests.

noi5e:
Yeah, this is strange that the tests aren't running. @jywarren

@imajit, something you could do is `git rebase` so that this PR is built off of the most current `main` branch. There might be some CI updates in the most current `main` that could help this one out. 

If you've never done a rebase, I just wrote up [some instructions](https://github.com/publiclab/plots2/pull/10142#issuecomment-921986628) for that this week.

cesswairimu:
I will try to close and reopen one more time

PR #10090, opened 2021-08-24
Refined design of delete button on delete page

### LINKED ISSUE #10087: Add tooltip to delete button on posts
It's hard to understand what the delete button in the `Tools` section in a note is for. The delete button looks like any button and there's no visual cue to understand that the contributor is about to commit a dangerous action that cannot be reverted ( a contributor only knows this after clicking the button). I'd suggest separating the delete button from other normal buttons and using a red background for the button. It would also be really helpful if words like `Delete Note` are added to the button.

Screenshot 2021-08-24 at 14 44 56

I'm not good at design but I think something like this works better
Screenshot 2021-08-24 at 15 05 23

@ebarry @jywarren @noi5e @Tlazypanda What do you think?

TildaDares:
Fixes #10087 

Screenshot 2021-08-24 at 16 32 42

### DIFF
--- a/app/views/like/_like.html.erb
+++ b/app/views/like/_like.html.erb
@@ -92,9 +92,6 @@
                     <i class='fa fa-user-plus' aria-hidden='true'></i>
                 </li>
                 <% end %>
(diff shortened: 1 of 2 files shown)


noi5e (approved):
Looks great, nice work @TildaDares!

PR #10135, opened 2021-09-05
run tests in vscode with a plugin!

jywarren:
https://www.gitpod.io/docs/languages/ruby#ruby-test-explorer


jywarren:
Hmm, too bad, not working:

image

The demo image looks great!

image

PR #10162, opened 2021-09-16
Update fr.yml with translations

comic-dev:
Added the translation as suggested in issue #10161 

* [-] screenshots/GIFs are attached 📎 in case of UI updation

### DIFF
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -1,5 +1,12 @@
 ---
 fr:
+  community:
+    community_title: "Communauté"
+    get_started: "Se lancer"
+    people: "Individus"
+    places: "Régions"
+    projects: "Projets"
+    events: "Evènements"
   users:
     reset:
       reset_password: "Réinitialisez votre mot de passe"


noi5e:
Thank you @comic-dev! This looks good so far!

@imajit Can you review this one? It will only take a few mins.

noi5e (approved):
On second thoughts, even though I'm not familiar with the translation files, I think this seems simple enough that it's easily mergeable. Merging now!

noi5e:
Congrats @comic-dev!

imajit:
Hi @noi5e , Sorry for the late reply 
Each translation file is linked with its corresponding Transifex resources. We export new strings from en.yml files to the project and the translation of corresponding strings is imported into language yml files. So only transifex bot must add commits to `language.yml`(by language i mean to say all other languages except English)  and strings to be translated can be added to `en.yml` file.
Adding translation directly to language files except en.yml can make things a bit tricky. 
Transifex project syncs the en.yml file regularly but the other language files are only synced when they are completely reviewed,  or we initiate a manual sync. So if we directly add strings to the language files like in this case we add to fr.yml 
- If the key is not present in en.yml file for instance let's say `community.community_title` is not present in en.yml file and we add a translation to it in `fr.yml` file, syncing becomes difficult because we would have to add all the keys missing in en.yml, wait for it to sync and then rerun the entire thing, else we might lose these translations.
- Till we don't sync the files these translations will not be reflected on the Project dashboard so the Translation team cannot review it and in the worst case, we might end up having different translations for the same string - one by the Translation team and one added by direct contributions. 
So, while I'm glad @comic-dev wants to help with the Translation project and appreciate the work, adding translations straight to the language file is not the best way to go about it. 
Here is a link to the Translation wiki https://publiclab.org/wiki/translation , you can add yourself to the Transifex project and add translation to the Transifex project.

noi5e:
@imajit Thanks so much for the detailed explanation! 

That means I'll have to revert the changes in this PR and will close the other issues that are open with similar updates to the translation files.

jywarren:
Screenshot_20210918-134944

Reverting! Pic for ref. No big deal, thanks!!

PR #10192, opened 2021-09-21
Add tallies to tag stats page

jywarren:
We'd look at https://publiclab.org/tag/water-quality/stats (or equiv) to test, at the bottom under the last graph!


jywarren:
This worked in GitPod!

PR #10211, opened 2021-09-28
add lazy load conditional on filename == "blob"

jywarren:
https://github.com/jassa/lazyload-rails/#features

fixes #10210

PR #10213, opened 2021-09-28
Lazy blob images fix 2

jywarren:
follow-up to https://github.com/publiclab/plots2/pull/10211/


jywarren:
So this is working on this page: https://stable.publiclab.org/tag/infragram but not on https://stable.publiclab.org/tag/infragram-upload, where i'm seeing an error! :-///

jywarren:
Small follow-up fix in a1b105fb6a50c939a9a303c1efb384ca7bcf389e to address the tag page!

PR #10322, opened 2021-10-12
Add a hover text on the lock and "+" buttons

### LINKED ISSUE #9978: Add a hover text on the lock button
This is subtask to #9888
Add a popover on hover instead of click?(explore if this is possible) When a contributor sees the `disabled` sign first, they might not bother to click and read why they can't add tags.

It would also be really helpful if contributors could hover over the lock icon on a locked page to see why it has that icon.

Screenshot 2021-07-09 at 10 06 24

_Originally posted by @TildaDares in https://github.com/publiclab/plots2/issues/9888#issuecomment-877031292_

DalilaBA:
Fixes #9978 

* [+] PR is descriptively titled 📑 and links the original issue above 🔗
* [+] tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR -- or run tests locally with `rake test`
* [+] code is in uniquely-named feature branch and has no merge conflicts 📁

Therefore, there was and old commit that I forgot, I didn't know how to delete it, it is the one about "updated the _flag_list and the translation en.yml file".

### DIFF
--- a/app/views/tag/show/_sort.html.erb
+++ b/app/views/tag/show/_sort.html.erb
@@ -1,5 +1,5 @@
 <div class="dropdown pull-right">
-  <button class="btn btn-outline-secondary dropdown-toggle" id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" style="margin-left:14px;">Sort by</button>
+  <button class="btn btn-outline-secondary dropdown-toggle" id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" style="margin-left:14px;"><%=translation('tag.index.sort_by')%></button>
   <div class="dropdown-menu dropdown-menu-right mt-1" aria-labelledby="dropdownMenuButton">
   <% if params[:action] == "show" %>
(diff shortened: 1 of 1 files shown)


DalilaBA:
I resolved the conflict on the _flag file by removing my modifications

TildaDares:
Hi @DalilaBA, I think you need to rebase to remove those old commits. I use this https://gist.github.com/jordanbtucker/17b6c3c7cdaca12327d0 when I get old commits in my PR.

The `<commit>` should be the hash of the first commit you'd like to start including commits from. In your case this would be `262d3c0`. The `<branch>` would be `main`.

DalilaBA:
I wanted to remove an older commit, and I found another solution check this out : https://stackoverflow.com/questions/9725156/remove-old-git-commits
I used the "drop" methode

TildaDares:
@DalilaBA Just ignore the message

TildaDares:
@DalilaBA, I think this PR needs to be broken into two. Work on just adding the hover text to the lock icon and I'll create another issue for the '+' button. That way we know exactly what is causing the problem.

PR #10353, opened 2021-10-18
display contributor from most recent

### LINKED ISSUE #10016: Display contributors and followers from most recent to least recent
**What this affects**

This is regarding tag pages, specifically the Contributors tab that shows both contributors and followers. Here is this tab for the tag `water-quality`: https://publiclab.org/contributors/water-quality

**Please describe the current behavior:**

Currently, the contributors are displayed from the oldest at top to newest at the bottom. This is also true for the followers table. 
The consequences of this is that the earliest Public Labbers always appear "above the fold" while the newest ones may never even been seen by those who are curating the topic area, such as @bhamster. 

contributor-order

It would be an improvement to put the newest contributors and newest followers at the top (instead of at the bottom). Thanks!

max-keviv:
Display contributors most recent to least recent

Fixes #10016

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -119,7 +119,9 @@ class TagController < ApplicationController
       end
     end
 
-    order_by = if params[:order] == 'views'
+    order_by = if params[:order] == 'newest'
+                 'node_revisions.timestamp ASC'
+               elsif params[:order] == 'views'
                  'node.views DESC'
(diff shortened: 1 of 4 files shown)


TildaDares:
Hi @max-keviv, instead of adding another filter we could just make sorting the users from the most recent be the default behavior. What do you think?

max-keviv:
@TildaDares thanks for messaging 
It is a good idea, trying to implement that

Issue #10548, opened 2021-12-01
Lead image uploading to new wikis is failing


ebarry:
### Please describe the problem (or idea)

I have verified the following report from Jeanette: 
> Adding a main image to a new wiki doesn't seem to be working for me. When I try to drag-and-drop an image, the grey image box gets a little darker, but nothing happens after. And when I try to select in image from file, the file dialog box comes up and I can select a file, but then nothing happens.

I created this gif: 
![wiki-image upload](https://user-images.githubusercontent.com/161439/144252478-9b783a9e-3919-4fee-b77b-01a9297a64a6.gif)

> What did you expect to see that you didn't?
I expected to see the lead image that i dragged onto the upload box to turn into a preview image, and then upon saving, i expected to see it appear as lead image on the wiki.  

### Please show us where to look

https://publiclab.org/wiki/foo or any wikiname that doesn't yet exist

### What's your PublicLab.org username?

liz

TildaDares:
I think this has to do with the change @jywarren made in #10522 specifically line 189
https://github.com/publiclab/plots2/blob/9a08a054cb0b695dd1928aae6fb534870a742e6d/app/assets/javascripts/editorToolbar.js#L189

Changing `document` to `window` fixes the problem and still shows lead images on notes when logged out.

jywarren:
Thanks for reporting!

Yes @TildaDares, good catch, that seems likely. We don't have good test protection for this feature actually, and I was trying to do some tests in #10517. That'd give us better reliability across this set of features. 

OK awesome, let's go for `window`, thanks!

ebarry:
cool sleuthing @TildaDares !!! catching two unexpected behaviors, so great!

PR #10556, opened 2021-12-07
fix wiki lead image upload

jywarren:
fixes https://github.com/publiclab/plots2/issues/10548


jywarren:
Manually tested in Gitpod. Good!

Issue #10219, opened 2021-09-30
more pagination in research notes 


max-keviv:
when In research notes of few topics the last few pages in pagination don't contain the research notes

there may be extra pagination 

here you can see
https://publiclab.org/tag/climate-change?page=2
or
https://publiclab.org/tag/indoor-air-quality?page=3

https://user-images.githubusercontent.com/59287619/135364762-b6b9ecb5-0aaf-4b3f-9c5a-796d393c6e1e.mp4

ebarry:
Confirming that I am also experiencing this pagination bug. Thanks for reporting @max-keviv !

jywarren:
Hi all, the issue is here: we're paginating before we filter out questions:

https://github.com/publiclab/plots2/blob/883e081e0bbbd7cd25332798a4b27404f67f1035/app/controllers/tag_controller.rb#L132-L143

@max-keviv, would you like to try to fix this bug? It might be as simple as moving the pagination lines down below, but there could also be unexpected complications. Happy to help in either case!

max-keviv:
@jywarren thank you, 
yes I would like to fix this bug 
so if I am right then, we need to move the line 132-133 after 143
please correct me if I'm missing something

KarishmaVanwari:
Can I take up this issue? @jywarren

TildaDares:
Hi @max-keviv, you can go ahead and move those lines. You can also check for unexpected complications as @jywarren said.

jywarren:
Yes @max-keviv thank you - hopefully we just pass all tests no problem. But if a test fails, let's try to see why!

@KarishmaVanwari i think we have one open here if you'd like to leave a comment claiming it: https://github.com/publiclab/plots2/issues/10368 

Thank you everyone!

tichnas:
Hi @TildaDares @jywarren

Is this issue still open? If yes, I'd love to work on it :)

Thanks

jywarren:
Hi @tichnas i believe @max-keviv is working on it in https://github.com/publiclab/plots2/pull/10569 now. If you'd like another issue to work on, have you completed one at https://code.publiclab.org yet? Thank you!

PR #10569, opened 2021-12-14
fixing more pagination in research notes

max-keviv:
As discussed in issue #10219
I move the line 132-133 after 143

### DIFF
--- a/app/controllers/tag_controller.rb
+++ b/app/controllers/tag_controller.rb
@@ -129,9 +129,6 @@ class TagController < ApplicationController
                  'node_revisions.timestamp DESC'
                end
 
-    @pagy, nodes = pagy(nodes.order(order_by), items: 24)
-    @paginated = true
-
     @qids = Node.questions.where(status: 1)
                .collect(&:nid)
     if @qids.empty?
@@ -141,6 +138,9 @@ class TagController < ApplicationController
       @notes = nodes.where('node.nid NOT IN (?)', @qids) if @node_type == 'note'
       @questions = nodes.where('node.nid IN (?)', @qids) if @node_type == 'questions'
     end
+    
+    @pagy, nodes = pagy(nodes.order(order_by), items: 24)
+    @paginated = true
 
     @wikis = nodes if @node_type == 'wiki'
     @wikis ||= []


TildaDares (changes requested):

app/controllers/tag_controller.rb
```diff
@@ -141,6 +138,9 @@ def show
       @notes = nodes.where('node.nid NOT IN (?)', @qids) if @node_type == 'note'
       @questions = nodes.where('node.nid IN (?)', @qids) if @node_type == 'questions'
     end
+    
+    @pagy, nodes = pagy(nodes.order(order_by), items: 24)
+    @paginated = true
```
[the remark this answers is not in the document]
max-keviv: so @paginated on line 143 is right or it should also be needed to move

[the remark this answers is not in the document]
TildaDares: The only thing that needs to change is `nodes` being sorted before line 138. The rest can stay.

app/controllers/tag_controller.rb
```diff
@@ -141,6 +138,9 @@ def show
       @notes = nodes.where('node.nid NOT IN (?)', @qids) if @node_type == 'note'
       @questions = nodes.where('node.nid IN (?)', @qids) if @node_type == 'questions'
     end
+    
+    @pagy, nodes = pagy(nodes.order(order_by), items: 24)
```
jywarren, line 142: So I think we need to make a new line on 137:

`nodes = nodes.order(order_by)`

Then we can change this line to:

```suggestion
    @pagy, nodes = pagy(nodes, items: 24)
```

That way the sorting still happens above, but the pagination happens below. Sound good?

app/controllers/tag_controller.rb
```diff
@@ -129,9 +129,6 @@ def show
                  'node_revisions.timestamp DESC'
                end
 
-    @pagy, nodes = pagy(nodes.order(order_by), items: 24)
-    @paginated = true
-
     @qids = Node.questions.where(status: 1)
```
jywarren, line 132:
```suggestion
    nodes = nodes.order(order_by)
    @qids = Node.questions.where(status: 1)
```

jywarren:
I added two suggestions you can accept. @TildaDares feel free to merge this once those are complete!

app/controllers/tag_controller.rb
max-keviv: @jywarren  thank you for suggestions

max-keviv:
open pull request at #10588

PR #10618, opened 2022-01-04
Change sorting of Tag.contributors to date of contribution, desc

jywarren:
Fixes 1/2 of https://github.com/publiclab/plots2/issues/10016

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -83,7 +83,7 @@ class Tag < ApplicationRecord
     uids.sort_by do |t|
       -t[1]
     end
-    uids = uids.uniq(&:0) # eliminate those without unique uids
+    uids = uids.uniq(&:first) # eliminate those without unique uids
       .collect(&:first) # flatten to an array of just uids
     User.where(id: uids)
         .where(status: [1, 4])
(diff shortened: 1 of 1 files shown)


app/models/tag.rb
```diff
@@ -65,13 +65,26 @@ def self.contributors(tagname, start: Time.now-100.years, finish: Time.now+1.yea
...
+      uids += n.revision
+        .where(timestamp: start.to_i..finish.to_i)
+        .order(timestamp: :desc)
+        .pluck(:uid, :timestamp)
     end
-    uids = uids.uniq
+    uids.sort_by do |t|
+      -t[1]
+    end
+    uids = uids.uniq(&:0) # eliminate those without unique uids
```
jywarren, line 86:
```suggestion
    uids = uids.uniq(&:first) # eliminate those without unique uids
```

jywarren:
Hm, no change visible...

jywarren:
So the gist here is that it doesn't care AT ALL the order of uids we pass to `User.where(uid: uids)`... so we might as well skip all this and just reverse the order of the `User` query.

PR #10745, opened 2022-03-01
Test PR from main to stable branch

jywarren:


jywarren:
Testing this as a way to publish to stable branch. But it looks like it doesn't allow FF -- it insists on rebasing or squashing. So, we'd better not.

PR #10793, opened 2022-03-18
Fixes #10218: Resolved CodeClimate issues on models/user.rb

### LINKED ISSUE #10218: CodeClimate issues reported on models/user.rb

This is not a bug, but a wip-tracking issue about the `user.rb` file.
Accordingly to codeclimate we have multiple issues there that could be addressed one by one.

https://codeclimate.com/github/publiclab/plots2/app/models/user.rb/source#issue-f685bcbb6f822884dec234767db1861a

KarishmaVanwari:
Fixes #10218 

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -38,10 +38,10 @@ class User < ActiveRecord::Base
     c.validates_format_of_login_field_options = { with: Authlogic::Regex::LOGIN, message: I18n.t('error_messages.login_invalid', default: "can only consist of alphabets, numbers, underscore '_', and hyphen '-'.") }
   end
 
-  has_attached_file :photo,
-                    styles: { thumb: '200x200#', medium: '500x500#', large: '800x800#' },
-                    url: '/public/system/profile/photos/:id/:style/:basename.:extension',
-                    path: ':rails_root/public/system/public/system/profile/photos/:id/:style/:filename'
+  has_attached_file :photo, 
+    styles: { thumb: '200x200#', medium: '500x500#', large: '800x800#' },
+    url: '/public/system/profile/photos/:id/:style/:basename.:extension',
+    path: ':rails_root/public/system/public/system/profile/photos/:id/:style/:filename'
 
   do_not_validate_attachment_file_type :photo_file_name
   # validates_attachment_content_type :photo_file_name, :content_type => %w(image/jpeg image/jpg image/png)
@@ -161,10 +161,10 @@ class User < ActiveRecord::Base
   end
 
   def normal_tags(limit = false)
-    tags(limit).select{ |tag| ! tag.name.include?(':') }
+    tags(limit).select { |tag| !tag.name.include?(':') }
   end
 
-  def tagnames(limit = 20, defaults = true)
+  def tagnames(limit = 20, _defaults = true)
     tagnames = []
     Node.includes(:tag).order('nid DESC').where(type: 'note', status: 1, uid: id).limit(limit).each do |node|
       tagnames += node.tags.collect(&:name)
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -161,10 +161,10 @@ def tags(limit = 10)
   end
 
   def normal_tags(limit = false)
-    tags(limit).select{ |tag| ! tag.name.include?(':') }
+    tags(limit).select { |tag| !tag.name.include?(':') }
   end
 
-  def tagnames(limit = 20, defaults = true)
+  def tagnames(limit = 20, _defaults = true)
```
brunoocasali, line 167: I'm not sure if this is related to something else, but `_defaults` is not used in the method, why keep it? 😄

app/models/user.rb
```diff
@@ -408,15 +409,15 @@ def generate_token
 
   class << self
     def search(query)
-      query = query.gsub('@',' ') # @ is a special char in full text search in MYSQL, and cannot be escaped; https://github.com/publiclab/plots2/issues/8344
-      query = query + '*' unless query.empty?
-      User.where('MATCH(bio, username) AGAINST(? IN BOOLEAN MODE)', "#{query}")
+      query = query.tr('@', ' ') # @ is a special char in full text search in MYSQL, and cannot be escaped; https://github.com/publiclab/plots2/issues/8344
+      query += '*' unless query.empty?
+      User.where('MATCH(bio, username) AGAINST(? IN BOOLEAN MODE)', query.to_s)
```
brunoocasali, line 414: A good practice that improves the readability of your code is breaking a line between the body of the function and their response.

```rb
def search(...)
   query = query.tr('@', ' ')
   query += '*' unless query.empty?
   
   User.where('MATCH(bio, username) AGAINST(? IN BOOLEAN MODE)', query)
end
```
When another developer reads, they will be aware that this is the implied return of the function.

brunoocasali: btw, I think you don't need the `.to_s` in the query call!

app/models/user.rb
```diff
@@ -493,8 +494,8 @@ def watching_location(nwlat, selat, nwlng, selng)
     end
   end
 
-  def recent_locations(limit = 5)
-    recent_nodes = self.nodes.includes(:tag)
+  def recent_locations(_limit = 5)
```
brunoocasali, line 497: In this case, you have the opportunity to use your argument:

```rb
...
.order(created: :desc)
.limit(limit)
```

brunoocasali (commented):
Hi, @KarishmaVanwari I made some suggestions, congratulations by the job :D

app/models/user.rb
KarishmaVanwari: Done!

app/models/user.rb
KarishmaVanwari: Codeclimate reported to use `to_s` instead of string interpolation, however, I've now removed it.

app/models/user.rb
KarishmaVanwari: Done!

KarishmaVanwari:
@brunoocasali I've made the changes as suggested. Hope it works. Thanks!

brunoocasali (approved):
Awesome @KarishmaVanwari :)
I'll let the real maintainers check out your branch.

And I think this does not fix entirely the issue, so I think you could remove the "Fixes xxxx" part :)

KarishmaVanwari:
@brunoocasali Done. Thanks!

jywarren (approved):
Very nice, thank you @KarishmaVanwari and @brunoocasali both!!!

PR #10794, opened 2022-03-18
Fixes #6304: Added eye icon for contributors that are following too

### LINKED ISSUE #6304: Reorganize /contributors/___ pages
(follow-up to https://github.com/publiclab/plots2/pull/5632) cc @IshaGupta18 🙌 

We now show 2 types of users on our contributor pages like https://publiclab.org/contributors/water-quality

1. contributors - who have posted using this topic or commented on this topic
2. followers - who are following this topic and will get notified when new content is posted to it

They're a bit messily next to each other:

image

Let's:

* [ ] show contributors in a table on top, but in each row, mark if that person is ALSO following -- maybe, with an eye icon? In a column called `Following`, maybe?)
* [ ] then let's show people who are NOT in the first list, but ARE subscribed. This can be in a second table with the title `Following`. 

Should we have a short text explanation of the distinction?

KarishmaVanwari:
An eye icon now appears right beside the name of the contributor if it is following too.

Fixes #6304 

Screenshot:
image

### DIFF
--- a/app/views/tag/_contributors.html.erb
+++ b/app/views/tag/_contributors.html.erb
@@ -46,7 +46,7 @@
           <% if @users %>
             <% @users.each do |user| %>
               <tr class="users-row">
-                <td><a href='/profile/<%= user.name %>'><i class="fa fa-user"></i> <%= user.name %></a> <% if user.following(params[:id])%><i class="fa fa-eye"></i><% end %> </td>
+                <td><a href='/profile/<%= user.name %>'><i class="fa fa-user"></i> <%= user.name %></a> <% if user.following(params[:id])%><a aria-label="Following" data-toggle="tooltip" title="Following" data-placement="top"> <i class="fa fa-eye"></i></a><% end %> </td>
                 <td><a href='/tag/<%=h params[:id] %>/author/<%= user.name %>'><%= translation('tag.contributors.notes') %> &raquo;</a></td>
               </tr>
             <% end %>
--- a/app/views/tag/_contributors.html.erb
+++ b/app/views/tag/_contributors.html.erb
@@ -46,7 +46,7 @@
           <% if @users %>
             <% @users.each do |user| %>
               <tr class="users-row">
-                <td><a href='/profile/<%= user.name %>'><i class="fa fa-user"></i> <%= user.name %></a></td>
+                <td><a href='/profile/<%= user.name %>'><i class="fa fa-user"></i> <%= user.name %></a> <% if user.following(params[:id])%><i class="fa fa-eye"></i><% end %> </td>
                 <td><a href='/tag/<%=h params[:id] %>/author/<%= user.name %>'><%= translation('tag.contributors.notes') %> &raquo;</a></td>
               </tr>
             <% end %>


TildaDares:
This looks really good @KarishmaVanwari but I don't know if users would be able to understand what the eye icon means. Can we add a tooltip or something that explains what the icon is used for? What do you think?

KarishmaVanwari:
@TildaDares That's a great idea! I can do that. Thanks!

KarishmaVanwari:
@TildaDares I've added the tooltip saying "Following". How does it look?
image

KarishmaVanwari:
@TildaDares I have a suggestion. The "Notes" link beside every contributor looks repetitive (especially when there are a large number of contributors), how about replacing it with a notes icon, using something like this `<i class="fa-solid fa-note-sticky"></i>` ? Does it sound like a potential change?

TildaDares:
@KarishmaVanwari I think you're right about the repetition but I believe the `Notes` text was added to serve as a link to all of the user's notes for that particular tag. Using an icon probably has the same effect as the `Notes` text since we'll have to repeat it for each contributor row. Also, an icon won't properly communicate to the user that this links to the user's content. 

Do you see any other way we could improve this?

Also, another possible improvement would be to add a `See More` link at the bottom of the `People who've contributed` and `Following` sections when the number of contributors exceeds a particular limit. What do you think about this?

TildaDares (approved):
Great work @KarishmaVanwari. Thank you!!

KarishmaVanwari:
> @KarishmaVanwari I think you're right about the repetition but I believe the `Notes` text was added to serve as a link to all of the user's notes for that particular tag. Using an icon probably has the same effect as the `Notes` text since we'll have to repeat it for each contributor row. Also, an icon won't properly communicate to the user that this links to the user's content.
> 
> Do you see any other way we could improve this?
> 
> Also, another possible improvement would be to add a `See More` link at the bottom of the `People who've contributed` and `Following` sections when the number of contributors exceeds a particular limit. What do you think about this?

Yes @TildaDares, I get your point of repetitive nature of icon too, maybe we can work out something else. I'll think over this. Thanks for considering my suggestion :)

Also, the idea of a `See More` link at botton seems a nice idea to be implemented. I'd love to implement it!

KarishmaVanwari:
@TildaDares Another thought that striked me is that on this page, contributors and followers column seem a little non-uniform:
image


Maybe we can either:
1. Add `Contributors` title and remove both `People who've contributed` and `People who are following` subtitles since the words `Contributors` and `Followers` are self explanatory (or, if needed we can add a toolip or hover text to explain better) or,
2. We can simply remove `Followers` title.


Implementing either of the above two will bring uniformity to both columns. What do you think?

TildaDares:
> 1. Add `Contributors` title and remove both `People who've contributed` and `People who are following` subtitles since the words `Contributors` and `Followers` are self explanatory

I like this! But we already have a `Contributors` heading so it'd be repetitive.

TildaDares:
I think this is a better option!
> 2\. We can simply remove `Followers` title.

jywarren:
Thank you!!!

PR #10896, opened 2022-04-01
added a catch to check for search term before results are loaded

### LINKED ISSUE #10882: Search by questions returns no result for empty string even before typing the question

When you go to the search page and you try to select the by type dropdown then select questions, the page loads and return a no result for an empty string response to the page.

Screenshot 2022-03-31 at 09 42 03

This idea will solve the issue of searching for an empty string

I expect the page to only return a response when the search icon is clicked or the field is not empty

https://publiclab.org/search/questions/

> Many bugs are related to these -- please help us track it down and reproduce what you're seeing!

****

## Thank you!

kunmi02:
Added a check to the profiles and question html file to return a message asking for the user to enter a search term if non is provided
Fixes #10882 

### DIFF
--- a/app/views/search/profiles.html.erb
+++ b/app/views/search/profiles.html.erb
@@ -11,7 +11,9 @@
 
       <div class="row">
         <div class="col-lg-7">
-        <% if @profiles.present? %>
+        <% if params[:query].blank?%>
+          <p>Please enter a search term.</p>
+        <% elsif @profiles.present? %>
 
           <h4>Results for "<b><%= params[:query] %></b>":</h4>
           <p>Searched profiles by <i>username</i> and <i>bio</i></p>
@@ -44,7 +46,9 @@
         <% end %>
     </div>
     <div class="col-lg-4 col-lg-offset-1" >
-      <% if @tag_profiles.present?%>
+      <% if params[:query].blank? %>
+        <p>&nbsp;</p>
+      <% elsif @tag_profiles.present?%>
         <h4>Results for "<b><%= params[:query] %></b>":</h4>
         <p>Searched profiles by <i>profile tags</i></p>
         <% @tag_profiles.each do |profile| %>
--- a/app/views/search/questions.html.erb
+++ b/app/views/search/questions.html.erb
@@ -9,7 +9,10 @@
 
     <%= render partial: "search/form", locals: { searchType: "questions/" } %>
 
-    <% if @questions.present? %>
+    <% if params[:query].blank?%>
+      <p>Please enter a search term.</p>
+      
+    <% elsif @questions.present? %>
       <h4>Results for "<b><%= params[:query] %></b>":</h4>
 
       <div class="related-tags-container">


TildaDares:
Hi @kunmi02, this looks really great. Can you try visiting `search/questions/` on your local env and check if it displays an error? Mine keeps displaying an error when I try to use the site with the changes in this PR.

kunmi02:
I just did that @TildaDares . Find attached the screenshot of what it looks likE
Screenshot 2022-04-03 at 16 16 57
e

kunmi02:
@TildaDares Please reveiew again

TildaDares:
@kunmi02 I'm still getting the same error when I try searching without a query

Screenshot 2022-04-04 at 19 02 57

kunmi02:
@TildaDares I did not even touch the node file but I will look through it and update the PR. Thanks

jywarren:
Hi, i wonder if there's a difference sometimes because we have separate handling for MySQL based installs vs. non-MySQL... the idea was that the non-mySQL will not try a more complex full text search... so the logic for running it with just sqlite (for example) is much simpler: 

https://github.com/publiclab/plots2/blob/7c3af440498624f5dca5ce9ee07f04bc3ad90e73/app/models/node.rb#L70-L72

@TildaDares it looks like the error you're posting is in that non-mySQL block, and it's not handling a nil input. We could add this just above the query lines to ensure no nils:

```ruby
query ||= "" # set non-nil default for non-mySQL queries
```

It's a short line and a bit mysterious so the explanatory comment is probably a good idea.

TildaDares:
@kunmi02 This looks really great!! Can you do the same for [Notes](https://github.com/publiclab/plots2/blob/main/app/views/search/notes.html.erb), [Wikis](https://github.com/publiclab/plots2/blob/main/app/views/search/wikis.html.erb) and [Tags](https://github.com/publiclab/plots2/blob/main/app/views/search/tags.html.erb)? Thanks a lot!

kunmi02:
@TildaDares  [Notes](https://github.com/publiclab/plots2/blob/main/app/views/search/notes.html.erb), [Wikis](https://github.com/publiclab/plots2/blob/main/app/views/search/wikis.html.erb) and [Tags](https://github.com/publiclab/plots2/blob/main/app/views/search/tags.html.erb) are already captured. The bug is only on questions page

TildaDares:
@kunmi02 On the tags page, it returns the result of nothing.
Screenshot 2022-04-11 at 13 37 49

kunmi02:
@TildaDares added check for tags

kunmi02:
@jywarren Kindly review PR please

app/views/search/tags.html.erb
```diff
@@ -4,7 +4,13 @@
 
 <div class="col-md-10">
   <div class="container">
-    <% if @tags.present? %>
+    <h2>Search tags</h2>
```
TildaDares, line 7: @kunmi02 This line should only be displayed when the query is empty.

TildaDares (changes requested):

Issue #11185, opened 2022-06-14
GSoC '22: "Legacy Code Deprecation" Project Planning


anirudhprabhakaran3:
Hello everyone! 🎉

This is the planning issue for my project "Legacy Code Deprecation" which I am going to work on over the summer. Comments, suggestions and improvements to everything mentioned in this are welcome and will be very useful for me!

My project proposal can be found [here](https://publiclab.org/notes/anirudhprabhakaran3/04-13-2022/gsoc-2022-proposal-legacy-code-deprecation).

### Before Coding

- [ ] Ask for an anonymized version of the production database.
- [ ] Talk to Sebastian about regular backups before every migration script 

### DrupalNodeImage
- [ ] Migration to migrate to native Image records and test on local system
- [ ] Database Backup (or at least tables that are to be changed)
- [ ] Test on unstable
- [ ] Test on stable

### Gallery Code
- [ ] Script to refactor legacy Gallery Code into Node Revision Body and test on local system
- [ ] Database Backup (or at least tables that are to be changed)
- [ ] Test on unstable
- [ ] Test on stable

### Answers model
- [ ] Check for any references missed out and test on local system
- [ ] Database Backup (or at least tables that are to be changed)
- [ ] Test on unstable
- [ ] Test on stable

### Map notes
- [ ] Script to convert map notes to normal notes
- [ ] Database Backup (or at least tables that are to be changed)
- [ ] Test on unstable
- [ ] Test on stable

### DrupalFile and DrupalUpload
- [ ] Migration to remove the two tables and test on local system
- [ ] Database Backup (or at least tables that are to be changed)
- [ ] Test on unstable
- [ ] Test on stable

jywarren:
Hi @anirudhprabhakaran3 this is looking great. I like how you've broken things up into groups, and see that you're using the existing deprecation milestone :tada:

In some places there are notes on whether records have been converted already, and I'm not sure about DrupalImages -- have you seen anything? As we go through these, probably best to link to the last known notes/status of these migrations. In some cases I think we have already moved the records and now just need to remove the tables and/or code. 

Were you thinking to go through these as they're written, top to bottom? Thanks!!!

anirudhprabhakaran3:
Yes, I'll add links to the previous issues, so that anyone viewing it can easily see the current status.

Regarding the DrupalImages, this was the last comment I found: https://github.com/publiclab/plots2/issues/76#issuecomment-481749324

[...]

PR #11225, opened 2022-06-25
Convert map nodes to note nodes

### LINKED ISSUE #4072: Begin converting Map nodes into normal note nodes
We have a range of nodes at https://publiclab.org/maps which are of type `map`, so `node.type == 'map'`. See one here: https://publiclab.org/map/chandeleur-islands-louisiana/2010-06-09

Let's migrate them to a simpler form of node, type `note` - but copy all the extra fields into the note body (actually part of the note's Revision record). We could do this in a migration kind of like this:

```ruby
Node.where(type: 'map').each do |node|
  revision = node.latest
  text = node.map.field_jpg_url_value
  revision.body = text + node.latest.body
  revision.save
  node.type = 'note'
  node.save({})
end
```

See how we can use Markdown to display all the relevant data in a simpler format, but still roughly in the arrangement we'd had in the original custom template: 

https://github.com/publiclab/plots2/blob/f9f09488560f92bb390ba762231e80f8fe61bdc6/app/views/map/show.html.erb#L34-L46

Let's not forget also the different forms of export we can display: 

https://github.com/publiclab/plots2/blob/f9f09488560f92bb390ba762231e80f8fe61bdc6/app/views/map/show.html.erb#L26-L29

Finally, let's embed a map as in the original display, at the top of the body content, using this line:

```ruby
<iframe style="width:100%;border:none;" height="375" src="https://publiclab.github.io/plots-leaflet-viewer/?tms=<%= @node.map.tms %>&lon=<%= @node.lon %>&lat=<%= @node.lat %>&zoom=<%= @node.map.min_zoom+1 %>"></iframe>
```

Once we can demonstrate that this works (please upload a screenshot of your script working; you can test it out on the rails console by running it with `rails console` interactively), we can turn it into a migration like one of these:

https://github.com/publiclab/plots2/blob/master/db/migrate/20180707065151_change_status_of_comments.rb

Follow-ups: removing the maps type code, including ones like this: https://github.com/publiclab/plots2/blob/master/app/models/drupal_content_field_map_editor.rb

and stuff inside https://github.com/publiclab/plots2/blob/master/app/views/map/

This is a bit complex but is a great exercise in Rails-style models and migrations! It'd also help resolve https://github.com/publiclab/plots2/issues/956

anirudhprabhakaran3:
Fixes #4072 

This MR aims to convert all map nodes into note nodes. It consists of a db migration, which can be run using `rails db:migrate` to be run. The script does the following:

1. Get the necessary values and data and add it to the revision body in the form of HTML
2. Update the type of node to `note`
3. Create a new path for the node to reflect its new type, using `generate_path`

## Screenshots

This is the console output before running the migration.

Screenshot from 2022-06-25 20-09-06

We can see that before the migration, it is being rendered as a map. We can see that from the URL as well.

Screenshot from 2022-06-25 20-11-08

After running the migration, we note that all the maps have been converted to notes:

Screenshot from 2022-06-25 20-12-46

It is also reflected when we go to the notes page:

Screenshot from 2022-06-25 20-13-10

Individually, we can see the map in the note template.

Screenshot from 2022-06-25 20-13-37

### DIFF
--- a/db/migrate/20220625142729_convert_map_to_note_nodes.rb
+++ b/db/migrate/20220625142729_convert_map_to_note_nodes.rb
@@ -2,14 +2,14 @@ class ConvertMapToNoteNodes < ActiveRecord::Migration[5.2]
   def change
     Node.where(type: 'map').each do |node|
       revision = node.latest
-      iframe_link = "<iframe style='width:100%;border:none;background:#ddd;margin-bottom:8px;' height='375' src='https://publiclab.github.io/plots-leaflet-viewer/?tms=#{node.map.tms}&lon=#{node.lon}&lat=#{node.lat}&zoom=#{node.map.min_zoom+1}'></iframe>"
+      iframe_link = "<iframe style='width:100%;border:none;background:#ddd;margin-bottom:8px;' height='375' src='https://publiclab.github.io/plots-leaflet-viewer/?tms=#{node.map&.tms}&lon=#{node.lon}&lat=#{node.lat}&zoom=#{node.map&.min_zoom+1}'></iframe>"
       conversion_buttons =
         "<div class='btn-toolbar'>"\
           "<div class='btn-group'>"\
-            "<a class='btn btn-sm btn-danger' rel='tooltip' title='View the map in a full-screen web viewer' href='https://publiclab.github.io/plots-leaflet-viewer/?tms=#{node.map.tms}&lon=#{node.lon}&lat=#{node.lat}&zoom=17'>Web viewer</a>"\
-            "<a class='btn btn-sm btn-success' rel='tooltip' title='' href='#{node.map.field_jpg_url_value}'>JPG (#{node.map.field_jpg_filesize_value.to_i} MB)</a>"\
-            "<a class='btn btn-sm btn-info' rel='tooltip' title='' href='#{node.map.field_geotiff_url_value}'>GeoTiff (#{node.map.field_geotiff_filesize_value.to_i} MB)</a>"\
-            "<a class='btn btn-sm btn-warning' rel='tooltip' title='Tiled Map Service (for developers)' href='#{node.map.field_tms_url_value}'>TMS</a>"\
+            "<a class='btn btn-sm btn-danger' rel='tooltip' title='View the map in a full-screen web viewer' href='https://publiclab.github.io/plots-leaflet-viewer/?tms=#{node.map&.tms}&lon=#{node.lon}&lat=#{node.lat}&zoom=17'>Web viewer</a>"\
+            "<a class='btn btn-sm btn-success' rel='tooltip' title='' href='#{node.map&.field_jpg_url_value}'>JPG (#{node.map&.field_jpg_filesize_value.to_i} MB)</a>"\
+            "<a class='btn btn-sm btn-info' rel='tooltip' title='' href='#{node.map&.field_geotiff_url_value}'>GeoTiff (#{node.map&.field_geotiff_filesize_value.to_i} MB)</a>"\
+            "<a class='btn btn-sm btn-warning' rel='tooltip' title='Tiled Map Service (for developers)' href='#{node.map&.field_tms_url_value}'>TMS</a>"\
           "</div>"\
         "</div>"
 
@@ -17,7 +17,7 @@ class ConvertMapToNoteNodes < ActiveRecord::Migration[5.2]
       "<div class='map-details'>"\
         "<small>"
 
-      if node.map.authorship
+      if node.map&.authorship
         map_details += "<p><b>By</b> #{node.map.authorship}</p>"
       else
         map_details +=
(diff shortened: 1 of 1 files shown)


anirudhprabhakaran3:
Also, part of the bigger deprecation effort being tracked at https://github.com/publiclab/plots2/issues/11185

(Added the comment so that MR shows up there too. This comment can be ignored)

db/migrate/20220625142729_convert_map_to_note_nodes.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+
+      old_body = revision.body
+
+      revision.body = iframe_link
+      revision.body += conversion_buttons
+      revision.body += map_details
+      revision.body += old_body
+      revision.body += notes
+      revision.body += cartographer_notes
```
cesswairimu, line 67: Love this arrangement :heart:

db/migrate/20220625142729_convert_map_to_note_nodes.rb
```diff
@@ -0,0 +1,76 @@
...
+  def change
+    Node.where(type: 'map').each do |node|
+      revision = node.latest
+      iframe_link = "<iframe style='width:100%;border:none;background:#ddd;margin-bottom:8px;' height='375' src='https://publiclab.github.io/plots-leaflet-viewer/?tms=#{node.map.tms}&lon=#{node.lon}&lat=#{node.lat}&zoom=#{node.map.min_zoom+1}'></iframe>"
+      conversion_buttons =
+        "<div class='btn-toolbar'>"\
+          "<div class='btn-group'>"\
+            "<a class='btn btn-sm btn-danger' rel='tooltip' title='View the map in a full-screen web viewer' href='https://publiclab.github.io/plots-leaflet-viewer/?tms=#{node.map.tms}&lon=#{node.lon}&lat=#{node.lat}&zoom=17'>Web viewer</a>"\
+            "<a class='btn btn-sm btn-success' rel='tooltip' title='' href='#{node.map.field_jpg_url_value}'>JPG (#{node.map.field_jpg_filesize_value.to_i} MB)</a>"\
+            "<a class='btn btn-sm btn-info' rel='tooltip' title='' href='#{node.map.field_geotiff_url_value}'>GeoTiff (#{node.map.field_geotiff_filesize_value.to_i} MB)</a>"\
```
cesswairimu, line 11: We might need to think on how to handle instances where some of the values here are nil nil e.g `field_geotiff_filesize_value` and  `field_geotiff_filesize_value`. Maybe if we can get unstable running we can see how these links look like.  I ran this for one map locally and some of this were nil this  is what I get 
```
"<a class='btn btn-sm btn-danger' rel='tooltip' title='View the map in a full-screen web viewer' href='https://publiclab.github.io/plots-leaflet-viewer/?tms=https://archive.publiclaboratory.org/leaflet/?    tms=https://archive.publiclab.org/2013/2013-04-15-us-massachusetts-plum-    island/tms/&lon=-70.80848&lat=42.7952&zoom=16&lon=23.0&lat=70.0&zoom=17'>Web viewer</a><a class='btn btn-sm btn-success' rel='tooltip' title='' href=''>JPG (0 MB)</a><a class='btn btn-sm btn-info' rel='tooltip' title='' href=''>GeoTiff (0 MB)</a>"
```
Everything else is looking great . Good job!

anirudhprabhakaran3: Hmm yes. When the values are nil, running `.to_i` on it returns `0` I believe. That's why the size for the JPG and GeoTiff is displayed as 0MB

jywarren: Sounds good - yes, perhaps we add a conditional to avoid adding them, OR we add a "disabled" class version, and/or a tooltip saying it's not available?

jywarren:
Hi @anirudhprabhakaran3 i'll be offline for 2 days but let me or Tilda or Cess know when this needs another review or when it's ready! Thanks!

db/migrate/20220625142729_convert_map_to_note_nodes.rb
anirudhprabhakaran3: I've added a "disabled" class to the button when the value is `nil` (i.e. 0MB)

anirudhprabhakaran3:
The migration is working succssfully on the unstable server!
https://jenkins.laboratoriopublico.org/job/Plots-Unstable/lastBuild/console

However, the build is not completing because of another error in the Uglifier gem,

jywarren:
@anirudhprabhakaran3 can you check that this seems to be working OK on stable before we push it to production? Thanks!

anirudhprabhakaran3:
Hi! Yeah, I'll try to check that. On my local laptop, I used to just run `Node.where(type: 'map').count` to see if it was a success. On the browser, I'll have to figure out a way to find map nodes. 

If you have access to the console to stable, could you run that maybe once and get the count? Thank you!

PR #11308, opened 2022-07-26
Deprecate answers model

anirudhprabhakaran3:
Fixes https://github.com/publiclab/plots2/issues/956


app/controllers/questions_controller.rb
```diff
@@ -90,8 +90,7 @@ def recently_commented
       .where(status: 1)
     @pagy, @questions = pagy(filter_questions_by_tag(@questions, params[:tagnames])
       .joins(:comments)
-      .order('comments.timestamp DESC')
-      .group('node.nid'), items: 24)
```
anirudhprabhakaran3, line 94: I have removed the `group` function from here for two reasons:

1. The `@questions` we are filtering on are already being grouped by `node.nid`.
2. The MySQL issue that is coming up was that ordering was happening on a variable on which grouping is not happening.

I believe that because of this, the behaviour of the code should not change, and we should get the same results.

anirudhprabhakaran3, line 94: I have removed the `group` function from here for two reasons:

1. The `@questions` we are filtering on are already being grouped by `node.nid`.
2. The MySQL issue that is coming up was that ordering was happening on a variable on which grouping is not happening.

I believe that because of this, the behaviour of the code should not change, and we should get the same results.

anirudhprabhakaran3:
Also commenting so that there is a track on the planning issue.

https://github.com/publiclab/plots2/issues/11185

anirudhprabhakaran3:
Hmm, I notice (nearly) all of the tests have failed. Interesting. Let me check what precisely the logs are,

anirudhprabhakaran3:
Looks like it is a CodeCov issue?
```
x> No token specified or token is empty
```

@jywarren could you help me out here, please? Thanks!

jywarren:
So strange. But CodeCove does get flaky sometimes. I'll try restarting it?

jywarren:
Ah wait, i now see this error:

```
Minitest::UnexpectedError:         ActiveRecord::StatementInvalid: Mysql2::Error: Table 'plots.answers' doesn't exist: SHOW FULL FIELDS FROM `answers`
```

jywarren:
I think we need to continue searching for where we depend on answers code! I mean, there are a lot: https://github.com/publiclab/plots2/search?q=answers

jywarren:
If you're not sure what to do in one of these places, feel free to leave a comment at the end of the line, and I can make a code suggestion inline in this PR!

app/controllers/home_controller.rb
```diff
@@ -145,7 +145,7 @@ def activity
     # group by day: http://stackoverflow.com/questions/5970938/group-by-day-from-timestamp
     comments = comments.group('DATE(FROM_UNIXTIME(timestamp))') if Rails.env == 'production'
     comments = comments.to_a # ensure it can be serialized for caching
-    answer_comments = Comment.joins(:answer, :user)
+    answer_comments = Comment.joins(:answer, :user) # Test failing here
```
anirudhprabhakaran3, line 148: I am not very sure what this code is doing, so I'm a bit stuck here.

test/functional/api/search_api_test.rb
```diff
@@ -178,7 +178,7 @@ def app
   end
 
   test 'search Tag Nearby Nodes functionality with a valid query and specific period' do
-    get '/api/srch/taglocations?nwlat=171.0&selat=0.0&nwlng=0.0&selng=174.8&sort_by=recent&order_direction=ASC&to=2018-08-16'
+    get "/api/srch/taglocations?nwlat=171.0&selat=0.0&nwlng=0.0&selng=174.8&sort_by=recent&order_direction=ASC&to=2018-08-16" # Very weird Test failing here
```
anirudhprabhakaran3, line 181: Yeah, this is failing, and I'm not very sure why. 
This is the error message:

```
ERROR SearchApiTest#test_search_Tag_Nearby_Nodes_functionality_with_a_valid_query_and_specific_period (131.05s)
Minitest::UnexpectedError:         ArgumentError: wrong number of arguments (given 2, expected 1)
            test/functional/api/search_api_test.rb:181:in `block in <class:SearchApiTest>'

```
No clue why this is happening.

test/functional/home_controller_test.rb
```diff
@@ -11,7 +11,7 @@ def setup
 
     get :home
     assert_response :success
-    assert_select "title", Sanitize.clean('&#127880;') + (" Public Lab: #{title}")
+    assert_select "title", Sanitize.clean('&#127880;') + (" Public Lab: #{title}") # Test failing here
```
anirudhprabhakaran3, line 14: This is giving a Nokogiri error, which I'm unable to debug.

```
ERROR HomeControllerTest#test_should_get_home (26.22s)
Minitest::UnexpectedError:         TypeError: wrong argument type Nokogiri::HTML5::DocumentFragment (expected Data)
            test/functional/home_controller_test.rb:14:in `block in <class:HomeControllerTest>'

```
Is this some version issue? The newer version experts some other type of data, and the tests haven't been updated to reflect that?

test/functional/wiki_controller_test.rb
```diff
@@ -484,7 +484,7 @@ def teardown
 
     assert_response :success
     assert_template :index
-    assert_select "title", Sanitize.clean('&#127880;') + (" Public Lab: Popular wiki pages")
+    assert_select "title", Sanitize.clean('&#127880;') + (" Public Lab: Popular wiki pages") # Test failing here
```
anirudhprabhakaran3, line 487: Similar Nokogiri issue here

test/unit/doc_result_test.rb
```diff
@@ -9,7 +9,7 @@ class DocResultTest < ActiveSupport::TestCase
       doc_type: 'QUESTIONS',
       doc_url: question.path(:question),
       doc_title: question.title,
-      score: question.answers.length
+      # score: question.answers.length    What should I do about this particular field?
```
anirudhprabhakaran3, line 12: For running locally, I've commented this out. But that isn't the right way to do it. What exactly is score? Should I replace this with the number of comments on a particular question?

test/functional/api/search_api_test.rb
anirudhprabhakaran3: Okay, I'm seeing this error in multiple test suites now, and I decided to actually go to the pages that these endpoints are hitting. The Passenger logs reveals the error: 
```
Completed 500 Internal Server Error in 44ms (ActiveRecord: 12.7ms)
  
ActiveRecord::ConfigurationError (Can't join 'Comment' to association named 'answer'; perhaps you misspelled it?):
  
app/controllers/home_controller.rb:154:in `activity'

```

I believe fixing that function will help in fixing all the corresponding errors.

app/controllers/home_controller.rb
jywarren: OK, here we can just remove the whole line, and lines 148-154, since those are all just collecting comments upon answers, and then also remove `answer_comments` from line 155 and 163. All comments are now just regular comments, so we don't need any of this! Thanks!

app/controllers/questions_controller.rb
jywarren: I think we are doing this to order it by the comment timestamps, but because nodes can have multiple comments, we had been seeing them appear more than once, so we re-added the group to ensure each node appears only once. But if you can confirm for sure that the duplication isn't happening, I like a simpler query!

app/views/home/home.html.erb
```diff
@@ -126,6 +125,6 @@
 </style>
 
 <% cache('feature_home-footer', skip_digest: true) do %>
-  <%= feature("home-footer") %>
```
jywarren, line 129: Hi, was this done incorrectly somehow? It might just be a feature we're not currently using. I think footer-notice is the topics table, so it might display twice if we add it here? But I could be wrong.

db/migrate/20220726161355_drop_answers.rb
```diff
@@ -0,0 +1,5 @@
+class DropAnswers < ActiveRecord::Migration[5.2]
+  def change
+    drop_table :answers
```
jywarren, line 3: yay!

test/functional/api/search_api_test.rb
jywarren: Great, good solving work!

test/unit/doc_result_test.rb
jywarren: Yes, comments sounds great! Answers were migrated to be comments.

test/functional/home_controller_test.rb
jywarren: Hmm, what version change are we making that could result in this? Maybe this will also just resolve because the home controller will render properly. But if not, that sanitized string is an emoji so maybe that's causing the problem. Let's try fixing the home controller first though.

app/controllers/questions_controller.rb
anirudhprabhakaran3: Hmm...on the data that I have I'm unable to see any duplication, but I think it is safer to keep it in that way then.

anirudhprabhakaran3:
All tests passing!! 🥳 🎉

app/views/home/home.html.erb
jywarren: Can you confirm this? Then I think we are ready to merge, thanks!

test/functional/home_controller_test.rb
anirudhprabhakaran3: Resolved now!

app/views/home/home.html.erb
anirudhprabhakaran3: Hey! So in the older ```home.html.erb```, this is there so I haven't changed it yet. Would you like me to remove this?

Reference:

https://github.com/publiclab/plots2/blob/5aa872006c56507b78b3641952a853f01e882b72/app/views/home/home.html.erb#L128-L130

test/functional/home_controller_test.rb
jywarren, line 14: Oops just need to remove this comment!

test/functional/wiki_controller_test.rb
jywarren, line 487: Here too!

jywarren (commented):
all good except 2 comments! Great work!

app/views/home/home.html.erb
```diff
@@ -1,7 +1,6 @@
 <% cache('feature_home-intro', skip_digest: true) do %>
   <%= feature("home-intro") %>
 <% end %>
-
```
jywarren, line 4: Ah i see in the final diff nothing has changed. Thank you, just checking! Merging!

jywarren:
Great work, a big PR!!!

PR #11332, opened 2022-08-05
Fixes #11331: Created random generator to fetch nodes randomly as cards

### LINKED ISSUE #11331: Create a Random Generator to fetch nodes randomly as cards

KarishmaVanwari:
Fixes #11331  

### DIFF
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -60,6 +60,11 @@ class WikiController < ApplicationController
     if !@node.nil? # it's a place page!
       @tags = @node.tags
       @tags += [Tag.find_by(name: params[:id])] if Tag.find_by(name: params[:id])
+      
+      tag1, tag2 = @node.normal_tags(:followers).includes(:tag).pluck(:name).first(2)
+ 
+      # get recommendations
+      @recommendations = Tag.get_recommendations(tag1, tag2)
     else # it's a new wiki page!
       @title = I18n.t('wiki_controller.new_wiki_page')
       if current_user
--- a/app/views/dashboard_v2/_sidebar.html.erb
+++ b/app/views/dashboard_v2/_sidebar.html.erb
@@ -56,18 +56,17 @@
   </div>
 </div>
 <div>
-<% if !@recommendations.nil? %>
+<% @recommendations ||= Tag.get_recommendations("blog", "three") %>
   <% @recommendations.each do |node| %>
-    <div class="card" style="width: 18rem;">
-    <img class="card-img-top" src="#" alt="Card image cap">
-    <div class="card-body">
-    <h5 class="card-title"><%= node.nid %></h5>
-    <p class="card-text">Post content here</p>
-    <a href="#" class="btn btn-primary">Read more button here</a>
-    </div>
+    <div class="card recommendations" style="width: 18rem;">
+      <img class="card-img-top" src="#" alt="Card image cap">
+      <div class="card-body">
+        <h5 class="card-title"><%= node.nid %></h5>
+        <p class="card-text">Post content here</p>
+        <a href="#" class="btn btn-primary">Read more button here</a>
+      </div>
     </div>
   <% end %>
-<% end %>
 </div>
 
 <style>
--- a/app/views/dashboard_v2/_sidebar.html.erb
+++ b/app/views/dashboard_v2/_sidebar.html.erb
@@ -55,7 +55,20 @@
     </div>
   </div>
 </div>
-
+<div>
+<% if !@recommendations.nil? %>
+  <% @recommendations.each do |node| %>
+    <div class="card" style="width: 18rem;">
+    <img class="card-img-top" src="#" alt="Card image cap">
+    <div class="card-body">
+    <h5 class="card-title"><%= node.nid %></h5>
+    <p class="card-text">Post content here</p>
+    <a href="#" class="btn btn-primary">Read more button here</a>
+    </div>
+    </div>
+  <% end %>
+<% end %>
+</div>
 
 <style>
 
--- a/test/unit/tag_test.rb
+++ b/test/unit/tag_test.rb
@@ -298,5 +298,6 @@ class TagTest < ActiveSupport::TestCase
   test 'Tag.get_recommendations' do
     nodes = Tag.get_recommendations("blog", "test")
     assert_not_nil nodes
+    assert nodes.length > 0
   end
 end
--- a/test/unit/tag_test.rb
+++ b/test/unit/tag_test.rb
@@ -294,4 +294,9 @@ class TagTest < ActiveSupport::TestCase
     assert lat.location_tag?
     assert lon.location_tag?
   end
+
+  test 'Tag.get_recommendations' do
+    nodes = Tag.get_recommendations("blog", "test")
+    assert_not_nil nodes
+  end
 end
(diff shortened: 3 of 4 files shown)


app/controllers/wiki_controller.rb
```diff
@@ -60,6 +60,11 @@ def show
     if !@node.nil? # it's a place page!
       @tags = @node.tags
       @tags += [Tag.find_by(name: params[:id])] if Tag.find_by(name: params[:id])
+      
+      tag1, tag2 = @node.normal_tags(:followers).includes(:tag).pluck(:name).first(2)
+ 
+      # get recommendations
+      @recommendations = Tag.get_recommendations(tag1, tag2)
```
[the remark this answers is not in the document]
jywarren: What if we put it in a reusable named methodin application_controller so it can be used in multiple places?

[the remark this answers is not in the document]
KarishmaVanwari: I tried placing this piece of code on line 54 in `home_conroller` but it gives the following error - 
`NoMethodError in HomeController#dashboard_v2`
`undefined method normal_tags for nil:NilClass `. 

@TildaDares @jywarren

jywarren:
Ah I believe it's using a private method from the bottom of the previous
controller. You could maybe move that method into the application
controller, which would make it available to all controller methods!

KarishmaVanwari:
Hi @jywarren, where do I need to shift the method since when I shift the `dashboard_v2` method in `application_controller`, it gives an error saying `The action dashboard_v2 could not be found for HomeController`? So maybe I need to add it in an already existing method in `application_controlller`, but which one -- how do I figure it out?

app/views/dashboard_v2/_sidebar.html.erb
```diff
@@ -55,7 +55,20 @@
     </div>
   </div>
 </div>
-
+<div>
```
jywarren, line 58: could we move this into its own partial - to load with `<%= render partial: "nodes/random" %>` and the partial template would be `_random.html.erb`.

app/controllers/wiki_controller.rb
```diff
@@ -60,6 +60,11 @@ def show
     if !@node.nil? # it's a place page!
       @tags = @node.tags
       @tags += [Tag.find_by(name: params[:id])] if Tag.find_by(name: params[:id])
+      
+      tag1, tag2 = @node.normal_tags(:followers).includes(:tag).pluck(:name).first(2)
```
jywarren, line 64:
```suggestion
      puts @node.inspect
      tag1, tag2 = @node.normal_tags(:followers).includes(:tag).pluck(:name).first(2)
```

app/views/dashboard_v2/_sidebar.html.erb
```diff
@@ -55,7 +55,20 @@
...
+<% if !@recommendations.nil? %>
```
jywarren, line 59:
```suggestion
<% @recommendations ||= Tag.get_recommendations("water-quality", "air-quality") %>
```

jywarren:
Just commenting here with notes from our call before, suggested pseudocode for making remote calls to open sub-topics:

```ruby
Tag…..each do |subtopic|

    <script>
      $(‘#expand-btn-<%= index %>’).click(function(e) {
        $.ajax(‘/tags/subtopics?subtopic=<%= subtopic %>’).done(function(response) {
            $(‘#contents-UNIQ’).append(response.responseText);
        });
      });
    </script>

end
```

Controller:

```ruby
def subtopics
  Tag.find…… params[:subtopic]
  render “tags/subtopic”, layout: false
end
```

KarishmaVanwari:
Hi @jywarren, I've opened a PR #11380. Thanks!

app/views/dashboard_v2/_sidebar.html.erb
KarishmaVanwari: @jywarren, I tried this out, but it still doesn't seem to work as I am unable to find cards as expected on the dashboard.

jywarren:
My apologies for mixing these all up, but i hid the subtopics discussion from this thread to try to clear it up. Looking at the issues here now!

jywarren:
OK, so I had to make some edits to the dashboard code to make it work, but ultimately I found that `Tag.get_recommendations("blog", "three")` is not returning anything (i used "blog" because that's in the db seeds file so it exists).

So let's debug why that's not working. I think we may want to write a unit test for that code. That will let us test `Tag.get_recommendations("blog", "three")` in isolation and ensure it returns the right thing. Why don't I write the unit test and you can work on figuring out why it's not returning anything!

test/unit/tag_test.rb
```diff
@@ -294,4 +294,9 @@ def setup
     assert lat.location_tag?
     assert lon.location_tag?
   end
+
+  test 'Tag.get_recommendations' do
```
jywarren, line 298: OK, here's the unit test. Let's ensure it returns something! We could even test how many nodes it returns, or something else. But most importantly that it returns nodes at all!

test/unit/tag_test.rb
```diff
@@ -294,4 +294,9 @@ def setup
...
+    nodes = Tag.get_recommendations("blog", "test")
+    assert_not_nil nodes
```
jywarren, line 300: Here we are asserting that we are returning SOMETHING. Next we should be sure that the model code for `get_recommendations` actually does return something, and if not, debug that. Does that make sense?

jywarren:
Generally unit tests test individual functions in isolation, not in the context of a big running application all at once. They're useful for finding narrow reasons things may have gone wrong, and that individual pieces function as expected. 

By contrast, integration tests, or system tests, are "full-stack" tests which start up the whole application and help us see how things might go wrong when everything is trying to work in synchrony, all at once. 

Both types of test -- the narrow ones and the broad ones -- are useful for catching errors. Sometimes both get tripped! But I hope that way of thinking of different kinds of tests makes sense!

test/unit/tag_test.rb
jywarren: Now that I added this test, we are actually seeing it return success (not failure!) in the test logs. You can click on "Details" for Unit tests below to see the output. Here's a deep link:

https://github.com/publiclab/plots2/runs/8193928002?check_suite_focus=true#step:5:90

What I wonder though, is if we are getting [] back -- and i just checked, `[].nil?` evaluates to `false` in ruby! So, we may need to test `nodes.length > 0` to be sure SOMETHING is returned and not an empty array.

jywarren: Let's try!

```suggestion
    assert_not_nil nodes
    assert nodes.length > 0
```

test/unit/tag_test.rb
jywarren: OK, see below where I made a further suggestion about the unit test: https://github.com/publiclab/plots2/pull/11332#discussion_r964248316

test/unit/tag_test.rb
```diff
@@ -294,4 +294,10 @@ def setup
...
+    assert nodes.length > 0
```
jywarren, line 301: Aha! This assertion failed! So, looks like we have been getting an empty array back:

https://github.com/publiclab/plots2/runs/8217990567?check_suite_focus=true#step:5:85

```
TagTest#test_Tag.get_recommendations (23.30s)
        Expected false to be truthy.
        test/unit/tag_test.rb:301:in `block in <class:TagTest>'
```

jywarren: Great, so now, we can try to fix that. Once we push a fix up as a commit to this PR, the test will start to pass again -- that's actually "Test Driven Development" -- we set up a specific enough test condition (or "assertion") that our test will pass once it's working properly. The test actually precedes the final solution!

PR #11338, opened 2022-08-08
Added unique ids for comments table row

Manaswee03:
I have successfully resolved the issue of "Adding unique ids for comments table below" by making the required changes in the _comments.html.erb file.

edit


Manaswee03:
Thank you @TildaDares, your support means a lot. I learned many new things and the credit goes to you! Just one last favor, as mentioned earlier I need some recognition in order to show a proof to my professor. So, can you please tweet mentioning my handle?

PR #11380, opened 2022-09-02
Fixes #11379: Made the subtopics method

### LINKED ISSUE #11379: Make the subtopics method

KarishmaVanwari:
Fixes #11379 


config/routes.rb
```diff
@@ -383,6 +383,7 @@
   post 'comment/react/update/:id' => 'comment#react_update'
 
   get 'topic-tree' => 'tag#topic_tree'
+  get 'tags/:subtopic' => 'tag#subtopics'
```
jywarren, line 386:
```suggestion
  get 'tags/subtopic/:topic' => 'tag#subtopics'
```
Here, Rails routing takes some getting used to - the parts with `:symbol`-style prefixes will capture anything written there in the URL, while all other text becomes literally part of the URL: (like `tags/subtopic/dinosaurs' will match `tags/subtopic/` but then will capture `dinosaurs` and create a variable called `params[:dinosaurs]` which you will then be able to reference from the controller method).

By setting `:topic` here, we give a useful name to the parameter it will capture from the URL.

jywarren:
Hi @KarishmaVanwari ok, so, this actually relates to this PR, on topic trees: https://github.com/publiclab/plots2/pull/11230#discussion_r908539690

I'm so sorry i mixed these up. But I've made some suggestions and explanations to refine this PR a bit, and you have to expand the "Show resolved" sections above to read them -- please do! Let me know if they make sense?

app/controllers/tag_controller.rb
```diff
@@ -568,4 +568,8 @@ def fetch_counts
   end
 
   def topic_tree; end
+
+  def subtopics
+    Tag.find_by(name: params[:subtopic])
```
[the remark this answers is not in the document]
jywarren: This had a little more explanation in this comment: https://github.com/publiclab/plots2/pull/11230#discussion_r908539690

app/controllers/tag_controller.rb
```diff
@@ -568,4 +568,10 @@ def fetch_counts
   end
 
   def topic_tree; end
+
+  def subtopics
+nodes = Tag.find_by(name: "parent:#{params[:topic]}") # here, we look for this special tag marking a node as having a parent
+  .nodes
+  .collect(&:slug) # collect the "slugs" i.e. the unique part of the URL, which should correspond to a tagname
+    render "tags/subtopic", layout: false
```
jywarren: OK, so here, i think we also have to actually make the partial template, at: `/app/views/tags/_subtopics.html.erb`, and for now it can just print out the subtopics. If you'd like, you can look at the HTML you made in #11230 and use that?

KarishmaVanwari:
> Hi @KarishmaVanwari ok, so, this actually relates to this PR, on topic trees: [#11230 (comment)](https://github.com/publiclab/plots2/pull/11230#discussion_r908539690)
> 
> I'm so sorry i mixed these up. But I've made some suggestions and explanations to refine this PR a bit, and you have to expand the "Show resolved" sections above to read them -- please do! Let me know if they make sense?

Hi @jywarren, I've gone through the explanations and understood my shortcomings and the respective explanations. Thank you so much Jeff for the detailed explanation and for bearing with me while I try to figure out stuff, thanks again! ❤️

app/controllers/tag_controller.rb
KarishmaVanwari: @jywarren : 
I created the file `_subtopics.html.erb` with the following code so as to only print the subtopics when pointed at a URL say `http://localhost:3000/tags/subtopic/one` or `http://localhost:3000/tags/subtopic/two`:

```ruby
<% @subtopics.each do |subtopic| %>
    <h1><%= subtopic %></h1>
<% end %>
```

However, it gives an error -

image

Is it because the tags "one" or "two" do not have any specified subtopics as such? Let me know if I am wrong. 

Thanks!

jywarren:
Hi, Karishma - i'm kind of surprised that your local doesn't show a full
development error log, which would highlight the line number of the error.
How are you starting your development server? And, even if we don't see it
in the browser debug output, we should be able to look in the console logs
to find the exact reason it failed. Let's look!

app/controllers/tag_controller.rb
```diff
@@ -568,4 +568,13 @@ def fetch_counts
...
   def topic_tree; end
+
+  def subtopics
+    # here, we look for this special tag marking a node as having a parent
+    # collect the "slugs" i.e. the unique part of the URL, which should correspond to a tagname
+    subtopic_names = Tag.find_by(name: "parent:#{params[:topic]}")
+      .nodes
+      .collect(&:slug)
+    @subtopics = Tag.find_by(name: subtopic_names)
+    render "tags/subtopics", layout: false
```
jywarren, line 579:
```suggestion
    render "tags/subtopics", layout: false
  end
```

jywarren:
Aha - the subtopics method is in the `private` area of the controller. I've made changes and will push them in a moment. 

As much as possible please do push up your local changes to the PR so I can see exactly what you're working with when i open it. Thanks!

app/controllers/tag_controller.rb
```diff
@@ -525,6 +525,17 @@ def comments
     render 'show'
   end
 
+  def subtopics
+    # here, we look for this special tag marking a node as having a parent
+    # collect the "slugs" i.e. the unique part of the URL, which should correspond to a tagname
+    subtopic_names = Tag.find_by(name: "parent:#{params[:topic]}")
+      &.nodes
+      &.collect(&:slug)
```
jywarren, line 533: OK, I pushed a couple commits to fix minor issues. See how I moved this up out of the "private" section.

Also I added `&.` which fails gracefully if the value is nil, instead of erroring - it just doesn't continue trying. 

I think it should work now, but I think it's worth writing a functional test for this. That will mean adding a tag to a wiki page with name `parent:TAGNAME` such that the code actually gets used. 

The reason is that for most tags we might test this with, there won't be any subtopics. We need to create some test data to make that situation appear so that we can write a test around it. Make sense? I'm happy to help with the test data, let me know if you'd like some support on that!

jywarren: And great work getting this far. It's complex!!

KarishmaVanwari: @jywarren 
I tried opening the page: `/tags/subtopic/one`, I am able to successfully render the page (blank although), and I can render topic name also on the page by extracting it from the URL. I created a wiki page with tag `parent:one` but I am still unable to fetch subtopics on the page. I doubt if `@subtopics` actually carries the subtopics in it. 

Also, a mention, on line 531 here, we have  

```ruby
Tag.find_by(name: "parent:#{params[:topic]}")
```

Here, tag is compared with `parent:#{params[:topic]}`, right? So I also tried creating a wiki page with tag `parent:#one` but still can't fetch it on the `/tags/subtopic/one` page. Conversely, I also tried to change the line 531 to this - 

```ruby
Tag.find_by(name: "parent:{params[:topic]}")
```

and created a wiki page with tag `parent:one`, but still I can't obtain desired result.

KarishmaVanwari: So, do you think I have this test data (the wiki pages I created), so as to write a test? Also, I am totally new to writing tests, could you please guide me with it? PS: Sharing resources also would do.

Thanks!

jywarren: Hi @KarishmaVanwari ok so you are doing really well then, if it rendered a blank page. We just need to debug the controller. 

A controller test is known as a functional test -- there is documentation here, but it's pretty dense: https://guides.rubyonrails.org/testing.html#functional-tests-for-your-controllers

I might start by looking at other tests:

https://github.com/publiclab/plots2/blob/4204a62b5bfff77f0e2bb3ce3bcb68a4b3a6d984/test/functional/tag_controller_test.rb#L206-L214

And at how we create test data -- by creating entries in https://github.com/publiclab/plots2/blob/main/test/fixtures/

For tags, we'll have to create not only a tag fixture entry in https://github.com/publiclab/plots2/blob/main/test/fixtures/tags.yml  

but also a record to link that tag to a node: https://github.com/publiclab/plots2/blob/main/test/fixtures/node_tags.yml

Does that make sense? Let's start with that, can you add that to this PR? Once it's in place, I can help you write a simple test, then narrow the test assertions until we find what was wrong with our assumptions. How does that sound?

jywarren: And, as far as I can tell, your `parent:one` tag should work! And here is hopefully a helpful tip - the `"line of text with #{something} in the middle"` syntax is a way to insert Ruby code into a string. It must use " double quotes, and the inserted Ruby code to be evaluated is surrounded by `#{  }`. It's a little weird looking but it is smoother than adding a bunch of separate strings together.

KarishmaVanwari: @jywarren : I created test data in https://github.com/publiclab/plots2/blob/main/test/fixtures/tags.yml and https://github.com/publiclab/plots2/blob/main/test/fixtures/node_tags.yml. Let me know if I got that right. Thanks!

test/fixtures/node_tags.yml
```diff
@@ -284,3 +284,9 @@ sub_tag:
   tid: 37
   nid: 1
   uid: 2
+
+subtopic_tag:
+  tid: 38
+  uid: 2
+  nid: 41
```
jywarren, line 291: This looks good, but it's applied to this page, which is a note, not a wiki:

https://github.com/KarishmaVanwari/plots2/blob/subtopics/test/fixtures/nodes.yml#L497

Will that affect the logic, and possibly cause it not to return things?

jywarren: Well, it should get the slug for this note, and try to use it as a subtopic tagname, according to this logic:

https://github.com/publiclab/plots2/pull/11380/files#diff-5d60373322d4baf080e5f3957fc30d0e8058337c202f5641e1775402776428b1R533

Aha - yes, so, then it takes the slug from that page, which is "jeff-12-20-2020-ouija-board" - and tries to find a tag with that name. It doesn't find one, so it returns null. 

Let's change this to instead point at this node, which is a page, with slug `about`. Then let's add a new tag with the name `about` so that there is actually a tag that can be returned. 

We could give these more realistic tag names so it makes more sense... but, maybe it's not worth it for now. Let's just get it working first.

```suggestion
  nid: 2
```

So now you'll also have to add a new tags.yml entry, but it doesn't need a node_tags.yml entry, since we only look if the tag exists, not if we used it anywhere. Let's see how that works!

app/controllers/tag_controller.rb
```diff
@@ -525,6 +525,17 @@ def comments
...
+    @subtopics = Tag.find_by(name: subtopic_names)
```
cesswairimu, line 534: I was trying this locally and this line returns nil,,, the `subtopic_names` variable above returns  slugs of the nodes' title downcase & hyphenated -- from the comment seems we are expecting this to be tagname slugs that maybe the issue

PR #11387, opened 2022-09-04
Migrate DrupalFile content to Revision of node

### LINKED ISSUE #4076: Deprecate DrupalFile and DrupalUpload tables and remove associated code to simplify codebase
As part of the #956 work, we should move the records for all DrupalFile entries into the node revision body. This will be similar to https://github.com/publiclab/plots2/issues/4072, where we're getting rid of needlessly customized infrastructure and database tables, and just building the content into the node body content. 

So we'd run a migration that'd run on each node, which would fetch `node.files` and for each one, add a link to the bottom of the node content, maybe in the format:

```
* https://...link here
* https://...link here
* https://...link here
```

This will allow us to remove these files:

https://github.com/publiclab/plots2/blob/master/app/models/drupal_file.rb

https://github.com/publiclab/plots2/blob/master/app/models/drupal_upload.rb

And make a migration to drop the corresponding tables.  

We'd also get rid of these lines:

https://github.com/publiclab/plots2/blob/master/app/models/node.rb#L79

https://github.com/publiclab/plots2/blob/1ee45090f0f1d56f4a8cb196bb719d1ae68af5d6/app/models/node.rb#L228-L230

We'd love help with this!

anirudhprabhakaran3:
Fixes #4076 

Ref: #11185

### DIFF
--- a/db/migrate/20220904060656_migrate_drupal_files_to_node_revision.rb
+++ b/db/migrate/20220904060656_migrate_drupal_files_to_node_revision.rb
@@ -1,0 +1,14 @@
+class MigrateDrupalFilesToNodeRevision < ActiveRecord::Migration[5.2]
+  def change
+    Node.all.each do |node|
+      new_revision = node.revision.last
+      file_links = "<ul>"
+      node.files.each do |file|
+        file_links << "<li><a href='#{file.filepath}'>#{file&.filename}</a></li>"
+      end
+      file_links << "</ul>"
+      new_revision.body << file_links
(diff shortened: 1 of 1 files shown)


anirudhprabhakaran3:
For this issue as well, should I keep the migration and deletion of tables as different MRs?

jywarren:
Yes please!

jywarren:
Yes I believe so I think it's better to confirm it's working before we do a
destructive database action. Thanks!

PR #11444, opened 2022-09-28
Update images_controller.rb

yaegareren:
Updated images_controller.rb


TildaDares:
Hi @yaegareren, contributors are only allowed to work on one first-timers only issue. Please take a look at other issues without the `first-timers-only` label for something to work on. Thank you!

PR #11869, opened 2026-03-24
Fix: subheading hidden under fixed header

### LINKED ISSUE #11052: UI update: making people's boxes equal in height
Hello community members. While navigating through the Publiclab.org website I stumbled upon the people's page. The page is very well drafted. however the div containing each users details are visually not equal in height due to badges assigned to each user.
To give you all a better idea, please find attached a photo explanation of the above issue.

image

An inspection of the above shows that the div are of equal size. but the cards inside it are not utilizing 100% height of their div. proof attached below

image

This can be fixed by making changes in the stylesheet file. For example I would like to propose we can specify height attribute to "peoplecard" class and its value can be set to 100%. I have attached a reference photo to show how the website will look after making the changes. for the sake of eg, I have specified height attribute to "card" class. In future this can be avoided to prevent unwanted changes across the overall website as class "card" seems to be more general.

image

I would suggest doing this change to bring about a touch up of symmetricity on the website. Do let me know your thoughts on the same and suggest changes if required. waiting for your feedback. Thank you 😄

Nazina-M-19:
Fixes #11052 

summary:
     - Ensured all project cards have equal height
     - used flexbox for consistent layout.


Nazina-M-19:
Hi, I have implemented a fix using flexbox to ensure equal height for projrct cards. Kindly review. Thank you!

PR #11875, opened 2026-04-27
Fix crash when editing profile for non-existent user

mowems:
### Problem
Accessing `/profile/:id/edit` with a non-existent username caused a server error due to `@user` being nil.

### Solution
Added a guard clause in `UsersController#edit` to handle cases where the user is not found.

### Result
The application now redirects safely with an error message instead of crashing.

### Testing
- Visited `/profile/fakeuser/edit`
- Confirmed no server error occurs

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -81,10 +81,14 @@ class UsersController < ApplicationController
   def edit
     @action = "update" # sets the form url
     @user = if params[:id] # admin only
-              User.find_by(username: params[:id])
-            else
-              current_user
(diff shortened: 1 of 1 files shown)


mowems:
It looks like the CI failures might be coming from deprecated GitHub Actions (for example `actions/upload-artifact@v2`), rather than anything introduced in this PR.

The changes here are focused on handling non-existent users in the profile edit flow, and everything is working as expected locally.

Happy to take a look at the workflow files and open a separate PR to help fix the CI if that would be useful.

PR #11879, opened 2026-06-13
fix: open all links in new tab in embed view - fixes #4214

### LINKED ISSUE #4214: Links in embeddable grids should open new browser tab
(heavily edited by @jywarren)

Embeds allow embedding of inline grids, and are available when clicking the faint grey link to the bottom right of each grid:

image

The generated iframe uses a URL like: https://publiclab.org/embed/grid/nodes:purpleair

Once embedded on another site, when you click links or buttons, the iframe loads an entire instance of the plots2 website within the iframe. It can be quite cumbersome on the browser and the user could get confused very quickly. 

Here's an example of this issue:
ezgif com-video-to-gif 3

Link:
https://store.publiclab.org/collections/featured-kits/products/lego-spectrometer-kit?variant=8187504787563

## Solution:

We should ensure all links in the `/embed/...` views are adjusted to have `target="_blank"` so they open in a new window/tab. We could do this in JavaScript; we'd have to detect the controller action with `params[:action] == 'gridsEmbed'`, because they're run out of:

app/controllers/tag_controller.rb:  def gridsEmbed

We'll want to only do this in the embed view, but not for normal inline grids code. The templates are in:

https://github.com/publiclab/plots2/tree/master/app/views/grids/

Any help appreciated!!!

faratabassum09-a11y:
## What does this PR do?
Fixes #4214

## Changes made
- Added JavaScript to `app/views/tag/gridsEmbed.html.erb` that sets `target="_blank"` and `rel="noopener noreferrer"` on all anchor tags
- This ensures links in embedded grids open in a new tab instead of loading inside the iframe
- The fix only applies to the embed view, not normal inline grids

## Checklist
- [x] Tested the change locally
- [x] No existing functionality broken
- [x] The fix is scoped only to the embed view


faratabassum09-a11y:
Hi maintainers! 
I'm a new contributor working toward GSoC 2027.
Could someone please review this PR when they get a chance?
The failing tests appear to be pre-existing Ruby/Rails test failures 
unrelated to my JavaScript change in gridsEmbed.html.erb.
Thank you!

faratabassum09-a11y:
This PR fixes #4214

Adds JavaScript to gridsEmbed.html.erb to ensure all links 
open in a new tab when viewed in embed mode, preventing 
the iframe from loading entire pages within itself.

